PR #6, opened 2022-10-17
Fixed almost all warnings

sonicdcer:
There are a few that i couldn't wipe because of fake matching issues.


Xeeynamo:
very much needed. Thanks!!

PR #8, opened 2022-10-17
func_801B2830 NON_MATCHING

sonicdcer:
this may be the lack of aspsx fault
image

https://decomp.me/scratch/sKMmw


Xeeynamo (changes requested):

src/st/np3/3246C.c
```diff
@@ -44,7 +44,38 @@ void EntityCandle(Entity *entity) {
...
+typedef struct {
+    /* 0x00 */ char pad00[0x2C];
+    /* 0x2C */ u16 unk2C;
+    /* 0x2E */ char pad2E[0x4E];
+    /* 0x7C */ s8 unk7C;
+    /* 0x7D */ s8 unk7D;
+    /* 0x7E */ s8 unk7E;
+} UnkStruct11; // size = 0x7F
+
+void func_801B2830(UnkStruct11 *arg0) {
```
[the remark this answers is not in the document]
sonicdcer: i'm missing one member in entity (unk7D), i can turn the u16 unk7C into two u8s, but functions func_8016B8E8 and func_8016B92C inside RIC break up and i'm unable to fix them, seems unk7C has to be u16 from the looks of it:

image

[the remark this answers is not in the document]
sonicdcer: i think i'm gonna solve this with another union after this gets merged (to avoid Entity struct conflicts)

include/dra.h
```diff
@@ -123,8 +132,8 @@ typedef struct {
     /* 0x70 */ s32 unk70;
     /* 0x74 */ s32 unk74;
     /* 0x78 */ s32 unk78;
-    /* 0x7C */ u16 unk7C;
-    /* 0x7E */ u16 unk7E;
+    /* 0x7C */ unkUnion3 unk7C;
+    /* 0x7E */ u8 unk7E;
```
Xeeynamo, line 136: Is the `u8` intentional? It should be either `u16 unk7E;` or `u8 unk7E; u8 unk7F;`

src/ric/1AC60.c
```diff
@@ -343,9 +343,15 @@ INCLUDE_ASM("asm/ric/nonmatchings/1AC60", func_8015F96C);
 
 INCLUDE_ASM("asm/ric/nonmatchings/1AC60", func_8015F9F0);
 
-#ifndef NON_MATCHING
-// This function matches with GCC 2.7.2 using -01 compilation flags
+<<<<<<< HEAD
```
Xeeynamo, line 346: 🔕 git is getting furious 😂

Xeeynamo (changes requested):

include/dra.h
sonicdcer: yes, because unkUnion3 has three u8s or one u16, in this last case the offset will keep in the correct position because the compiler will fill the alignment pads automatically

sonicdcer: with s16 unk7E, the real 0x7E was being moved to 0x7F because the u16 didn't fit

sonicdcer: image

Xeeynamo (approved):

PR #17, opened 2022-10-24
more NO3 simple functions and libpgu.h comments

frog404:


src/st/no3/377D4.c
```diff
@@ -5,25 +5,33 @@ void ReplaceCandleWithDrop(Entity*);
...
+extern u16 D_80180E80[];
+extern s16 D_801D7D2A;
+extern u16 D_801D7D2C;
+extern s16 D_801D7D2E;
+extern s16 D_801D7D30;
+extern s16 D_801D7D32;
+extern s16 D_801D7D36;
 extern s16 D_801D7D60;
 extern s16 D_801D7D62;
+extern s16 D_801820C4[];
```
Xeeynamo, line 28: We should aim to order the field as their offsets. That includes renames variables such as `PfnEntityUpdates`. This will helps to understand better the layout of the overlay we are trying to decompile.

src/st/no3/377D4.c
```diff
@@ -475,12 +492,20 @@ Entity* AllocEntity(Entity* start, Entity* end) {
     return NULL;
 }
 
-INCLUDE_ASM("asm/st/no3/nonmatchings/377D4", func_801C5534);
+#ifndef NON_MATCHING
+INCLUDE_ASM("asm/st/no3/nonmatchings/377D4", func_801C5534); // not mathching only locally: https://decomp.me/scratch/6ZStN
```
Xeeynamo, line 496: Please do not put comments at the same line of `INCLUDE_ASM` as `make format` will mess up the line and `make extract` will break. Put the comments above `INCLUDE_ASM`

Xeeynamo (changes requested):
Hey I requested some minor changes. The decompiled code looks good!

src/st/no3/377D4.c
frog404: I'm affraid I'm not sure to understand what you mean by that, sorry. Could you elaborate please ?

Xeeynamo: For example `s16 D_801D7D62;` should be after `s16 D_801820C4[];`

frog404: from what i understand we should order strictly by variable name, so i commited this:
is that correct , or should we reorder by type first and then name of var alphabetical second ?
image

PR #18, opened 2022-10-27
mostly DRE and some MAD, WRP and RIC functions

sonicdcer:


include/dra.h
```diff
@@ -459,7 +492,7 @@ extern s32 g_playerLevel;
 extern s32 g_playerExp;
 extern s32 g_playerGold;
 extern s32 g_killCount;
-extern s32 g_playerHeart;
+extern playerHeart g_playerHeart[];
```
Xeeynamo, line 495: `g_playerHeartMax` (0x80097BAC) comes right after `g_playerHeart` (0x80097BA8). So either  `g_playerHeart` is really a `s32` or we need to replace `g_playerHeartMax` with `g_playerHeart->unk4`. On top of that the fields of `struct playerHeart` might be renamed as `current` and `max` if we decide to pursue that route.

include/dra.h
```diff
@@ -533,7 +566,7 @@ extern s32 D_801375FC;
 extern s32 D_80137614;
 extern s32 D_80137618;
 extern u8 D_8013761C[]; // can't use `extern MenuContext D_8013761C[];` as it's
-                        // 2-byte aigned
+                        // 2-byte signed
```
Xeeynamo, line 569: Good catch of that mis-spelling. But I think it was `aligned` and not `signed`

include/macros.h
```diff
@@ -1,2 +1,11 @@
...
-#define ABS_ALT(x) ((x) < 0 ? -(x) : (x))
\ No newline at end of file
+#define ABS_ALT(x) ((x) < 0 ? -(x) : (x))
+#define DECR(x) ((x) == 0 ? 0 : --(x))
+
+#define CLAMP(x, min, max) ((x) < (min) ? (min) : (x) > (max) ? (max) : (x))
+#define CLAMP_MAX(x, max) ((x) > (max) ? (max) : (x))
+#define CLAMP_MIN(x, min) ((x) < (min) ? (min) : (x))
+
+#define ARRAY_COUNT(arr) (s32)(sizeof(arr) / sizeof(arr[0]))
```
Xeeynamo, line 10: I find the name `ARRAY_SIZE` as slightly more appropriate

src/st/dre/11A64.c
```diff
@@ -134,79 +448,459 @@ s16 func_8019A718(void) {
...
 
+// This function matches with PSYQ4.0 GCC 2.7.2 with -01 and -02 Optimization
+// at -v0 reg swap
+// flags https://decomp.me/scratch/dlcph
+#ifndef NON_MATCHING
 INCLUDE_ASM("asm/st/dre/nonmatchings/11A64", func_8019AC78);
+#else
+extern s16 D_80180D80[];
+
+s32 func_8019AC78(u8 arg0, s16 arg1) { return D_80180D80[arg0] * arg1; }
```
Xeeynamo, line 488: Pretty sure if you merge `master` to this one, you can have a match here and at `func_8019ACA4` thanks to the recent implementation of `ASPATCH`!

Xeeynamo (changes requested):
Mindblowing work!!! I just left some minor comments, but overall crazy amount of work in here. Good job!

include/macros.h
sonicdcer: I'm sorry but i don't agree, the MACRO "counts" how many "whatevers" are in the array based on it's size

Xeeynamo (approved):

PR #21, opened 2022-10-31
refactor a few 377D4 funcs

Rainchus:
refactors a few goto loops to be while (true) loops instead


Xeeynamo (approved):
LGTM! I am merging

PR #25, opened 2022-11-15
[TOOL] edit  find_duplicates.py to autogenerate tags for duplicate functions

frog404:


frog404:
This should behave like before, with a few edge cases fixed, and also another argument "--allcomment" that instead of dumping the duplicates in a .txt file, comment the .c files directly. 
I know this can be dangerous/not desired to edit C files directly so, i added a prompt that ask for user to confirm she/he is sure to proceed.

Xeeynamo:
It work magically. Thanks!

PR #26, opened 2022-11-16
[CEN] cleared obvious duplicates

frog404:
used the dupes tool to spot the low hanging fruit in CEN


Xeeynamo (approved):

Xeeynamo:
@frog404 A suggestion for next time is to also add the symbols to let function have proper names :)

PR #28, opened 2022-11-30
[TOOL] endianess fix, cleaning the .txt format, adding two args modes (comment, uncomment), various fixes

frog404:


Xeeynamo (changes requested):
Before merging, it would be ideal to enable this tool to continuously report duplicated functions through our pipeline. It was previously broken due to some python error and I am curious if we can bring it back. The lines to uncomments are located here: https://github.com/Xeeynamo/sotn-decomp/blob/master/.github/workflows/report.yml#L59

Xeeynamo (approved):

Xeeynamo:
I am merging it into a separate branch to test the automated dups functionality before merging to master

PR #29, opened 2022-12-09
Several DRA functions and Readme update

sonicdcer:


include/dra.h
```diff
@@ -433,6 +444,7 @@ extern s8 D_8005436D;
 #define LBA_STAGE_NP3_VH 0x91FF
 #define LBA_STAGE_NP3_BIN 0x9235
 
+extern s32 aBu1d1dS;
```
Xeeynamo, line 447: I think you can change the type of this from `s32` to `const char aBu1d1dS[]`

Xeeynamo (approved):
I left a single comment. But overall you did an incredible job in there 👏 . Thank you!!

PR #33, opened 2022-12-30
mostly renamings, one func

sonicdcer:
Matched func_8010E27C from previous NON_MATCHING state

Entity struct members:
facing, hitPoints,

It's worth mentioning that those names are not definitive, but rather intended to make up some context while decompiling functions, more accurate names will come once we decomp more.

NZ0 function names:
EntityTransitionDoor
EntityBloodSkeleton

I also updated the permuter configuration yaml file but there's still work left to do with it for the decomp.me import function to work properly.


Xeeynamo:
All very good! How were you able to find the BloodySkeleton?
PS. I see the GH pipeline says it's not matching, I'll wait it to pass before merging

PR #42, opened 2023-01-09
Decompile func_800E9610

sozud:
Decompile func_800E9610

### DIFF
--- a/include/dra.h
+++ b/include/dra.h
@@ -1186,6 +1186,7 @@ extern s16 D_8013B698;
 extern s16 D_8013AEE0;
 extern u8 D_8013AEEC;
 extern s32 D_8013B158;
+extern Unkstruct_8013B160 D_8013B160[];
 extern s32 D_8013B3D0;
 extern s16 D_8013B3E8[];
 extern s32 D_8013B5E8;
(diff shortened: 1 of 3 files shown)


include/dra.h
```diff
@@ -621,6 +621,10 @@ typedef struct {
     /* 8003C8B4 END*/
 } GameEngine; /* size=0x140 */
 
+typedef struct {
+    u8 data[0x278];
+} UnkStruct11;
+
```
sonicdcer, line 627: Hello and welcome to the project!
Just a small nit to keep our project structure consistent.

This struct should go inside unkstruct.h as we are in the process of identifying and merging them all, work it's being done while we import real data to figure out the rest of the project structure so we are putting them in that file in the meantime. It's name should be Unkstruct_8013B160 to state which symbol it corresponds to. We also state the name in both ends in case there's a pointer to itself which we don't know.
```suggestion
typedef struct Unkstruct_8013B160  {
    u8 data[0x278];
} Unkstruct_8013B160;
```
I would also recommend to read docs/style.md to familiarize yourself to our style :)

Thank you for contributing to this Dream project, we are happy to see you working among us!

sonicdcer (changes requested):

include/dra.h
sozud: Thanks, I believe the comments are addressed now. If it looks good I can rebase.

sonicdcer: Looks good to me!

Xeeynamo (approved):
Thank you for your contribution and congrats for your first PR to be merged in the project! Let us know if you have any question for the future of the project.

PR #43, opened 2023-01-10
 Decompile func_8011BD48 

sozud:
Not totally sure how to fix the `(*(var_v1 - 5))` part, open to suggestions there.

### DIFF
--- a/src/dra/42398.c
+++ b/src/dra/42398.c
@@ -4806,7 +4806,19 @@ INCLUDE_ASM("asm/dra/nonmatchings/42398", func_8011B5A4);
 
 INCLUDE_ASM("asm/dra/nonmatchings/42398", func_8011BBE0);
 
-INCLUDE_ASM("asm/dra/nonmatchings/42398", func_8011BD48);
+bool func_8011BD48(Entity* entity) {
+    s32 i = 0x10;
+    s16 objId = entity->objectId;
+    s16 subId = entity->subId;
+    Entity* e = &g_EntityArray[i];
+    for (; i < 0x40; i++, e++) {
+        if (objId == e->objectId && subId == e->subId && e != entity) {
+            return 1;
+        }
+    }
+
+    return 0;
+}
 
 INCLUDE_ASM("asm/dra/nonmatchings/42398", func_8011BDA4);
 


Xeeynamo (approved):
You can simplify `(*(var_v1 - 5))` as `var_v1[-5]`. The rest matches and looks good. I am approving but not merging yet so you can refactor that line you mentioned.

BTW we have a couple Discord servers where we keep in sync about the project and giving each other code suggestions. Let me know if you would like to take part of it.

sozud:
Thanks, I made that change. I joined the CV Romhacking discord now, let me know if there's another one I should be on.

Xeeynamo (changes requested):
The function looked very weird, especially that negative array access. It looks very similar to the [negative struct offset loop pattern](https://github.com/pmret/papermario/wiki/GCC-2.8.1-Tips-and-Tricks). What the compiler does under the hood is shifting the alignment of arrays of structs in a loop by the first most accessed field. In this case that is `subId`, located at the offset `0x30`. So in reality `var_v1 = &g_EntityArray[16].subId;` should be `var_v1 = &g_EntityArray[16];`, but then you have to shift back everything else by `0x30` to basically reverse what the compiler optimised. It's the most complicated pattern you will ever find. This is the same identical function after I explored the meaning of it:

```c
bool func_8011BD48(Entity* entity) {
    s32 i = 0x10;
    s16 objId = entity->objectId;
    s16 subId = entity->subId;
    Entity* e = &g_EntityArray[i];
    for (; i < 0x40; i++, e++) {
        if (objId == e->objectId && subId == e->subId && e != entity) {
            return 1;
        }
    }

    return 0;
}
```

I hope it makes a bit sense. You might want to compare the two functions side by side to get an idea.

Also note how I changed the returning type from `s32` to `bool`. We try to always put it when we know the returning values are either `1` or `0`. Finally you do not need to do `struct Entity*`: a `Entity*` will work just fine.

Be free to ping me in CV Romhacking if you have more specific questions. I will be glad to help!

sonicdcer (changes requested):
-

Xeeynamo (approved):

PR #44, opened 2023-01-11
Decompile func_800EDB08

sozud:
Based on examining func_801093C4, which calls this function, I think the argument is an entry of D_80086FEC (POLY_GT4). However this struct layout doesn't seem to fit. I saw that other functions have the same issue so I left it for now.

### DIFF
--- a/src/dra/42398.c
+++ b/src/dra/42398.c
@@ -1380,7 +1380,21 @@ void func_800EDAE4(void) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


Xeeynamo (changes requested):
`Unkstruct_800974AC` is `DR_ENV`

Xeeynamo (approved):

PR #48, opened 2023-01-12
coupe of funcs

sonicdcer:
I decompiled func_800EDB08 without knowing @sozud did it first (forgot to pull), i used my own struct but looking at his version and discussing it with him, we came to the conclusion that both are fake, mine is obviously fake because i'm using an incomplete and invented struct, but in his version, he used POLY_G4 and DR_ENV as types and we don't know if even one of those is actually real, there are several reasons to it, to begin with:

* The casting to ```*(u32*)&poly->r1 = ptr;``` is very weird, the assembler is looking for a s32 but r1 is u_char.
* It doesn't make sense for that pointer to be assigned to it.
* The only hint we have to suspect POLY_G4 is the function FreePolygons, which also looks fake.
* The only hint we have to suspect DR_ENV is it's size (0x40)

Either one of them is fake, or both of them are.

In my version, i didn't even try with those types, as i did my own struct.

So the real question right now is which version should we have in the repo, because his is generating a warning that i don't know how to clear honestly, maybe we should keep the unkstruct until we have more information or until func_801093C4 (which calls this function) is decompiled? Let us know about your thoughts @Xeeynamo .


Xeeynamo:
I'd say there is a strong possibility that the type is really `POLY_GT4` due to the struct size and the fact I recognise `poly->code = n`, which is often used in the rest of the codebase. The `DR_ENV` is a weaker assumption as it's just based on the size and the usage of the parameters, but `D_800974AC` not used anywhere else in the code and I would feel less disruptive to change it down the line. I'd keep `func_800EDB08` as it is.

`func_800FB0FC` looks fine.

Why did you put the parenthesis to `func_8011BD48`? They are unnecessary.

sozud:
To explain my reasoning for the original version better, I selected `POLY_GT4` because of how `func_801093C4` calls `func_800EDB08`. Here is a decompilation excerpt:

```
    var_s0 = (D_800734F8 * 0x34) + &D_80086FEC;
    do {
        var_s0 = *var_s0;
        var_a0 -= 1;
    } while (var_a0 >= 0);
....
    } else if (func_800EDB08(var_s0) != 0) {
        var_s0->unk7 = 7;
        var_s0->unk32 = 8;
        temp_s0_2 = var_s0->unk0;
        func_800EDB08(temp_s0_2);
```

So it seem it loads an offset of `D_80086FEC`, which is currently an array of `POLY_GT4`.

PR #49, opened 2023-01-13
Decompile func_800E912C

sozud:


src/dra/42398.c
```diff
@@ -885,7 +885,17 @@ s32 func_800E908C(void) {
...
+s32 func_800E912C(void) {
+    if (TestEvent(D_80073068) == 1) {
+        return 1;
+    } else if (TestEvent(D_8007306C) == 1) {
+        return 2;
+    } else if (TestEvent(D_80073070) == 1) {
+        return 3;
+    }
+
+    return (TestEvent(D_80073078) == 1) * 4;
```
Xeeynamo, line 897: Did you try if the following works?

```c
if (TestEvent(D_80073078) == 1) {
    return 4;
}
return 0;
```

Xeeynamo (changes requested):
Leaving a suggestion to keep in sync on the logic found in the function `_peek_event`

src/dra/42398.c
sozud: That also matches. I rebased.

Xeeynamo (approved):

PR #53, opened 2023-01-15
Match func_800E96E8 (with @sonicdcer)

sozud:


src/dra/42398.c
```diff
@@ -949,8 +949,33 @@ s32 func_800E9640(s32 arg0, s32 arg1, s32 arg2, s32* readBufferAddress,
...
 
-// https://decomp.me/scratch/6dWJG very close (branching)
-INCLUDE_ASM("asm/dra/nonmatchings/42398", func_800E96E8);
+s32 func_800E96E8(s32 arg0, s32 arg1, s32 arg2, void* arg3, s32 arg4,
+                  s32 arg5) {
+    s8 sp18[32];
+    s32 new_var;
+    s32 device;
+    s32 ret;
+    sprintf(&sp18, &g_strMemcardSavePath, arg0, arg1, arg2);
```
Xeeynamo, line 958: No need of using `&sp18`, just do `sp18`

src/dra/42398.c
```diff
@@ -949,8 +949,33 @@ s32 func_800E9640(s32 arg0, s32 arg1, s32 arg2, s32* readBufferAddress,
...
+    if (arg5 == 1) {
+        device = open(&sp18, (arg4 << 0x10) | 0x200);
```
Xeeynamo, line 960: :no_bell: same here

src/dra/42398.c
```diff
@@ -949,8 +949,33 @@ s32 func_800E9640(s32 arg0, s32 arg1, s32 arg2, s32* readBufferAddress,
...
+        if (device != (-1)) {
+            close(device);
+        } else {
+            return -2;
+        }
+    }
+    new_var = arg4 << 0xD;
+    device = open(&sp18, 0x8002);
```
Xeeynamo, line 968: :no_bell: same here

src/dra/42398.c
```diff
@@ -949,8 +949,33 @@ s32 func_800E9640(s32 arg0, s32 arg1, s32 arg2, s32* readBufferAddress,
...
+    if (device != (-1)) {
```
Xeeynamo, line 969: In order to avoid code nesting, you can flip the condition and return earlier if `device` is `-1` like the following:

```c
if (device == (-1))
    return -1;

D_80137474 = device;
func_800E91B0();
write(device, arg3, new_var);
return 0;
```

I am suggesting this because I observed it is common practice to return the earliest from a function whenever there is an error. In that case, `device = open` returns the error `-1`, so we want to exit the earliest. Otherwise we continue with the happy path.

src/dra/42398.c
```diff
@@ -949,8 +949,33 @@ s32 func_800E9640(s32 arg0, s32 arg1, s32 arg2, s32* readBufferAddress,
     return ret;
 }
 
-// https://decomp.me/scratch/6dWJG very close (branching)
-INCLUDE_ASM("asm/dra/nonmatchings/42398", func_800E96E8);
+s32 func_800E96E8(s32 arg0, s32 arg1, s32 arg2, void* arg3, s32 arg4,
+                  s32 arg5) {
+    s8 sp18[32];
```
Xeeynamo, line 954: I kind-of recognise this function. It looks like it's creating a string with the path of where the save file is actually found. You can rename `sp18` as `buffer` or `filePath` or `savePath` or whatever name you find more suitable

Xeeynamo (changes requested):
Left minor :nail_care:

sozud:
I addressed the comments and rebased.

Xeeynamo (approved):
Thank you for the precious work!

PR #55, opened 2023-01-16
func_8011B334 & func_80115C50 + partial struct + offset comments

sonicdcer:


Xeeynamo (approved):
`func_801A8620` it's like dreams slowly coming true

PR #61, opened 2023-01-18
Decompile func_800FD7C0

sozud:
I think this is matching except for an issue where the jump table is not being linked correctly.


Xeeynamo (approved):
This is all very good, thank you for your contribution! We have to resolve the conflict before merging it though.

sozud:
The conflicts should be resolved now.

PR #63, opened 2023-01-19
Split ST/LIB.BIN

sozud:
I'm not sure what `config/symbols.stlib.txt` does so it's empty at the moment. I'm not able to run the duplicates script locally without edits. `root_directory` seems like it ought to be `root_dir`. `get_symbol_length` also seems to need:

```
if not sym_name in map_offsets:
  return 0
```
otherwise it crashes for me.

I filled out one duplicate (func_801D1904) to make sure that adding code works.


Xeeynamo:
Unfortunately I cannot merge this PR. A while ago we decided to not allow, for now, any new overlays to be decompiled in the repo. It's a fault on my side as I did not put such information in the README yet.

I will go through the reasons, in no particular order:

- It would increase the scope of the project, which right now is something we do not really want to have.
- We are still on the way to understand the current overlays, which is required to manage new ones.
- It will be much easier to decompile new overlays once we will be able to be close to fully decompile the existing ones.
- I would like to to keep the focus on what we do not have decompiled yet, rather than adding more.

I would say once we understand NO3, NP3 and NZ0, we can really start messing around with all the remaining overlays one by one! Let me know your opinion here or on Discord.

PR #81, opened 2023-01-28
func_80124164 & func_80123F78 decompiled

sonicdcer:


Xeeynamo (changes requested):
I checked the caller `func_801243B0` to see where the parameter of your decompiled `func_80124164` would come from and it seems an argument from `&D_80086FEC[arg0->firstPolygonIndex]`. That implies `arg0` is a `POLY_GT4*`!

src/dra/5D874.c
```diff
@@ -2145,7 +2145,107 @@ INCLUDE_ASM("asm/dra/nonmatchings/5D874", func_80123B40);
 
 INCLUDE_ASM("asm/dra/nonmatchings/5D874", func_80123F78);
 
-INCLUDE_ASM("asm/dra/nonmatchings/5D874", func_80124164);
+void func_80124164(POLY_GT4* poly, s32 arg1, s32 arg2, s32 arg3, s32 arg4) {
+    s16 temp = arg2 - arg3;
```
Xeeynamo, line 2211: `s16 top = arg2 - arg3;`

src/dra/5D874.c
```diff
@@ -2145,7 +2145,107 @@ INCLUDE_ASM("asm/dra/nonmatchings/5D874", func_80123B40);
...
+    s16 temp2 = arg2 + arg3;
```
Xeeynamo, line 2212: `s16 bottom = arg2 + arg3;`

src/dra/5D874.c
```diff
@@ -2145,7 +2145,107 @@ INCLUDE_ASM("asm/dra/nonmatchings/5D874", func_80123B40);
...
+        poly->y1 = 0;
+        poly->y0 = 0;
+    }
+
+    if (poly->y0 > 240) {
+        poly->y3 = 240;
+        poly->y2 = 240;
+    }
+
+    if (arg4 == 0) {
```
Xeeynamo, line 2231: `arg4` might be of `bool` type

src/dra/5D874.c
```diff
@@ -2143,9 +2143,171 @@ void func_80123A60(Entity* entity) {
...
+        temp3 = arg1 * D_800AE270[0];
+        if (temp3 < 0) {
+            temp3 += 255;
+        }
+        poly->r3 = (u32)temp3 >> 8;
```
Xeeynamo, line 2236: You can remove those `(u32)` casts by changing the type of `temp3` into `u32`. Also I think `temp3` can be renamed as `color`, `colorUnit` or `colorChannel`

src/dra/5D874.c
```diff
@@ -2143,9 +2143,171 @@ void func_80123A60(Entity* entity) {
...
+        D_80097F3D = D_80097F4A;
+        D_80097F48 = D_80097F4A;
+        D_80097F44 = D_80097F4A;
+        D_80097F40 = D_80097F4A;
+        D_80097F3C = D_80097F4A;
+        break;
+    }
+}
+
+void func_80124164(POLY_GT4* poly, s32 arg1, s32 arg2, s32 arg3, s32 arg4) {
```
Xeeynamo, line 2210: `arg2` can be renamed as `y`, while `arg3` is a bit more ambiguous... maybe something like `halfHeight` or `radius`? While `arg1` might be something like color intensity

Xeeynamo (changes requested):
I left another review for some minor variable renaming. I find the meaning of this function a bit odd. It looks like some kind of glowing algorithm. It would be curious to know in-game what it does.

Xeeynamo (approved):

PR #99, opened 2023-02-18
more small CEN functions

frog404:


frog404:
did a PR before i pushed on my repo...

PR #123, opened 2023-02-28
 Decompile func_801027C4

sozud:
I commented out the `// asm volatile("move $16, $2");` since they break the permuter even when `#ifdef`-ed out


sozud:
The first return seems to be necessary to match, but I changed the other two returns to breaks. To me it seems better to leave `default` in so that it's explicit, but I can remove it if we don't like it.

PR #132, opened 2023-03-04
Split remaining DRA jump tables

sozud:
This should split out the remaining DRA jump tables. This was done with a script but I looked over most of the files and it seems correct to me. The purpose is making decompiling jump table functions a little quicker. 

I'm thinking about writing a script to split all the C files accordingly, do we agree with that? There's about 95 remaining jump table functions in DRA, so there would be about that many file splits. This would save time when working on jump table functions since all the splitting work would already be done.

### DIFF
--- a/config/splat.us.dra.yaml
+++ b/config/splat.us.dra.yaml
@@ -37,42 +37,140 @@ segments:
       - [0x8000, data]
       - [0x8900, data]
       - [0xCEB0, data]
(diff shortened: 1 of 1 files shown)


sozud:
Ok, then I think I may make a script to automate the C file split for a particular function. Unless you see an issue with the splits here this PR can be merged.

PR #134, opened 2023-03-05
Add script to split jump tables in splat yaml

sozud:
I cleaned up the script used in the previous PR. The script reads in the current yaml, finds all the jump tables, then merges that into the original yaml. The `subsegments:` part of the yaml will be printed to the console, which can then be pasted back into the original yaml. Usage looks like this:

```
# split DRA (uses jpt_ prefix)
# python3 tools/split_jpt_yaml.py config/splat.us.dra.yaml asm/us/dra/data/ jpt_

# split NO3 (uses jtbl_ prefix)
# python3 tools/split_jpt_yaml.py config/splat.us.stno3.yaml asm/us/st/no3/data/ jtbl_
```


sozud:
Ok, I moved the unit tests to `test.py` and moved the script to its own folder.

sozud:
I added a basic explanation of the process to add jump table functions, I don't think it's enough to get a beginner 100% through the process but it's a start.

PR #147, opened 2023-03-13
Match func 8018ef78

SestrenExsis:
Match the following functions, which are all duplicates of previously decompiled functions:
- RWRP - func_8018EF78
- ST0  - func_801B61DC
- ST0  - func_801B6314


sonicdcer:
The whole file is different, this is probably a problem caused by the windows filesystem, you should use WSL remote connection in vscode to avoid this problem:
image

SestrenExsis:
Thanks for the tip, @sonicdcer, I'm now using WSL in vscode. However, I'm unclear what to do about this issue now. When looking at 27d64.c in vscode, it appears all the lines end in `\n`, which I thought was correct for Linux.

PR #154, opened 2023-03-18
Split BIN/WEAPON0.BIN

Xeeynamo:
Splits `BIN/WEAPON0.BIN` in its own individual C files and assets. There are 59 overlays 59 PNGs extracted as part of it. The `.data`, `.text` and `.sbss` section are correctly split. I did not yet split the `.rodata` but I assume it will be so small we can do it on-the-go.

This is fully integrated in the build process, it gives an :ok: once built back and it is fully integrated in the CI. We did not yet know how weapons work in-game and we already have a few instances in the code where we call `D_8017A000();` and similar. I think it is time to start documenting them.

Thanks to @bismurphy for addressing `analyze_calls.py` for this specific PR.

### DIFF
--- a/tools/analyze_calls.py
+++ b/tools/analyze_calls.py
@@ -86,6 +86,8 @@ def handle_jal_call(full_file, call_index, known_func_list):
             return "PlaySfx"
         if "0xB8($a2)" in callreg_setline or "-0x20($s0)" in callreg_setline:
             return "UnknownEntityFunction"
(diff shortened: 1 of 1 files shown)


sttng:
Would it make sense to check the weapon files of the PSP/hdbin folder w0_...bin, which looks like there is one file per weapon  ?

sozud (approved):
Great work, looks good to me once analyze_calls is updated.

PR #161, opened 2023-03-25
Improve jump table function documentation

sozud:


sozud:
I moved the info to the wiki and left a link in the README.

PR #164, opened 2023-03-25
Add function finder script

sozud:
The idea behind this script is making it easier to keep track of what is remaining to work on in the project. Remaining functions are sorted by the length and number of branches, and whether they have jump tables or not. This gives a sense of the difficulty of the function and should make it easier for new contributors to find a function to work on. Decomp.me WIP links are also included. The script looks for comments with the following format and includes them in the output:
```
DECOMP_ME_WIP func_800F9E18 https://decomp.me/scratch/VmuNt
```

The output of the script looks like this: https://gist.github.com/sozud/da6744b3fe58309e2d989ca4198e9515

It might be nice to integrate this with the CI as well. A further improvement would be listing whether the function is a duplicate or not.


sozud:
Thanks, I went with the comment approach because my understanding is that we'd have to use an actual C parser and preprocessor otherwise. The script could be modified to output CSV to import into a spreadsheet. I figure that having a single source of truth on the state of the codebase is the easiest rather than trying to keep the source and spreadsheet in sync? I'm not opposed to a spreadsheet though if it has advantages I'm not thinking of.

sozud:
Comments should be resolved now.

PR #166, opened 2023-03-26
Extract memory card icons

Xeeynamo:
Extract the memory card icons out of DRA.BIN and SEL.BIN.

image

I also took the opportunity to extract `g_MemcardPortMask` into its own source file `save_mgr_pre.c`. Unfortunately this is required to keep the original data order. This suggests that most likely the icon palette and bitmap were baked into the original `save_mgr.c` as byte arrays. I decided to take a different approach and extracted them as PNGs for better moddability.

I had to spin-off Splat into a new fork due to some breaking changes on 0.18.0.


tools/splat_ext/grey4.py
```diff
@@ -0,0 +1,48 @@
+import n64img.image
```
sozud, line 1: Am I missing something or is grey4 unused?

sozud (approved):

tools/png2s.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/python3
```
sozud, line 1: What's the tradeoff of png2s vs a splat extension?

PR #195, opened 2023-04-17
Entity refactor proposal 3

sonicdcer:


sonicdcer:
> I can see three major disadvantages:
> 
> * `self->ext.equipItemDrop.timer` names will become long, like the proposal [Entity refactor proposal 2 #188](https://github.com/Xeeynamo/sotn-decomp/pull/188)
> * Duplicate `Entity` structure, like in the proposal [Entity refactor proposal 1 #187](https://github.com/Xeeynamo/sotn-decomp/pull/187)
> * The biggest concern I have: each entity struct will bloat `objects.h` and even couple different overlays between each other

The Entity struct duplicate will disappear once we replace it in all Entities, like i said before. Also i think objects.h need to be overlay specific because they don't share the ids after 15, and the prototypes don't even need to be there.

PR #196, opened 2023-04-18
Match func_801B12E8

SestrenExsis:
This pull request matches NZ0 `func_801B12E8`. This particular function is responsible for handling one of the backdrops of the Alchemy Laboratory that scrolls with the camera. It appears as a repeating purplish-red brick pattern behind the foreground.

It is comprised of 15 textured primitives laid out in a 5-tall by 3-wide grid.

- NZ0  - `func_801B12E8` - Renamed to `EntityPurpleBrickScrollingBackground`

### DIFF
--- a/src/st/nz0/30958.c
+++ b/src/st/nz0/30958.c
@@ -97,7 +97,7 @@ void func_801B12E8(Entity* self) {
             return;
         }
         prim = &g_PrimBuf[firstPrimIndex];
-        self->firstPolygonIndex = (s32) firstPrimIndex;
+        self->firstPolygonIndex = (s32)firstPrimIndex;
         *(s32*)&self->unk7C = prim;
         self->flags |= 0x800000;
         if (prim != NULL) {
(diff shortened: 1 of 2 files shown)


src/st/nz0/30958.c
```diff
@@ -78,8 +78,73 @@ void EntityRedEyeBust(Entity* self) {
...
+                prim->u1 = 0xFF;
+                prim->v1 = 0x80;
+                prim->v0 = 0x80;
+                prim->v3 = 0xBF;
+                prim->v2 = 0xBF;
+                prim->priority = 0x20;
+                prim->blendMode = 2;
+                prim = prim->next;
+            } while (prim != NULL);
+        }
```
sonicdcer, line 121:
```suggestion
while (prim != NULL) {
                prim->tpage = 0xF;
                prim->clut = 4;
                prim->u2 = 0x80;
                prim->u0 = 0x80;
                prim->u3 = 0xFF;
                prim->u1 = 0xFF;
                prim->v1 = 0x80;
                prim->v0 = 0x80;
                prim->v3 = 0xBF;
                prim->v2 = 0xBF;
                prim->priority = 0x20;
                prim->blendMode = 2;
                prim = prim->next;
            } 
the if statement is not necessary, this is a common pattern for Entities like this, you will find it very often :)

sonicdcer (changes requested):
Amazing! I still have the scratch for this function opened in my browser, since i tried it for several days now. I'm glad you found the solution!

src/st/nz0/30958.c
sonicdcer: the if statement is not necessary, this is a common pattern for Entities like this, you will find it very often :)

PR #197, opened 2023-04-18
Debug module

sonicdcer:
Bring back some things


sonicdcer:
> I do not agree on the change in `entity_spawn.c` as we already have that information on the top right:
> 
> image
> 
> image

Except that it doesn't work, it only works sometimes when you're in bat form, but not always. And even if it did worked, it is not clear to read with that sort of translucency effect, and the title should be where everything else is.
image

sonicdcer:
Nearly unreadable
image

sonicdcer:
We don't need to duplicate anything, i'd say we drop the fancy translucent sign and came back to just adding one word ST/DRA) to the printf, but if you insist on keeping the fancy translucent sign, i suggest you merge this PR until you can make it work, because as of now, it doesn't work properly and people just started using this debug module.

sonicdcer:
It's not that i'm against it, it's that it doesn't work, it doesn't show the text 99.9% of the time.

PR #232, opened 2023-05-22
WIP saturn cue/bin parsing

sozud:
The Saturn cue/bin I have looks like this:

```
FILE "Akumajou Dracula X - Gekka no Yasoukyoku (J)(Saturn).BIN" BINARY
  TRACK 01 MODE1/2352
    INDEX 01 00:00:00
  TRACK 02 AUDIO
    PREGAP 00:02:00
    INDEX 01 45:30:63
```

I attempted to add  MODE1/2352 support. It's able to extract some of the files, which seem to be correct, but some are missing. I'm not sure why that's happening. Maybe the Saturn disk uses some filesystem features the PSX one doesn't? 

I had to change the track adding code to add more than one track per file, otherwise track 1 gets replaced with the audio track.


sozud:
It's rejecting sectors that don't meet the `offset+bufSafe > secSize` requirement here. Maybe the missing files are there. Could you explain what the fix to this code would be?

```
		// horrible hack as it can read unnecessary bytes, but it works
		if offset+bufSafe > secSize {
			sec, err := readSector(file.reader, location(chloc), file.mode, false)
			if err != nil {
				return nil, err
			}
			data = []byte(sec)
			offset = 0
			chloc++
		} else {
			fmt.Print("rejected sector\n")
		}
```

PR #233, opened 2023-05-24
Integrate splitting for saturn

sozud:
This PR does some initial integration work for splitting the Saturn version. It adds T_BAT.PRG (Bat overlay) and GAME.PRG (similar to DRA.BIN). These are extracted to asm/saturn/* and src/saturn/*. Right now there's no attempt to compile. I haven't been able to figure out the issues with sotn-disk so this requires pre-extracted files in disks/saturn. Something like this is able to do it:

```
bchunk saturn.bin saturn.cue saturn.iso
7z x saturn.iso 
```

The Saturn version has the code split up differently so I think putting it in a separate folder to begin with makes sense. For example the Random function is in GAME.PRG rather than in the stage overlays. There's also an Alucard overlay rather than it being included in GAME.PRG. See here for some comparisons: https://github.com/Xeeynamo/sotn-decomp/wiki/Saturn-Matching

I didn't attempt to install Rust as part of the project setup like with Go.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "tools/m2c"]
 	path = tools/m2c
 	url = https://github.com/matt-kempster/m2c.git
+[submodule "tools/saturn-splitter"]
+	path = tools/saturn-splitter
+	url = https://github.com/sozud/saturn-splitter.git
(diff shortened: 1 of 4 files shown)


sonicdcer (approved):

DerpPrincess (approved):

sozud:
Ok, bchunk is in the requirements and we extract with 7z. 7z gives an error but the files are OK as far as I know. I added a few fixes for game.prg and updated the splitter version.

PR #234, opened 2023-05-25
Add Saturn toolchain

sozud:
This builds on the previous PR but I've separated it since I'm expecting a bunch of comments. The new changes are in https://github.com/Xeeynamo/sotn-decomp/commit/b9aebb1d5fc430bc8739f43802a8f070a828bb55

This adds some new Makefile commands:

build_saturn_toolchain: Downloads GCCSH, and builds two docker containers. One container is binutils for SuperH, and the other has Dosemu. 

build_saturn: Copies GCCSH, the source code, and asm into the build directory and runs the GCCSH via Dosemu in the docker container. Everything is kept in the same directory since it's hard to use Dosemu otherwise.

check_saturn: Dumps the object files to binary using the binutils Docker container, then checks the hashes.

diff_saturn: Compares the binaries and outputs a diff to build/saturn/$FILENAME-diff.txt. Used like FILENAME=GAME.PRG make diff_saturn.

I've added the source for game.c and t_bat.c. I've replaced one of the asm functions in t_bat.c to show that it's possible to match simple functions. I'm not able to generate linker scripts yet so functions that rely on symbols from other overlays generally can't be matched yet.

This workflow is not all that clean so I'm open to ideas on improving it. I tried using various versions of https://github.com/decompals/old-gcc to try to get a native compiler but couldn't get a matching build. I'm not 100% sure that this version of GCCSH is correct but initial results seem promising.

As of right now, it's not possible to run this in CI since the dosemu container changes the USER and Github won't let you write to disk if you do that. https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#user


sozud:
I rebased to fix conflicts.

sozud:
I don't understand how you're supposed to compile old gcc git revisions since there's no configure script and no configure.in or whatever to generate one. I wrote a script to try and figure out what revisions of gcc have a configure script. https://gist.github.com/sozud/a1953c6d72f89a0f18a36707da6d8d41 

The results are here: https://gist.githubusercontent.com/sozud/b3e381874a9283fa24c22fa2da80288e/raw/726f9773c9e26c4c95ac19f0c6bc5122324b44ff/gistfile1.txt

The oldest revision with a configure script is 861bb6c1b0958236ad93717f98d347aa6152bd09 (1997-08-11 09:56:58 -0600) 

I tried using configure scripts from released versions on a revision around 1996-09-04 but wasn't able to figure out how to get it to compile. Also based on the changelog it's possible this compiler is branched from a particular git revision but has an assortment of patches on it that doesn't make it equivalent to any exact git revision in the gcc repo. https://github.com/sozud/saturn-compilers/blob/af27f4aa6566c1b0fa5332b488e63d6ffd28bd48/cygnus-2.7-96Q3/GCCSH/README/GCC.LOG#L51

I tried to see if there's old gcc mailing lists to figure out what these patches are but the archive only goes back to May 1998 https://gcc.gnu.org/pipermail/gcc-patches/1998-May/thread.html

Is a native compiler required?

sozud:
It seems like release tarballs have configure scripts but the git commits don't as far as I can tell. Here's the commit for 1996-09-04. https://github.com/gcc-mirror/gcc/tree/98d283063056ace6070cddf0c2e23f4ca95e2f60/gcc

I see config.guess, config.sub, and configure.bat. I can't seem to find any build or install instructions. This msdos script seems to resemble a configure script I guess? https://github.com/gcc-mirror/gcc/blob/98d283063056ace6070cddf0c2e23f4ca95e2f60/gcc/config/msdos/configur.bat

sozud:
I updated the PR to enable linking between different overlays. The splitter now generates linker scripts and the objects are linked with modern ld. I added a new function to t_bat.c, func_80173F30, which demonstrates using a function from 0.BIN.

PR #247, opened 2023-06-10
Update splat

mkst:
Thank you in advance for contributing to sotn-decomp.
Before submitting this PR, please make sure:

- [x] The PR is small and focuses to address a single task
- [x] `make all` reports all `OK`
- [x] `make format` reports no files changed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ define link
 	$(LD) -o $(2) \
 		-Map $(BUILD_DIR)/$(1).map \
 		-T $(1).ld \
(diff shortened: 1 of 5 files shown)


mkst:
Splat creates an undefined_syms_auto.txt (and undefined_funcs_auto.txt) and puts any symbols it doesnt know about in them. For a bunch of different reasons, it doesn't always have the entire list (I guess it thinks symbols *have* been defined, but for some reason they dont end up being included in the resulting objects that `ld` is linking) - so these missing symbols are added to a manual undefined_syms.txt list so `ld` knows their address.

PR #248, opened 2023-06-11
Set up Saturn CI

sozud:
This was working on my fork but may require a little tweaking for this repo since I can't test the dependency fetching. The CI build uses a regular install of dosemu2 since you can't run it in a container on GitHub actions. I rearranged the Makefile a bit to parallelize some actions and reduce duplication for the two workflows.


sozud:
Conflict fixed and also fixed a naming issue where two build steps had the same name. Tested on my fork and seems OK.

PR #252, opened 2023-06-16
Extract Saturn adpcm music

sozud:
This adds a new tool that extracts the Saturn ADPCM music files to WAV. @sonicdcer pointed me to an old winamp plugin that had support for this format, so I reworked the relevant part into a standalone tool. This doesn't try to re-encode, since I would want to examine the original decompression code to make sure that it's 1:1 with this implementation first. Not all the files seem to be in this format, so I explicitly list the ones that are in the Makefile. I think the other PCM files are the voice tracks, which would make sense to use a voice-oriented codec.

### DIFF
--- a/.github/workflows/saturn-build.yml
+++ b/.github/workflows/saturn-build.yml
@@ -57,6 +57,8 @@ jobs:
         run: make -j saturn_ci_tasks
       - name: Extract asm
         run: make extract_saturn
+      - name: Extract pcm
+        run: make extract_saturn_pcm
       - name: Build saturn binaries
         run: make build_saturn_ci
       - name: Check saturn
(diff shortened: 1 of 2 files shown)


sonicdcer:
And to think i found that source buried very deep in the way back machine, i'm very happy that you decided to do something like this with it, this is amazing man.

sttng:
Just out of curiosity: ffmpeg does not support extraction of Saturn ADPCM ?

sozud:
Which codec? The Sega ADX one seems to be something different. https://github.com/FFmpeg/FFmpeg/blob/d26ea03f60ac78476766bfcc115c0ec050c68198/libavcodec/adxdec.c#L258

sttng:
According to vgmstream:

```
vgmstream-cli-m SD0A.PCM

metadata for SD0A.PCM
sample rate: 44100 Hz
channels: 2
loop start: 995328 samples (0:22.570 seconds)
loop end: 4677632 samples (1:46.069 seconds)
stream total samples: 4677632 (1:46.069 seconds)
encoding: Intel DVI 4-bit IMA ADPCM (mono/interleave)
layout: interleave
interleave: 0x4 bytes
metadata from: Konami KCEN DVI. header
bitrate: 352 kbps
play duration: 8800936 samples (3:19.568 seconds)
```

sozud:
Interesting, thanks for finding that!

PR #257, opened 2023-06-17
Split Sega GFS, SPR and BUP libraries

sozud:
I wrote a tool that objdumps the Sega libraries and matches the bytes to find them within 0.BIN. This PR splits the sections for the Sega GFS, SPR and BUP libraries and labels the functions that were found. GFS and SPR were distributed with source so the parts where it's not matching may be edited or a different version. They apparently shipped the debug version of GFS.

### DIFF
--- a/config/saturn/zero.bin.yaml
+++ b/config/saturn/zero.bin.yaml
@@ -16,7 +16,7 @@ segments:
         type: data
         file: zero
         end: 0xae07
-
+  
       # ratan2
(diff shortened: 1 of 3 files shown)


sonicdcer:
This is huge!

sozud:
The autoformatter broke this since my regex to find INCLUDE_ASM can't tolerate newlines, working on a fix.

PR #258, opened 2023-06-17
Fix #257

sozud:
The splitter regex to find INCLUDE_ASM macros expects it all to be on one line, so after clang-format was applied, the asm got put in the wrong directories. This should fix the previous. Sorry about the issues.


sozud:
I'm not sure why this is still failing, it seems fine locally.

PR #266, opened 2023-06-19
Dockerless Saturn build

sozud:
@sonicdcer was having trouble with Docker so here is a Dockerless build. I've changed the CI to use this since it's faster.

### DIFF
--- a/.github/workflows/saturn-build.yml
+++ b/.github/workflows/saturn-build.yml
@@ -35,9 +35,9 @@ jobs:
       - name: Clone dependencies
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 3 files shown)


sozud:
It looks like it ran master again. Should we not use `github.head_ref` in the script?

"The head_ref  or source branch of the pull request in a workflow run. This property  is only available when the event that triggers a workflow run is either pull_request or pull_request_target."

sozud:
Yeah I think what you are saying is correct. My understanding is that the github actions yaml in the previous run was from master, but rest is the contents of this PR, which is why it failed.

PR #272, opened 2023-06-20
Import GameApi from game.h to symbols.us.txt

bismurphy:
Currently, there are many functions in GameApi that do not exist in symbols.us.txt. This rectifies this by taking everything from GameApi and copying it into symbols.us.txt. This was done with the Python script that is added in the tools directory here. Note that this tool is run manually and its output is copied into the symbols; it does not automatically write to the file and will not automatically update it if game.h has more symbols change names.

The benefit of this is improved assembly. For example, line 411 of asm/us/st/nz0/nonmatchings/47CF0/EntityMagicallySealedDoor.s now shows up as `lw         $v0, %lo(g_api_PlaySfx)($v0)` where previously it was `lw         $v0, %lo(D_8003C7DC)($v0)`. The updated name now means that the assembly will have useful names in all the places that game.h's GameApi struct has useful names.

Thank you in advance for contributing to sotn-decomp.
Before submitting this PR, please make sure:

- [x] The PR is small and focuses to address a single task
- [x] `make all` reports all `OK`
- [x] `make format` reports no files changed
- [x] Have Actions enabled in your fork to run the auto-formatter


bismurphy:
Yes, I had basically assumed the tool would not be particularly useful. But I assume the gameapi struct will continue to change, and it is not an obvious conclusion that the two files will remain synchronized, so I figured having it available would be more useful than not. It can also be adapted for other similar extraction symbols from header files, because I suspect this is not the only place where symbols are not maintaining pace with headers.

PR #276, opened 2023-06-22
Decompile+rename func_800F4D38; extract Accessory fields

bismurphy:
This started with simply decompiling func_800F4D38. But then I found it referencing an unk08 within Accessory, which I now know to be attBonus.

I ended up going through accessory.json (which only just recently started existing, with one of my other pull requests) and classified all the equippables that change the player's stats. This includes, for example, the Gauntlet which gives ATT+5, or Lapis Lazuli which gives LCK +20.

Therefore this PR does not really address only a single task, since it now extracts all the accessory fields. However, because they were not extracted previously, there is no existing code which references these fields of Accessory, and therefore there should be no issues with this.

Note that the decompilation references Equipment.damageScale; my other current pending PR renames that field to itemCategory, and therefore these two PRs will have a small conflict. Hopefully this is not too much of an issue.

Finally, func_800F4D38 has been renamed as CalcAttack. This is because it calculates how much Attack power the player has, taking into account the equipped weapon, any stat-boosting armor, etc. All references to this function have been changed to match.

- [ ] The PR is small and focuses to address a single task
- [ ] `make all` reports all `OK`
- [ ] `make format` reports no files changed


src/dra/5298C.c
```diff
@@ -47,14 +47,69 @@ bool IsAlucart(void) {
...
+        totalAttack += strengthStat / 2;
+    }
+
+    totalAttack += equipmentAttackBonus;
+
+    if (equipId == 0x7D) { // Badelaire sword
+        totalAttack += g_Status.timerHours;
+    }
+    if (equipId == 0x8D) { // Muramasa sword
+        totalAttack += SquareRoot0(g_Status.D_80097C40);
```
sozud, line 86: D_80097C40 -> numMuramasaBloodHeals or similar. I checked by changing 97c00 in main ram to 0x8d, re-equipping in the menus to load the sword, then killing an enemy to heal. D_80097C40 goes up by one each time.

bismurphy: Agree, though I'd like to focus on getting this PR in first, then allow a check in names-discussion to make sure others agree. I'm also looking into anywhere else in the code that references this address, would be nice to have all functions that use it dealt with before names start changing. For now I'd like to leave it as the D_ value.

PR #281, opened 2023-06-24
Decompile func_801989AC and duplicates

bismurphy:
This function appears to deal with a bunch of collision logic. I won't pretend to understand it, but maybe someone after me can identify its purpose and make any naming tweaks.

This function appears in 9 different overlays; code is identical in all of them.

The function calls MoveEntity, which is also heavily duplicated. However, for some reason MoveEntity within rwrp was not decompiled - I've taken the liberty of copying the decompilation of all the other MoveEntity instances into rwrp, and adding it to config/symbols.us.strwrp.txt, and changing the other calls to use the new name MoveEntity. Therefore, this PR does not strictly only address a single task, but given how common MoveEntity is in the other overlays, updating rwrp to match the others seems like a relatively innocent change.

- [ ] The PR is small and focuses to address a single task
- [ ] `make all` reports all `OK`
- [ ] `make format` reports no files changed

### DIFF
--- a/config/symbols.us.strwrp.txt
+++ b/config/symbols.us.strwrp.txt
@@ -4,6 +4,7 @@ Random = 0x8018A168;
 Update = 0x8018A198;
 CreateEntityWhenInVerticalRange = 0x8018BE1C;
 CreateEntityWhenInHorizontalRange = 0x8018BF34;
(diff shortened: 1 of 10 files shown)


sonicdcer (approved):
Excellent work! Thank you for this!

PR #284, opened 2023-06-26
Rename hitboxOffX

sozud:
Having issues getting the repo to build properly locally for some reason, checking this on CI.


sozud:
I think I will just keep the existing PRs and fix the conflicts. This one should be good to go, if the others get conflicted after merging I will fix them. Next time I have to do something like this I will see if there are adjacent lines and batch those renames together.

PR #286, opened 2023-06-27
Rename rotAngle

sozud:


sozud:
I rebased this, should be OK now.

PR #291, opened 2023-06-28
Fix saturn CI

sozud:
This PR installs a version of dosemu2 from about a month ago from debs from https://github.com/sozud/dosemu-deb
The Dockerfile shows how do do it if we need to make new builds in the future. I tested this on my branch and it should be OK.


sozud:
The build failed because it uses the yaml from master rather than the PR.

PR #293, opened 2023-06-29
Enable some functions fixed by maspx

sozud:


mkst:
If you `cd tools/maspsx && git pull origin master` you will put the latest commit to maspsx which fixes the `sltu` issue here. 

maspx commit for reference: https://github.com/mkst/maspsx/commit/4c06a8b9149b55944c33cf049c61e4988d5b81b4.

I've made an assumption that this expansion only happens for negative numbers. I did a clean build and everything is OK with the change.

sozud:
I pulled, seems to be fixed now. Thanks for looking into this.

PR #301, opened 2023-07-01
Add data_tracer.py

bismurphy:
This is the script demonstrated in Discord, with both graphical and command line interfaces.
I won't say it's particularly good Python, but it gets the job done for now :)

This traverses all functions in /asm (after doing `make force_extract`), and logs all functions called by every function. This takes several seconds, so the result of this traversing is saved to a text file called sotn_calltree.txt and used for future runs. I've added this to the gitignore. As mentioned in the comments at the top of the Python file, this file can be deleted if the user would like to re-generate the call tree (mainly if symbols have been renamed, which would change the generated assembly files).

This is very much meant to be a first draft; it may have bugs and it definitely has ugly, low-quality code (I had to do a whole bunch of special cases for assembly that calls functions indirectly, for example), but at least for now this could be a useful starting point. Hopefully myself or others will make improvements into the future!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ disks/
 
 .vscode/settings.json
 tools/go
+tools/sotn_calltree.txt
 tools/saturn_toolchain/GCCSH
 config/saturn/game_syms.txt
 config/saturn/stage_02_syms.txt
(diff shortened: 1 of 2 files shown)


bismurphy:
Yes, I think this could reasonably replace analyze_calls. I didn't want to replace your work but if you think this is better then sure.

If removing `analyze_calls`, I might suggest renaming _this_ script to be `analyze_calls`, since I do think that's a better name. Thoughts on that?

PR #308, opened 2023-07-02
Add Saturn symbols for Entity, Game and Warp

Xeeynamo:
I discovered a handful amount of symbols by comparing the WARP overlay between the PSX and the Saturn version.

I updated some game engine functions such as `MoveEntity` (which I refactored), `AllocPrimitive` and more.

I also updated the `Entity` struct in `sattypes.h`. You will see that `unk1E` comes before `unk1A` and `unk1C`. That is not a typo. I imported the same header in Ghidra and all the three struct fields are used correctly. I believe the Saturn developers just moved those fields around to keep everything aligned. Note that `primIndex` on Saturn is `s16` and not `s32` like the PSX version.

Last, but not least, I was able to identify all the functions from the WARP program. I discovered `EntityWarpRoom` is broke down into three different functions, which makes sense considering how large the function is on PSX. Maybe that is an hint that on PSX they were also three different functions but two of them got inlined.

That said, the code between PSX and Saturn seems to be vastly identical. I will proceed at decompiling more functions in a separate PR.

### DIFF
--- a/src/saturn/game.c
+++ b/src/saturn/game.c
@@ -630,8 +630,7 @@ s32 Random(void) {
 }
 
 void CreateEntityFromCurrentEntity(u16 objectId, Entity* entity);
-INCLUDE_ASM("asm/saturn/game/f_nonmat", f607B318,
-            CreateEntityFromCurrentEntity);
+INCLUDE_ASM("asm/saturn/game/f_nonmat", f607B318, func_0607B318);
 INCLUDE_ASM("asm/saturn/game/f_nonmat", f607B374, func_0607B374);
 INCLUDE_ASM("asm/saturn/game/f_nonmat", f607B3D0, func_0607B3D0);
 INCLUDE_ASM("asm/saturn/game/f_nonmat", f607B448, func_0607B448);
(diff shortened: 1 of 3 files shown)


config/saturn/warp_user_syms.txt
```diff
@@ -0,0 +1,4 @@
+EntityWarpRoom = 0x060DC040;
```
sozud, line 1: Just FYI, the linker already knows where to put these functions since they're directly compiled into the file. When I fix the splitter to use user defined symbols these would be useful though. 

We have to add user symbols for stuff that's referenced by functions but is extern. So for example if I had a function like:
```
extern bool g_bool;
void example()
{
  g_bool = 1;
}
```

`g_bool` would need a user symbol added.

If these were unknown to the linker, they wouldn't link because they need an underscore in front. So we would eventually want `_EntityWarpRoom` etc.

src/saturn/game.c
```diff
@@ -521,9 +521,9 @@ INCLUDE_ASM("asm/saturn/game/f_nonmat", f6079A2C, func_06079A2C);
 INCLUDE_ASM("asm/saturn/game/f_nonmat", f6079AF0, func_06079AF0);
 INCLUDE_ASM("asm/saturn/game/f_nonmat", f6079B74, func_06079B74);
 
-void func_06079B9C(s32* param_1) {
-    param_1[1] += param_1[3];
-    param_1[2] += param_1[4];
+void MoveEntity(Entity* entity) {
+    LOW(entity->posX) += entity->accelerationX;
```
sozud, line 525: Just FYI I'm planning on doing the `f32` refactor eventually but this is fine for now.

sozud (approved):

src/saturn/game.c
```diff
@@ -629,7 +629,9 @@ s32 Random(void) {
     return g_randomNext >> 0x18;
 }
 
-INCLUDE_ASM("asm/saturn/game/f_nonmat", f607B318, func_0607B318);
+void CreateEntityFromCurrentEntity(u16 objectId, Entity* entity);
+INCLUDE_ASM("asm/saturn/game/f_nonmat", f607B318,
```
sozud, line 633: Oh, actually it seems like there's some build breakage due to renaming these. I'd add a comment for now, my tool doesn't do as much work as splat does.

PR #315, opened 2023-07-04
Decompile SetStageOverlayAddress

sozud:
This match seems odd since I would have expected the addresses to be literals. With @sonicdcer 

I'll rebase when the other PR is merged.

https://decomp.me/scratch/KYz2v
https://decomp.me/scratch/MyAL


sozud:
This is rebased now.

PR #320, opened 2023-07-06
Dockerfile to check setup instructions

sozud:
I'm not sure this actually needs to be in the repo or not but I made this so we can easily check the PSX setup instructions from a clean slate. This does a fully in-container build using the host filesystem as little as possible for that reason. setup.sh should be pretty much exactly what a user would need to do from a fresh system. The Dockerfile assumes that sotn.us.cue is in the "sotncue" folder. I'm thinking maybe not adding something like this to the CI since it would introduce another thing that can go wrong there. I figure if someone says the instructions are wrong we run this to double check.


sozud:
Fine

PR #330, opened 2023-07-07
DRA DebugCaptureScreen, DebugCaptureVideo

Xeeynamo:
Matches the two orphan functions `DebugCaptureScreen` and `DebugCaptureVideo`. I also fully imported the `.rodata` into `42398.c` and I am proud to say the file is now fully decompiled! I also renamed two other functions that are used by the two I decompiled. They are now all fully documented. Note the unused `u8 bmp[0x100];` in `DebugCaptureVideo`, probably a copy&pasta left-over. The function does not match without it.

I am waiting for https://github.com/mkst/maspsx/pull/6 and https://github.com/mkst/maspsx/pull/7 to be merged as currently maspsx breaks the two functions.


src/dra/42398.c
```diff
@@ -47,16 +56,193 @@ void func_800E2398(const char* str) {
...
+        }
+    }
+    if (close(fid) < 0) {
+        FntPrint(D_800DB3E4);
+        return;
+    }
+    DebugInputWait(fileName);
+}
+
+void DebugCaptureVideo(void) {
```
sonicdcer, line 167: I think this should be DebugCaptureDemo, a video is something different. I'm assuming this was used to capture the strategy demos for defeating the bosses? That's what japan devs usually call Demos.

sonicdcer (changes requested):

sonicdcer (approved):

mkst:
I have merged the 2 PRs into `maspsx` - thank you very much :)

PR #333, opened 2023-07-08
Decompile DRA func_800F4FD0

bismurphy:
This has similar flow to the recently-decompiled CalcAttack, and therefore I think it should be CalcDefense.

The main differences are that they have different arguments and return forms, but their overall flow is the same.

If naming this CalcDefense is reasonable, let me know and I will include that renaming into this PR before you merge it.


src/dra/5298C.c
```diff
@@ -192,7 +192,88 @@ void func_800F4F48(void) {
...
+        g_Status.D_80097C2C |= 0x200;
+    }
+    if (D_80139854 != 0) {
+        g_Status.D_80097C2A |= 0x800;
+    }
+
+    totalDefense += (SquareRoot0(g_Status.statsTotal[STAT_CON]) - 2);
+    // Walk armor
+    if (CheckEquipmentItemCount(0x13, 2) != 0) {
+        totalDefense += (g_roomCount / 60U);
```
sonicdcer, line 263:
```suggestion
totalDefense += g_roomCount / 60;

Comment (sonicdcer) [reply src/dra/5298C.c:263]: you need to change g_roomCount type to be u32 for it to match.

**Review by sonicdcer — CHANGES_REQUESTED** [review]

File: src/dra/5298C.c [hunk 2]
```diff
@@ -192,7 +192,88 @@ void func_800F4F48(void) {
...
+
+    if (*D_80139828 != 0) {
+        totalDefense += 0x14;
+    }
+    if (totalDefense < 0) {
+        totalDefense = 0;
+    }
+    if (totalDefense >= 1000) {
+        totalDefense = 999;
+    }
```
sonicdcer, line 274:
```suggestion
if (totalDefense > 999) {
       totalDefense = 999;
    }

sonicdcer (changes requested):

sonicdcer (approved):

PR #334, opened 2023-07-08
EntityElevator

sonicdcer:
Gotta admit, this one gave me a hard time.
I included struct Primitive* inside the extension of all the entities that use only that member, like this one.


include/entity.h
```diff
@@ -246,6 +246,7 @@ typedef struct {
 } ET_StageTitleCard;
 
 typedef union {
+    /* 0x7C */ struct Primitive* prim;
```
sozud, line 249: I think I want Xeeynamo's input on this change.

src/st/cen/D600.c
```diff
@@ -613,7 +614,164 @@ s16 func_801904B8(Primitive* prim, s16 arg1) {
...
+            self->posY.i.hi = player->posY.i.hi;
+            player->posX.i.hi = self->posX.i.hi;
+            self->animCurFrame = 10;
+            g_Entities[1].ext.stub[0x00] = 1;
+            SetStep(3);
+        }
+        break;
+
+    case 1:
+        if (*(u8*)&self[-1].ext.stub[0x4]) {
```
sozud, line 669: I think we'd want to update this when Entity1B is decompiled but I think it's fine for now

sozud (approved):
This looks good to me but needs Xeeynamo's input in my opinion.

PR #335, opened 2023-07-08
Decompile DRA func_8010EB5C

bismurphy:
Honestly not sure what this one does, but it's decompiled now. Something to do with using subweapons.

Fake variable is frustrating to have there, but necessary to make the match.

Note that this function calls 5 other functions; all 5 of those are already decompiled. At this point it might be worth someone determining what all these functions do and what their effect is together, and renaming them to proper descriptive names.


sonicdcer (changes requested):
Made some changes and cleaned it up a little.
https://decomp.me/scratch/5aY7h

PR #339, opened 2023-07-09
Fix PfnEntityUpdates address across overlays

Xeeynamo:
Thanks to the disassembly of the Saturn version (kudos to @sozud), I realised how we got some fake matches in the PSX all along.

https://github.com/Xeeynamo/sotn-decomp/blob/e567920005b4dd7a23a6c39e4bfbe7951c9e163d/src/saturn/game.c#L716

Following the pattern from the Saturn version I was able to confirm my theory to use `objectId - 1` and fix some oddities I found while I was importing the `.data` section from PSX WRP:

https://github.com/Xeeynamo/sotn-decomp/blob/e567920005b4dd7a23a6c39e4bfbe7951c9e163d/src/st/wrp/6FD0.c#L200


sonicdcer (approved):
This is a great finding, good work! And special thanks to @sozud.

sozud (approved):

PR #340, opened 2023-07-09
Improve some equipment function matches based on saturn port

sozud:


sozud:
I renamed based on the discord discussion.

PR #354, opened 2023-07-13
Update symbols for PLAYER

bismurphy:
Got annoyed with seeing PLAYER_unk2E in my decomps when that's step_s. Symbols should now load in with their fully-updated names. Updated a few other symbols too.


sozud (approved):
This seems like a simple merge

PR #360, opened 2023-07-15
Ghidra project generation WIP

sozud:
I'm putting this up as a draft so those interested can try it out and bugs in the setup can be fixed. This is Ghidra project generation for the Saturn port. This automatically defines functions and data types without having to do it manually. This could be extended for PSX but I'm not sure there's any reason to do so.

image

Usage:
```
cd tools/ghidra
GHIDRA_ROOT=/home/user/ghidra_10.3.2_PUBLIC sh generate.sh
```
A project will be generated in tools/ghidra/out/sotn.gpr. Press Analyze in the GUI to disassemble the functions.

The way this works is:
`tools/ghidra/symbols` uses a C parser to collect function definitions from the C files, and addresses from the map files, and emits `tools/ghidra/out/symbols.json`
`tools/ghidra/create-project.py` runs several other Ghidra scripts:
```
rec2-ghidra-apply-symbols.py  - Use symbols.json and create functions
rec2-ghidra-import-types.py - Import headers for type information
apply-mem.py - Create High work ram region and load overlays
```

The python scripts are based on stuff from the rec2 project. 

Right now this is hardcoded to load 0.BIN, GAME.PRG, T_BAT.PRG and STAGE_02.PRG to their respective locations. It could be extended in the future to load differently. Right now I don't parse data so that doesn't get automatically defined yet.

This doesn't work well in WSL since Ghidra locks a project to the user that created it or something like that, so you can't generate in WSL then open in Windows. Also if you open the project in WSL some of the GUI windows are blank.

I'm probably going to make this a submodule in the final PR.


sozud:
I don't have plans to resurrect this since I don't really think it's needed anymore. Closing

PR #362, opened 2023-07-15
Rename everything from Object to Entity for consistency

Xeeynamo:
Mass refactor to get rid of some inconsistencies. `objectId` becomes `entityId`. `LayoutObject` becomes `LayoytEntity` and so on.


sozud (approved):

sozud:
Looks like this has an issue? Not sure if this came from the merge commit I added. Was trying to avoid the "PR is not rebased" problem.
```
src/st/nz0/3E30C.c: In function `func_801C090C':
src/st/nz0/3E30C.c:574: structure has no member named `objectId'
```

PR #367, opened 2023-07-17
DRA HandlePlay matching

Xeeynamo:
Very happy of this one. I tried decompiling it two months ago without success but I decided to give it another try less than 24 hours ago. It is the function called on every frame while the normal gameplay is running. This also manages how to transition from a stage to another, which files to load, set specific random seeds when playing a Tactics recording.

Note that I refactored `func_800EAF28` as it is used by `HandlePlay`. I also moved `GameSteps` out of `gameover.c` as it is started to be used in another handler too. I do not know what `SET_UNK_0B` or `SET_UNK_12` do, but I added in case at some point we want to explore their meaning.

`func_800F298C` is the function that handles the whole gameplay. That is the biggest DRA function and the second biggest function of the game. I hope one day we will manage to decompile that too.

In `case Play_LoadStageChr` there seems to be a hidden loading screen populated by that `for` loop right after `D_8013640C = AllocPrimitives(PRIM_GT4, 16);`. It is hidden because `blendMode` is set to `0`. Setting it to either `8` or `0x88` shows a black screen rather than showing Alucard frozen in the CD room for those two seconds. I do not believe it was supposed to be black as X2, Y2, X3 and Y3 have some kind of rotation and the U/V coords hints the fact some textures were supposed to be loaded.

### DIFF
--- a/src/dra/play.c
+++ b/src/dra/play.c
@@ -40,7 +40,7 @@ void HandlePlay(void) {
         D_8003C9A4 = 0;
         D_8003C738 = 0;
         if (D_8003C730 == 0) {
-            for (i = 0; i < 0x300; i++) {
+            for (i = 0; i < LEN(D_8003BDEC); i++) {
                 D_8003BDEC[i] = 0;
             }
             D_8003BDEC[0xB9] = 1;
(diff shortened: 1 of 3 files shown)


sonicdcer (approved):

src/dra/play.c
```diff
@@ -21,16 +21,278 @@ void func_800E4970(void) {
...
+            for (i = 0, ptr = &g_MenuNavigation; i < sizeof(MenuNavigation);
+                 i++) {
+                *ptr++ = 0;
+            }
+
+            for (i = 0; i < BUTTON_COUNT; i++) {
+                g_Settings.buttonConfig[i] = i;
+                g_Settings.buttonMask[i] = D_800A0160[i];
+            }
+            for (i = 0; i < LEN(g_Settings.cloakExteriorColors) +
```
sozud, line 66: I think this suggests there was a single array of cloak colors but the current way we have it is probably more readable. I don't think this needs to be changed, just observing.

sozud (approved):

PR #368, opened 2023-07-17
DRA ApplyJosephsCloakPalette matching

Xeeynamo:
Simple and effective. Not sure why we never decompiled this before.

I renamed `g_josephsCloakLiningDark555` into just `g_JosephsCloak.liningDark` and I believe it is a bit simpler for the eyes to decode.

I also changed the type `g_Settings.cloak*` to be a single array of `6` rather than two of `3` given how I matched it in #367.

I also refactored `DarkenCloakColor` as it was a fake match.

Finally I synchronised the symbols between the US and the HD versions.


src/dra/5298C.c
```diff
@@ -546,19 +546,30 @@ void DrawMenuAlucardPortrait(MenuContext* ctx) {
...
+// Equivalent of floor(number * .75)
+// Cloak color components are 5 bit.
+// Examples: 31->23, 15->11
+s32 DarkenCloakColor(s32 color) { return color * 3 / 4; }
 
-// Apply cloak palette
 // Creates light and dark versions of cloak colors in BGR555 format
-INCLUDE_ASM("asm/us/dra/nonmatchings/5298C", ApplyJosephsCloakPalette);
+void ApplyJosephsCloakPalette(void) {
+    g_JosephsCloak.lingingDark =
```
sozud, line 556: "lining". https://www.merriam-webster.com/dictionary/lining

sozud (approved):
Do we know if this also handles the reverse cloak colors?

PR #374, opened 2023-07-19
Decompile DRA func_800FA7E8 to CheckWeaponCombo

bismurphy:
Been working closely with folks in Discord on this one.

Nothing to report besides what's in the code; let me know if this name is not reasonable.


src/dra/5298C.c
```diff
@@ -1337,9 +1337,37 @@ INCLUDE_ASM("asm/us/dra/nonmatchings/5298C", func_800FA3C4);
...
+// If you use both attack buttons at once, see if something special happens.
+// Applies to Shield Rod + Shield, or dual Heaven Swords
+void CheckWeaponCombo(void) {
+    s32 weapon0;
+    s32 weapon1;
+    s32 combo1;
+    s32 combo2;
+    s32 comboBits;
+    s32 i;
+    s32 oddComboCheck;
```
sozud, line 1349: What's the meaning of `odd` here? That the code for the check is strange?

sozud (approved):

Xeeynamo:
Merging master into it to check if the HD build works with your decompiled function

config/symbols.us.dra.txt
```diff
@@ -956,6 +956,7 @@ DrawSpellMenu = 0x800F7B60;
 DrawSystemMenu = 0x800F8374;
 BlinkMenuCursor = 0x800F8C98;
 DrawConsumableCount = 0x800F8E18;
+CheckWeaponCombo = 0x800FA7E8;
```
Xeeynamo, line 959: You need to rename `func_800FA7E8` from `symbols.dra.hd.txt` too

src/dra/5298C.c
bismurphy: Yes, that's correct. Hopefully when we know the proper usage of D_8013AEE4 we can rename the variable to something that usefully describes the purpose of that.

PR #376, opened 2023-07-19
Do not strip debug symbols and start to support modern GCC

Xeeynamo:
A few tweaks to the Makefile:

* `PSXCC_FLAGS` now contains all the GCC flags that only work with the old versions
* Add debug symbols in the PSX ELFs by adding `-g` and by removing `-s`
* Remove `--no-check-sections` as the reallocation would break the executable if we add too much data

I did not remove `-s` from the Saturn version. We might want to do that but we also need to add the `-g` flag in `tools/saturn-splitter/scripts/compile_dosemu/compile_dosemu.sh` (cc. @sozud)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,10 +15,12 @@ LD              := $(CROSS)ld
 CPP             := $(CROSS)cpp
 OBJCOPY         := $(CROSS)objcopy
 AS_FLAGS        += -Iinclude -march=r3000 -mtune=r3000 -no-pad-sections -O1 -G0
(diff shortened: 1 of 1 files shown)


sozud (approved):
Unless it's blocking you I think leaving the `-s` and `-g` options unchanged for the moment is fine.

Xeeynamo:
> Unless it's blocking you I think leaving the `-s` and `-g` options unchanged for Saturn for the moment is fine.

It is not a blocker but potentially something good to have for the Saturn version too. I will merge this one for the time being.

PR #379, opened 2023-07-20
Decompile func_060705CC

sozud:
Co-authored-by: Luciano Ciccariello <Xeeynamo@users.noreply.github.com>


sozud:
This one is just a copy-paste from the PSX version so I don't think it's going to be controversial, merging to avoid conflicts.

Xeeynamo:
> This one is just a copy-paste from the PSX version so I don't think it's going to be controversial, merging to avoid conflicts.

I was meant to check how similar it looked with the PS1 counterpart and forgot to review. I am happy to have that merged!

PR #384, opened 2023-07-21
DRA HD cross-referencing 47BB8 and 5298C

Xeeynamo:
HD is a japanese build. The menu size and the font spacing is slightly different in the Japanese version. The Relic menu scrolling is also different to accommodate the extra two familiars that did not make through the US release.

The HD build seems to also be older than the US counterpart, hinting being potentially something between JP 1.0 and JP 1.1 or between JP 1.1 and the E3 build. There are new functions that only appear in the US release that are responsible of printing strings in the menu. The US build also has a slightly more complex `DrawMenuStr`.

Not sure if those PRs are interesting to see. I am making them for visibility.


sozud (approved):
Looks good, interesting to see these version differences

PR #385, opened 2023-07-21
Split saturn libs, add function

sozud:
This splits the saturn libs into individual files and adds one function to test the workflow. The splitter can't split multiple files at the moment so the linker file is made manually. I plan to fix this eventually. I updated the splitter to not overwrite linker files that already exist. I'm open to suggestions on the directory structure but this seemed like the easiest way given the dosemu copying stuff.


sozud:
I had stale asm left over which was causing this to work locally, I need to fix a couple of issues with the splitter. The concept of the PR is going to be the same. I'll just be updating the subrepo hash.

sozud:
I fixed the splitter issues.

Xeeynamo (approved):
Thank you for decompiling those juicy functions in `zero` and to name them accordingly. In particular I find the naming of both functions and parameters very on-point with what the functions are really intended to do under the hood. The new file organisation is also concise and coherent with their content. This was spectacularly executed. Me and the people from the decomp appreciates the effort you took for this PR. I know you worked for weeks on this one. We will always be grateful.

PS. There are too many parentheses for my tastes... but I will let it go just for this time!!!!!! 🛸

PR #386, opened 2023-07-21
Rename Acceleration variables to velocities

bismurphy:
Entities have two variables for their velocity, called accelerationX and accelerationY. These are wrong names, because they are velocities and not accelerations.

This PR replaces every instance of these variables in all the code and symbols files. I began with a simple find-and-replace which handled most of it, and then searched the codebase for instances of "Accel" in other forms, which led me to several local variables. I changed many of these locals from accelX to velX, and similar adjustments.

The function AccelerateX has also been renamed to SetSpeedX.

More places may be found in the future where accelerations and velocities are mixed up, but hopefully with this PR those mixups will no longer continue to be created.


Xeeynamo (approved):
ggwp

PR #391, opened 2023-07-23
TimeAttack and Stage documentation

Xeeynamo:
We talked about and it is time to have some action.


include/game.h
```diff
@@ -261,6 +201,69 @@ typedef enum {
     Game_99 = 99,
 } GameState;
 
+#define STAGE_INVERTEDCASTLE_MASK 0x1F
+#define STAGE_INVERTEDCASTLE_FLAG 0x20
+typedef enum {
```
sozud, line 206: I think the manual numbering here is fine for now although there's various disagreements about style. This PR adds a lot of useful documentation and I wouldn't hold it up for this.

sozud (commented):
Other than fixing the Saturn version I think this looks good.

sozud:
Saturn is fixed now so going to go ahead and merge. I believe the enum style used here was what won in that poll a while ago so let's go with that.

PR #397, opened 2023-07-25
Decompile sin functions

SestrenExsis:
This pull request matches the following functions, which are responsible for returning the mathematical sine of a value:

- MAIN - `rsin`
- MAIN - `sin_1`


sozud (commented):
This looks good to me, thanks! It seems like the arg0 stuff can be removed like this https://decomp.me/scratch/WRbb5

Xeeynamo (approved):
Thank you Sestren!

PR #398, opened 2023-07-25
Sound effect extraction WIP

sozud:
It seems like there's some interest in figuring out more about the sound effects lately so I tried making a tool to extract them. They're in a standard Sony VAB format, split into the header (.VH) and a bundle of VAG files (.VB). VAG files are ADPCM encoded. Eventually the idea is being able to round-trip these files but this just extracts to WAV at the moment. I may be wrong but some of the sound effects appear to be unused.

Usage:
`sh test.sh`
Look at output in `./out`

The playback speed is hardcoded since I haven't figured out yet if the pitch is part of these files or somewhere else. It can be changed by altering `samp_freq`.


sozud:
I fixed a couple of issues, I think this is working for all the currently extracted overlays.

sozud:
I added tone attribute parsing and some unit tests. I think this is good enough for a first pass at the problem. There's a pitch shift value in the tone attributes but I'm not sure how to translate that to a playback speed or if that is what the game uses.

sozud:
This is now able to serialize the VH file to JSON and read it back to matching bytes. Example here https://gist.github.com/sozud/f802c0ec3d96af04ce3b51ad575ab907

sozud:
I think this tool is in a reasonable state as a first pass at the problem, merging.

PR #399, opened 2023-07-25
Function finder refactoring

Xeeynamo:
Merges some code between the PSX and the Saturn function finders. The only functional change is the PSX function finder now exports in markdown. This is to prepare the integration of #396 in the CI.

### DIFF
--- a/tools/function_finder/function_finder_psx.py
+++ b/tools/function_finder/function_finder_psx.py
@@ -4,45 +4,9 @@
 from pathlib import Path
 from tabulate import tabulate
 import os
-import requests
-import json
-import time
 import concurrent.futures
 
-
-# search for scratches with the name on decomp.me
-def find_scratches(name):
-    try:
-        response = requests.get(f"https://decomp.me/api/scratch?search={name}")
-        response.raise_for_status()
-        scratches = json.loads(response.text)
-    except requests.exceptions.HTTPError as http_err:
-        return None
-    except Exception as err:
-        return None
-
-    best_result = None
-    best_percent = 0
-
-    for result in scratches["results"]:
-        # seems to give approximate matches, skip these
-        if result["name"] != name:
-            continue
-        if result["platform"] != "ps1":
-            continue
-
-        score = result["score"]
-        max_score = result["max_score"]
-        percent = (max_score - score) / max_score
-
-        if percent > best_percent:
-            best_percent = percent
-            best_result = result
-
-    if best_result:
-        return [f"https://decomp.me{best_result['url']}", round(best_percent, 3)]
-
-    return None
+from helpers import find_scratches
 
 
 # look in asm files, read in the text and check for branches and jump tables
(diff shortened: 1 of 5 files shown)


tools/function_finder/helpers.py
```diff
@@ -0,0 +1,39 @@
...
+    best_result = None
+    best_percent = 0
+
+    for result in scratches["results"]:
+        if not "name" in result:
+            continue
+        # seems to give approximate matches, skip these
+        if result["name"] != name:
+            continue
+        if result["platform"] != "saturn":
```
sozud, line 25: I think the platform needs to be a parameter?

tools/function_finder/function_finder_saturn.py
```diff
@@ -176,22 +105,21 @@ def print_github_flavored_table(data):
...
-            obj["%"] = ""
+            row[4] = results[i]["link"]
+            row[5] = results[i]["percent"]
 
-        to_print.append(obj)
+        to_print.append(row)
 
-    print_github_flavored_table(to_print)
+    headers = ["Filename", "Function Name", "Length", "Branches", "WIP", "%"]
+    print(tabulate(to_print, headers=headers, tablefmt="github"))
```
sozud, line 125: I thought I tried this exact thing and couldn't get it to display right on GitHub. Do you have a branch pushed with the result?

tools/function_finder/function_finder_psx.py
```diff
@@ -176,8 +146,8 @@ def find_wip(o):
     # Update output with the results
     for i, o in enumerate(output):
         # keep the in-source results as definitive
-        if o[4] == "":
```
sozud, line 179: Is the (Scraped) part being handled?

tools/function_finder/helpers.py
Xeeynamo: I am glad reviewing brings light to oversights like this. Fixed this and the other change to link the scraper result to the function list

tools/function_finder/function_finder_saturn.py
Xeeynamo: It seems to work just fine. I originally tested it with a subset. This is the full dump after running the tool from the latest `make extract_saturn`: https://gist.github.com/Xeeynamo/a41713c9b2a610a67c95199641a19394

tools/function_finder/function_finder_psx.py
sozud: Based on the result here it looks like we aren't indicating whether it's an in-source DECOMP_ME_WIP or scraped anymore. I don't mind much either way, just thought I should point this out. https://gist.github.com/Xeeynamo/a41713c9b2a610a67c95199641a19394

sozud (approved):
Seems ok to me

sozud:
I'm going to go ahead and merge this to reduce the PR queue, will fix any breakage that comes up.

PR #401, opened 2023-07-27
func_801B3C38 Matching

sozud:


include/unkstruct.h
```diff
@@ -286,3 +286,12 @@ typedef struct {
     s16* coords;
     s16 unk8;
 } Unkstruct_80102CD8;
+
+typedef struct {
+    /* 0x0 */ s16 unk0; // Entity::animSet
```
sozud, line 291: @sonicdcer Should we go ahead and rename the known ones?

src/st/nz0/30958.c
```diff
@@ -1175,4 +1175,137 @@ void func_801B3B78() {
...
+            self->velocityX = temp_s0 * rsin(self->ext.generic.unk84.S16.unk0);
+            self->velocityY =
+                -(temp_s0 * rcos(self->ext.generic.unk84.S16.unk0));
+            self->step_s++;
+        }
+
+        if (self->animFrameIdx >= 13) {
+            velocityX = self->velocityX;
+            if (velocityX < 0) {
+                var_v0 = velocityX + 3;
```
sozud, line 1276: @sonicdcer var_v0 -> adjustedVelocityX or similar? Same for var_v0_2. Other than these two comments looks good.

include/unkstruct.h
sonicdcer: Yes, it would be alright.

src/st/nz0/30958.c
sonicdcer: yes, whatever you think is right i'm fine with it.

sonicdcer: changes pushed

sozud:
I think with the revisions this is pretty clean. Merging to reduce the queue and since I doubt there's going to be objections.

Xeeynamo:
We should avoid using `.ext.generic` and `.ext stub`. We decided to make separate structs for every entity for a reason. We should strictly avoid piling up code on legacy stuff. Let's keep that in mind for every new PR.

sonicdcer:
> We should avoid using `.ext.generic` and `.ext stub`. We decided to make separate structs for every entity for a reason. We should strictly avoid piling up code on legacy stuff. Let's keep that in mind for every new PR. @sozud @sonicdcer
IMO is not the right approach, because by forcing every entity to be decompiled with a new extension, you're effectively increasing the number of Unknown Entity Extensions, which will need to be replaced anyway in the future when the function is documented, entity.h will become so populated it will be gigantic, it also increases the workload on the contributor. I think using the generic extension or stub makes much more sense in the meantime because you don't need to create extensions for every entity you decompile, which will be needed to be modified again anyway when documented, avoiding unnecessary clutter inside the header, as well as saving time for the person decompiling the function.
@Xeeynamo

sozud:
@Xeeynamo Ok, I've added that to the PR guidelines page. https://github.com/Xeeynamo/sotn-decomp/wiki/Pull-Request-Guidelines#entities

I made the merge decision on this so I'll put up a fix PR.

PR #402, opened 2023-07-27
Decompile DRA func_800FF7B8

bismurphy:
This is a big one. This is the main function that runs to initialize the player's stats and equipment. It takes into account the bonuses that the player gets for Prologue completion (completing faster, or with certain items, gives bonuses of stats and items). It also checks for the famous X-X!V''Q and AXEARMOR savename codes. I therefore called it InitStatsAndGear.

This function takes only one argument. In the real game, both calls to this function have an argument of 0, but one of the first things in this function is `if (arg0 == 1) {`. I therefore assume this is a debug parameter, where developers could have called this with 1, in order to do some sort of testing. I decided to call the argument debugMode.

I chose to call the two savename codes "c_strLuckModeCode" and "c_strAxeArmorCode"; I believe this matches the existing formatting for strings that exist in the game, but let me know if I've made a mistake there.

I removed many D_8009XXXX which are within g_Status, to make sure nobody uses these in their D_ form in the future.

There is a loop in this function which iterates, in blocks of 3, over the last 21 entries in g_Status. I noticed that this overlaps with the Sword Familiar level check in CalcAttack, so I believe this to be the stats for each of the 7 available familiars. One stat is the level, another should be the EXP, and the other is something else. I will leave this as a matter of future work, since the details of familiars are not key to the decompiling of this particular function.

I tried to comment this function fairly heavily. Many variables are touched by this and there are a large number of if-statements which are highly relevant to the game logic. Given that different people reading the code have different levels of familiarity with behavior of the game, being clear about the circumstances where a variable holds a certain value seems useful.

This function was a DECOMP_ME_WIP previously; I did not use that preexisting code and instead decompiled this function from scratch myself.

I think that covers all the things worth mentioning!


src/dra/5D6C4.c
```diff
@@ -798,282 +754,268 @@ void func_800FF7B8(s32 arg0) {
...
+                g_Status.relics[15] = 3;
+                g_Status.relics[0] = 3;
+                g_Status.relics[1] = 3;
+                g_Status.relics[4] = 3;
+                g_Status.relics[5] = 3;
+                g_Status.relics[6] = 3;
+                g_Status.relics[7] = 3;
+                g_Status.relics[12] = 3;
+                g_Status.relics[13] = 3;
+                AddToInventory(0x6F, 0); // Firebrand
```
sozud, line 971: This function has so many item IDs I'm wondering if it's a good time to create an enum?

config/symbols.us.txt
```diff
@@ -1027,25 +1027,8 @@ g_Status_relics_27 = 0x8009797F;
 g_Status_relics_28 = 0x80097980;
 g_Status_relics_29 = 0x80097981;
 g_Status_spells = 0x80097982;
-D_8009799A = 0x8009799A;
```
sozud, line 1030: Should we have `g_Status_statsFamiliars_0_level` etc to improve the asm?

src/dra/5298C.c
```diff
@@ -169,8 +169,9 @@ s32 CalcAttack(s32 equipId, s32 otherEquipId) {
     if (equipId == 4 && D_800A4B04[otherEquipId].itemCategory == ITEM_SHIELD) {
         totalAttack += 5;
     }
-    if (equipId == 0x7E) {                  // Equippable Sword Familiar
-        totalAttack += g_Status.D_80097C74; // Level of sword familiar
+    if (equipId == 0x7E) { // Equippable Sword Familiar
+        totalAttack +=
+            g_Status.statsFamiliars[4].level; // Level of sword familiar
```
sozud, line 174: enum for familiar IDs?

sozud (commented):
Nice work on a long and difficult function!

config/symbols.us.txt
bismurphy: Probably not necessary at least for now - I don't think many places in the game will really care about the level of familiars? Maybe we can revisit that later but for now it seems like that would be a lot of symbols for not much benefit. Open to other suggestions though.

src/dra/5298C.c
bismurphy: Probably a good idea, but not for now for this function - Determining each familiar's ID would be its own task. I think someone working out the 3 parameters of statsFamiliar (level, probably xp, and something else), along with the ID for each familiar, would be a good task to tackle. I might do that some time soon, or someone else can if it interests them.

src/dra/5D6C4.c
bismurphy: Possibly, depending on how many places reference item IDs. Implementing that and replacing all the existing uses of item ID's sounds like a full separate PR worth of work though.

Maybe we should make a Wiki page of "Things someone should look into that would help with the decompilation but are not actually decompiling assembly" - including constructing enums like this one. We track all sorts of progress on decompiled code/functions, but these other side tasks that don't increase percentage tend to be forgotten. Determining the purpose of D_80XXXX variables and giving them useful names could be another task of that type.

src/dra/5D6C4.c
```diff
@@ -592,199 +592,155 @@ void func_800FF708(s32 arg0, s32 arg1) {
...
         func_800FF60C();
         return;
     }
-    if (arg0 == 1) { // First block fully matching
+    if (debugMode == 1) {
+        // Remove Alucard Sword from left hand
         if (g_Status.equipment[0] == 0x7B) {
             g_Status.equipment[0] = 0;
+            // Same from right hand
         } else if (g_Status.equipment[1] == 0x7B) {
```
sonicdcer, line 611:
```suggestion
} else if (g_Status.equipment[1] == ITEM_ALUCARD_SWORD) {

File: src/dra/5D6C4.c [hunk 5]
```diff
@@ -592,199 +592,155 @@ void func_800FF708(s32 arg0, s32 arg1) {
...
             g_Status.equipment[1] = 0;
-        } else if (D_80097A05 != 0) {
-            D_80097A05--;
+            // Same from inventory
+        } else if (g_Status.equipHandCount[0x7B] != 0) {
```
sonicdcer, line 614:
```suggestion
} else if (g_Status.equipHandCount[ITEM_ALUCARD_SWORD] != 0) {

File: src/dra/5D6C4.c [hunk 6]
```diff
@@ -592,199 +592,155 @@ void func_800FF708(s32 arg0, s32 arg1) {
...
         }
+        // Same logic, for Alucard Shield
         if (g_Status.equipment[0] == 0x10) {
             g_Status.equipment[0] = 0;
         } else if (g_Status.equipment[1] == 0x10) {
             g_Status.equipment[1] = 0;
-        } else if (D_8009799A != 0) {
-            D_8009799A--;
+        } else if (g_Status.equipHandCount[0x10] != 0) {
+            g_Status.equipHandCount[0x10]--;
```
sonicdcer, line 624:
```suggestion
} else if (g_Status.equipHandCount[ITEM_ALUCARD_SHIELD] != 0) {
            g_Status.equipHandCount[ITEM_ALUCARD_SHIELD]--;

**Review by sonicdcer — APPROVED** [review]
Terrific work on this one!!!
I have no other suggestion other than using the items enum inside /src/include/items.h, i've pointed out some of them but not all, you get the idea :)

File: src/dra/5D6C4.c [hunk 7]
```diff
@@ -592,199 +592,155 @@ void func_800FF708(s32 arg0, s32 arg1) {
...
-    int new_var;
+    u8* namePtr;
 
     if (D_8003C730 != 0) {
         func_800F53A4();
         func_800FF60C();
         return;
     }
-    if (arg0 == 1) { // First block fully matching
+    if (debugMode == 1) {
```
sonicdcer, line 606:
```suggestion
if (debugMode == true)

sonicdcer: since it's boolean, or just if (debugMode) if you prefer, just a nit without too much importance

sozud:
@bismurphy The PR queue is getting big again, are you planning to respond to any of @sonicdcer 's feedback? Just checking so I don't merge this prematurely. I think this is good work regardless of additional documentation or not so I'm inclined to merge to reduce the queue.

sozud:
Got confirmation in Discord, merging to keep the queue moving.

PR #404, opened 2023-07-27
Move Primitive handling functions away from POLY_GT4

bismurphy:
Per discussion in Discord, renamed func_80107360 to SetTexturedPrimRect.

This raised the issue that both SetPolyRect and this function should use Primitive as their argument, and not POLY_GT4. Therefore, these are changed. Additionally, everywhere in the codebase that calls these functions now uses Primitive for their local variables which get passed to these functions.

Finally, we update every case of setting blendMode to 8, to use the BLEND_VISIBLE flag. Hopefully in the future we can identify the rest of the available blendModes.


sozud (approved):
This looks reasonable to me but Xeeynamo should probably approve before merging.

src/dra/5298C.c
```diff
@@ -459,7 +459,7 @@ void DrawMenuSprite(
     MenuContext* context, s32 x, s32 y, s32 width, s32 height, s32 u, s32 v,
     s32 clut, s32 tpage, s32 arg9, s32 colorIntensity, s32 argB) {
     u32* ot = g_CurrentBuffer->ot;
-    POLY_GT4* poly = &g_CurrentBuffer->polyGT4[g_GpuUsage.gt4];
+    Primitive* poly = &g_CurrentBuffer->polyGT4[g_GpuUsage.gt4];
```
Xeeynamo, line 462: We have to change it back to `POLY_GT4*`. I know we are still passing it to `SetTexturedPrimRect` but it is fine as the order of fields is the same.

src/dra/gameover.c
```diff
@@ -31,12 +31,12 @@ void func_800E5358(void) {
 }
 
 void func_800E5498(void) {
-    POLY_GT4* poly = &g_CurrentBuffer->polyGT4[g_GpuUsage.gt4];
+    Primitive* poly = &g_CurrentBuffer->polyGT4[g_GpuUsage.gt4];
```
Xeeynamo, line 34: Here too, we want to revert it back as `POLY_GT4*`

Xeeynamo (approved):
I am glad you took the time to refactor this ancient part of the decompilation. It was something that had to be done sooner or later.
I noticed just two spots where `POLY_GT4*` needs to be used instead of `Primitive*` and it is essentially when the struct pointer is taken from `g_CurrentBuffer->polyGT4`. Once these two spots are reverted back I will be happy to merge this in!

sozud:
I'm not sure what the build error is about but I'm going to try merging master into this.

sozud:
Actually looks like a couple of spots were missed
```
src/dra/5298C.c: In function `DrawMenuSprite':
src/dra/5298C.c:471: structure has no member named `type'
src/dra/5298C.c:474: structure has no member named `type'
src/dra/5298C.c:476: structure has no member named `type'
```

sozud:
Looks good, merging to try to get all these conflicted PRs through the pipeline.

PR #405, opened 2023-07-27
Decompile NO3 EntityRoomTransition2

bismurphy:
I'm still not super confident decompiling jump table/rodata functions, but I think I did this one right. Took a while fighting the splat to get things to match.

Two things of note with this function. Thing 1: In my initial decompilation, I had the local variables `newEntity`, `i`, and `primIndex`. I found that `i` and `primIndex` used the same register, and in my code I was seeing them on different registers. I ended up lumping them into one variable. Since it serves two totally unrelated purposes, I don't have a good name for it, so I just called it localVar.

Thing 2: The function starts by preloading pointers for `g_Entities` and `g_CurrentRoomTileLayout`. I couldn't get the compiler to do this by itself, so I ended up creating two local variables that hold these locations. It's ugly, but I guess it works.

This ends up matching. Note that this calls my other recently-decompiled function func_800FF7B8/InitStatsAndGear (which is how I wound up decompiling this one in the first place), so these two PR's will create a conflict since this one still calls that function by its old name.


src/st/no3/3FF00.c
```diff
@@ -0,0 +1,1208 @@
...
+
+void EntityRoomTransition2(Entity* self) {
+    Entity* newEntity;
+    Primitive* prim;
+    // Used as both a loop variable and a temp
+    s32 localVar;
+
+    // These are weird fake vars needed for matching
+    Entity* gents = g_Entities;
+    Unkstruct8* layout = &g_CurrentRoomTileLayout;
```
sonicdcer, line 11: These are not fake, it's a common pattern for entity functions that use g_CurrentRoomTileLaout.

src/st/no3/3FF00.c
```diff
@@ -0,0 +1,1208 @@
...
+
+    switch (self->step) {
+    case 0:
+        InitializeEntity(D_80180AD0);
+        layout->unk40 = 0xFC;
+        g_Player.D_80072EF4 = 0x2000;
+        D_8003C8B8 = 0;
+        g_Player.D_80072EFC = 1;
+        D_801D7DD0 |= 0x100;
+        return;
```
sonicdcer, line 27:
```suggestion
break;

File: src/st/no3/3FF00.c [hunk 3]
```diff
@@ -0,0 +1,1208 @@
...
+        if (gents[0].posX.i.hi < 0x34) {
+            g_Entities[1].ext.generic.unk7C.S8.unk0 = 1;
+            g_Player.D_80072EF4 = 0x2000;
+        } else {
+            g_Player.D_80072EF4 = 0;
+            D_801D7DD0 |= 0x80;
+            self->step++;
+        }
+        g_Player.D_80072EFC = 1;
+        return;
```
sonicdcer, line 39:
```suggestion
break;

File: src/st/no3/3FF00.c [hunk 4]
```diff
@@ -0,0 +1,1208 @@
...
+                }
+                CreateEntityFromEntity(0x59, gents, newEntity);
+                newEntity->params = localVar;
+            }
+            g_Player.D_80072EF4 = 0x10000;
+            self->ext.generic.unk7C.s = 0x10;
+            self->step++;
+        }
+        g_Player.D_80072EFC = 1;
+        return;
```
sonicdcer, line 57:
```suggestion
break;

File: src/st/no3/3FF00.c [hunk 5]
```diff
@@ -0,0 +1,1208 @@
...
+                prim->r0 = 0;
+                prim->blendMode = 0x35;
+                self->step++;
+            } else {
+                self->ext.generic.unk7C.s++;
+            }
+        }
+        g_Player.D_80072EF4 = 0;
+        g_Player.D_80072EFC = 1;
+        return;
```
sonicdcer, line 81:
```suggestion
break;

File: src/st/no3/3FF00.c [hunk 6]
```diff
@@ -0,0 +1,1208 @@
...
+        if (prim->r0 >= 0xF0) {
+            self->step++;
+            DestroyEntity(&g_Entities[208]);
+            g_CurrentRoom.bg[0].D_800730F4 |= 1;
+            g_api.PlaySfx(0xA1);
+            g_api.PlaySfx(0x30F);
+        }
+        g_Player.D_80072EF4 = 0;
+        g_Player.D_80072EFC = 1;
+        return;
```
sonicdcer, line 94:
```suggestion
break;

File: src/st/no3/3FF00.c [hunk 7]
```diff
@@ -0,0 +1,1208 @@
...
+            DestroyEntity(self);
+            gents = &g_Entities[192];
+            layout->unk40 = 0;
+            D_8003C8B8 = 1;
+            DestroyEntity(gents);
+            CreateEntityFromCurrentEntity(0x17, gents);
+        }
+        g_Player.D_80072EF4 = 0;
+        g_Player.D_80072EFC = 1;
+        return;
```
sonicdcer, line 108:
```suggestion
break;

File: src/st/no3/3FF00.c [hunk 8]
```diff
@@ -0,0 +1,1208 @@
+#include "no3.h"
+
+void EntityRoomTransition2(Entity* self) {
+    Entity* newEntity;
+    Primitive* prim;
+    // Used as both a loop variable and a temp
+    s32 localVar;
+
+    // These are weird fake vars needed for matching
+    Entity* gents = g_Entities;
```
sonicdcer, line 10:
```suggestion
Entity* player = &PLAYER;

File: src/st/no3/3FF00.c [hunk 9]
```diff
@@ -0,0 +1,1208 @@
...
+        InitializeEntity(D_80180AD0);
+        layout->unk40 = 0xFC;
+        g_Player.D_80072EF4 = 0x2000;
+        D_8003C8B8 = 0;
+        g_Player.D_80072EFC = 1;
+        D_801D7DD0 |= 0x100;
+        return;
+    case 1:
+        // Evil use of local 'gents' instead of PLAYER
+        if (gents[0].posX.i.hi < 0x34) {
```
sonicdcer, line 30:
```suggestion
if (player->posX.i.hi < 0x34) {

File: src/st/no3/3FF00.c [hunk 10]
```diff
@@ -0,0 +1,1208 @@
...
+                self->primIndex = localVar;
+                self->flags |= 0x800000;
+                prim->u0 = 0xFF;
+                prim->v0 = 0xFF;
+                prim->priority = 0x50;
+                prim->x0 = 0;
+                prim->y0 = 0;
+                prim->g0 = 0;
+                prim->b0 = 0;
+                prim->r0 = 0;
```
sonicdcer, line 72:
```suggestion
prim->r0 = prim->b0 = prim->g0 = prim->y0 = prim->x0 = 0;

File: src/st/no3/3FF00.c [hunk 11]
```diff
@@ -0,0 +1,1208 @@
...
+        return;
+    case 3:
+        if (D_801D7DD0 & 0x40 && !(--self->ext.generic.unk7C.s & 0xFFFF)) {
+            localVar = g_api.AllocPrimitives(PRIM_TILE, 1);
+            if (localVar != -1) {
+                prim = &g_PrimBuf[localVar];
+                self->primIndex = localVar;
+                self->flags |= 0x800000;
+                prim->u0 = 0xFF;
+                prim->v0 = 0xFF;
```
sonicdcer, line 66:
```suggestion
prim->v0 = prim->u0 = 0xFF;

**Review by sonicdcer — CHANGES_REQUESTED** [review]
Great work on this one!

File: src/st/no3/3FF00.c [hunk 12]
```diff
@@ -0,0 +1,1208 @@
...
+    Entity* newEntity;
+    Primitive* prim;
+    // Used as both a loop variable and a temp
+    s32 localVar;
+
+    // These are weird fake vars needed for matching
+    Entity* gents = g_Entities;
+    Unkstruct8* layout = &g_CurrentRoomTileLayout;
+
+    if (self->ext.generic.unk80.modeS32 == 0 && self->step < 4) {
```
sozud, line 13: We are trying to phase out usage of `ext.generic`. Would you mind making a struct? See this PR https://github.com/Xeeynamo/sotn-decomp/pull/409

src/st/no3/3FF00.c
Xeeynamo: While I agree on `prim->r0 = prim->b0 = prim->g0 = 0;` (it is a pattern commonly seen in the decomp) I do not for `prim->y0 = prim->x0 = 0;`. The struct field order should be from the left to the right and from the top to the bottom whenever it is matching regardless the order. This is also valid for the `prim->v0 = prim->u0 = 0xFF;` suggested/requested below.

src/st/no3/3FF00.c
```diff
@@ -0,0 +1,1208 @@
...
+        }
+        g_Player.D_80072EFC = 1;
+        return;
+    case 3:
+        if (D_801D7DD0 & 0x40 && !(--self->ext.generic.unk7C.s & 0xFFFF)) {
+            localVar = g_api.AllocPrimitives(PRIM_TILE, 1);
+            if (localVar != -1) {
+                prim = &g_PrimBuf[localVar];
+                self->primIndex = localVar;
+                self->flags |= 0x800000;
```
Xeeynamo, line 64:
```suggestion
                self->flags |= FLAG_HAS_PRIMS;
```

src/st/no3/3FF00.c
bismurphy: I prefer "return" in cases like this, because when reading the code you know the function is done, rather than being the style of function where something "extra" happens after the switch statement. If tracing program flow, reaching the return means you don't need to scroll to the end of all the cases to know you've seen everything that happens for a given case. I think I will leave these as return, unless there is a competing reason not to.

src/st/no3/3FF00.c
bismurphy: This doesn't work though because on line 100 we have ` gents = &g_Entities[192];` - I definitely don't want to name the variable `player`, because then that line would have to say `player = &g_Entities[192]` which doesn't make much sense.

sozud:
This is currently conflicted but once it gets rebased I think it's ready to be merged.

include/entity.h
```diff
@@ -268,6 +268,14 @@ typedef struct {
     /* 0x80 */ s32 unk80;
 } ET_RoomTransition2;
 
+    /* 0x7C */ u8 pad0[0x4];
```
sozud: Looks like this needs `typedef union {`

include/game.h
```diff
@@ -966,7 +966,7 @@ typedef struct {
     /* 8003C848 */ void (*func_800FE044)(s32, s32);
     /* 8003C84C */ void (*AddToInventory)(u16 itemId, s32 itemCategory);
     /* 8003C850 */ RelicOrb* D_800A8720;
-    /* 8003C854 */ void (*func_800FF7B8)(s32 arg0);
+    /* 8003C854 */ void* InitStatsAndGear;
```
sozud: Probably `void (*InitStatsAndGear)(s32 arg0);`

sozud:
Looks like this is ready now, going to merge to continue trying to get all the conflicted PRs moving.

PR #406, opened 2023-07-28
Decompile DRA func_80112BB0

bismurphy:
Decompiled by request of Mazenk in Discord.

Appears to handle a lot of player movement logic. I think this might also involve the dive-kick. Honestly not entirely clear what this is for. Seems like determining g_Player.unk44 is might help understand.

D_800ACF7C is kind of weird, it seems like it's just a table of 8 bytes where, depending on the player's Y velocity, we read from bytes 0, 1, and 2 or bytes 4, 5, and 6. The first two bytes we read are an s16 and the third is a u8. Not sure if this really needs to be a structure, but it works.

I think I handled all the hex values that have enums (particularly the PAD_whatever values), but if I missed any I'm sure someone will let me know :)

Proud of my speed on this one. From the time of the request to submitting this PR was only one hour. And this is a jump table instruction too!


sozud:
I tried fixing this but haven't been able to figure it out. I think the issue is that the new file split isn't reflected in the `hd` version.

From the hd symbols file we see `func_80112BB0 = 0x801128A0;` 

We need to get the physical address of the new file split from taking the difference of the load address and that symbol in the `hd` version:
`0x801128A0 - 0x800A0000 = 0x728A0`.

So we set the new file to that address:
```
       - [0x71574, c, 71830]
+      - [0x728A0, c, 72BB0]
       - [0x7379C, c, 73AAC_hd]
```
However I get various undefined symbol issues when doing this so I'm not sure what I'm missing.

Xeeynamo (approved):
There are some magic values for speed and velocity. They will probably be handled after #408 after that one is merged. Just sharing it for transparency. Good job on this one! I will look into what's wrong with HD too

Xeeynamo:
image
LEFT: HD
RIGHT: decompiled US

as you can see some symbols are different. What in HD is `D_801393F0`, in US is `D_80139824`. The two symbols needs to be cross-referenced. In this case it will become `D_80139824 = 0x801393F0;` and it needs to be added into `symbols.hd.dra.txt`. There will be times though where the symbol `D_80139824` is already used somewhere else and the two names would overlap. In that case you just do `D_80139824_hd = 0x80139824`, where `_hd` means no functions have been found to cross-reference that symbol yet.

For this PR, you can fix it with the following:

### config/splat.hd.dra.yaml

```yaml
# line 99
- [0x414A4, .rodata, 72BB0]
```

```yaml
# add a new line after 172
- [0x7291C, c, 72BB0]
```

### config/symbols.hd.dra.txt

```
D_800ACF7C = 0x800AD040;
D_80138FC8 = 0x80138B94;
D_80138FCA = 0x80138B96;
D_80139824 = 0x801393F0;
```

sozud:
This one has also gotten conflicted and needs a rebase.

sozud:
Looks like this one is clean now, merging.

PR #407, opened 2023-07-29
Decompile DRA func_800FE044

bismurphy:
This function is responsible for upgrading the player's stats. This is called upon collection of a Life Max Up, Heart Max Up, or when killing an enemy to get experience. This function gives experience to the player and their active familiar.

HUGE thanks to @mkst for the maspsx update that allowed this one to match!

The arguments to this function are a little unusual. arg0 is mostly used as the amount to increase the heartMax, hpmax, or exp, but in the case where we're getting a relic, is used as the relic ID. Similarly, arg1 is mostly used to identify which of these traits is being increased, but in the case where we're getting exp, is used to define the enemy's level to determine how much exp to award. It seems odd that increasing each of these stats is baked into this one function, instead of all being their own functions.

This is one of the first functions I looked at when I joined the decomp several months ago and it's very cool to have it finally working, now using the extra knowledge I've picked up over that time.


src/dra/5D6C4.c
```diff
@@ -240,7 +240,122 @@ void func_800FDE00(void) {
...
+        return 0;
+    }
+
+    // Heart Max Up
+    if (type == 0x4000) {
+        if (g_CurrentPlayableCharacter != 0) {
+            return 1;
+        }
+        oldHeartMax = g_Status.heartsMax;
+        if (oldHeartMax == 0x270F) {
```
sozud, line 277: 0x270F is 9999

sozud (approved):

sozud:
Just noticed this has conflicts, other than than I think it looks about ready to merge

sozud:
Looks good now, merging before more conflicts appear

PR #408, opened 2023-07-29
Document some fixed point numbers, add script

sozud:
I'm putting this up as a proof of concept for documenting fixed point numbers. This does `self->velocityX` in NZ0 to get started. I'm including a script to do this although it's currently hard-coded just for velocityX. I think it can be extended for more situations. I chose `SOTN_FIXED` for the macro because the Saturn SDK already uses `FIXED`. I'm open to suggestions on the macro name. I chose 4 digits of precision since that's the highest precision currently seen so far with `SOTN_FIXED(0.1875);`

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -35,6 +35,6 @@
 
 int sprintf(char* dst, const char* fmt, ...);
 
-#define FIX(x) (((x)*65536.0))
+#define FIX(x) ((u32)((x)*65536.0))
 
 #endif
--- a/include/common.h
+++ b/include/common.h
@@ -35,4 +35,6 @@
 
 int sprintf(char* dst, const char* fmt, ...);
 
+#define FIX(x) (((x)*65536.0))
+
 #endif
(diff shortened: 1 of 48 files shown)


Xeeynamo:
This is such a cool idea. I would like to see it applied everywhere in the code base. There might be some places where this technique might not work as expected.

For the macro name what do you think about just using `F`? Like `F(0.125)`

Looking forward to the future, I would like to have this one and many more helpers integrated straight in our formatting pipeline. Creating code to help us to write code 🤓

sozud:
Is `FIX` fine? I'm thinking `F` may end up being hard to grep later on. I updated the script to handle more cases and linted all of `src`. The script is setup to return nonzero if it finds differences so this should be possible to set up in CI to check.

Xeeynamo:
I like `FIX`, yes! Not too long but also not as ambiguous as `F`.

Xeeynamo (approved):

sozud:
I rebased, going to go ahead and merge since this is going to cause more conflicts the longer it's up.

PR #409, opened 2023-07-29
Fix func_801B3C38

sozud:
This fixes up func_801B3C38 to PR guidelines regarding usage of ext and stub as discussed in https://github.com/Xeeynamo/sotn-decomp/pull/401#issuecomment-1656491828


Xeeynamo (approved):
Thank you. It looks so much better!

PR #410, opened 2023-07-29
Initial work on items script

sozud:
This is the start of a script to reformat item constants. This sort of stuff seems to be a regular pain point in PR reviews so I'm trying to figure out ways to automate more of it. It seems like calls to AddToInventory with a category other than 0 add an offset to the items array, but the function isn't decompiled yet.


sozud:
The enum is apparently split into two halves for AddToInventory, hand items and everything else. I'll think more about how to handle this in a cleaner way.

sozud:
Closing this since it got done manually

PR #411, opened 2023-07-30
Item Enum macro PR

sonicdcer:
@sozud and I were discussing about Items Enum usage since the decompilation of **InitStatsAndGear** gave us new clues about how these were likely used in the original code. This is how i think it works.
The HAND and HAND_COUNT macros are mostly unnecessary but they are there for consistency, i don't know if people prefer them gone.

Please let me know what you think, obviously the names are up for debate.


include/items.h
```diff
@@ -1,4 +1,39 @@
-typedef enum {
+#define ITEM_HEAD_START 0xC3
+#define ITEM_ARMOR_START 0xA9
+#define ITEM_CAPE_START 0xD9
```
sozud, line 3: These should use the enum values like `#define ITEM_CAPE_START ITEM_NO_CAPE`

sonicdcer: Done

sozud (approved):
This seems about the best possible solution as far as I can tell. I'm not sure why they decided to handle item ID values this way.

Xeeynamo:
image

It is not the right approach. I already mentioned the enum overflows into the next structure. We do not need all these macros.

sozud:
So the correct approach is a separate enum per item type, with the numbering starting at 1? Then there's a NULL_ITEM define or similar?

sonicdcer:
I think he meant the first value would be NO_ARMOR which would be 0. 
This is line 421 of D600.c, CEN overlay.
Can you explain how your theory works here?

```suggestion
if (D_8019D424 & 8) {
            CreateEntityFromCurrentEntity(E_EQUIP_ITEM_DROP, &g_Entities[204]);
------>     g_Entities[204].params = ITEM_HOLY_GLASSES;
            g_Entities[204].step = 5;
            g_Entities[204].flags = 0;
            self->step++;
        }

**Comment by Xeeynamo** [comment]
image
 setting everything to `01`. Also check how `CheckEquipmentItemCount` works.

**Comment by sozud** [comment]
Unless I'm misunderstanding something there seems to be a difference between how items are stored in-memory and the IDs that are used when calling AddToInventory.

For example:
```
AddToInventory(1, 2); // Cloth Tunic
AddToInventory(6, 0); // Knight Shield
```

Cloth tunic would be `/* 0x001 */ ITEM_MONSTER_VIAL_1` if the IDs were consistent. So it seems like `AddToInventory` uses different separated enums vs. `CheckEquipmentItemCount`, which uses the current big sequential enum.

**Comment by Xeeynamo** [comment]
What I think it is simply happening is that the Hand/Weapon enum finishes with `HAND_END`, the `PlayerStatus` has `u8 equipHandCount[HAND_END];` and a simple macro `WEAR(x) ((x) + HAND_END` would solve the `EntityEquipItemDrop` problem:

```c
            if (itemId < HAND_END) {
                g_api.LoadEquipIcon(g_api.D_800A4B04[itemId].icon,
                                    g_api.D_800A4B04[itemId].palette, index);
            } else {
                itemId -= HAND_END;
                g_api.LoadEquipIcon(g_api.D_800A7718[itemId].icon,
                                    g_api.D_800A7718[itemId].palette, index);
            }
```

the only reason why it works that way is because an entity parameter can hold a single value and there is no way to specify both Equip ID and Equip Type. So they come up with `itemId -= HAND_END;` to circumnavigate the problem.

sozud:
What do you think about the count arrays? 

```
        // Same logic, for Dragon Helm
        if (g_Status.equipment[2] == 0x2D) {
            // ID for Unequip in head slot
            g_Status.equipment[2] = 0x1A;
        } else if (g_Status.equipHeadCount[0x13] != 0) {
            g_Status.equipHeadCount[0x13]--;
        }
```
https://github.com/Xeeynamo/sotn-decomp/blob/60e5c27d3854ef4bc6fddfe75be6bf73452e7ab1/src/dra/5D6C4.c#L627

Here we have 0x2D for the "full ID", and 0x13 for the ID within the count array. In the full enum we have:
```
    /* 0x0D6 */ ITEM_DRAGON_HELM,
```

HAND_END would be 0xA9, so 0xD6-0xA9 would be 0x2D. so that explains the `g_Status.equipment[2] == 0x2D`, but `g_Status.equipHeadCount[0x13]` seems to require more than just `WEAR(x)`?

Xeeynamo:
> equipHeadCount

I think `equipBodyCount`, `equipHeadCount`, `equipCloakCount` and `equipOtherCount` should be part of the same array. That is the structure I came up with when I still did not have enough knowledge on how the Item IDs worked:

```c
        for (i = 0; i < 90; i++) {
            g_Status.equipBodyCount[i] = 0;
            g_Status.equipBodyOrder[i] = i;
        }`
```

This is the proof that they were originally one single array. If you sum the count of each individual item count of the various equips, you will get exactly the number `90`.

```
      - [0x4B04, equipment, equipments]
      - [0x7718, accessory, accessory_data]
```
This is another proof. The `equipment` contains 169 rows, while `accessory` contains 90 rows.

sozud:
@sonicdcer Would you be able to try combining the count arrays like Xeeynamo is suggesting?

sonicdcer:
@sozud Okay, i'll try in a few hours.

sonicdcer:
Continues in PR #440

PR #413, opened 2023-07-30
Add tool to automatically perform symbols cross referencing

Xeeynamo:
I wanted to make this tool for a long time. Given the fact HD recently became more dense into DRA, people who are willing to contribute to that overlay should not have the burden of manually cross-referencing symbols as I have been doing for the past weeks. Instead of creating a tutorial I created a tool that cross-reference symbols between functions. The two functions can be either between two different revisions of the game or between two overlays. The latter is especially important to finally remove all the duplicate functions.

I also took the opportunity to add some unit tests (finally!) in `tools/tests` instead of creating the dedicated folder `tools/symbols` with tools and test in it. Let me know if `tools/tests` is a pattern we want to keep following for any future tool that needs to be tested.

I also added an assembly parser written in Python. As for today it lives in `tools/symbols.py`. It is barebone but it can parse the majority of assembly lines. I also added tests around it in case we want to expand it or extract the parser into its own utility. I've been thinking to ideally use it for a V2 of `find_duplicates.py` if we find the current tool to be limiting.

### Demo

```
sotn-decomp$ ./tools/symbols.py cross asm/us/dra/nonmatchings/71830/func_80112BB0.s asm/hd/dra/nonmatchings/71830/func_80112BB0.s
D_80139824 = 0x801393F0;
D_800ACF7C = 0x800AD040;
D_800ACF7E = 0x800AD042;
D_80138FC8 = 0x80138B94;
D_80138FCA = 0x80138B96;
```

these symbols can be automatically copy&pasted into `symbols.hd.dra.txt` or the relevant symbol list.


sozud (approved):
I don't understand all the details about how this works but it looks cool. I think my preference would be putting tests alongside the tool rather than a single test directory but I don't have a particularly strong opinion about it. Some in-source documentation explaining how to use the tool might be nice.

sozud:
This looks good to me, merging.

PR #412, opened 2023-07-30
Relic documentation and linter

sozud:
This adds a script to fix relic magic numbers and applies it.

PR #414, opened 2023-07-30
Lint followup + Relics docs

Xeeynamo:
Follow up to #412 . This also integrates the two new linters within the formatter. By the way, are we happier with `RELIC_END` or `RELIC_COUNT`?


sozud (approved):
Seems good to me. I think `NUM_RELICS` sounds the most natural but any of these is fine.

PR #415, opened 2023-07-31
Document familar IDs

sozud:
I checked this by editing the level of the familiar and looking at the familiar status menu. The JP only ones still get shown in the menu if you have the relic.


sozud:
I made the `NUM_FAMILIARS` changes, going to go ahead and merge.

PR #416, opened 2023-07-31
Document CastSpell function on PSX and Saturn

Xeeynamo:
As per title, plus an enum for the Spell IDs.

Side note: I see there is a lot of copy & paste between the PSX and Saturn, especially on the headers. While it is fine for the small tasks as this one, I believe at some point we should at least converge the headers to reduce maintenance.


sozud (commented):
Thanks for updating the Saturn version. Seems like it doesn't like the `u8` to `s32` change. Here's a scratch https://decomp.me/scratch/zHZsy Other than that looks good

Regarding the Saturn headers, I originally wanted to combine more code but there was a lot of pushback on Discord so I left it alone.

PR #418, opened 2023-07-31
HD DRA cross-reference 73AAC and 953A0

Xeeynamo:
More cross-referencing between US and HD. The good news is with this one only 4 files remains to be merged. The bad news is that I left the biggest functions as last. #413 should massively help.


sozud (approved):
Looks good to me except it seems like the PSX CI didn't run?

PR #420, opened 2023-07-31
HD DRA cross reference 5d6c4

Xeeynamo:
Follows #402 , the functions in `5d6c4.c` are almost all similar with the exception of `InitStatsAndGear`, which I had to re-decompile as the tail of the function happened to be completely different.

The changes in HD are super-weird, especially the lazy way the developers chosen to assign the equipment. There are two goto labels I was not able to remove no matter what. If you want to remove them, please go ahead. I have a scratch uploaded, although it seems to behave completely different despite matching locally: https://decomp.me/scratch/wLROQ


src/dra/5D6C4.c
```diff
@@ -1126,6 +1149,37 @@ void InitStatsAndGear(bool DeathTakeItems) {
...
+                    g_Status.subWeapon = (rand() % 9) + 1;
+                }
+                do {
+                loop_check_equip_id_1:
+                    equipId = rand() % 169;
+                    if (equipId == 216) {
+                        goto loop_check_equip_id_1;
+                    }
+                } while (D_800A4B04[equipId].itemCategory > 4);
+                g_Status.equipment[0] = equipId;
```
sozud, line 1168: It apparently gives a random item for each hand? Makes me curious what the demos are like in that version.

sozud (approved):
Nice work

PR #421, opened 2023-07-31
Decompile DRA func_800EA5E4

bismurphy:
I will be completely honest: I have no idea what this function does or how it works. Something to do with CLUTs.

The function returns 1, 0, or -1, but all the current decompiled functions that call it do not use the return value, so the function was previously marked with a return type of void.

I was not able to get a match with the argument being a s32 as it was before.

This function relies on an array of structs which starts at D_8006C3C4. Previously, a neighboring function, func_800EA538, accessed what it believed to be an array of structs at D_8006C3CC. However, it only accessed a single member of these structs. It was assumed that that member was the first (unk0), but I have now determined that instead, that function accesses unk8 of this same struct.

I have therefore renamed the unkstruct to reflect being at C3C4 instead of C3CC.

There is one very powerful section of code here:
```
    if (temp_v0 & 0x8000) {
        clut = g_api.o.cluts[arg0];
    } else {
        clut = D_800A3BB8[arg0];
    }
```
This indicates that both D_800A3BB8 and g_api.o.cluts hold the same type. Therefore, we update g_api.o.cluts to hold the new UnkStructClut pointer, instead of a generic s32 pointer.

UnkStructClut is a new struct I've made. I do not know anything about it and the only indication of its purpose is that it is set by g_api.o.cluts. The members in it are just copied based on accesses and there could be more members which come after the ones I have found.

Unkstruct_8006C3C4 has been modified to reflect the fact that it appears to have an array that is 0x30 members in size, initialized in this function. I do not know why it had so many members previously, especially since the only other function that uses this struct is the aforementioned one that only touches unk8.

Overall, this function is still in the early days of being understood. But at least it matches.


src/dra/47BB8.c
```diff
@@ -709,73 +707,61 @@ void func_800EA5AC(u16 arg0, u8 arg1, u8 arg2, u8 arg3) {
...
+    }
+    if (clut->unk0 == -1) {
+        return 1;
     }
-    var_v1 = 0;
-    var_t0 = D_8006C3C4;
-    for (; var_v1 < 0x20; var_t0++, var_v1++) {
-        if (var_t0->unk8 != 0)
+    unkStruct = &D_8006C3C4;
+    for (j = 0; j < 32; unkStruct++) {
```
sozud, line 735: Possible change would be `LEN(D_8006C3C4)` for here and other loops

sozud (approved):
Looks good to me, thanks!

Edit: Missed HD build issue
```
mipsel-linux-gnu-ld: build/hd/src/dra/47BB8.c.o: in function `LM529':
src/dra/47BB8.c:(.text+0x2a80): undefined reference to `D_800A3BB8'
mipsel-linux-gnu-ld: src/dra/47BB8.c:(.text+0x2a88): undefined reference to `D_800A3BB8'
make: *** [Makefile:138: build/hd/dra.elf] Error 1
```

PR #430, opened 2023-08-02
RBO3 & MAR extraction and decomp

sonicdcer:
So i've been decompiling from these two /BOSS overlays, the primary reason was that Mott and I were thinking about having at least one boss decompiled to try doing a custom boss for his Boss Rush mod, so we needed to see the code and overlay structure of one of them, for shiftability and memory concerns. We choose the simplest and smallest of them, which is the Medusa (RBO3). As for MAR, i decided to work on it by pure curiosity, i was curious as to why did the developers decided to put this cutscene overlay in a /BOSS folder, maybe there was some leftover code from what became the Maria Boss?  It was a long shot but i decided to check anyway, spoiler alert, and sorry to disappoint but there's no such thing.

So even though we're not ready to expand the scope of the project to other overlays, why have i decided to submit these? I had a chat with @sozud where he recommended me to submit them, and i personally think it's worth it for the following reasons: These overlays are very small and most of the functions are duplicates from the overlays we currently have extracted, so most of it is already decompiled, i checked every function and left a comment on each and everyone of them, to have a complete panorama of the overlay contents. I decided to focus on the unique functions these overlays have to offer, and i decompiled them all, except one or two that i couldn't get a match. So i don't think it's really increasing the scope of the project, since almost all the functions are duplicates and the ones that are not are almost all decompiled.


src/boss/mar/CF70.c
```diff
@@ -0,0 +1,862 @@
...
+        }
+    } else if (status->timerMinutes & 1) {
+        g_Statues[LEFT_STATUE] = false;
+    } else {
+        g_Statues[LEFT_STATUE] = true;
+    }
+
+    switch (self->step) {
+    case 0:
+        if (g_blinkTimer == ((g_blinkTimer / 60) * 0x3C)) {
```
sozud, line 348: 0x3C is 60 which probably makes more sense

src/boss/mar/CF70.c
```diff
@@ -0,0 +1,862 @@
...
+        if (--self->ext.birdcage.timer == 0) {
+            self->ext.birdcage.timer = 64;
+            self->zPriority = 0;
+            self->step++;
+            g_api.PlaySfx(0x608);
+        }
+        break;
+
+    case 3:
+        self->posX.val += 0x2000;
```
sozud, line 539: Try `FIX(0.125)`? I should adjust the script to fix these too

sozud (approved):
In general, my opinion is that we should focus on finishing the existing overlays. However, there's a lot of great work here and it's better for this code to evolve along with the repo rather than sit outside of it and get out of sync, so I think it should be merged. The ultimate goal is decompiling the whole game, so this still gets us further, even though the immediate goal is finishing some of the existing overlays. Nice job!

sonicdcer:
> I disagreed many times in the past to add new overlays. I ultimately bent on CEN and DRE but then said they would have been the last ones until at least one overlay would have reach at least 90% of completion. We already have 10 stage overlays and still plenty of functions that are either still in progress or left to decompile. Despite everyone in the project knowing that, you decided to propose not one, but two new overlays. I understand you have your own personal goals and personal preferences. But this is not compatible to what we agreed. I am not sure if I should turn a blind eye on this specific PR. But I shall me this clear once and for all:
> 
> ### **NO NEW STAGE OR BOSS OVERLAYS WILL BE MERGED UNTIL FURTHER NOTICE.**
> 
> Any PR related to new overlays can sit until the pre-requisites we already agreed are satisfied.
> 
> Not to add fuel on the fire, but just two weeks ago you agreed the same thing. So why changing your mind?
> 
> image
> 
> ---
> 
> That said, I would have much preferred to have the two overlays to be a separate PR. The integration with the progress script is also missing. You are also forgetting the stage graphics in the Makefile `BOSS/{ovl}/F_{ovl}.BIN`. If a new overlay should be added in the project it should at least respect the core fundamentals than having something half-baked.
> 
> Why adding comments like `// 0.990 - ('ST0', 'func_801A8CB0') - (decompiled)` since we already have a page that holds the same exact information and that it is always up to date? They should not be hard-coded in the code.
> 

Again, I wasn't going to submit this because I knew you would reject it, even though the reason we weren't including new overlays was to avoid increasing the scope of the project, this submission doesn't do that and yet you're refusing it anyway, I'm not surprised at all tbh. I just did it because it was recommended to me. Good to know you appreciate the work.

PR #434, opened 2023-08-04
More libsnd, header refactor

sozud:


sozud:
Thanks for the suggestions, applied

PR #435, opened 2023-08-04
Items refactor

sozud:
This does the refactor suggested here https://github.com/Xeeynamo/sotn-decomp/pull/411#issuecomment-1657267022
I'm pretty sure this is correct based on GetEquipOrder and GetEquipCount https://github.com/Xeeynamo/sotn-decomp/blob/6050bb51542257efaa660f8d46380a6b9dd6df36/src/dra/5D6C4.c#L44
This is just the refactoring part and not the labeling.


sonicdcer (approved):

sozud:
This is Xeeynamo's suggestion so I don't think there's going to be any objections. Merging so @sonicdcer can work on the labeling part.

PR #436, opened 2023-08-04
Decompile _SsReadDeltaValue

sozud:
I'm not sure about `struct SeqStruct* temp_a1 = &_ss_score[arg0][arg1];` since it doesn't seem to make sense with the struct definition. I got the idea from https://github.com/AliveTeam/sound_rev/blob/7a9223139c3375bf10e96a4ac17d77b973979e20/psx_seq_player/lib_snd.cpp#L201

Co-authored-by: Mc-muffin <8714476+Mc-muffin@users.noreply.github.com>


sozud:
Going to merge this to avoid conflicts, will do a follow-up if there's any issues.

PR #439, opened 2023-08-05
Rewrite lints, lint more FIX 

sozud:
I noticed that my lints are already pretty slow and hard to maintain so I moved them to rust and optimized them. All the files are now linted in parallel, and we can do as many transforms to a single line as we want. I think this provides a good baseline where we can have a big suite of lints if we want to.

I added more cases to the FIX linter as well. Rust always puts a newline at the end of a file so we picked up some changes due to that. I think git wants you to do that anyway so it seems like an improvement to me.


sozud:
I ended up with some typos during cleanup, will fix.

sozud:
Fixed now

PR #440, opened 2023-08-05
Items Enum rework

sonicdcer:
Follow up to PR #411


src/dra/5298C.c
```diff
@@ -40,8 +40,9 @@ bool func_800F483C(void) {
 }
 
 bool IsAlucart(void) {
-    if (CheckEquipmentItemCount(0xA8, 0) && CheckEquipmentItemCount(0xA7, 0) &&
-        CheckEquipmentItemCount(0x59, 2))
+    if (CheckEquipmentItemCount(ITEM_ALUCART_SWORD, HAND_TYPE) &&
+        CheckEquipmentItemCount(0xA7, HAND_TYPE) &&
```
sozud, line 44: ITEM_ALUCART_SHIELD

src/dra/5D6C4.c
```diff
@@ -1024,24 +1026,24 @@ void InitStatsAndGear(bool isDeathTakingItems) {
...
                     if (i == 8) {
-                        AddToInventory(0x19, 2);
+                        AddToInventory(ITEM_AXE_LORD_ARMOR, ARMOR_TYPE);
                     }
                 }
             } else {
                 // This Else block is for this function if called outside
                 // the "Entrance (first visit)" overlay. Applies to demo mode.
-                for (i = 0; i < 0x20; i++) {
+                for (i = 0; i < 32; i++) {
```
sozud, line 1046: LEN(g_Settings.timeAttackRecords)

src/st/np3/3DEF4.c
```diff
@@ -442,13 +442,13 @@ void EntityEquipItemDrop(Entity* self) {
 
         g_api.PlaySfx(NA_SE_PL_IT_PICKUP);
 
-        if (itemId < 169) {
+        if (itemId < ITEM_HAND_END) {
```
sozud, line 445: The two enum strategy seems to work well except for this and `ITEM_HAND_END + ITEM_HOLY_GLASSES;`. I'm not sure what the cleanest solution is. It seems like there's many more places where two enums work well, than places like this where it doesn't work well. So maybe that's the best choice.

sonicdcer: This is what @Xeeynamo suggested previously, not sure if we want to have a HAND(x) (x - ITEM_HAND_END) macro for the Holy Glasses but it seems that's the only occurrence of an item being passed as a parameter in the project for now, we can probably leave it like this until we find more evidence. For now unifying the arrays and using overlapping enums seems to make the more sense, IMO he was totally right.

sonicdcer: The LIB overlay might provide more hints to this in the future.

sonicdcer:
> ```c
>  statBonus = D_800A7718[g_Status.equ
> ```

I don't know, what you want for that for loop looks weird to me, for instance the NUM_SLOTS thing is too much of an assumption to use there, we don't really know if they did the NUM_SLOTS thing, the + 2 in the array access makes more sense because it's skipping the first two hand slots, CheckEquipItemCount and AddToInventory functions provide absolute and faithful use of the enum values that is enough proof for me that they are right.

Regarding NUM_WHATEVER at the end of every enum list, feels weird to me the fact that the name starts different than the rest of the members:

```
typedef enum Consoles {
ITEM_MEGA_DRIVE,
ITEM_SNES,
ITEM_SATURN,
ITEM_PSX,
NUM_CONSOLES, <---- Weird
} Consoles;
```
Because when you see NUM_CONSOLES used alone in the code it feels a little ambiguous to me.

My recommendation would be something like:

```
typedef enum Consoles {
ITEM_MEGA_DRIVE,
ITEM_SNES,
ITEM_SATURN,
ITEM_PSX,
ITEM_MAX,
} Consoles;
```
I understand that it's a matter of preference but i wanted to state my opinion on it.

PR #444, opened 2023-08-06
Fully cross-reference DRA.BIN between US and HD

Xeeynamo:
This completes the work I started a few months back when I discovered the unique PlayStation 1 build I found between the files of Castlevania Dracula X Chronicles. It has been a longer journey than expected but we have now all the tools and knowledge in place to do it again with either the stage overlays of HD and new builds like JP11, EU and so on (which will not begin any time soon).

My conclusion is this build is an earlier Japanese debug build earlier than the US release. I do not know where to place it in the timeline when compared to JP10 and JP11. The only way to find out is to cross-reference these two Japanese builds too.

The HD has some chunks of code missing, which I assume the US build has as bug fixes. It will be interesting for speed-runners to understand these differences better and exploit them.


sozud (approved):
This is a cool milestone to hit, nicely done!

PR #445, opened 2023-08-06
DRA func_800F9E18

sonicdcer:
Co-authored-by: @nim-ka


sozud:
I'm pretty sure this is operating on an array of structs that defines the relic menu data. I'm not sure how to match it though. Here's my attempt https://decomp.me/scratch/6eTHE
Xeeynamo can probably help here.

sonicdcer:
> The function is fake. After exploring @sozud suggestion I made the following scratch matching: https://decomp.me/scratch/js3m0

Good work, I'm doing the changes.

sozud:
The requested changes are applied and it sounds like we will do the HD version later, so merging.

PR #446, opened 2023-08-06
NO3 EntityUnkId50 & EntityUnkId52

sonicdcer:
I'm including both functions in the same PR because they are basically copy pasted from one another and next to each other.

I'm providing the scratches in case somebody wants to try removing the labels.
https://decomp.me/scratch/T8Lwy
https://decomp.me/scratch/lpft1


src/st/no3/3C4EC.c
```diff
@@ -131,12 +131,121 @@ void EntityUnkId53(Entity* entity) {
...
+    s32 var_v0;
+    u16* ptr;
+
+    switch (self->step) {
+    case 0:
+        InitializeEntity(D_80180AD0);
+        ptr = D_801814EC;
+        temp = &g_Entities[192];
+        self->unk68 = 0xC0;
+        self->ext.generic.unk7C.s = 0;
```
sozud, line 148: Should we make an extension?

src/st/no3/3C4EC.c
```diff
@@ -131,12 +131,121 @@ void EntityUnkId53(Entity* entity) {
...
+        self->flags |= FLAG_UNK_08000000;
+    label:
+        while (*ptr <= 288) {
+            newEntity = AllocEntity(temp, &temp[64]);
+            if (newEntity != NULL) {
+                CreateEntityFromCurrentEntity(0x11, newEntity);
```
sozud, line 154: entity ID enum?

src/st/no3/3C4EC.c
```diff
@@ -131,12 +131,121 @@ void EntityUnkId53(Entity* entity) {
...
+    u16 temp_s3;
+    s32 var_v0;
+    u16* ptr;
+
+    switch (self->step) {
+    case 0:
+        InitializeEntity(D_80180AD0);
+        ptr = D_801817F8;
+        temp = &g_Entities[192];
+        self->ext.generic.unk7C.s = 0;
```
sozud, line 206: extension

sozud (approved):

src/st/no3/3C4EC.c
sonicdcer: For just one member? I think it's better to do that at documentation time because that can be a member used by other function related to this one and we don't have those decompiled yet.

PR #447, opened 2023-08-06
DRA func_800F9F40 

sonicdcer:
Co-authored-by: @Mc-muffin

US version:
https://decomp.me/scratch/GdX90

HD version:
https://decomp.me/scratch/SoE2S

I tried including the HD version but i couldn't, i keep getting weird errors about unreferenced symbols and rodata even after including the references in symbols.hd.dra.txt (from another function that isn't even decompiled...) so i don't know what's going on, also there's a function that takes 2 arguments instead of 3. @Xeeynamo Can you do it yourself later? Don't worry the CI is not broken, i put an #if defined(VERSION_HD) with the assembly path so it will still build ok.


src/dra/5298C.c
```diff
@@ -1411,7 +1411,35 @@ void func_800F9DD0(u8* arg0, u8* arg1) {
 
 INCLUDE_ASM("dra/nonmatchings/5298C", func_800F9E18);
 
+#if defined(VERSION_HD)
```
sozud, line 1414: Since we aren't scraping decomp.me for the HD version, could you add your link if it doesn't get solved?

src/dra/5298C.c
```diff
@@ -1411,7 +1411,35 @@ void func_800F9DD0(u8* arg0, u8* arg1) {
...
+    s32 i;
+
+    for (i = 0, ptr = D_800DC70C, idx = 0x80; i < 8; i++) {
+        spellId = D_80097902[idx];
+        if (spellId & 0x80) {
+            spellId ^= 0x80;
+            var_a3 = (Unkstruct_800F9F40*)buffer;
+            var_a2 = (Unkstruct_800F9F40*)&D_800DC6EC;
+            do {
+                __builtin_memcpy(var_a3++, var_a2++, 0x10);
```
sozud, line 1433: Could be                 
```
__builtin_memcpy(var_a3++, var_a2++, sizeof(Unkstruct_800F9F40));
```
Not sure what's going on with the 0xB

sozud (approved):

sonicdcer:
I don't think it's fair to force me to work on the HD version you know? So if I submit a function I am obligated to decompile the HD version too? I just did the work for the US version, and I'm doing you a favour by providing the scratch for the HD version because I couldn't make it happen, so I'm not trying to push any work to you, don't do it if you don't want to, I'm just doing the work for the US version ATM.

sonicdcer:
> > I don't think it's fair to force me to work on the HD version you know? So if I submit a function I am obligated to decompile the HD version too? I just did the work for the US version, and I'm doing you a favour by providing the scratch for the HD version because I couldn't make it happen, so I'm not trying to push any work to you, don't do it if you don't want to, I'm just doing the work for the US version ATM.
> 
> I do not think it is fair to force me to do work on it too, don't you think? I do not have as much context as you do on the function you submit. I spent months to bring parity between the two versions and I do not want corners to be cut now that the work is finally done.

Again, i'm not forcing you to do anything or pushing any work to you, i can barely understand this function so we are on the same page in that regard, i took a look at your tool but i don't even know how to use it.

So let me be clear with my question here:

Is it MANDATORY to decompile the HD version of every function i decompile in the US version now? There are functions whose HD version is completely different and they are huge, is it mandatory to decompile them too?. I don't think that's fair at all.

I can tell you in advance, if that's gonna be mandatory from now on (i didn't see any notice btw), i won't work on DRA anymore.

src/dra/5298C.c
sonicdcer: Probably the struct needs to be the right size, but again, it's weird because they are using the same struct and the number is different in both instances, i don't understand the function enough to tell if they hardcoded those values or we have two different structs there, which sounds to me is very unlikely.

sozud:
@sonicdcer Are you willing to use the tool and match the symbols if it gives the right answer, or is it a 100% no on any HD work?

sonicdcer:
> @sonicdcer Are you willing to use the tool and match the symbols if it gives the right answer, or is it a 100% no on any HD work?
If it works 100% of the time and people tell me how to use it yeah, i have no problem doing that, i just want to know if decompiling the HD version of the function is mandatory or not, because even when the majority of times the function is practically the same, there are occasions where it's completely different, look at func_800F9E18 for example, the HD version is a completely different monster i don't even know if even the symbols reference tool would work with it, unless it's badly referenced or something i don't see myself decompiling that HD function to being able to submit it's US function. You know what i mean? Like i said i tried referencing the symbols myself as i said in the description of this PR, it is something i'm familiar with since it's the same as working with duplicate functions, i tried and i couldn't make that happen because some other function was complaining about it and i have no idea why that's happening.

sozud:
I tried writing a tutorial on how to do this but I'm not able to get it to match. Maybe Xeeynamo can point my error out? https://github.com/Xeeynamo/sotn-decomp/wiki/HD-Verison-Matching

sozud:
Merging since both versions are decompiled now.

PR #448, opened 2023-08-06
Some dups progress

sozud:
After attempting this de-duplication process I see that it's pretty slow, tedious and error-prone. I think it would make sense to have an intermediate phase where we split stuff into #includes like this PR does. Eventually this will be combined back into C files with proper file splits. I think a process that is incremental and would get us there eventually could be like this:

Split single function into a header, #include it.
Rename any D_XXXXXXX variables to UNK_Whatever#, add to symbols.*.txt
Once the above is done for all instances of the function, perform file splits (if we know there's a sequential block of functions, wait to split until the whole block is done)
Combine individual.h files into a c file, integrate building st/shared_#.c into Makefile

I'm probably just going to do a little bit on this each day to try and chip away at it, it's too tedious to do a big push.


sonicdcer:
i suggest using .inc instead of .h if possible

sozud:
> i suggest using .inc instead of .h if possible

I guess either name is fine? The plan is to rename to c files but it requires a few more steps.

PR #449, opened 2023-08-08
DRA+RIC CanTeleportToOtherCastle matching

Xeeynamo:
Not a very special function per se. I struggled at de-duplicating the four variables at the top of the function as using just one would give me some minor reg swaps. I thought adding an enum as a returning value but I think it is a bit over killing for how small the function is. It is only called by `EntityAlucard`.

Interestingly, the function in DRA is different between HD and US. But the HD version is identical to the duplicate found in RIC US. What I suspect it happened is one of the developers while scrolling the various functions, found the `- 14407` in an odd spot relative to the rest of the function and moved it to make it consistent. But as some code from DRA was copy&pasted to RIC, the same developer did not think to apply the same change there.

EDIT: Note the `FIX_TO_I`. It is a new macro I added for another function I failed to decompile.


sozud (approved):

sonicdcer (commented):
Great work!
I'm a little skeptic about the function name, to me CanTeleportToOtherCastle makes me think that the answer should be a boolean true or false, but yet the possibilities are 0, 2, 4. I don't have another name to propose yet but i think we should give it some more thought.

src/ric/1AC60.c
```diff
@@ -21,7 +21,43 @@ void func_80156C60(Entity* entity) {
...
 
-INCLUDE_ASM("asm/us/ric/nonmatchings/1AC60", func_80156CCC);
+s32 CanTeleportToOtherCastle(void) {
+    s32 xCheckTop;
+    s32 yCheckTop;
+    s32 xCheckRTop;
+    s32 yCheckRTop;
+
+    // Is player in the pose when pressing UP?
+    if (LOW(PLAYER.step) != 0x10000) {
```
sonicdcer, line 31:
```suggestion
if ((PLAYER.step != 0) || (PLAYER.step_s != 1)) {

File: src/dra/692E8.c [hunk 2]
```diff
@@ -224,8 +224,51 @@ void func_8010A3F0(void) {
...
-s32 func_8010A4A4(void);
-INCLUDE_ASM("dra/nonmatchings/692E8", func_8010A4A4);
+s32 CanTeleportToOtherCastle(void) {
+    s32 xCheckTop;
+    s32 yCheckTop;
+    s32 xCheckRTop;
+    s32 yCheckRTop;
+
+    // Is player in the pose when pressing UP?
+    if (LOW(PLAYER.step) != 0x10000) {
```
sonicdcer, line 234:
```suggestion
if ((PLAYER.step != 0) || (PLAYER.step_s != 1)) {

**Review by sonicdcer — CHANGES_REQUESTED** [review]

**Review by sonicdcer — CHANGES_REQUESTED** [review]
.

**Comment by sonicdcer** [comment]
I can't seem to do a Request Changes. Why is that?

File: include/player.h [hunk 3]
```diff
@@ -0,0 +1,10 @@
+#ifndef PLAYER_H
+#define PLAYER_H
+
+typedef enum {
+    TELEPORT_CHECK_NONE = 0,
+    TELEPORT_CHECK_TO_RTOP = 2,
+    TELEPORT_CHECK_TO_TOP = 4
+} TeleportCheck;
```
sonicdcer, line 8: Everything looks good.

One last nit, if you declare it like this the enum name gets picked up by vscode like this:
image
otherwise it displays it like <unnamed>
image
i found this out relatively recently.
```suggestion
typedef enum TeleportCheck {
    TELEPORT_CHECK_NONE = 0,
    TELEPORT_CHECK_TO_RTOP = 2,
    TELEPORT_CHECK_TO_TOP = 4
} TeleportCheck;

sozud:
New changes look good to me

PR #450, opened 2023-08-08
Split files at Random()

sozud:
The point of this PR is splitting overlays at the `Random()` function in preparation for making it a C file eventually. After this I will do a split at `Update()`. I have a script to partially automate this that I will include in a separate PR.


sonicdcer (approved):

sozud:
This is just a file split with no other changes, concerned about things getting conflicted so going to go ahead and merge.

PR #452, opened 2023-08-08
NP3 func_801B3704

sonicdcer:
First function in the project to use GTE functions i think, i had a hard time with the loop near the end and the types, i cleaned it up the best i could.

I'm leaving the scratch here in case somebody wants to mess with it and maybe get to remove the padding i had to use for the stack pointer:
https://decomp.me/scratch/s3BEy

### DIFF
--- a/include/psxsdk/libgte.h
+++ b/include/psxsdk/libgte.h
@@ -6,17 +6,17 @@ void InitGeom();
 typedef struct {
     short m[3][3]; // 3 x 3 matrix coefficient value
     long t[3];     // Parallel transfer volume
-} MATRIX;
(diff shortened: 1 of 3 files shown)


sozud (approved):
I don't see any GTE instructions in the asm? Anyway looks good
Edit: I misread your and thought your description said "GTE instructions" not "GTE functions"

sonicdcer:
> I don't see any GTE instructions in the asm? Anyway looks good Edit: I misread your and thought your description said "GTE instructions" not "GTE functions"

Yeah, i meant GTE functions from libGTE

PR #453, opened 2023-08-08
Rename UpdateStageEntities

sozud:
I wasn't able to get this to work for MAD for some reason. Splat wouldn't extract the asm under the new filename.


sozud:
Merging since this is simple, will do a follow-up for MAD if someone knows what's going on

PR #456, opened 2023-08-08
Split on TestCollisions

sozud:
This splits on TestCollisions, giving a block of the following 3 functions in a separate file and that are fully decompiled with the exception of MAD.

```
#include "../random.h"
#include "../update.h"
#include "../update_stage_entities.h"
```

It should now be possible to change the build system to make that a C file that is shared among the overlays.


sozud:
Merging to avoid conflicts

PR #457, opened 2023-08-09
Add TeleportCheck to enum

sozud:
@sonicdcer brought up that I merged https://github.com/Xeeynamo/sotn-decomp/pull/449 before this was addressed


sonicdcer:
Not a mandatory request, but I was waiting for an answer/opinion.

PR #460, opened 2023-08-09
NP3 NO3 EntityShuttingWindow

sonicdcer:


src/st/no3/377D4.c
```diff
@@ -85,7 +85,99 @@ void EntityUnkId16(Entity* self) {
 INCLUDE_ASM("asm/us/st/no3/nonmatchings/377D4", EntityBackgroundLightning);
 
 // window that opens and shuts in the background
-INCLUDE_ASM("asm/us/st/no3/nonmatchings/377D4", EntityShuttingWindow);
+void EntityShuttingWindow(Entity* self) {
```
sozud, line 88: We should see what Xeeynamo thinks before you do it but I think we should deduplicate like in this PR https://github.com/Xeeynamo/sotn-decomp/pull/459

include/psxsdk/libgte.h
```diff
@@ -18,11 +18,27 @@ typedef struct {
     short pad;        // System reserved
 } SVECTOR;            // size = 0x8
 
+typedef struct {
+    /* 0x00 */ SVECTOR* v0;
+    /* 0x04 */ SVECTOR* v1;
+    /* 0x08 */ SVECTOR* v2;
+    /* 0x0C */ SVECTOR* v3;
+} SVEC4; // size = 0x10
```
sozud, line 26: Is this a libgte type or something you made? If you made it then it should probably be in game-specific header.

src/st/no3/377D4.c
sozud: Yeah I don't think we are going to see this one outside of those two overlays, so just between those two. I just mean that if we incrementally tackle duplicates rather than letting them build up it should be easier in the future. Right now the de-duplication is a huge task since we've let it build up.

sonicdcer: > You mean de-duplicating it between NO3 and NP3? I think it sounds reasonable. Contributors should start to feel accustomed to know how to do it and greatly increase the decomp coverage in one go. The learning curve is steep at the beginning but then it's easy to tackle down. How do you feel about it @sonicdcer ?

Well, you know how I dislike duplicates and I wanted to tackle this issue in the past, however, there are a few things that concern me with this.

First and less important is that putting code in headers feels wrong to me. Can't we at least use .c.inc?

Now the most concerning part is that we are going to need to import data sooner or later, and I'm afraid that doing something like this is going to make that task difficult when that time comes, there's also the problem with the different symbols, even if I use #ifdef STAGE where numbers are different I will also need to force addresses to one version only, which can become confusing later and we can have symbol collision, this is a problem we still need to think how we're going to solve before starting deduplicating functions between overlays.

Are we sure we want to do this right now? I would say no.

sozud: The symbols aren't different though? You just set the data to the right address. For example:

```
PfnEntityUpdates = 0x80180A94;
vs.
PfnEntityUpdates = 0x801803E4;
```

Can you give a specific example of the issues you are expecting to run into?

My main concern is if multiple files change how it compiles or not. I wrote a test to see what the difference would be with a single file vs. multiple files. Using a custom linker file, the results are mostly the same except for padding that I'm not sure how to get rid of in the multi file version. I guess we have to figure out if split files can be equivalent to a single file first. https://github.com/sozud/sotn-decomp/tree/data_test

sonicdcer: To give you an example:
InitializeEntity(D_80180AC4);
InitializeEntity(D_80180A54);

What would you do there? 
D_80180A54 = 0x80180AC4; ?
or
D_80180AC4 = 0x80180A54; ?

What criteria should we use to make that decision?, then if you're in the phase of importing data, you first have to make sure the Symbol is pointing to the same address or not same goes with the functions, it's confusing and makes the process even slower, that's one of the reasons i think this should be postponed until we decompile at least one overlay at 100% and import the data, these functions were in different files and the only clue we have to isolate them correctly is the data itself, so my choice would be decompile the overlay entirely first, import the data later, and after that, we can de-duplicate entity functions like this.

sozud: I would do something like:
```
InitializeEntity(ShuttingWindowData);
ShuttingWindowData = 0x80180AC4;

InitializeEntity(ShuttingWindowData);
ShuttingWindowData = 0x80180A54;
```
It's the same symbol at a different address. This is what I've been doing for the other functions I've de-duplicated.

sonicdcer: What about functions then? Because the problem is you might have a name for initialized data but if you're gonna go with the same approach on everything, then you need to come up with every name.

sozud: > First and less important is that putting code in headers feels wrong to me. Can't we at least use .c.inc?

The plan is to remove these files. I'm close to being able to remove 3 of them but it needs some build system assistance. https://github.com/Xeeynamo/sotn-decomp/blob/master/src/st/wrp/81E8.c

c.inc or whatever is fine with me.

> What about functions then? Because the problem is you might have a name for initialized data but if you're gonna go with the same approach on everything, then you need to come up with every name.

I've just been picking names for functions, I don't think there's a better solution.

sonicdcer: > > First and less important is that putting code in headers feels wrong to me. Can't we at least use .c.inc?
> 
> The plan is to remove these files. I'm close to being able to remove 3 of them but it needs some build system assistance. https://github.com/Xeeynamo/sotn-decomp/blob/master/src/st/wrp/81E8.c
> 
> c.inc or whatever is fine with me.
> 
> > What about functions then? Because the problem is you might have a name for initialized data but if you're gonna go with the same approach on everything, then you need to come up with every name.
> 
> I've just been picking names for functions, I don't think there's a better solution.

Well that's another rabbit hole, because that implies looking at all the functions, try to understand what they do, give it a name that could be wrong or some other people will complain or not agree with, it just increases the change of the PR being stalled.

Can we merge this for the time being?

sozud:
@sonicdcer says he is willing to do a follow-up after https://github.com/Xeeynamo/sotn-decomp/pull/464 is fixed, so going to merge.

sonicdcer:
Yeah i have no problem coming back to it to do the de-duplication when the method is sorted out, i just wanted this merged because i need to work on some functions that use those GTE protos and types. Thanks for the merge.

PR #461, opened 2023-08-09
Decompile DRA func_800FDE20

bismurphy:
This is a cool one! This runs every frame, and checks if you have enough EXP to level up. If you do, then it increases your level and gives you the stat bonuses for it. Therefore, I think `CheckAndDoLevelUp` might be a good name? I've held off on renaming; if reviewers are okay with the name then I'll update the PR with this.

I don't know what the 3 values being checked at the beginning are. I might dig into those next.

D_800A2EC0 is a table which tells how much to increase max HP for each level. Its values are 1, 3, 6, 10, 20, 30, 40, 50, 100, 200. So the first 10 levelups you get, you'll gain one max HP, then the next 10 give you 3 each, etc. I think this should be called LevelUpHPIncreaseTable, but again, holding off on renaming. I'm not sure if there's a way to include this as an array in the code rather than as an extern.

I've also included updating the repo to the latest maspsx as part of this PR, let me know if that should be removed.


src/dra/5D6C4.c
```diff
@@ -231,7 +231,61 @@ void func_800FDE00(void) {
...
+    if (c_arrExpNext[g_Status.level] <= g_Status.exp) {
+        g_Status.level++;
+        statsGained = 0;
+        g_Status.mpMax += 4 + (rand() & 1);
+        g_Status.hp += D_800A2EC0[(s32)g_Status.level / 10];
+        g_Status.hpMax += D_800A2EC0[(s32)g_Status.level / 10];
+        g_Status.heartsMax += 2;
+        // Run again, in case we have enough EXP to level up twice
+        func_800FDE20();
+        for (i = 0; i < 4; i++) {
```
sozud, line 263: LEN(statsBase)

src/dra/5D6C4.c
```diff
@@ -231,7 +231,61 @@ void func_800FDE00(void) {
...
+            // Flip a coin to decide if you will gain a stat here
+            statgain = rand() & 1;
+            g_Status.statsBase[i] += statgain;
+
+            if (g_Status.statsBase[i] > 99) {
```
sozud, line 268: Should we have a STATS_MAX define?

sozud (approved):

src/dra/5D6C4.c
bismurphy: Seems unnecessary to me but I can add it if you want

PR #463, opened 2023-08-10
Decompile DRA func_800FE97C

bismurphy:
Several comments on this one, it's weird.

This appears to handle some of the logic involved when the player takes damage. It is called by 3 functions. One is EntityAlucard, and of the other two one is decompiled and one is not. I have decompiled the other one, and will submit that PR next.

The first argument is a pointer to a structure which holds some of the results of this function. This is similar to CheckCollision which stores the results in a Collider struct. Most of the members of this struct are still unknown.

The logic with the ballroom mask is extremely strange and is worth further research. There appears to be a bug since it tests whether the player is wearing two ballroom masks, which should be impossible.

Overall lots of unknowns being used here, but hopefully this function and others like it will be helpful in identifying those. I hope I successfully replaced everything that has an enum.

The line which has `if ((g_Status.defenseEquip > 99)` was causing an issue with SLTI vs SLTIU, so the type of defenseEquip has been changed to match this.

### DIFF
--- a/include/unkstruct.h
+++ b/include/unkstruct.h
@@ -309,9 +309,9 @@ typedef struct {
     /* 0x0E */ s16 unk0E;
 } RelicDesc;
 
-typedef struct{
+typedef struct {
     s32 unk0;
     s32 unk4;
-    s32 unk8;
+    s32 damageTaken;
     s32 unkC;
 } Unkstruct_800FE97C;
--- a/include/unkstruct.h
+++ b/include/unkstruct.h
@@ -308,3 +308,10 @@ typedef struct {
     /* 0x0C */ s16 unk0C;
     /* 0x0E */ s16 unk0E;
 } RelicDesc;
+
+typedef struct{
+    s32 unk0;
+    s32 unk4;
+    s32 unk8;
+    s32 unkC;
+} Unkstruct_800FE97C;
--- a/src/dra/5D6C4.c
+++ b/src/dra/5D6C4.c
@@ -474,7 +474,7 @@ void AddHearts(s32 value) {
     }
 }
 
-//Note: Arg3 is unused, but is given in the call from func_80113D7C
+// Note: Arg3 is unused, but is given in the call from func_80113D7C
 s32 func_800FE97C(Unkstruct_800FE97C* arg0, s32 arg1, s32 arg2, s32 arg3) {
     s32 ret;
     s32 var_s1;
@@ -488,7 +488,7 @@ s32 func_800FE97C(Unkstruct_800FE97C* arg0, s32 arg1, s32 arg2, s32 arg3) {
         var_s1 *= 2;
     }
     if (g_Status.D_80097C2A & arg0->unk0) {
-        var_s1 /=  2;
+        var_s1 /= 2;
     }
     if (g_Status.D_80097C2C & arg0->unk0) {
         if (!(g_Status.D_80097C2C & arg0->unk0 & 0x200)) {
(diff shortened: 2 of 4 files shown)


src/dra/5D6C4.c
```diff
@@ -474,7 +474,160 @@ void AddHearts(s32 value) {
...
+            if (g_Status.hpMax < g_Status.hp) {
+                g_Status.hp = g_Status.hpMax;
+            }
+        }
+        return 5;
+    }
+
+    // Ballroom mask???
+    itemCount = CheckEquipmentItemCount(ITEM_BALLROOM_MASK, HEAD_TYPE);
+    if ((itemCount != 0) && (arg0->unk0 & 0xF980)) {
```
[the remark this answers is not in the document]
bismurphy: If I try that, the compiler optimizes it wrongly. No idea why. The proper instruction is an `andi` with 0xF980, but if I do `~0x67F` then I get a `li v1,-0x680` followed by `and` with v1. One instruction becomes two.

src/dra/5D6C4.c
```diff
@@ -474,7 +474,160 @@ void AddHearts(s32 value) {
...
+        }
+        // Check for player wearing a Talisman
+        itemCount = CheckEquipmentItemCount(ITEM_TALISMAN, ACCESSORY_TYPE);
+        if (itemCount != 0) {
+            if (itemCount * g_Status.statsTotal[3] >= (rand() & 0x1FF)) {
+                return 2;
+            }
+        }
+        if (arg0->damageTaken > 0) {
+            if (arg0->unk4 < 2U) {
```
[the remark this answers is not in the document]
bismurphy: Good point, I will make `unk0` and `unk4` both be u32 since they are bit flags

src/dra/5D6C4.c
```diff
@@ -474,7 +474,160 @@ void AddHearts(s32 value) {
...
 }
 
-INCLUDE_ASM("dra/nonmatchings/5D6C4", func_800FE97C);
+// Note: Arg3 is unused, but is given in the call from func_80113D7C
+s32 func_800FE97C(Unkstruct_800FE97C* arg0, s32 arg1, s32 arg2, s32 arg3) {
+    s32 ret;
+    s32 var_s1;
+    s32 itemCount;
+
+    var_s1 = arg2;
```
sonicdcer, line 483: you can remove var_s1 by replacing it entirely by arg2
https://decomp.me/scratch/Qyv5F

sonicdcer (approved):

include/unkstruct.h
```diff
@@ -308,3 +308,10 @@ typedef struct {
     /* 0x0C */ s16 unk0C;
     /* 0x0E */ s16 unk0E;
 } RelicDesc;
+
+typedef struct {
+    u32 unk0;
+    u32 unk4;
+    s32 damageTaken;
+    s32 unkC;
+} Unkstruct_800FE97C;
```
sonicdcer, line 317:
```suggestion
typedef struct {
    /* 0x0 */    u32 unk0;
    /* 0x4 */    u32 unk4;
    /* 0x8 */    s32 damageTaken;
    /* 0xC */    s32 unkC;
} Unkstruct_800FE97C;

File: src/dra/5D6C4.c [hunk 5]
```diff
@@ -474,7 +474,158 @@ void AddHearts(s32 value) {
...
+            func_800FE8F0();
+            g_Status.hp += arg0->unkC;
+            if (g_Status.hpMax < g_Status.hp) {
+                g_Status.hp = g_Status.hpMax;
+            }
+        }
+        return 5;
+    }
+
+    // Ballroom mask???
```
sonicdcer, line 532: remove this comment since you already added the enum value?

bismurphy: I want to call it out as a very unusual thing. The ballroom mask is not supposed to have any special properties and the code should not have to check if you are wearing two of them.

sonicdcer: i think It would be great if those implications are stated in the comment instead of just the question marks. Do you think you can do a brief multiline comment explanation?
/*
 * Explain unusual behaviour
 */

bismurphy: Done, thank you for the feedback.

PR #464, opened 2023-08-10
Start sharing some common stage code

Xeeynamo:
Following up #454 , #453 and related from @sozud , I started to convert some previously defined headers into proper shared C code. The advantage from the header is the C file is completely independent, it will have its own `.data` and `.rodata` and it will be compiled once.

I also took the opportunity to move `include/st/AnimateEntity.h` to `src/st/animate_entity.h` for consistency. I think the change is small enough to be included as part of this PR.


config/splat.us.stmad.yaml
```diff
@@ -62,7 +62,9 @@ segments:
       - [0xD874, rodata]
       - [0xD884, rodata]
       - [0xD8C8, c]
-      - [0xE964, c]
+      - [0xE964, c, ../random]
+      - [0xE994, c, ../update]
+      - [0xEC90, c, CollectGold]
```
sozud, line 67: Hmm if the 3rd function is not UpdateStageEntities that weakens my theory about how the overlays are structured. MAD is sort of a special case though. I think either they just copy-pasted or did something like:
```
stXXX:
  stXXX.c shared1.c shared2.c shared3.c etc.
```

sozud (approved):
Thanks for figuring out these build system changes. I think this should make working with stage overlays a lot more sustainable.

src/st/update_stage_entities.c
```diff
@@ -1,3 +1,7 @@
+#include <stage.h>
+
+u16 UNK_Invincibility0[];
```
sozud, line 3: Actually we seem to have this declared multiple times? See update.c. Also I thought these needed to be extern but apparently not? That's a mistake I made on an earlier PR.

sozud:
I've filed a ticket for the splat issue https://github.com/ethteck/splat/issues/266

sozud:
Filed a ticket for the splat issue: https://github.com/ethteck/splat/issues/266
(Accidentally deleted my previous post)

sozud:
I attempted a fix for this with https://github.com/ethteck/splat/pull/271/
Adding `force_c_asm_files_in_nonmatchings: yes` turns it on.

PR #466, opened 2023-08-12
Enhance Makefile for WEAPON0.BIN

Xeeynamo:
## What is this

Required for #465 . I re-wrote how the build process works for the `BIN/WEAPON0.BIN` ~~overlay~~ file.

## A bit of a background

On the PSP build I found `WEAPON0.BIN` is just a blob with `f0_000.bin`, `w0_000.bin`, `f0_001.bin` and so on concatenated as a single file. `f0` contains the graphics (or CHR) and `w0` contains the actual overlay (or PRG). The individual overlays share a great amount of functions I documented in `config/symbols.us.weapon.txt`.

## The problem

Splat creates its own `weapon.ld` linker script, insisting to link all the overlays as a single binary file. While this is not technically correct, it creates a bigger problem: symbols cannot be shared. In 58 overlays there are 58 `DestroyEntity`. Other decomps are circumnavigating the issue by adding a prefix. Following the same pattern we could have `W000_DestroyEntity`, `W001_DestroyEntity` and so on. But this is not maintainable and it does not reflect our long-term vision of having shared code between overlays.

## The solution

I created the dedicated linker script `weapon0.ld` to output the individual overlay `w0_000.bin` and similar. `f0_000.bin` is also created from the correspondent PNG. Once all the overlays and graphics are created I am simply concatenating them back to `WEAPON0.BIN`. This is much more efficient in terms of separation of concerns, maintainability, build speed and it also allows to use the same symbol names across overlays. Also very important, the generated individual overlays matches the structure of what is found in the PSP build.

## Side note

I thought it was wise to prepare the Makefile and the linker script to also allow rebasing and compiling for WEAPON1.BIN. The generated file does not currently match and it will probably need a bit more research. I left the build infrastructure there. To compile the non-matching WEAPON1 just modify `$(BUILD_DIR)/WEAPON0.BIN: $(WEAPON0_FILES)` to `$(BUILD_DIR)/WEAPON0.BIN: $(WEAPON0_FILES)  $(WEAPON1_FILES)`.


sozud (approved):
Looks good to me!

PR #467, opened 2023-08-12
Add duplicates report to dups tool

sozud:
This adds an equivalent of find_duplicates.py to tools/dups. This runs in about 2 minutes on my machine so it should be significantly faster on the CI. The algorithm isn't exactly the same so the report is a little different. Here's an example:

https://gist.github.com/sozud/503fd3b3014668e6644fb2dfae51d5e5

This works by grouping all the functions in to clusters, basically:

```
if levenshtein_similarity > threshold
  cluster.append(current_function)
```

Memoization gives a little speedup to avoid computing the levenshtein distance for the same pairs over and over again. This is still a brute-force algorithm. I did some research and there's a lot of similar problems but didn't find something that seemed like it would be a good fit. I think this is probably fast enough to last for a while.

### DIFF
--- a/tools/dups/Cargo.toml
+++ b/tools/dups/Cargo.toml
@@ -7,3 +7,4 @@ edition = "2021"
 
 [dependencies]
 clap = { version = "4.3.19", features = ["derive"] }
+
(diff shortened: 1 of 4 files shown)


sozud:
I fixed an issue where instructions were being parsed with the wrong endianness, do we think this is good enough to replace the current script? https://gist.github.com/sozud/a2682e07b2ca4fd619d8140fc110480b

sozud:
I changed it to ignore empty functions, the output looks like this now: https://gist.github.com/sozud/4d9280c16f46f55eaa3a4d19f47c0dd2

PR #471, opened 2023-08-14
Rewrite analyze_calls

bismurphy:
I have rewritten much of the overall logic flow of analyze_calls. This should make it run faster. Hopefully the program logic is also easier to understand.

The output of this script is mostly identical to the previous version. It will now include calls through g_api to a given function, which is a nice upgrade (for example, a call to g_api.func_800FE97C will now properly show up on func_800FE97C's call graph).

Important change: I am now using the argparse library to parse arguments. Running without arguments will generate all the call trees, and save the files along with the HTML page that links to all of them. There are two optional arguments:

`--dry` will render all call trees, but will not save them to files. The bytes of the SVG are created, but are then simply forgotten.

`--ultradry` will perform the analysis of every function, what it calls, and what calls it, but will stop there without generating any trees.

I believe that `--ultradry` will be sufficient to automatically run in PRs, since the majority of problems are in the function analysis stage, and not in the call tree generation stage. Therefore, for the CI, I would recommend that we set PRs to run with `--ultradry`, and the main repo to run with no arguments. In the event that we find a situation where `--ultradry` misses an error that `--dry` catches, we can revert to using `--dry` on PRs.

For some benchmarking: Normal runs and --dry runs take approximately 30 seconds on my computer. It turns out that writing to files is a relatively small amount of the execution time. On the other hand, `--ultradry` takes less than 1 second. This is the main reason I suggest using `--ultradry` for PRs.

This will hopefully be applicable to the stalled `weapon0-weapon1-makefile` branch.

Note: The previous run modes of this script (running with command line interface, or running generating single flow charts one at a time) are now removed. If anyone was relying on them, I can reimplement them, but I assume the web interface is more convenient for everyone.


sozud:
It looks like this isn't running in CI since the file permissions changed `100755 → 100644`

sozud:
I think this needs to be merged to continue testing it since the CI script from master gets used. Is it ready otherwise?

bismurphy:
Yes, should be ready to merge, let's give it a try.

PR #475, opened 2023-08-14
Decompile DRA func_80101A80

bismurphy:
Decompiled this one a while ago but forgot to make it a PR.

I believe this does a lot of the graphical computations for the normal HUD that appears in the upper left of the screen. It does the establishing math for the HP number, heart number, MP bar, and subweapon display.

The new variable displayHP is the value that appears on the HP meter, which lags behind the internal HP value. Not sure it's the greatest name, but seems good enough.

I've held off on naming it so far because I want to make sure I match conventions. Do we want DrawAlucardHUD, perhaps? Notably, I don't think this draws the static "frame" that holds the status elements, it only does the determination for the status elements themselves. If we are happy with DrawAlucardHUD, please let me know and we can include that name before merging.

I created two new UnkStruct entries for single s32 and u32 variables, which are referenced by address and are therefore likely struct members, but for which it is not clear what the struct boundaries might be. I think this makes more sense than creating a dedicated UnkStruct when none of the other members are known.


bismurphy:
(working on the hd fix :) )

config/symbols.hd.dra.txt
```diff
@@ -890,7 +893,7 @@ D_80137954 = 0x80137528;
 D_80137960 = 0x80137534;
 D_80137964 = 0x80137538;
 D_80137968 = 0x8013753C;
-D_8013796C = 0x80137540;
+g_displayHP = 0x80137540;
```
sonicdcer, line 896:
```suggestion
g_DisplayHP = 0x80137540;

Comment (sonicdcer) [reply config/symbols.hd.dra.txt:896]: same with the other globals

File: include/unkstruct.h [hunk 2]
```diff
@@ -314,4 +314,10 @@ typedef struct {
...
     /* 0xC */ s32 unkC;
-} Unkstruct_800FE97C;
\ No newline at end of file
+} Unkstruct_800FE97C;
+typedef struct {
+    s32 unk0;
+} Unkstruct_oneWord;
+typedef struct {
+    u32 unk0;
+} Unkstruct_oneUword;
```
sonicdcer, line 323: You can probably use u32 and s32 arrays for these until we find the right structs

src/dra/5D6C4.c
```diff
@@ -1576,7 +1576,264 @@ void func_8010189C(void) {
...
 
-INCLUDE_ASM("dra/nonmatchings/5D6C4", func_80101A80);
+void func_80101A80() {
+    SubweaponDef subwpn;
+    RECT rect;
+    Primitive* prim;
+
+    s32 stat_xpos;
+    s32 hpdiff;
+    s32 mp_fill_steps;
```
sonicdcer, line 1586:
```suggestion
s32 mpFillSteps

camelCase, same with the rest of them

sonicdcer (changes requested):
Good work on this one!

bismurphy:
I will do DrawHudAlucard. I actually didn't realize how close the functions are as well, they are basically right after each other. So that sounds great. I'll change the name and then I think this should be ready to merge!

PR #488, opened 2023-08-21
Update ET_WaterEffects extension + function names

sonicdcer:


sozud (approved):

sozud:
Actually, why are we changing the function names? The old names seem to be more descriptive?

sonicdcer:
> Actually, why are we changing the function names? The old names seem to be more descriptive?

I'm just using the commented names that were there, do you prefer the names that were in NO3 ?

PR #489, opened 2023-08-21
NO3 & NP3 EntitySurfacingWater

sonicdcer:
https://decomp.me/scratch/ULVFw


src/st/no3/46684.c
```diff
@@ -1265,8 +1265,171 @@ INCLUDE_ASM("asm/us/st/no3/nonmatchings/46684", EntityAlucardWaterEffect);
...
+    s16 temp;
+    s16 temp3;
+    s16 tempv0;
+    s16 tempa0;
+    s16 x, y;
+    s32 i;
+
+    switch (self->step) {
+    case 0:
+        InitializeEntity(D_80180B00);
```
sozud, line 1294: Can we synchronize the data names for these? e.g
```
D_80180B00 -> EntitySurfacingWaterInitData
D_80180A90 -> EntitySurfacingWaterInitData
```

This applies to all the data in these PRs. It will make deduping them easier later.

sozud (approved):

src/st/no3/46684.c
sonicdcer: It will make data migration more tedious and difficult, i rather not deduplicate entity functions until we finish the overlay at 100% and migrate the data first.

sonicdcer: I know what he asked for, D_80180B00 is an Init Chain used not only by this function but 5 others too, i think it's impractical to lose the address in the name and having to look it up at data migration time. All of these Init Chain data entries seem to be at the beginning of the overlay for some reason, they probably are part of the same array or something. I think we are engaging into patterns we don't yet know for sure. I've seen you guys deduplicate entity functions and put them into headers, i don't think that's right at all, the solution needs to be found with data migration in mind, don't get me wrong but i don't think it's time to do this right now, in my opinion is time wasting and will probably make things difficult later.

sonicdcer:
@Xeeynamo I managed to get it matching with your suggestions and a while loop, great team work here. Let me know if everything is ok.

PR #490, opened 2023-08-21
NO3 & NP3 EntitySideWaterSplash

sonicdcer:
https://decomp.me/scratch/jpF5j


src/st/no3/46684.c
```diff
@@ -1268,8 +1268,93 @@ INCLUDE_ASM("asm/us/st/no3/nonmatchings/46684", EntitySplashWater);
...
+                prim->priority = self->zPriority + 2;
+                prim->blendMode = 0x37;
+                prim = prim->next;
+            }
+            params = self->params;
+            temp_s0 = params & 0xF;
+            if (temp_s0 == 0) {
+                g_api.PlaySfx(D_801813A8[0]);
+            }
+            angle = LOH(D_801838E4[(params >> 3) & 0x1E]);
```
sozud, line 1312: Have you tried D_801838E4 as a struct?

sozud (approved):

src/st/no3/46684.c
sonicdcer: Why would i? this is an AND operation, not a multiplication, & 0x1E does not suggest a struct of any kind. If it's something that i'm not seeing please make a scratch with the suggested theory.

tools/analyze_calls.py
```diff
@@ -329,25 +329,25 @@ def render_svg(self):
             style="filled",
             fillcolor=graph_colors[self.decompile_status],
         )
-        # We sort the items so they will be alphabetized. This reduces changes from
-        # one run to the next. Any differences should be due to changes in the repo.
-        for callee, flags in sorted(self.callees.items()):
```
sozud, line 334: I missed this, could you put up a version of this PR without this change?

PR #503, opened 2023-08-22
EntitySideWaterSplash

sonicdcer:


tools/analyze_calls.py
```diff
@@ -329,25 +329,25 @@ def render_svg(self):
             style="filled",
             fillcolor=graph_colors[self.decompile_status],
         )
-        # We sort the items so they will be alphabetized. This reduces changes from
```
sozud, line 332: Remove all changes to analyze_calls.py please.

PR #505, opened 2023-08-22
Import w_000 data

Xeeynamo:
The first overlay 100% decompiled. All the functions and data is imported in `w_000.c`, symbols are all names and assets are extracted into their appropiate files.

So far the `animset` always comes at the beginning of the data. This overlay also has CLUTs and the animation definitions. I am not yet sure if these resources can be extracted as assets. I am waiting to decompile more weapon overlays to understand if this is a pattern that can be used or the developers really baked this information as byte arrays. If this data can be extracted as assets we might want to come up with a simple viewer/editor in the future.

I could not yet get rid of `data` and `sbss` due to how the Makefile is structured. Therefore I flagged the empty sections in the YAML. Once they are all empty they can be removed all at once.


sozud (approved):
Nice work!

PR #506, opened 2023-08-23
Deduplicate shared code between weapon overlays

Xeeynamo:
As per title.

Interestingly, I could get w_029 matching only with `-O1`. I added an override in the Makefile. There were also a very few overlays that needed some `#if !defined(W_0xx)` as few functions were absent in some of them.

I could not rename `shared.h` into `shared.c` and compile it for technical reasons. Compiling it into an object `shared.o` means I could not use `#if !defined(W_0xx)` to exclude some functions. And compiling it as `shared_0xx.o` would kinda defeat the purpose of having it shared in the first place.

### DIFF
--- a/src/weapon/setweaponproperties.h
+++ b/src/weapon/setweaponproperties.h
@@ -1,1 +1,0 @@
-#include "weapon_private.h"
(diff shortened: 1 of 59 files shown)


sozud:
Looks good, the #ifdefs seem to give more data in favor of the "everything is copy pasted" theory though. I still think it makes sense to deduplicate as long as the final result is matching.

sozud (approved):

PR #507, opened 2023-08-24
Decompile Shield Rod

Xeeynamo:
Given the fact `func_ptr_80170020` is called as an effect of the Shield Rod, I decided to decompile it. Unfortunately I couldn't find any extra code to handle that logic. It is almost identical to `w_000`


sonicdcer (approved):
Great work!

sozud (approved):

PR #510, opened 2023-08-24
Decompile all w_040 and w_042

Xeeynamo:
As w_042 is similar to w_040 I decided to decompile them both.

### w_040 - Monster Vial 1

f_040

### w_042 - Monster Vial 3

f_042

### EntityWeaponAttack

`EntityWeaponAttack` is the vial dropping to the ground. When `CheckCollision` detects a collision with the solid pavement, the vial gets immediately destroyed. There is some wild magic happening, which I think it's happening when the code does `g_api.func_8011AAFC(`. That might trigger the function `func_ptr_80170004`, which will summon the vial monster.

### func_ptr_80170004

The state machine of the summon itself. We have `self->unk1A` and `self->unk1C`, which if I recall correctly they indicates ScaleX and ScaleY respectively. `self->unk19 |= 3` will tell to use both horizontal and vertical scaling. When the scaling reaches 100% (or `0x100`) after 64 frames, the summon will get inherit the properties of... `self->ext.weapon.parent->ext.weapon.equipId;`? That is the potential bug that the folks from the SOTN Speedrun server described to me. `g_api.func_8011AAFC` seems to invoke the projectile.

### func_ptr_80170008

This seems to be the projectile function. Again, this is also calling `SetWeaponProperties`, which will copy the projectile attack properties from a dubious source. On `w_040` the logic is pretty simple as it is just keep adding the X coord until something is hit. On `w_042` both X and Y are involved as the thrown bone will form an arch. The thrown bone seems to not check the stage collisions.

### func_ptr_8017000C

I think this is the light effect at the base of the summon when it is still spawning. The function looks so simple I cannot imagine otherwise.

### DIFF
--- a/include/entity.h
+++ b/include/entity.h
@@ -121,11 +121,12 @@ typedef struct {
 } ET_EquipItemDrop;
 
 typedef struct {
-    /* 0x7C */ s32 unk7C;
+    /* 0x7C */ s16 unk7C;
+    /* 0x7E */ s16 unk7E;
     /* 0x80 */ s32 unk80;
     /* 0x84 */ s32 unk84;
     /* 0x88 */ s32 unk88;
-    /* 0x8C */ s32 unk8C;
+    /* 0x8C */ struct Entity* parent;
     /* 0x90 */ s32 unk90;
     /* 0x94 */ s32 unk94;
     /* 0x98 */ s32 unk98;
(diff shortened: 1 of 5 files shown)


sozud (approved):
Looks good to me, having some of the documentation you wrote up in the source might make sense, but not essential to get this merged.

PR #511, opened 2023-08-25
NO3 & NP3 EntityStairwayPiece + merman rock Docs

sonicdcer:
image

I've been trying to match this one since two weeks now, i very happy to finally being able to decompile it.
https://decomp.me/scratch/AZMMV

I'm including a few duplicates and doc corrections since they are in the same file for consistency.


src/st/no3/377D4.c
```diff
@@ -1001,7 +1144,7 @@ void EntityMermanRockRightSide(Entity* self) {
             tilePos += 0x30;
         }
 
-        if (D_8003BDEC[51] & 2) { /* 0 0 0 0 0 0 1 0 = Half broken */
+        if (D_8003BDEC[jewelSwordRoomUnlock] & rockBroken) {
```
sozud, line 1147: A lot of changes unrelated to decompiling EntityStairwayPiece are showing up in this PR, this is probably going to slow down PR review. Could you make a second PR for the other stuff?

sonicdcer: Just checked, pretty sure you can read through it in one minute, I only documented merman rock left and right and included two duplicates, it isn't that much.

sozud (approved):

PR #512, opened 2023-08-25
Decompile SEL TestCollisions (fake name)

bismurphy:
This is very strange.

All of the "normal" overlays have a function called TestCollisions which is extremely large and complicated. I noticed this one in the functions.md list, and specifically that this one is fewer instructions than the others. I thought decompiling this one might help as a simplified version to solve the other ones.

I was surprised to find that this is a completely different function, and does nothing related to collisions at all. It only does a bunch of Primitive handling. It is not relevant for the real TestCollisions, but I figured I should decompile it anyway. Here it is.

I do not know what this does. The function call graph https://raw.githubusercontent.com/Xeeynamo/sotn-decomp/gh-duplicates/function_calls/sel.TestCollisions.svg does not indicate any callers. We should either determine what this does and give it a useful name, or revert it to `func_801B410C`.

The behavior with the "scaled_b014" variable is unusual; I'm not sure if it's possible to change some types somewhere to eliminate its usage. 

I think those are all the main important points, more than happy to make changes to this very weird function.


src/st/sel/3410C.c
```diff
@@ -0,0 +1,539 @@
...
+        if (g_pads[0].tapped & 0x800) {
+            g_api.PlaySfx(0x63D);
+            D_8003C9A4++;
+        }
+        return;
+    case 3:
+        D_8003C9A4 = 6;
+        return;
+    case 4:
+        if (g_pads[0].tapped & 0x6000) {
```
sozud, line 236:
```
        if (g_pads[0].tapped & (PAD_DOWN | PAD_RIGHT)) {
```

This is for the file select / name change screen.

src/st/sel/3410C.c
```diff
@@ -0,0 +1,539 @@
...
+            g_api.PlaySfx(0x688);
+            if (++D_800987B4 >= 2) {
+                D_800987B4 = 0;
+            }
+        }
+        if (g_pads[0].tapped & 0x9000) {
```
sozud, line 242:
```
        if (g_pads[0].tapped & (PAD_LEFT | PAD_UP)) {
```

src/st/sel/3410C.c
```diff
@@ -0,0 +1,539 @@
...
+            if (i != 0x10) {
+                continue;
+            }
+            if (prim->tpage == 0x18) {
+                prim->u1 = 0x78;
+            } else {
+                prim->u1 = 0x70;
+            }
+        }
+        if (g_pads[0].tapped & 0x800) {
```
sozud, line 227:
```
PAD_START
```

Also this line is checking for pressing start on the title screen. We can probably rename this function

sozud (approved):
Excellent work!

src/st/sel/3410C.c
bismurphy: Thanks! Knew I was forgetting something! All button enums are now in.

PR #514, opened 2023-08-26
Resolve remaining duplicates for CollectGold

bismurphy:
Most overlays have the same CollectGold function (st0 is an odd one out, there may be others). I noticed in the duplicates report (https://raw.githubusercontent.com/Xeeynamo/sotn-decomp/gh-duplicates/duplicates.txt) that there were 3 overlays which still had placeholders in. This PR resolves that and finishes decompiling all of them.

The C code in all of these is copied from the `wrp` version, if it matters. Symbols have been changed to account for different data locations across overlays. I have attempted to be as formulaic as possible with these.

In most of these, there is a function which is called at the end of the function. In the `cen` version, it is `func_8019BBA4`. I have also decompiled that function in every overlay, so once this PR is in, I will submit another for those functions.


sozud:
Looks good, if you want to take this further you can rename the function and D_ data and pull these into .h files. Example https://github.com/Xeeynamo/sotn-decomp/pull/487

PR #515, opened 2023-08-26
Decompile ST0 func_801BA7D8

bismurphy:
This is an unknown function. I don't know what it does. It is duplicated across most overlays. It is called in CollectGold, EntityEquipItemDrop, and TestCollisions.

It seems to take in a string, and then do a bunch of things with primitives, but I don't know what. We don't directly reference that array, we only read and write to it based on a pointer which moves through the array. I tried to remove the pointer but didn't get anywhere.

Fully-matching scratch is here: https://decomp.me/scratch/UV2K2

I am starting with only this one version of the function, so that if changes need to be made (variable names, control flow, etc), I can update them once here, and then copy them to the other places, instead of making changes in every place. Once we are happy with the state of this function, I will add the duplicates to this PR prior to merging.


sozud:
I tried `nop`ing this out and the function is the logic for displaying this 
image

bismurphy:
As far as I know, this should be in good shape now. No more placeholder names. Happy to hear feedback if any names should be changed. If we're happy with this, I will copy this to all the other overlays and then it should be good to merge.

sozud (approved):
Looks good to me, I can merge this as it is or if you want to add the duplicates in this PR that's also fine

bismurphy:
Okay, all the versions are in, this should be ready to merge any time.

PR #518, opened 2023-08-27
NO3 EntityCastleDoorTransition

sonicdcer:
This is the Entity that pushes the player through the castle door at the entrance.

I decompiled this a long time ago, but it wasn't matching because of an instruction reorder, oddly, it was matching in PSYQ 4.0 for some weird reason. Today i discovered why, the problem was the function prototype.
https://decomp.me/scratch/a1jYQ


src/st/no3/3C4EC.c
```diff
@@ -20,7 +21,8 @@ void EntityPushAlucard(Entity* entity) {
     case 1:
         player->posX.val += 0x60000;
         if (roomLayout->unkA > 0x800) {
-            g_Entities[UNK_ENTITY_1].ext.generic.unk7C.S8.unk0 = 0;
+            g_Entities[alucardController].ext.alucardController.unk7C =
```
sozud, line 24: Why do you use `g_Entities[UNK_ENTITY_1]` and `g_Entities[alucardController]`?

sonicdcer: Sorry, that was a mistake, i'll take care of it.

sozud: Why is using a local variable better than a define? I see there's many usages of `g_Entities[1] across the codebase.

sonicdcer: You have a point there, corrected.

PR #520, opened 2023-08-28
WRP EntityRelicOrb matching

Xeeynamo:
Took me a while. I had to move a good chunk of the data from 6FD0 due to `const char* D_80180F84[] = {"Obtained "};` as it declares data in both `.data` and `.rodata`. Due to its strategic location, I did not yet find a clean way to have it matching in other overlays without starting to import data.

The function is not very clean and there are loads of temp variables and fake stuff. I ask to spend a bit of time to make improvements where possible. I want to get it in a good shape before committing myself to de-duplicating it for all the other overlays.
https://decomp.me/scratch/kyBr8

### DIFF
--- a/include/entity.h
+++ b/include/entity.h
@@ -423,7 +423,7 @@ typedef struct {
 
 typedef struct {
     /* 0x7C */ s8 unk7C;
-} ET_AlucardController; 
+} ET_AlucardController;
 
 typedef struct {
     /* 0x7C */ u16 unk7C;


src/st/wrp/D020.c
```diff
@@ -477,7 +477,450 @@ void EntityEquipItemDrop(Entity* self) {
...
+            self->ext.relicOrb.unk90 = 0x40;
+            self->ext.relicOrb.unk94 = -self->ext.relicOrb.unk94;
+        }
+        MoveEntity();
+        UNK_Update0[self->ext.relicOrb.unk8C] = 0x10;
+        break;
+
+    case 5:
+        g_api.func_800FE044(relicId, 0x2000);
+        if (relicId > 0x15 && relicId < 0x14) {
```
sozud, line 701: I think  `RELIC_GHOST_CARD / RELIC_FAERIE_CARD`?

sozud (approved):
Nice work!

PR #443, opened 2023-08-06
Add dups tool

sozud:
This PR adds a tool that compares duplicates across two files, and prints them out in order. The idea is identifying patterns in the sequence of duplicates within two overlays. My thinking is that if we identify sequences of functions that are the same, we can split those out into individual files and share them, reducing the amount of copy-pasting. Usage is like this:

```
cargo run --release ../../asm/us/st/nz0/nonmatchings/ ../../asm/us/st/np3/nonmatchings/ .90
```

`.90` is the Levenshtein similarity threshold. 

Output looks like this:
https://gist.github.com/sozud/4ced23959c505d8527b3bcd691cb59c2

The first section shows found duplicates and their similarity:
```
func_801B3C38                            | func_801B65FC                            | 0.9039301310043668                      
```
The second section prints all of the functions in the first file out, and shows any duplicates found in the second column. 

```
func_801B8F94                            |                                         
Random                                   | Random                                  
Update                                   | Update                                  
func_801B9800                            | func_801B93E8                           
TestCollisions                           | func_801B94F0                           
EntityNumericDamage                      | EntityNumericDamage                     
CreateEntityFromLayout                   | CreateEntityFromLayout                  
CreateEntityWhenInVerticalRange          | CreateEntityWhenInHorizontalRange       
CreateEntityWhenInHorizontalRange        | CreateEntityWhenInHorizontalRange       
func_801BB3B8                            |                                         
```

For example the above section from `Random` to `CreateEntityWhenInHorizontalRange` may have originally been a file that's in every stage overlay, but this requires more investigation.

PR #527, opened 2023-08-29
func_801A910 Duplicates + Custom SEL version

sonicdcer:


include/game.h
```diff
@@ -988,6 +988,26 @@ typedef struct {
     /* 0x0E */ u16 unk0E;
 } RelicOrb; /* size=0x10 */
 
+typedef struct {
+    /* 0x00 */ const char* D_801C24CC;   // pointer to dialogue next character
```
sozud, line 992: I know this was pre-existing but seems like if we know what these do we ought to just name them?

src/st/sel/3642C.c
```diff
@@ -32,7 +32,33 @@ void func_801B68E0(s32 arg0) {
...
+void func_801B690C(u8 ySteps, Entity* self) {
+    s32 primIndex = g_Dialogue.D_801C24DA + 1;
+    Primitive* prim;
+    s32 i;
+
+    while (primIndex >= 5) {
+        primIndex -= 5;
+    }
+
+    if (self->step_s == 0) {
```
sozud, line 44: Is self vs. g_CurrentEntity the only difference?

sozud (approved):

src/st/sel/3642C.c
sonicdcer: Yes it is

include/game.h
sonicdcer: I rather leave them because @Xeeynamo was working on these functions and maybe he needs to conserve the addresses for something, and i don't want to disrupt the process he's going through to understand them.

src/st/sel/3642C.c
sonicdcer: prim->v0 = ySteps + prim->v0;
also this, it doesn't match with prim->v0 += ySteps;

sonicdcer: https://decomp.me/scratch/Bdq1K

include/game.h
sonicdcer: Alright! just renamed them, let me know if everything looks good.

sonicdcer:
> It looks good. In the future we need to put these into a single file like @sozud did. I cannot imagine to keep copy&pasting the code for the remaining 40 stage overlays.

I agree, but there is a catch, there seems to be a pattern between functions that are always together and repeated across overlays, i want to get them right before de-duplicating them, rather than creating a file for each function if you know what i mean.

sonicdcer:
> > > It looks good. In the future we need to put these into a single file like @sozud did. I cannot imagine to keep copy&pasting the code for the remaining 40 stage overlays.
> > 
> > 
> > I agree, but there is a catch, there seems to be a pattern between functions that are always together and repeated across overlays, i want to get them right before de-duplicating them, rather than creating a file for each function if you know what i mean.
> 
> We already know what the pattern is #443

Interesting app, it seems i missed it.

sozud:
This got conflicted

sonicdcer:
> This got conflicted

Resolved.

PR #528, opened 2023-08-30
WRP BlitChar

Xeeynamo:
image

Used by `EntityRelicOrb` and `EntityUnkId0E` to pre-render the text to show as a texture.

I am pretty sure the function is bugged. When I use it in the debug module it runs fine. But when I copy&paste the matching decompiled function and use it, the game crashes here:
```c
        for (i = 0; i < FontHeight; i++) {
            ptr = chPix;
            ptr += i * FontStride;
            for (j = 0; j < 5; j++) {
                ptr[0] = ptr[1];
                ptr++;
            }
            *ptr = 0;
        }
```
The debug module is using a much more modern compiler, so it is entirely possible that there is an undefined behaviour in the current logic. I am poorly convinced about the `if (ch == MINSCODE)` and skeptical of the `if (i != FontHeight)` condition. The crash happens when a non-wide character is used (e.g. `l`, `i`, `e`, `d`). It matches though, so there you go.

The font the game uses is found in the game BIOS and it is retrieved by `Krom2RawAdd`. According to [Gemini](https://www.romhacking.net/forum/index.php?topic=17691.20) the font width is resized in SOTN. This is done by `g_api.func_80106A28(ch, 1)`. Although good luck at decompiling it: https://decomp.me/scratch/Lulhn

Co-authored-by: @Mc-muffin for figuring out the `while(true)` loop
Co-authored-by: @sonicdcer for figuring out the regallocs


src/st/wrp/D020.c
```diff
@@ -475,7 +475,112 @@ void EntityEquipItemDrop(Entity* self) {
...
+            var_s3 = 2;
+        } else {
+            ch = *str++ | (ch << 8);
+            if (ch == MINSCODE) {
+                var_s3 = 2;
+            }
+        }
+    }
+
+    if (ch == 0x8168) {
```
sozud, line 511: Apparently
```
# RIGHT DOUBLE QUOTATION MARK
```
https://github.com/python/pythontestdotnet/blob/master/www/unicode/SHIFTJIS.TXT

src/st/wrp/D020.c
```diff
@@ -475,7 +475,112 @@ void EntityEquipItemDrop(Entity* self) {
...
+
+    ch = *str++;
+    var_s3 = 0;
+
+    if (ch >= 'a' && ch <= 'z') {
+        ch += 0x8220;
+    } else if (ch >= 'A' && ch <= 'Z') {
+        ch += 0x821F;
+    } else {
+        if (ch == 0x20) {
```
sozud, line 500: ascii space it seems

src/st/wrp/D020.c
```diff
@@ -475,7 +475,112 @@ void EntityEquipItemDrop(Entity* self) {
...
+    s32 letterWidth;
+    u8* chPix;
+    u8* ptr;
+    u8* dst;
+
+    ch = *str++;
+    var_s3 = 0;
+
+    if (ch >= 'a' && ch <= 'z') {
+        ch += 0x8220;
```
sozud, line 496: This function is sort of doing the opposite of this it seems.
https://github.com/wine-mirror/wine/blob/b4280a486f7308fd1da260cd5b48b7d9e3c747c5/dlls/msvcrt/mbcs.c#L534

Shift ascii char into shift-jis lowercase fullwidth range

src/st/wrp/D020.c
```diff
@@ -475,7 +475,112 @@ void EntityEquipItemDrop(Entity* self) {
...
+    } else if (ch >= 'A' && ch <= 'Z') {
+        ch += 0x821F;
```
sozud, line 498: shift into uppercase fullwidth range

sozud (approved):

sonicdcer:
Took me a while to figure out that regalloc, but yet again like you say, there's something weird going on with that logic, i'm not convinced that this is a real match but it will have to do, definitely something to go back to in the future. Great work!

sonicdcer (approved):

PR #529, opened 2023-08-30
EntityRelicOrb duplicates

Xeeynamo:
As I finished importing the data in WRP, I came with a possible new pattern when de-duplicating functions. I kept using a header file (easily indexable from VS Code) in `src/st`. But this time it gets included by `entity_relic_orb.c`. All the sections `.data`, `.rodata` and `.text` are migrated into `entity_relic_orb.c` but the file itself just includes `src/st/entity_relic_orb.h`. Thinking forward, once we will be able to fix the problem present in #464 it will just be a matter of renaming `entity_relic_orb.h` into `entity_relic_orb.c` and remove each individual `entity_relic_orb.c` that currently acts as a proxy.

~~This PR is a draft. If you agree this is a good pattern I will proceed to do the same with the remaining overlays after #528 is good to be merged.~~

### DIFF
--- a/src/st/mad/mad.h
+++ b/src/st/mad/mad.h
@@ -79,7 +79,7 @@ extern LayoutEntity* D_801997D8;
 extern u16* D_801997DC;
 extern u8 D_801997E0;
 extern s8 D_801997E4;
-extern u16 g_ItemIconSlots[];
+extern u16 g_ItemIconSlots[0x20];
 
 extern u16 g_eBreakableInit[];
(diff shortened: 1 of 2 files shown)


sozud:
To make sure I'm understanding:

1. Split EntityRelicOrb into it's own C file in the splat yaml
2. Include the shared header in that C file from the outer directory

This seems like a reasonable solution to me.

sozud (approved):
Looks good to me

PR #534, opened 2023-08-31
MAD EntityRelicOrb

Xeeynamo:
Similar to #520 and #529 but this decompiles the version in MAD. The biggest difference is how the character sequence is decoded. In the US version it looks like they splitted chunks of code into smaller functions. I tried to put these chunk of codes into functions using the `inline` keyword but I got reg swaps and instructions moved.

This confirms MAD was part of the Japanese build. Due to the file timestamp and the name `symbols.beta.txt` I decided to assign to it `VERSION_BETA`.


sozud (approved):
Interesting findings, nice work

PR #538, opened 2023-09-01
Remove PixPattern

Xeeynamo:
The structure `PixPattern` was a fluke. I think it was introduced due to `func_800EB4F8`, one of the first functions ever compiled. Since then I always assumed the structure was used everywhere when it is clearly not the case. Using the right type clearly simplifies the code and it will make future functions easier to decompile. Note I did not get rid of some `(PixPattern*)` as they are part of a NON_EQUIVALENT function.


src/dra/47BB8.c
```diff
@@ -70,17 +70,17 @@ s32 func_800E7E08(u32 arg0) {
         } while (i < 0x10);
         break;
     case 19:
-        LoadTPage((PixPattern*)0x80280000, 0, 0, 0x2C0, 0x100, 0x100, 0x80);
-        LoadTPage((PixPattern*)0x80284000, 0, 0, 0x2C0, 0x180, 0x80, 0x80);
+        LoadTPage((u8*)0x80280000, 0, 0, 0x2C0, 0x100, 0x100, 0x80);
```
sonicdcer, line 73: The SDK manual says LoadTPage takes a u_long ptr in the first argument. Should we change the prototype in libgpu.h and respect the original castings?

sonicdcer (approved):

sozud (approved):

PR #539, opened 2023-09-01
Remove find_duplicates

Xeeynamo:
As [we discussed](https://discord.com/channels/1079389589950705684/1135205782703570994/1146365208974086194), this tool is no longer needed since #467 was introduced.

### New script to generate the duplicates report:
```
cd tools/dups && cargo run --release -- --threshold .90 --output-file ../../gh-duplicates/duplicates.txt
```

### Always up-to-date duplicates list
[**generated report**](https://raw.githubusercontent.com/Xeeynamo/sotn-decomp/gh-duplicates/duplicates.txt)

### Alias to query duplicates of a given function
```
alias find_duplicates="curl https://raw.githubusercontent.com/Xeeynamo/sotn-decomp/gh-duplicates/duplicates.txt | grep -A 5 -B 5"
find_duplicates func_801C58E0
```


sonicdcer:
As we discussed in Discord, i'm completely against removing this tool because it is very handy and i use it everyday.

sonicdcer:
> > As we discussed in Discord, i'm completely against removing this tool because it is very handy and i use it everyday.
> 
> I provided three different options to circumnavigate your issue. I am not sure why you are so against moving forward.

Because it makes my work more difficult, and leaving a file alone doesn't make a difference. I use the command line to look for duplicates, and using an alias requires to be set every time you open a terminal, the script is also slower and it's output very bulky.

sonicdcer:
> > using an alias requires to be set every time you open a terminal
> 
> That is not true. Please look online how to set an alias permanent based on your system configuration.

Didn't know that, my other reasons stand anyway.

sozud (approved):
Originally I wasn't sure what to do with this since I don't want to break people's workflows, but since @sonicdcer is moving on I guess we can merge.

PR #541, opened 2023-09-02
Dialogue duplicates

sonicdcer:
A few of them were pretty different and i had to decompile some differences.

### DIFF
--- a/config/symbols.us.stno3.txt
+++ b/config/symbols.us.stno3.txt
@@ -147,3 +147,4 @@ EntityZombieSpawner = 0x801D6710;
 UnkPolyFunc0 = 0x801D7020;
 UnkLoopFunc = 0x801D704C;
 g_ItemIconSlots = 0x801D7DD8;
+g_Dialogue = 0x801D7D24;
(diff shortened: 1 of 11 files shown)


sonicdcer:
> There is a lot of duplication and copy & paste. You should start moving the shared code appropriately like in the other PRs from me, sozud and bismurphy.

Yes, I am committed to do it once i got the pattern right, don't worry.

sonicdcer:
Just to be clear, i'm on the same page as you, i don't see myself copy pasting functions across 40 overlays in the future, i agree that it needs to be done, and i will do it.

sonicdcer:
> Then I do not get why you cannot do it now and accept a promise?

As I said before, ill do it once I get the pattern right, which I haven't done yet because I'm focusing on something else, there's also the data problem to go through with deduplicating entity functions that will make everything worse. Please, I need this merged, have more NZ0 functions decompiled waiting for this.

sonicdcer:
> I do not get of what pattern are you talking about. 

Function and data grouping basically.

sozud:
My opinion is that this gets us closer to real deduplication, so it's still worth merging. @sonicdcer Could you reopen this?

PR #545, opened 2023-09-02
DRA func_800F2014

cram0:


cram0:
This function is interesting, it seems like it's checking if the player is visiting _enough_ space within a room to unlock it based on a smaller grid : `1 << ((3 - (x & 3)) * 2);` 

If you have any better var names be my guest 😄

PR #548, opened 2023-09-04
DRA func_801119C4

cram0:


sozud (approved):
Nice work, this one needs symbol addresses for
```
extern s32 D_80137FDC;
extern s8 aCharal02x;
```
Added to https://github.com/Xeeynamo/sotn-decomp/blob/master/config/symbols.hd.dra.txt to fix the HD build

PR #549, opened 2023-09-04
Rename placeholder to IsRelicActive

bismurphy:
Function was still using a placeholder even though its purpose is fairly obvious.

I also changed it to take the relicid as an enum for its argument, so that hopefully these will fill in more automatically when decompiling. I also went through all the existing calls to this function, and changed them to use the relic name enum in their calls, as well as treating them as booleans. Some lines will show all three of these changes. A good example is:

`func_800FE3A8(0xE) == 0` which becomes: `!IsRelicActive(RELIC_HOLY_SYMBOL)`. We change the function name, we change the argument to be the enum, and we change the ==0 to a simple !. Overall I think this makes the code for these relics much more readable, and definitely much more searchable.


sozud (approved):

sozud:
It looks that tools/maspsx is conflicted.

bismurphy:
> 

I could be mistaken, but I believe this is updating maspsx, rather than a conflict.

PR #552, opened 2023-09-05
Decompile DRA func_8012D3E8

bismurphy:
This function introduces some weirdness.

First, because it has a jump table, we split it out to a new file. We're now chopping up 75F4C.c. This function is at the top of 8D3E8.c (hidden in the diff by default due to splitting to new file).

This function seems to run a state machine based on the value of D_800B0914. It would likely be worthwhile to try to determine what state this is tracking, so we can name both that variable and this function.

This function calls `func_8012CA64`, but strangely, in one instance, passes a 0 argument, when that function is not supposed to take arguments. I have therefore commented out `func_8012CA64` from `dra.h`. This seems to work, at least for now.

This writes to PLAYER, at location 0xAC, which is in the entity extension. I couldn't find an existing extension for the player (which seems odd), but I've created one now, which just holds that value. Maybe other functions will flesh this out.

This uses a ton of `PAD_` values, as well as `FIX()` values. I think I got them all, but let me know if I missed one. There is a @bug on one of them, where a value being capped at -3 accidentally used a 3. I'm pretty sure this is a bug/typo, but let me know if you disagree.

I think that's everything. A surprisingly complicated function.


sozud (approved):
Nice work!

PR #553, opened 2023-09-05
DRA func_800F5904

cram0:
Looks like the function uses a MenuContext as its first argument everytime it is called which is why I changed its type instead of casting it from `void`.

### DIFF
--- a/src/dra/5298C.c
+++ b/src/dra/5298C.c
@@ -451,12 +451,10 @@ bool ScissorSprite(SPRT* sprite, MenuContext* context) {
 }
 
 void func_800F5904(MenuContext* ctx, s32 x, s32 y, s32 w, u32 h, s32 u, s32 v,
-                   s32 unk1, s32 unk2, bool disableTexShade, s32 unk4) {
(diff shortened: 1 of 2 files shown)


src/dra/5298C.c
```diff
@@ -450,7 +450,42 @@ bool ScissorSprite(SPRT* sprite, MenuContext* context) {
...
+    }
+    SetSemiTrans(sp, 0);
+    SetShadeTex(sp, disableTexShade);
+    sp->x0 = (s16)x;
+    sp->y0 = (s16)y;
+    sp->w = (s16)w;
+    sp->h = (s16)h;
+    sp->u0 = (u8)u;
+    sp->v0 = (u8)v;
+    if ((ctx == 0) || (ScissorSprite(sp, ctx) == false)) {
```
bismurphy, line 479: Probably best to say ctx == NULL instead of 0?

cram0: Updated it with NULL instead of 0 👍🏻

sozud (approved):
Looks good!

PR #555, opened 2023-09-07
Decompile DRA func_8012E040

bismurphy:
According to https://raw.githubusercontent.com/Xeeynamo/sotn-decomp/gh-duplicates/function_calls/dra.func_8012D28C.svg

This should finish the last of the callers to func_8012D28C.

It's extremely interesting that all three callers use exactly the same block of code to call this. Seems like a copy paste. The rest of these functions are very different though. I think I'll need to move up the call tree to see what calls these three.

Anyway, no major comments on this one.

### DIFF
--- a/src/dra/8D3E8.c
+++ b/src/dra/8D3E8.c
@@ -253,7 +253,128 @@ void func_8012DBBC(void) {
 
 INCLUDE_ASM("dra/nonmatchings/8D3E8", func_8012DF04);
 
-INCLUDE_ASM("dra/nonmatchings/8D3E8", func_8012E040);
+void func_8012E040(void) {
+    s32 vel_boost;
+    s32 var_s0;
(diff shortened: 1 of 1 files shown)


src/dra/8D3E8.c
```diff
@@ -253,7 +253,128 @@ void func_8012DBBC(void) {
...
+            PLAYER.velocityY = 0;
+            func_8012CED4();
+            return;
+        }
+        PLAYER.velocityY = 0;
+    }
+    if (g_Player.padTapped & PAD_SQUARE) {
+        func_8012CC30(1);
+    }
+    switch (D_800B0914) { /* irregular */
```
sozud, line 313: Should we keep the irregular comment?

sozud (approved):

PR #561, opened 2023-09-07
Decompile DRA func_8012EF2C

bismurphy:
This function is only called by EntityAlucard. It handles a bunch of different function calls, depending on what step_s is.

The two if-statements at the end might be better as FIX() values, but until we know the importance of `D_80138430` I'm going to leave these as raw hex.

This function has rodata, but because I've been going through and decompiling every function in 8D3E8.c, we can parse the rodata without needing to split out to a whole new C file, which is nice.

I will be decompiling all of the various functions called as a result of the switch statement in this function.

Take a look at D_800AFFB8 - I'm wondering if it's a struct of some sort, where D_800AFFB8 is an array of pairs, and we're writing to the first member of each pair. Not sure, hopefully other functions will help identify that.


bismurphy:
Oh! This is my first function which is different between US and HD! I think I did the #ifdef properly, but please double check me on this one! No idea why the difference exists, but at least I got a match.

sozud (approved):
I reran the linter and it's fine. Looks good to me!

src/dra/8D3E8.c
```diff
@@ -1319,5 +1398,3 @@ void func_80134E64(void) {
 
 void func_80134F50();
 INCLUDE_ASM("dra/nonmatchings/8D3E8", func_80134F50);
-
-const u32 rodataPadding_jpt_8012D4D8 = 0;
```
sozud, line 1323: I missed this originally, do you know why do we no longer need this padding?

bismurphy: That's because of pulling in the rodata for this new function. Before we only had the rodata for func_8012D3E8, but now we also have func_8012EF2C's rodata getting parsed in.

The jump table for func_8012D3E8 ends with 4 bytes of zeros, while the jump table for func_8012EF2C does not. I don't know the **exact** details of why that is or anything, but that's the reason. Before, our last rodata ended with a set of zeros, and now, our last rodata does not. 
image

PR #562, opened 2023-09-07
DRA func_8011B480

cram0:
I took the initiative to add a macro for the flag, tell me whether it's a good thing or not


src/dra/75F54.c
```diff
@@ -861,7 +861,24 @@ void func_8011B334(Entity* entity) {
     }
 }
 
-INCLUDE_ASM("dra/nonmatchings/75F54", func_8011B480);
+void func_8011B480(Entity* entity) {
+    if (PLAYER.step != 5 || PLAYER.step_s != 3) {
+        DestroyEntity(entity);
+    } else {
+        entity->flags = FLAG_UNK_60000;
```
sozud, line 868: Is `FLAG_UNK_20000 | FLAG_UNK_40000` equivalent? `0x6` is `0b110` so it's actually two flags.

sozud (approved):

src/dra/75F54.c
cram0: Yes it is, you've been using it with both flags instead of the one I added, I'll remove it and use them instead 😃 , I should've checked before.

cram0: Updated it 👍🏻

sozud: Thanks, just FYI there's places where they are used individually, which is why we think they are individual flags https://github.com/Xeeynamo/sotn-decomp/blob/19a71a51e5b166d7c6cfd877858a3708d69bb543/src/dra/75F54.c#L724

cram0: Exactly, there are other parts of the code that also use `FLAG_UNK_20000 | FLAG_UNK_40000` 👍🏻

PR #566, opened 2023-09-08
Make markdown page link to raw version of svg

bismurphy:
Small change here, just linking to the raw SVG so it's clickable. The github preview makes the bubbles no longer operate as links.

While we're at it, would it be okay if I moved the function_graphs.md up a directory, so that it sits alongside duplicates.md and functions.txt, rather than being mixed in the function_calls folder with all the svgs?


sozud (approved):
> While we're at it, would it be okay if I moved the function_graphs.md up a directory, so that it sits alongside duplicates.md and functions.txt, rather than being mixed in the function_calls folder with all the svgs?

I think that would be fine

PR #568, opened 2023-09-08
Decompile DRA func_8010A234

bismurphy:
A nice DECOMP_ME_WIP to resolve.

The behavior with D_8017A000 is really weird. I copied the handling of this from func_80109594 which is how I got the match to work.

It seems like D_8017A000 is sort of an API to the currently-held weapon. It reminds me of the gameApi overlay.

Anyway, it matches, so that's nice.

### DIFF
--- a/src/dra/692E8.c
+++ b/src/dra/692E8.c
@@ -158,55 +158,49 @@ void func_80109990(void) {
 void func_80109A44(s32 arg0);
 INCLUDE_ASM("dra/nonmatchings/692E8", func_80109A44);
 
-// regalloc
-// DECOMP_ME_WIP func_8010A234 https://decomp.me/scratch/rdeqb
-#ifndef NON_MATCHING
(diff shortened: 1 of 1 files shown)


src/dra/692E8.c
```diff
@@ -158,55 +158,50 @@ void func_80109990(void) {
...
+        PLAYER.unk19 &= 0xF3;
+        func_8010FAF4();
+        PLAYER.rotPivotY = 0;
+        PLAYER.rotPivotX = 0;
+        if (g_Player.pl_vram_flag & 1) {
+            func_8010E570(0);
+        } else {
+            func_8010E7AC();
+        }
+        g_Player.unk0C &= 0xFEFFFFFF;
```
sozud, line 198: Probably `~0x01000000` is more clear

sozud (approved):

PR #571, opened 2023-09-09
Decompile StoreSaveData

Xeeynamo:
This took me a monumental effort. The PR does:
* Decompile `StoreSaveData` for DRA (US and HD)
* Decompile `StoreSaveData` for SEL
* Match `MemcardFormat` in SEL
* Cross-reference symbols between DRA and SEL
* Bring common code between DRA and SEL to parity
* Discover `save_mgr.c`,  `save_mgr2.c` and  `save_mgr3.c` have the same layout in both DRA and SEL
* Deduplicate `save_mgr3.c`
* Import `.data` and `.rodata` from both DRA and SEL
* Convert UTF-8 Japanese characters into Shift-JIS while building
* Minor refactor with less hardcoded magic numbers

I made the following compromises:
* In DRA US `MakeMemcardPath` there is some dirty data at the end of the string. I had to create a byte array called `HACK_SaveName`. This does not occur on DRA HD or SEL
* I did not de-duplciate `save_mgr` as there are a good amount of functions and symbols I have no knowledge about
* I did not decompile `save_mgr2` and it will come as a separate PR.

### DIFF
--- a/src/dra/save_mgr3.c
+++ b/src/dra/save_mgr3.c
@@ -10,6 +10,10 @@ void MakeMemcardPath(char* dstSaveName, s32 saveSlot) {
     dstSaveName[0x11] += saveSlot % 10;
 }
 #elif defined(VERSION_US)
+// The US version has an extra 'n' character after the string terminator,
+// probably a result of some left-overs from a previous build. Defining the
+// string `SaveName` as an array to manually append 'n' at the end was the only
+// way to get a match.
 const char HACK_SaveName[] = {
     'B', 'A', 'S', 'L', 'U', 'S', '-', '0', '0',  '0',
     '6', '7', 'D', 'R', 'A', 'X', '0', '0', '\0', 'n',


src/dra/save_mgr3.c
```diff
@@ -0,0 +1,22 @@
...
+#include "../save_mgr3.h"
+
+#if defined(VERSION_HD)
+void MakeMemcardPath(char* dstSaveName, s32 saveSlot) {
+    STRCPY(dstSaveName, "BISLPM-86023DRAX00");
+    dstSaveName[0x10] += saveSlot / 10;
+    dstSaveName[0x11] += saveSlot % 10;
+}
+#elif defined(VERSION_US)
+const char HACK_SaveName[] = {
```
sozud, line 13: Maybe a comment explaining the issue?

sozud (approved):
Really cool to see the Japanese strings getting integrated, nice job!

src/dra/save_mgr3.c
sozud: I'm also wondering if either of these options match:

```
const char HACK_SaveName[] = "BASLUS-00067DRAX00\0n";
```

```
const char HACK_SaveName[] = "BASLUS-00067DRAX00";
const char padding[] = "n";
```

PR #577, opened 2023-09-10
Decompile DRA func_8010D800

bismurphy:
Notes:

1) We are pulling in more .data. Because this is just consecutive after the previous data, we just bump out the next segment in the splat.

2) Now we're using the unk1 member of the entity1 extension. Cool!

3) We turn out to be using parts of a previously-defined array. Generally, the code in this function feels a little clumsy. I don't like temp_t0, temp_t1, and temp_t2, but they're needed. Because I don't know what g_Entities[1] is, or what it's holding in its unk1 and unk2 values, I don't know better names for these variables. Same goes for `current_thing` which is a terrible variable name, but for now, it serves the purpose. In the future when we can work out more of how this works, we can give these useful names.


sozud:
SD previously added some stuff for entity 1 here: https://github.com/Xeeynamo/sotn-decomp/blob/ad95e38790c8830600c6a73a0027a21e89302ac1/src/st/no3/3C4EC.c#L4
It seems to have to do with taking control from the player / cutscenes.

src/dra/6D59C.c
```diff
@@ -71,7 +71,68 @@ void func_8010D59C(void) {
     }
 }
 
-INCLUDE_ASM("dra/nonmatchings/6D59C", func_8010D800);
+u8 g_D_800ACF2C[] = {4, 4, 4, 4, 6, 6, 6, 6, 8, 8, 16, 16, 16, 16, 16, 16};
```
sozud, line 74: If this is all 16s the shadows fade out very quick. `g_shadowOpacityReductionTable` or something?

src/dra/6D59C.c
```diff
@@ -71,7 +71,68 @@ void func_8010D59C(void) {
...
+    Unkstruct_800ECBF8_1* current_thing;
+    Primitive* prim;
+    s32 entNum;
+    s32 i;
+    u8 temp_t0;
+    u8 temp_t1;
+    u8 temp_t2;
+
+    i = 0;
+    prim = &g_PrimBuf[g_Entities[1].primIndex];
```
sozud, line 89: `g_Entities[ENTITY_ALUCARD_SHADOW].primIndex` or similar?

src/dra/6D59C.c
```diff
@@ -71,7 +71,68 @@ void func_8010D59C(void) {
...
+    temp_t2 = g_Entities[1].ext.ent1.unk1;
+    temp_t1 = g_D_800ACF2C[g_Entities[1].ext.ent1.unk2];
+    temp_t0 = g_D_800ACF3C[g_Entities[1].ext.ent1.unk2];
+
+    current_thing = &D_80097D1C[1];
+    for (i = 0; i < 6; prim = prim->next, i++) {
+        if (temp_t0 < prim->r0) {
+            prim->r0 -= temp_t1;
```
sozud, line 97: `prim->r0` seems to be opacity. I'm wondering if there's other meanings to these values we haven't figured out.

sozud (commented):
I tried `nop`ing this out and it seems to be the function that draws Alucard's "shadows" when walking around.

src/dra/6D59C.c
bismurphy: Could you try checking if this also applies to Richter when you do the dash move? I'd do it myself but I don't know how to nop out functions.

sozud: You can do it by going  to the beginning of the function and replacing the first 8 bytes with `0800E003 00000000`
```
/* 0800E003 */  jr         $ra
/* 00000000 */   nop
```

sozud: This function doesn't seem to execute at all during the prologue so I think Richter's shadows are handled somewhere else.

bismurphy: Perfect, thanks! Richter's shadows appear to be unaffected when changing this. In the prologue I saw some activity in the region of g_Entities[1], but this happened regardless of whether he had shadows happening, so I'm not sure how that works.

I'm a little concerned at how SD's function seems to think it's taking control from the player, but this one seems to be about shadows. I think overall we still need to learn more about g_Entities[1] before renaming it to ENITTY_ALUCARD_SHADOW or anything else.

PR #579, opened 2023-09-10
DRA DrawTimeAttackMenu

cram0:
This function is used to draw the time attack menu that is located in the pause menu.

Feel free to offer any opinions or ideas on how to name the variables.


sozud:
Do you have scratches for both versions? I think some things can be simplified

src/dra/5298C.c
```diff
@@ -784,7 +784,57 @@ void func_800F6BEC(MenuContext* context) {
...
+    s32 newCursorY;
+    s32 offsetY;
+    s32 i;
+    s32 time;
+
+    i = 0;
+    offsetY = 8;
+    while (i < 12) {
+        cursorX = ctx->cursorX;
+        temp_s0 = i + D_8003C9F0;
```
sozud, line 800: I think this may be an array?

src/dra/5298C.c
```diff
@@ -784,7 +784,57 @@ void func_800F6BEC(MenuContext* context) {
...
+        newCursorY = ctx->cursorY + offsetY;
+        DrawMenuInt(temp_s0 + 1, cursorX + 16, newCursorY, ctx);
+#if defined(VERSION_US)
```
sozud, line 803: I think the blocks of each version are the same except for some constant values so I think you could do something like
```
#if defined(VERSION_US)
#define POS 12345
#else
#define POS 23456
#endif
```
etc.

cram0:
First time trying to add more symbols and cross-referencing them, I'll gladly take any improvements if need be 😄

src/dra/5298C.c
```diff
@@ -784,7 +784,42 @@ void func_800F6BEC(MenuContext* context) {
...
+    s32 cursorY;
+    s32 entryIdx;
+    s32 offsetY;
+    s32 i;
+    s32 hours;
+    s32 seconds;
+
+    for (i = 0, offsetY = 8; i < 12; i++, offsetY += 12) {
+        cursorX = ctx->cursorX;
+        entryIdx = i + g_TimeAttackVerticalPageCursor;
```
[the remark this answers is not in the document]
cram0: I should've checked if it already had a symbol before, my bad ! 
correcting it right away, thanks !

PR #583, opened 2023-09-13
Decompile DRA func_801131C4

bismurphy:
This has a lot of similarities to my previous PR. Otherwise, not a ton of interesting content here.

This does appear to be the final function in 72BB0.c, which is nice.

### DIFF
--- a/src/dra/72BB0.c
+++ b/src/dra/72BB0.c
@@ -178,7 +178,264 @@ void func_80113148(void) {
     }
 }
 
-INCLUDE_ASM("dra/nonmatchings/72BB0", func_801131C4);
+void func_801131C4(void) {
+    s32 i;
+    s32 x_offset;
+    u16 local_flags;
+    bool condition = false;
+
+    local_flags = 0;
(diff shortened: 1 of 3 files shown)


src/dra/72BB0.c
```diff
@@ -178,7 +178,264 @@ void func_80113148(void) {
...
+                }
+                break;
+            } else if (g_Player.unk72 == 3) {
+                if ((u16)PLAYER.animFrameIdx >= 2U) {
+                    PLAYER.animFrameIdx = 2;
+                    PLAYER.animFrameDuration = 3;
+                }
+                break;
+            } else if (g_Player.unk72 == 4) {
+                if ((u16)PLAYER.animFrameIdx >= 3U) {
```
sozud, line 300: It looks like some of these casts and Us can be removed

```
            else if(g_Player.unk72 == 4){
                if (PLAYER.animFrameIdx >= 3) {
                    PLAYER.animFrameIdx = 3;
                    PLAYER.animFrameDuration = 3;
                }
                break;
            }
```

Probably the same for other spots

sozud (approved):

PR #590, opened 2023-09-13
Decompile DRA func_80116408

bismurphy:
Another random function that I don't know the purpose of. I'm mostly just trying to decompile everything called by EntityAlucard at this point.

This has some nice accesses of `g_Player.D_80072BD0`, which helps to confirm some of the details of what that is and how it works.

Note that I had a `(PLAYER.posX.i.hi >0xF9)` which I changed to use `(PLAYER.posX.i.hi > (u8)-8)) {`. I think that's a more reasonable way to write this if statement, given the if statement that comes right after it which tests against 8. If F9 is the proper value, let me know and I'll change it back.

We have two checks of inputs, one is checking PAD_RIGHT and the other is PAD_DOWN. It's strange that it doesn't check left at any point; this might be an oversight or maybe left is handled by a totally different function.


src/dra/75F54.c
```diff
@@ -142,7 +142,72 @@ void func_80116208(void) {
...
+            PLAYER.step_s++;
+        }
+        break;
+    case 2:
+        PLAYER.velocityX = 0;
+        if (func_8010E27C() != 0) {
+            if (g_Player.padPressed & PAD_RIGHT) {
+                if ((g_Player.D_80072BD0[2][0] & 0x8001) ||
+                    (g_Player.D_80072BD0[1][0] & 0x8000) ||
+                    (PLAYER.posX.i.hi > (u8)-8)) {
```
sozud, line 167: I'm thinking `(PLAYER.posX.i.hi > 248)` is more likely?

sozud (approved):

PR #592, opened 2023-09-14
DRA+SEL MemcardParse

Xeeynamo:
Decompiling this function the meaning of various functions, de-fake struct and symbols and document a previously unknown function.

This was the second last function from the `save_mgr` group. If possible, with the next PR I plan to merge `save_mgr`, `save_mgr2`, `save_mgr3` and importing the relevant `.data` and `.rodata` segments.

### DIFF
--- a/src/dra/5298C.c
+++ b/src/dra/5298C.c
@@ -1545,7 +1545,58 @@ void func_800F9F40(void) {
 
 INCLUDE_ASM("dra/nonmatchings/5298C", func_800FA034);
 
-INCLUDE_ASM("dra/nonmatchings/5298C", func_800FA3C4);
+void func_800FA3C4(s32 arg0, s32 arg1, s32 arg2) {
+    // FAKE: Should figure out how this actually works.
+    // Could be that 7676 is the start of another struct within MenuData.
+    s16* menuitem = &g_MenuData.D_80137676;
+    s32 limit;
+    s32 top_offset;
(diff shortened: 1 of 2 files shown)


src/st/sel/save_mgr2.c
```diff
@@ -1,20 +1,21 @@
...
 
-s32 func_801B8A8C(s32 arg0, s32 arg1) {
+s32 func_801B8A8C(s32 slot, s32 arg1) {
     s32 temp_s2;
-    Unkstruct_801B8A8C* unk;
-    temp_s2 = (arg0 * 0x10) + arg1;
-    unk = &D_801BC3F0[arg0];
-    switch (D_801D6B20) {
+    MemcardInfo* memcard;
+    temp_s2 = (slot * 0x10) + arg1;
```
sozud, line 11: Seems like `arg1` should be `cardNumber`?
`_card_info(long chan): Port number x 16 + Card number`

Maybe `chan = (slot * 0x10) + cardNumber;`

cardNumber maybe nonzero when a multitap is connected

sozud (approved):

src/st/sel/save_mgr2.c
sozud: I was just going off this 
image
I suppose I'd be inclined to go with the terminology from the manual: port_num and card_num or whatever, but I think the current ones are fine

PR #594, opened 2023-09-14
Fully decompile Save Manager for DRA (US+HD) and SEL

Xeeynamo:
What it is done:

* Decompile the last function from the `save_mgr` trio, `MemcardDetectSave`.
* Merge `save_mgr.c`, `save_mgr2.c` and `save_mgr3.c` into a single file
* Merge the respective headers
* Import all the `.rodata` segment
* Import all the `.data` segment
* DRA and SEL shares all `src/save_mgr.h`
* Remove orphan symbols

Removing the orphan symbols is very interesting. When certain functions or global variables are no longer referenced by the exported assembly code, there is no need to define them in the symbol list. This allows to freely relocate the code in case modders wants to add or remove code. I performed this step manually, but will make a script to do this automatically.


sozud (approved):
I think this is the first example of de-duplication with correct data and rodata? Nice work!

PR #595, opened 2023-09-14
Remove orphan symbols

Xeeynamo:
Allows to not hard code the location in-memory of decompiled functions and imported data if not required. This allows to relocate the referenced symbols when editing the original code or game data. If you were getting the offset of those symbols from the symbol list in `config/` to debug with PCSX Redux, I suggest to use the built `build/us/*.map` file instead.

### DIFF
--- a/.github/workflows/linter.yaml
+++ b/.github/workflows/linter.yaml
@@ -34,4 +34,4 @@ jobs:
       - name: Format code
         run: make format
       - name: Check for pending changes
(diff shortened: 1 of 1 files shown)


sozud:
I removed `--name-only` from the CI to diagnose this. I think the issue is that the symbol linter removes the line at the end, but the orphans script adds it. Since master doesn't run the orphans script, they won't match in the PR CI.

PR #599, opened 2023-09-16
SEL func_801B55C8

Xeeynamo:
I am not sure what this function does. I've been navigating through the main menu but nothing seems to trigger this function.


sozud:
I tried but it seems like several function call levels up at func_801B60D4 is also not called. Maybe left over debug code?

sozud (approved):

PR #600, opened 2023-09-16
Decompile DRA func_80116B0C

bismurphy:
This is a pretty cool function. It seems to handle most of the logic for the bat transformation, including the wing smash spell, fireballs, echos, and normal movement.

I am not sure what to call it - maybe ControlBatForm? Very open to suggestions.

This is a large one, so please do double check what I have here to make sure I'm not introducing code that will need to be fixed later. I suspect this function might be helpful for understanding some behavioral details.

There are a few really weird version differences between US and HD here. It makes parts of the code come out really ugly, but I guess that's the price we have to pay for representing two codebases in one.

Looking forward to any and all feedback for this one!


src/dra/75F54.c
```diff
@@ -400,7 +400,365 @@ s32 CheckWingSmashInput(void) {
...
+            PLAYER.facingLeft = 1;
+            func_8011690C(0x180);
+            if (PLAYER.velocityX < FIX(-1.25)) {
+                func_8010E1EC(0x1200);
+            } else {
+                PLAYER.velocityX = FIX(-1.25);
+            }
+            func_8010E234(0x1200);
+            var_s3 = 1;
+            if (D_80138000 == 0) {
```
sozud, line 580: I think this could be `g_BatNoScreetch` or something. At the beginning of the function, var_s3 is set to 0. When moving left and right, var_s3 is set to 1. Then at the end of the function that is assigned to `D_80138000`. So if we are moving left or right, make it so we don't screech, otherwise screech?

sozud (approved):
Great work as usual

src/dra/75F54.c
bismurphy: Sounds good, I'll rename the local and global vars. How do we feel about ControlBatForm for this function?

PR #603, opened 2023-09-17
SEL func_801ACFBC

Xeeynamo:
This function allowed me to have much more details about the `SaveSummary` struct, which also lead to different corrections here and there around the code. I also noticed the order of the symbols in `sel.h` was all over the place, so I moved things around to be more coherent with the rest of the codebase.

I am not yet confident of importing the rodata in this one, so I left the strings be.

### DIFF
--- a/src/st/sel/2C048.c
+++ b/src/st/sel/2C048.c
@@ -184,11 +184,11 @@ void func_801ACFBC(s32 port, s32 slot, s32 textId) {
         func_801B263C(playerName, textId);
         func_801B25D4(D_801A7620, textId + 1);
     } while (0);
-    if (g_SaveSummary[port].isRichter[slot] != 0) {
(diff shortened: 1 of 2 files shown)


src/st/sel/2C048.c
```diff
@@ -163,7 +164,37 @@ void func_801ACF7C(void) {
...
+        strcat(playerName, D_801803D0[func_801B2984(ch)]);
+    }
+
+    do {
+        func_801B263C(playerName, textId);
+        func_801B25D4(D_801A7620, textId + 1);
+    } while (0);
+    if (g_SaveSummary[port].isRichter[slot] != 0) {
+        strSaveKind = D_801A76A4;
+    } else if (g_SaveSummary[port].kind[slot] & 1) {
```
sozud, line 189: An enum or defines for the kind magic numbers?

sozud (approved):

PR #607, opened 2023-09-18
Decompile DRA DrawSettingsButton

bismurphy:
This is an old NON_MATCHING that I was able to finish. Not much more to it than that.


src/dra/5298C.c
```diff
@@ -906,41 +906,40 @@ void DrawTimeAttackMenu(MenuContext* ctx) {
...
     s32 buttonId;
     s32 btn1_x;
     s32 btn2_x;
-
+#if defined(VERSION_HD)
+    cursorX = 0x80;
+#define W 0x28
+#define XVAR cursorX
+#elif defined(VERSION_US)
     cursorX = 0x98;
```
sozud, line 921: My opinion is that having a macro for the value is a little cleaner for cursorX. `cursorX = CURSORX_VAL;` https://decomp.me/scratch/G4Qv4

sozud (approved):

src/dra/5298C.c
bismurphy: Sounds reasonable to me, change incoming

config/symbols.us.stsel.txt
```diff
@@ -1505,7 +1505,6 @@ func_801B2F50 = 0x801B2F50;
 func_801B3120 = 0x801B3120;
 func_801B3164 = 0x801B3164;
 func_801B367C = 0x801B367C;
-TryLoadSaveData = 0x801B3694;
```
sozud, line 1508: Did the formatter remove this? Seems like it's required

```
src/st/sel/2D260.c:(.text+0x3e98): undefined reference to `func_801B3694'
```

bismurphy: Yes, that was the formatter, not me. Seems to think it's an orphan symbol.

src/dra/5298C.c
bismurphy: I don't actually know the difference between a define and a const, if const is preferred I'll change it

PR #608, opened 2023-09-18
Decompile DRA DrawSpellMenu

bismurphy:
This one was pretty complicated, lots of weird differences between US and HD. Let me know if this could be handled better. Some of the differing function calls are particularly unusual. Anyway, it matches.


sozud (approved):

cram0:
Shouldn't coordinates (inside of DrawStr functions for example) be decimal values directly instead of hexadecimal values for better readability ?

Otherwise great job ! 👍🏻

PR #609, opened 2023-09-19
func_801177A0 matching

bismurphy:
More work in gathering data on the variables around D_800ACE##.

Not much progress on what any of those actually do, but this function seems to match the structures that I established in my previous PR. Hopefully with enough matching functions that use this data, we can work out what it's meant to do.

`else_cycles` isn't a great variable name, but... it counts how many cycles take the 'else' branch. Once we know what values we're comparing this could probably be renamed.


src/dra/75F54.c
```diff
@@ -760,7 +760,94 @@ void ControlBatForm(void) {
...
+            func_8010E83C(0);
+            PLAYER.posY.i.hi -= 3;
+            if (!(g_Player.pl_vram_flag & 0x8000)) {
+                PLAYER.velocityY = FIX(-1);
+            }
+            PLAYER.palette = 0x8100;
+#if defined(VERSION_US)
+            g_Player.unk20[0] = 0x18;
+#elif defined(VERSION_HD)
+            D_800ACE40 = 0x18;
```
sozud, line 835: I don't have the HD version built so I can't check, but do we know that this variable is outside the `g_Player` struct? This could be checked with the map file or `generated.symbols` file I think.

sozud (approved):

PR #623, opened 2023-09-22
Decompile DRA func_8010EDB8

bismurphy:
This one is an absolute beast and I'm amazed it matches!

This is one of the primary functions that runs when you have a weapon (or other item equipped in your hand) and you press the attack button (square or circle by default). It handles the special cases for many weapons, based on their `unk11` property. I have gone through `unk11` and labeled weapons in the places I was able to identify them. For example, this code includes the test on rand() & 7 that applies when you use the Red Rust, and causes it to fail to swing.

There is still a lot of `goto` in this. I suspect some of it could be cleaned up, but not all. My scratch is here if you would like to play around with cleaning up that control flow https://decomp.me/scratch/d1brK

I think documenting this function would be very worthwhile for someone in the future who is very interested in weapons; for now, I just want to get it in as it's working.


sozud (approved):
Excellent work, thinking I want to hear from Xeeynamo before merging

PR #625, opened 2023-09-23
Regenerate orphaned symbols in the CI

Xeeynamo:
[As mentioned](https://discord.com/channels/1079389589950705684/1135205782703570994/1154948845638254672), since #595 some tools are not able to pick up some function names as the symbol names are gone. This PR adds a quick tool to re-generate that list from a map file and integrates it in the CI flow.

EDIT: CI is failing because the running Linter is from `master`, where `mapfile-parser` is not installed.


sozud (approved):
Looks good to me assuming that merging will fix the CI.

PR #626, opened 2023-09-23
SEL CheckIfMemcardsCanBeUsed

Xeeynamo:
plus some orphan symbol clean-up from #623


src/st/sel/2D260.c
```diff
@@ -420,7 +420,39 @@ void func_801AECA0(void) {
...
-INCLUDE_ASM("asm/us/st/sel/nonmatchings/2D260", func_801AED48);
+void CheckIfMemcardsCanBeUsed(void) {
+    s32 isCard0Full;
+    s32 isCard1Full;
+    s32 i;
+
+    isCard0Full = true;
+    isCard1Full = true;
+    for (i = 0; i < BLOCK_PER_CARD; i++) {
+        if (g_SaveSummary[0].icon[i] != -2) {
```
sozud, line 431: To me it seems like `icon` is the status of each block or something like that? How did we pick the name `icon`? https://github.com/Xeeynamo/sotn-decomp/blob/5277b4f9c448c072b957dff2ad02a52135add7ba/src/st/sel/2D260.c#L425

It seems like `D_801BAF10` is `g_UnableToSave` or similar.

sozud (approved):

PR #627, opened 2023-09-23
SEL func_801B2E5C

Xeeynamo:
Another small step to decompile `func_801B3164`. It is still not exactly clear what this function is doing. I couldn't import `D_801A7AE4` from the `.rodata` without committing into a file split. I can import it if I am also able to decompile `sel.Update`, which is planned after `func_801B2F50`, `func_801B3164` and, hopefully, `func_801ADF94`.


include/game.h
```diff
@@ -775,22 +775,26 @@ typedef struct {
     /* 0x80 */ u8 Icon[3][128];
 } MemcardHeader; /* size=0x200 */
 
+typedef struct {
```
sozud, line 778: How did you figure out this is a separate struct? Not seeing a usage in the new function

src/st/sel/sel.h
```diff
@@ -63,17 +63,17 @@ typedef struct {
...
-    /* 801BCA48 */ u32 money[BLOCK_PER_CARD];
-    /* 801BCA84 */ u32 percentage[BLOCK_PER_CARD];
+    /* 801BCA48 */ u32 gold[BLOCK_PER_CARD];
+    /* 801BCA84 */ u32 nRoomsExplored[BLOCK_PER_CARD];
     /* 801BCAC0 */ u32 playHours[BLOCK_PER_CARD];
     /* 801BCAFC */ u32 playSeconds[BLOCK_PER_CARD];
     /* 801BCB38 */ u32 playMinutes[BLOCK_PER_CARD];
     /* 801BCB74 */ u32 kind[BLOCK_PER_CARD]; // 0: play, 1: clear, 2: replay
-    /* 801BCBB0 */ u32 isRichter[BLOCK_PER_CARD];
+    /* 801BCBB0 */ u32 character[BLOCK_PER_CARD];
```
sozud, line 76: It seems to me that if we want to rename to character we should use an `CHARACTER_RICHTER` enum here or similar? https://github.com/Xeeynamo/sotn-decomp/blob/57d47163ae0e093f97c0b6f8919a09a4a5c397ac/src/st/sel/2C048.c#L418

sozud (approved):

PR #628, opened 2023-09-23
Decompile DRA func_800FEEA4

bismurphy:
This is a cool one, this function is responsible for draining your MP when you're in a transformed state (bat, mist, or wolf).

arg0 specifies the relevant transformation, I think this should be 'form'. 0 = bat, 1 = mist, 2 = wolf. As far as I know, these values only exist in this function; if we find them used elsewhere they should become an enum but I'm inclined to hold off on that for the moment (let me know if I should make an enum right away though).

arg1 is a bool for whether we should actually drain MP, or just check if we have enough. Not sure what to call this. I've seen this kind of logic in a few places on a few functions, this could maybe be an enum with "CHECK_ONLY" or "REDUCE".

Finally, this function returns 0 if all is well, and -1 if there is a problem (for example, if we need to reduce MP, but the player doesn't have enough to stay transformed, or if the relevant relic is not active).

I'm not sure what to call this function. Maybe HandleTransformationMP? This kind of does three things, it checks our MP state being okay, it takes our MP away, and it reports whether we can stay transformed, so I think "Handle" is the best verb. Happy to take other opinions.


sozud (approved):
arg0 -> form
arg1 -> mode

I think an enum for arg0 makes sense whether it's used other places or not.

arg1  "CHECK_ONLY" or "REDUCE". sounds good except maybe add a prefix like MP_CHECK_ONLY or something

An enum for the return values would be good probably

HandleTransformationMP sounds good to me.

PR #634, opened 2023-09-24
Added several new sound defines

synamaxmusic:


include/sfx.h
```diff
@@ -241,6 +240,7 @@ typedef enum { MONO_SOUND, STEREO_SOUND } soundMode;
 
 // STAGE DRE
 #define NA_SE_SU_SHOOT_PINKBALLS 0x62C
+#define SE_SUC_REVEAL 0x637
```
sozud, line 243: Is updating the locations where these are used supposed to be a part of this? For example https://github.com/Xeeynamo/sotn-decomp/blob/140a5e6435cf317914aac89e5114322c806de5d2/src/st/dre/1697C.c#L44

synamaxmusic: I can start going through the codebase and updating those entries.

include/sfx.h
```diff
@@ -262,10 +263,52 @@ typedef enum { MONO_SOUND, STEREO_SOUND } soundMode;
...
+#define SE_TREE_BRANCH_SNAP 0x7A4
+#define SE_CASTLE_GATE_RISE 0x7A5
 
 // NZ1
 #define SE_HARPY_WING_FLAP 0x601
 
 // SHARED SOUNDS
 // These are sounds that are shared across multiple BIN files
 #define SE_WEAPON_STAB 0x62E
+#define SE_DOOR_OPEN 0x642
```
sozud, line 279: I think we already have one for this? Also are we sure these are shared? See the steps in ST0

```
#define NA_SE_EN_OPEN_DOOR 0x642
```
https://github.com/Xeeynamo/sotn-decomp/blob/ff1c3dc64cb207db2df4bd57142c72462bd5dab1/src/st/st0/2A8DC.c#L184

synamaxmusic: If you want to remove it, that's fine but I think SE_DOOR_OPEN is a better more concise name.  Not sure if the door effect was changed in the Japanese version though; but if it's not, do we need NA and EN?

To find if there's a sound call in a BIN file, I search for the sound ID in little-endian, followed by $0434.  For the vast majority of sound calls, this will work.  The only sound calls that I found that don't apply to this method are Alucard pain sounds, which are randomly triggered from a table lookup.

Searching for 0x642 by looking up "42060434" reveals that this sound shows up in 52 bin files.  Don't know how much redundancy you guys want with defines so I figured putting a shared sounds section in sfx.h would be okay.

And just because the sound ID is referenced in the code doesn't mean it'll actually get called in game.  0x602 is supposedly a sound for Beelzebub but I haven't gotten it to trigger yet in-game.

sozud: I think the problem with this technique is that 0x642 can probably mean different things in different overlays, so the ID may be used in multiple places but points to a different sound depending on what is loaded.

synamaxmusic: I think enemies are the only objects that I've encountered so far in game that have different sound effects depending on what stage is loaded.  Considering how many enemies are in the game, it makes sense that they would share common "attack" or "death" sound IDs and then the VAB associated with that stage loads the samples required for those enemies.

If we can make a list of sounds that use multiple VAB files then I can do some further investigating to see if I can find a pattern between sound IDs.

synamaxmusic: I also make sure the sound calls are actually valid by loading the stage and checking the SFXRingBuffer to see if the expected sound ID pops up.  That's how I figured out why this technique didn't work for the Alucard pain sound calls.

sozud:
Deferring to @Xeeynamo on this one, merge when you feel like it's ready.

synamaxmusic:
Sounds good!  Considering that CEN has so few sound calls, I'll go ahead and work on the defines for that overlay.

PR #636, opened 2023-09-24
SEL Update function matching

Xeeynamo:
I am proud to announce I decompiled the biggest function from the game detected so far! This operates the whole logic of the main menu and links together all the functions I've been decompiling in the past week.

I did my best to import the `.rodata` function. There's a big catch though. The symbol `D_801A7620` is used in this function, but I cannot import it because it is also used by `func_801ACFBC`, found much earlier than `Update`. This most likely means `2C048.c`, `2D260.c` and `2EED8.c` were a single source file.

Before committing myself into fully documenting the function, I want to finish to decompile `func_801ADF94` and `func_801AD66C`, which are necessary to merge the three files and finish importing the `.rodata`. Then I need to start to decompile `func_801B2108`, `func_801B3164`, `func_801B38B4`, `func_801B3A94` and `func_801B3F94` as they are dependencies of `Update`.


sozud:
Amazing work! Wasn't expecting to see this one done any time soon

PR #641, opened 2023-09-25
WRP and RWRP Sound IDs added to overlay headers / 6FD0.c now points to new sound names

synamaxmusic:


synamaxmusic:
Glad I could help!

PR #643, opened 2023-09-25
PlayerState collider restructure

bismurphy:
A recently decompiled function found that `s32 D_80072BD0[8][9]` was actually an array of Collider. This PR updates that to be correct.

It looks like the struct member right after it should be grouped in to have `colliders[22]` but for whatever reason that doesn't seem to work so I'm leaving it. Given that the struct size matches and everything, we probably have `colliders[8]` and then `colliders2[14]` but without more evidence I don't want to mark this down yet.

Some of these lines come out pretty long like if `((g_Player.colliders[1].effects & (EFFECT_SOLID_FROM_BELOW + EFFECT_SOLID)) == EFFECT_SOLID ||`. Here is what that looks like after I run the formatter:
image

Personally, I think the original code is a lot more readable since everything is stacked up. You can see the [1][2][3] much more easily than when the lines are interwoven with each other. This isn't the most egregious example of lines being broken apart, but it's not ideal. I think we should seriously talk about what we're going to do about the 80-column limit and how it impacts our work. It seems like people are open to expanding it to 120 columns or something, which would help a lot. Another alternative would be to remove `EFFECT_` from this enum so we just have `SOLID_FROM_BELOW` and `SOLID_` (because right now, the three instances of `EFFECT_` are eating 21 characters right away). We need to either expand the limit, shrink the names, or be okay with our code becoming ugly.

Anyway, if I should just run the formatter and call it good for now, let me know, otherwise we can keep this PR staged as a "once we resolve the column limit, then we put this one in".

### DIFF
--- a/include/game.h
+++ b/include/game.h
@@ -1227,7 +1227,7 @@ typedef struct {
 
 typedef struct {
     /* 80072BD0 */ Collider colliders[8];
-    /* 80072CF0 */ s32 D_80072CF0[14][9]; // probably also colliders? can't combine them though
+    /* 80072CF0 */ s32 D_80072CF0[14][9]; // Seems like another collider set
     /* 80072EE8 */ s32 padPressed;
     /* 80072EEC */ s32 padTapped;
     /* 80072EF0 */ s32 padHeld;
--- a/include/game.h
+++ b/include/game.h
@@ -1226,7 +1226,7 @@ typedef struct {
 } FgLayer; /* size=0x8 */
 
 typedef struct {
-    /* 80072BD0 */ s32 D_80072BD0[8][9];
+    /* 80072BD0 */ Collider colliders[8];
     /* 80072CF0 */ s32 D_80072CF0[14][9];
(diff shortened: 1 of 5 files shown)


sozud:
If you change this and reformat does it do what you want? https://github.com/Xeeynamo/sotn-decomp/blob/105ed4e4ef33b3889af352352581af1f51a2f560/.clang-format#L60

bismurphy:
> If you change this and reformat does it do what you want?
> 
> https://github.com/Xeeynamo/sotn-decomp/blob/105ed4e4ef33b3889af352352581af1f51a2f560/.clang-format#L60

Yes, changing that to 120 leaves my lines here untouched, but of course it also un-separates many, many other lines in the repo, as you can see here: 
image

sozud:
How about using this to turn off clang format?
```
// clang-format off
your code here
// clang-format on

```

bismurphy:
oops, i need to learn how to rebase properly, didn't expect all those commits...

PR #644, opened 2023-09-25
Decompile w_022 func_ptr_80170004

bismurphy:
The next least-complex one of these. This one is for the Magic Missile. Kind of neat to be able to recognize the behavior. This definitely seems like the logic for the missile itself, since it moves at a fixed X speed until it has a hit flag, at which point it changes its AnimationFrame, changes its `step` to 2, and then lets its animation run out and destroys itself.

### DIFF
--- a/src/weapon/w_022.c
+++ b/src/weapon/w_022.c
@@ -1,9 +1,50 @@
 #include "weapon_private.h"
 #include "shared.h"
 
+extern AnimationFrame D_9E000_8017A760[];
+extern AnimationFrame D_9E000_8017A800[];
+
 INCLUDE_ASM("weapon/nonmatchings/w_022", EntityWeaponAttack);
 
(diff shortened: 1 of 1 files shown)


src/weapon/w_022.c
```diff
@@ -1,9 +1,50 @@
...
 #include "shared.h"
 
+extern AnimationFrame D_9E000_8017A760[];
+extern AnimationFrame D_9E000_8017A800[];
+
 INCLUDE_ASM("weapon/nonmatchings/w_022", EntityWeaponAttack);
 
-INCLUDE_ASM("weapon/nonmatchings/w_022", func_ptr_80170004);
+void func_ptr_80170004(Entity* self) {
+    switch (self->step) { /* irregular */
```
sozud, line 10: We can remove the `/* irregular */`

sozud (approved):

PR #645, opened 2023-09-26
Add weapon names to w_0xx.c files

bismurphy:
The purpose of this is to make it easier to know what you're reading when you look at a w_0xx.c file.

I made a Python script which went through each file, got the `weaponID`, then looked through equipment.json for all weapons which have that ID. Then, for each one, if it had a `name_resolved` trait, we grabbed that name, and otherwise treated it as an Unknown (I believe these are all 'virtual weapons', things like the projectiles or particles created as a result of using a weapon, but which cannot be directly equipped). Then it prepended a summary of all the weapons it found. Let me know if the formatting for this should be changed, given that it's all generated by a script it should be easy to make tweaks if we want.


src/weapon/w_000.c
```diff
@@ -1,3 +1,8 @@
+// Weapon ID #0. Used by weapons:
+// Tyrfing, Namakura, Gladius, Scimitar, Cutlass, Saber, Falchion, Broadsword,
+// Bekatowa, Damascus sword, Hunter sword, Bastard sword, Talwar, Sword of
+// Hador, Luminus, Harper, Gram, Mormegil, Terminus Est, Dark Blade,
+// Mourneblade, Badelaire, Unknown
```
sozud, line 5: Would it be useful to make the Unknowns `Unknown#XX` or something with the equipment ID?

sozud (approved):

src/weapon/w_000.c
bismurphy: Good idea, done

PR #649, opened 2023-09-26
Decompile w_029 EntityWeaponShieldSpell

bismurphy:
Now that we fixed the `li` instruction, the other functions in this weapon file can come spilling out :)

There was a jump table that I changed to being rodata. Luckily that went smoothly.

Interestingly, this does the same %5 as the other function, but there are two spots where it uses g_Timer (the one that always runs) and one other where it uses g_GameTimer (the one that pauses when the game pauses). It's only palettes so it doesn't really matter, but it's neat.

One problem: This stores two bytes to self->7C and self->7D. But we already had a different weapon that used 7C as a s16. So we may end up needing to make different `ext` entries for the different weapons. For now, I'm leaving it generic.

### DIFF
--- a/src/weapon/w_029.c
+++ b/src/weapon/w_029.c
@@ -3,7 +3,9 @@
 #include "weapon_private.h"
 #include "shared.h"
 
+extern u16 D_CF000_8017ACF8[];
 extern u16 D_CF000_8017AD04[];
+extern AnimationFrame D_CF000_8017AD24[];
 
 INCLUDE_ASM("weapon/nonmatchings/w_029", EntityWeaponAttack);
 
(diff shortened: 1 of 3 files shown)


sozud:
@Xeeynamo are we still banning usage of `ext.generic`?

src/weapon/w_029.c
```diff
@@ -47,7 +49,151 @@ void func_ptr_80170014(Entity* self) {}
...
+    case 0:
+        self->primIndex = g_api.AllocPrimitives(PRIM_GT4, 1);
+        if (self->primIndex == -1) {
+            DestroyEntity(self);
+            return;
+        }
+        SetSpriteBank1(g_Animset);
+        if (g_HandId != 0) {
+            self->animSet = -0x7FEE;
+            self->ext.weapon.childPalette = 0x128;
```
sozud, line 68: I think it probably makes sense to use the weapon29 extension for everything in this function now that it exists

bismurphy: For now I'd like to keep weapon29 simplified. The biggest thing is that its only mismatch is right at the beginning of the struct. As we identify the other members of the overall weapon struct, it would be nice to not have to copy each one into weapon29. 7C/7D is the only oddball member of the struct so I think we can use the broader `weapon` struct to refer to the places where weapon29 behaves the same as a normal weapon. This is why I only defined the first two members for weapon29 in entity.h.

sozud: @Xeeynamo What's your preference here?

bismurphy: When I asked prior to updating this PR here: https://discord.com/channels/1079389589950705684/1087866009983139870/1156579697086439464 xeeynamo expressed no preference. Not trying to argue my case or anything, but if it's helpful for you, I think he already answered.

PR #650, opened 2023-09-28
Decompile w_029 EntityWeaponAttack

bismurphy:
Continuing to finish out w_029 now that we have g_GameTimer % 5 figured out.

There is a very unfortunate thing here. We are using `self->ext.weapon.lifetime`, but EntityWeaponShieldSpell just had a whole mess about 7C and 7D and needing to create ext.weapon29. Clearly, weapon29 references this location as both a s16, and as two u8's. I really don't know what to make of that. Happy to adjust things here if we want to handle that differently, but for now this seems to work. It will be cool if we can get this strange weapon fully decompiled.

### DIFF
--- a/src/weapon/w_029.c
+++ b/src/weapon/w_029.c
@@ -47,7 +47,7 @@ void EntityWeaponAttack(Entity* self) {
         unkAC_offset++;
     }
     switch (self->step) {
-    case 0:              
+    case 0:
         SetSpriteBank1(g_Animset);
         if (self->params & 0x8000) {
             self->animSet = -0x7FEE;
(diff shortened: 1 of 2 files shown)


src/weapon/w_029.c
```diff
@@ -2,12 +2,143 @@
...
+        self->animCurFrame = 0x3E;
+    }
+    if ((PLAYER.step == 2) && (PLAYER.step_s != 2)) {
+        unkAC_offset++;
+    }
+    switch (self->step) {
+    case 0:
+        SetSpriteBank1(g_Animset);
+        if (self->params & 0x8000) {
+            self->animSet = -0x7FEE;
```
sozud, line 53:
```
            self->animSet = 0x8012;
            self->ext.weapon.unk80 = 0x128;
            var_v0 = 0x66;
        } else {
            self->animSet = 0x8010;
```
We also have defines for these https://github.com/Xeeynamo/sotn-decomp/blob/bebb4789a139d322565206bd13ec930a8fcadadd/include/game.h#L191

sozud (approved):

src/weapon/w_029.c
bismurphy: Oh fantastic, thanks for informing me! I think I did it properly now but let me know if anything is still amiss.

PR #651, opened 2023-09-28
Decompile w_029 func_ptr_80170024

bismurphy:
This one was big and complicated. It rounds out w_029; there is no longer any INCLUDE_ASM in this, so that's cool.

I believe this function controls the shield rod spell with the fire shield, but I can't guarantee that. This required creating a new struct, which I called FireShieldDragon. I put it in the .c file since I don't expect anything else to use that struct so it doesn't need to be in a .h file.

Overall this function could probably use some study and cleanup, but for now it works and it's great to have the whole function decompiled. Nobody will ever have to deal with the `-O1` of this function ever again :)

### DIFF
--- a/src/weapon/w_029.c
+++ b/src/weapon/w_029.c
@@ -509,8 +509,8 @@ void func_ptr_80170024(Entity* self) {
     for (i = 0; i < 19; i++, prim = prim->next) {
         prim->clut =
             self->ext.weapon.childPalette + D_CF000_8017AD10[g_Timer / 2 % 10];
-        prim->x0 = (s16)prev_x2;
-        prim->y0 = (s16)prev_y2;
+        prim->x0 = prev_x2;
+        prim->y0 = prev_y2;
         prim->x1 = prev_x3;
         prim->y1 = prev_x4;
         var_s7++;
@@ -520,10 +520,10 @@ void func_ptr_80170024(Entity* self) {
         centerY = var_s7->posY.i.hi;
         angle1 = var_s7->unk8 - 0x400;
         angle2 = var_s7->unk8 + 0x400;
-        prev_x2 = prim->x2 = centerX + (((rcos((s32)angle1) >> 4) * size) >> 8);
-        prev_y2 = prim->y2 = centerY - (((rsin((s32)angle1) >> 4) * size) >> 8);
-        prim->x3 = prev_x3 = centerX + (((rcos((s32)angle2) >> 4) * size) >> 8);
-        prim->y3 = prev_x4 = centerY - (((rsin((s32)angle2) >> 4) * size) >> 8);
+        prev_x2 = prim->x2 = centerX + (((rcos(angle1) >> 4) * size) >> 8);
+        prev_y2 = prim->y2 = centerY - (((rsin(angle1) >> 4) * size) >> 8);
+        prim->x3 = prev_x3 = centerX + (((rcos(angle2) >> 4) * size) >> 8);
+        prim->y3 = prev_x4 = centerY - (((rsin(angle2) >> 4) * size) >> 8);
     }
 
     var_s7 = &D_CF000_8017C9A0[dragonNumber][0];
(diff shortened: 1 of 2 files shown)


src/weapon/w_029.c
```diff
@@ -326,7 +342,214 @@ void EntityWeaponShieldSpell(Entity* self) {
...
+        if (self->ext.weapon.unk9C > 0x100) {
+            self->ext.weapon.unk9C = 0x100;
+        }
+    }
+    firstPrim = prim = &g_PrimBuf[self->primIndex];
+    var_s7 = &D_CF000_8017C9A0[dragonNumber][0];
+    for (i = 0; i < 19; i++, prim = prim->next) {
+        prim->clut =
+            self->ext.weapon.childPalette + D_CF000_8017AD10[g_Timer / 2 % 10];
+        prim->x0 = (s16)prev_x2;
```
sozud, line 512: It looks like these casts can be removed.

src/weapon/w_029.c
```diff
@@ -326,7 +342,214 @@ void EntityWeaponShieldSpell(Entity* self) {
...
+        prim->x1 = prev_x3;
+        prim->y1 = prev_x4;
+        var_s7++;
+        size = var_s7->unk10 *
+               (rsin(var_s7->unkE) * var_s7->unk14 / 256 + 4096) / 4096;
+        centerX = var_s7->posX.i.hi;
+        centerY = var_s7->posY.i.hi;
+        angle1 = var_s7->unk8 - 0x400;
+        angle2 = var_s7->unk8 + 0x400;
+        prev_x2 = prim->x2 = centerX + (((rcos((s32)angle1) >> 4) * size) >> 8);
```
sozud, line 523: casts

src/weapon/w_029.c
```diff
@@ -326,7 +342,214 @@ void EntityWeaponShieldSpell(Entity* self) {
...
+    size = ((self->ext.weapon.unk9C * 54 / 256) *
+            (rsin(var_s7->unkE) * var_s7->unk14 / 256 + 4096)) /
+           4096;
+    nextX = (var_s7 + 1)->posX.i.hi;
+    nextX_copy = nextX;
+    nextY = (var_s7 + 1)->posY.i.hi;
+    nextY_copy = nextY;
+    angle1 = var_s7->unk8 - 0x199;
+    angle2 = var_s7->unk8 + 0x199;
+    firstPrim->x0 = nextX + (((rcos((s32)angle1) >> 4) * size) >> 8);
```
sozud, line 539: casts

src/weapon/w_029.c
```diff
@@ -326,7 +342,214 @@ void EntityWeaponShieldSpell(Entity* self) {
...
+    firstPrim->y0 = nextY - (((rsin((s32)angle1) >> 4) * size) >> 8);
+    firstPrim->x1 = nextX + (((rcos((s32)angle2) >> 4) * size) >> 8);
+    firstPrim->y1 = nextY - (((rsin((s32)angle2) >> 4) * size) >> 8);
+    angle1 = var_s7->unk8 + 0x999;
+    angle2 = var_s7->unk8 + 0x667;
+    firstPrim->x2 = nextX + (((rcos((s32)angle1) >> 4) * size) >> 8);
```
sozud, line 545: casts

sozud (approved):
Nice work, tried to NOP it out to check but I don't think `80170024` is where this is loaded in ram.

bismurphy:
> Nice work, tried to NOP it out to check but I don't think `80170024` is where this is loaded in ram.

Yes, this is the function which is pointed to by the pointer which is at 80170024. The Weapon files get swapped in and out similarly to the Overlay in g_api. Take a look at `include/weapon.h`.

sozud:
I figured this out, the right way is:
Weapon2 starts at `0x8017D000`. This function pointer is offset to 0x24. Look at the value at 0x8017D024, which is 0x8017EE78. That is the start of the function, which can be NOP-ed out. The red sphere remains but the dragons disappear.

PR #652, opened 2023-09-29
Decompile DRA CheckGravityBootsInput

bismurphy:
This is pretty cool. So there are only a few places left in the game with non-decompiled calls to IsRelicActive. I picked one of them at random and started going through it and it turns out it's the gravity boots check. This function acts as a state machine where state 0 is at rest, and pressing down initiates the transition to state 1. At state 1 it waits for an UP input, and if it gets it, moves to state 2 to wait for an X input. At each state, there is a timer limiting your time to put in the next button input.

Now, we have these two variables that work together (the state, which tracks how many correct buttons have been entered, and the timer), and they exist in an existing struct which had two values in it. I believe the array of this struct holds all the button counters and timers for all the different spells in the game. However, with my recent decompilation of the Wing Smash, we see that the Wing Smash is handled with its own logic. I think most of the other spells will use this same array to keep track of their overall button counters and timers.

When this function is called in func_80111830 (At the start of this file: https://github.com/Xeeynamo/sotn-decomp/blob/master/src/dra/71830.c), it's in a for loop which goes through several different checks and calls different functions. I'm betting these are all the spell button inputs. And none are decompiled. So I will start going through those one by one. Should be fun!

I had to change some existing accesses of the g_ButtonState which were set up before we really understood what that was, so let me know if you have any thoughts on those. I'm excited to be able to get the rest of these in this set decompiled!


src/dra/75F54.c
```diff
@@ -2911,7 +2911,8 @@ void func_8012CB4C(void) {
 void func_8012CC30(s32 arg0) {
     if (arg0 == 0) {
         D_80138444 = 1;
-        if (D_80138FC0[1].x == 255 && IsRelicActive(RELIC_SKILL_OF_WOLF) &&
+        if (g_buttonCombo[1].buttons_correct == (u8)(-1) &&
```
sozud, line 2914: It seems like each entry represents a spell, so we could do `g_buttonCombo[COMBO_WOLF_CHARGE].buttons_correct`

src/dra/6D59C.c
```diff
@@ -1370,7 +1370,47 @@ s16 func_80110394(void) {
     return 0;
 }
 
-INCLUDE_ASM("dra/nonmatchings/6D59C", func_801104D0);
+bool CheckGravityBootsInput(void) {
+    switch (g_buttonCombo[0].buttons_correct) {
```
sozud, line 1374: `g_buttonCombo[COMBO_GRAVITY_BOOTS].buttons_correct` etc

src/dra/6D59C.c
```diff
@@ -748,7 +748,7 @@ bool func_8010EDB8(void) {
 block_32:
     equipped_id = g_Status.equipment[hand];
     equipped_item = &g_EquipDefs[g_Status.equipment[hand]];
-    if (D_80138FC0[1].x != 0xFF) {
+    if (g_buttonCombo[1].buttons_correct != 0xFF) {
```
sozud, line 751: `g_buttonCombo[COMBO_WOLF_CHARGE].buttons_correct`

sozud (approved):

src/dra/75F54.c
bismurphy: Just to be clear, you'd like me to create a new enum to hold all the button combos?

sozud: Yeah I was thinking that would make sense to help document

PR #655, opened 2023-09-30
Decompile DRA CheckBackForwardInput

bismurphy:
Next one in the sequence.

This, like the previous one, needs a variable to keep track of what "forward" means from the start of the input sequence. It is right after the last one, but unfortunately they don't work as an array, so I just placed a 2 at the end, with a comment, instead of making the name overly long. Hope that's okay. Otherwise no further comments, hopefully since we're mostly just using the structs I already established this and further functions will come quickly :)


src/dra/6D59C.c
```diff
@@ -1478,7 +1478,56 @@ bool CheckQuarterCircleForwardInput(void) {
...
+            g_ButtonCombo[COMBO_BF].timer = 15;
+            g_ButtonCombo[COMBO_BF].buttonsCorrect++;
+        } else if (--g_ButtonCombo[COMBO_BF].timer == 0) {
+            g_ButtonCombo[COMBO_BF].buttonsCorrect = 0;
+        }
+        break;
+    case 1:
+        if (directionsPressed == forward) {
+            g_ButtonCombo[COMBO_BF].timer = 15;
+            g_ButtonCombo[COMBO_BF].buttonsCorrect = 0xFF;
```
sozud, line 1511: I'm thinking we should define 0xFF as `COMBO_COMPLETE` or similar

sozud (approved):

src/dra/6D59C.c
bismurphy: Added one, I just put it in this file for now since this file seems to host everything relating to those functions but we can move it to dra.h, either if you want it now, or if we find external uses?

sozud: There's also a usage here https://github.com/Xeeynamo/sotn-decomp/blob/141c8b22c726f8dec8db8863adc529263cb2c7a7/src/dra/75F54.c#L2914

Probably dra.h makes the most sense?

bismurphy: Good find, adding to dra.h alongside the ButtonComboState definition, seems best to group things that are related in usage

PR #660, opened 2023-10-03
Rewrite Saturn make rules

Xeeynamo:
## What

Enhance the existing `Makefile` to build new rules for the Saturn side of the decomp. This should allow more flexibility when adding new overlays or when tuning existing rules.

## Changes

I separated part of the Saturn build process in the separate file `Makefile.saturn.mk`. I realise that naming it `saturn.mk` would have been enough, but I pre-pended `Makefile.` so it can be found right below the main `Makefile` when listing files in an alphabetic order. I plan to do the same with the psx and psp toolchain, therefore you will find `include Makefile.*.mk` in the main `Makefile`.

I deleted all the game building process done with Docker. Now that we have an established way to do it natively I think it is no longer required. We can always run the entire buildchain within a Docker container instead of having `_native` and `_docker`. Now all the `_native` references are removed. `build_saturn_native` is now `build_saturn`.

`check_saturn` is no longer responsible of stripping the ELF into a binary. That is now part of `build_saturn`.

I removed the references to `_li.o` (e.g. `alucard_li.o`) and used `.elf` instead, which is closer to how the PSX build chain works. If `_li.o` was a better preference, please let me know.

I am no longer using `./compile_dosemu.sh`. Instead I am using the new `$(DOSEMU)` to directly invoke the tool within the Makefile. I have done that to reduce the amount of dependent files.

I tried minimising duplication as much as possible. We now have a list of overlays found in `SATURN_OVL_TARGETS`. Each expected output triggers a series of dependencies so seamlessly build everything. `Makefile` is smart enough to call `$(SATURN_TOOLCHAIN)` only once. If the game was already built but just one source or symbol file changed, triggering a new `build_saturn` will only compile the modified overlay.

The Saturn ADPCM files are now extracted in `assets/` instead of `build/`. I think `assets/` should contain all the uncompressed and uncooked files. The list of PCM file is not hardcoded. Instead I am now using `$(wildcard disks/saturn/SD/*.PCM)`. This now means the tool tries to convert PCMs from `SDD0.PCM` to `SDF0.PCM` with no success. As the tool is silently failing I decided to leave it as I wrote it.

## Problems

I rewrote everything thinking about concurrency in mind. But `make -j build_saturn` gives some unexpected output on `stdout`. I did not dig too much into it. I suspect it might be dosemu. This is not a stopper as we were not using `-j` when building the game anyway.

I also noticed doing `VERSION=saturn make build` calls `mipsel-linux-gnu-ld` for `stage_02`. I suspect it is calling the rule `$(MAIN_TARGET).elf: $(MAIN_O_FILES)` and simply moving `include Makefile.*.mk` above it should fix it. But then it would cause the same problem if I split the PSX rules into their own separate file. We never used `make build` by setting the env variable `VERSION`, so this is not either a breaking change or a stopper.

## Post thoughts

I am happy with what I achieved so far. I used the knowledge I accumulated when maintaining the PSX counterpart. Since I now better understand how `make` works, I was able to make some better decisions in the Saturn counterpart. For example triggering a new build when the symbol list changes is something the PSX build chain lacks of. I think in the future it would be nice to trigger `make extract` when either the YAML or the symbol list changes.

### DIFF
--- a/Makefile.saturn.mk
+++ b/Makefile.saturn.mk
@@ -116,8 +116,3 @@ $(SATURN_SPLITTER_APP):
 assets/saturn/SD/%.wav: disks/saturn/SD/%.PCM $(SATURN_SPLITTER_APP)
 	mkdir -p $(ASSET_DIR)/saturn/SD
 	$(SATURN_ADPCM_EXTRACT_APP) $< $@
-
-# TODO: Remove me. Temporary backward compatibility with the CI
-build_saturn_native: build_saturn
-check_saturn_native: check_saturn
-build_saturn_toolchain_native: extract_disk_saturn


sozud (approved):
Thanks for looking into this, seems to work OK for me locally. I think there's still more cleanup that could be done but this is a big improvement.

I think the docker stuff here can be removed:

https://github.com/Xeeynamo/sotn-decomp/blob/07c03548e501e2d68858b441e927f4f9c245c41a/Makefile#L460

I think strip.sh and the dockerfiles can be removed
https://github.com/Xeeynamo/sotn-decomp/tree/master/tools/saturn_toolchain

sozud:
Seems to be working OK for me locally, thanks for doing this cleanup

PR #662, opened 2023-10-04
Enhance make extract on PSX

Xeeynamo:
Previously if either the YAML or the symbol list were changed, we were forced to either manually delete the `asm` folder of the specific overlay or to invoke `make clean`, where the latter forced us to re-extract all the overlays via `make -j extract`.

That changes with this PR. How `make` works is `target: dep1 dep2` where if one of the dependencies has the last modified date greater than the target, the rule is triggered again. Previously we were extracting overlays doing `make extract_stcen`. But since every overlay extraction generates a linker script I now changed the rules to do `make build/us/stcen.ld` to extract the same overlay. As I explained above, if either the YAML or one of the related symbols changes, the linker modified date will be older and the extraction for that overlay to be trigger again.

The above allowed me to stop polluting the repo root with the linker scripts as they are now moved into `build/us` or `build/hd` depending of what you're trying to build. If you do `make extract` twice in a row you will be welcomed with a `make: Nothing to be done for 'extract'.`.

There are still some instances where you **might** need `make clean` beforehand, especially when modifying the `segment` section in the YAML or renaming the symbols. A `rm -rf asm/$(VERSION)/$*` can help but I want to see if the current solution will be enough.

### Please try removing `make clean` from your workflow, once this is merged, to quickly detect possible problems with this new approach.

I am also planning to make more substantial changes on our build-chain like this PR or #660 in the future. I am aiming to small incremental changes over time in case I break someone's flow or detect design flaws. I am also considering the breaking changes introduced in make 4.4, which will most likely be included in Ubuntu 24.04 LTS. I set March 2024 as a deadline to finish all the new build-chain work.

### DIFF
--- a/tools/decompile.py
+++ b/tools/decompile.py
@@ -71,7 +71,7 @@ def get_c_context(src_file) -> str:
     return m2ctx.import_c_file(src_file)
 
 
-def decompile(func: NonMatchingFunc, ctx_str: str):
+def run_m2c(func: NonMatchingFunc, ctx_str: str):
     with tempfile.NamedTemporaryFile(
         mode="w", encoding="utf-8", suffix=".c"
     ) as tmp_ctx:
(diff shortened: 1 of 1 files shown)


sozud (approved):
I did a basic check of the workflow locally and it seems ok

PR #663, opened 2023-10-06
SEL import data/rodata (part 1)

Xeeynamo:
I started importing the data in SEL also as a proof of concept we can also import the SOTN-specific formatted strings. For that I integrated the new tool `tools/sotn-str.py` into the build chain. It works like the following:

* `./tools/sotn-str.py parse disks/us/ST/SEL/SEL.BIN 27598`: read the binary content in a specific offset and parse it as a readable Unicode string.
* `cat src/st/sel/2C048.c | ./tools/sotn-str.py process` converts the Unicode string into a byte sequence that can be understood by the SOTN engine.

There is a dirty hack I did to make this work:
```c
#ifndef SOTN_STR
#define _S(x) (x) // Strings processed by tools/sotn-str.py
#endif
```
When invoking `mipsel-linux-gnu-cpp` I am passing -DSOTN_STR, effectively ignoring `_S` in the pre-processor. Then I am passing it into `sotn-str` before `iconv` to resolve any string surrounded by `_S`.
I am a bit of afraid this might be annoying when using the permuter. I  tried doing `$(SOTN_STR) -f $< | $(CPP)` but apparently `cpp` does not work well with content passed via stdin.


src/st/sel/2C048.c
```diff
@@ -35,24 +149,17 @@ void SetupFileChoose(void) {
...
-        prim->x0 = D_80180068[i] + 0x68;
-        prim->y0 = D_80180088[i] + 0x58 + ypos;
-        prim->u0 = D_801800A8[i];
-        prim->v0 = D_801800C8[i];
-        prim->u1 = D_801800E8[i];
-        prim->v1 = D_80180108[i];
+
+    for (i = 0; i < NUM_MENU_UNK_084; i++) {
+        prim->x0 = g_MenuUnk084X[i] + 104;
+        prim->y0 = g_MenuUnk084Y[i] + 0x58 + ypos;
```
sozud, line 158: Should this one also be decimal?

include/game.h
```diff
@@ -192,6 +192,9 @@ typedef struct Primitive {
 #define ANIMSET_DRA(x) (x)
 #define ANIMSET_OVL(x) ((x) | ANIMSET_OVL_FLAG)
 
+#ifndef SOTN_STR
+#define _S(x) (x) // Strings processed by tools/sotn-str.py
```
sozud, line 196: I guess this is long enough since if you grep for `_S(` you will only get these, but if you just search for `_S` there's a lot of hits. Was wondering if it should be a little longer. It's probably fine as-is

sozud (approved):

include/game.h
bismurphy: I'm not totally clear on what this does since i don't know enough about the build chain. It might be nice if there was a writeup about what exactly sotn-str.py does (with an example going from binary to Unicode to the SOTN byte sequence), and why the _S( is necessary since it seems like it's just returning x unchanged. I don't know the details of what is happening here and overall it would probably be nice to have a reference available since I'm sure I'm not the last person who will be confused about what's happening here.

bismurphy: Yes, I think so. I think the important thing is that the _S is meaningless in terms of C code, and we're using it as a hack, to be a signal flag that goes to a custom python script that we're inserting in the build chain. Is that a correct understanding?

PR #666, opened 2023-10-07
Extract g_EnemyDefs with new system

bismurphy:
Now that assets.py exists, it's a whole lot easier to extract assets!

In order to make g_EnemyDefs get added to assets/dra/enemydefs.json, only a few changes are needed.

1) Realign the bounds of the splat to add a separate line for the end of g_EnemyDefs. g_EnemyDefs is exactly 400 elements in size. We have a new data segment at 0xC780.

2) Change the yaml to extract it as "assets", named enemydefs.

3) Previously we would output name_resolved and desc_resolved to the .json assets; since g_EnemyDefs don't have a description, we now move that to an if statement that first checks if it exists in the config file.

4) Create the config file for enemydefs!

Overall this creates a very smooth workflow for extracting more assets. The biggest limitation we have with this approach is that the config file is selected based on the name in the third space in the yaml (for example, `[0x4B04, assets, equipment]` loads `tools/splat_ext/equipment_config.json` by name). Therefore entries like `layoutobj` which is used many times in `splat.us.stwrp.yaml` are not possible to extract using this method, since there are many of them which use the same config. I'll see if I can work out a solution to this. Ultimately we're using the same `self.name` property to control both what config file gets loaded, and what file we output to in `assets`, but there's no reason those need to be the same parameter. I just need to figure out how to pass an extra variable from the splat into the parsing script (`assets.py`). I'll work it out somehow at some point, and then we'll be even more flexible. For now, it's nice to be able to browse all the enemies in the game!


bismurphy:
Ah, looks like splat does have support for passing extra arguments, I'll figure out how to add support for that in my next PR :)

sozud:
Could you post the output to a gist or whatever?

bismurphy:
Here is a gist of `assets/dra/enemydefs.json` : https://gist.github.com/bismurphy/7be584f6f48c9869f4f9d43265a587e6

bismurphy:
I could add an ability for the json to specify each of the bit for those members which are bits. Let's hold off on merging this, and I'll add an implementation for that to this repo.

bismurphy:
Okay, new version has support for bitfields. Here is an example of what ends up in assets/dra/enemydefs.json, for the Guardian enemy:
```
{
        "id": 396,
        "id_hex": "18C",
        "ram_addr": "800AC6E0",
        "name_resolved": "Guardian",
        "name_addr": "0x800E05D8",
        "hitPoints": 500,
        "attack": 50,
        "attackElement": [
            "field_att0",
            "HIT"
        ],
        "defense": 34,
        "hitboxState": 35,
        "weaknesses": [],
        "strengths": [
            "CUT"
        ],
        "immunes": [
            "ICE",
            "THUNDER",
            "FIRE"
        ],
        "absorbs": [
            "DARK"
        ],
        "level": 60,
        "exp": 1500,
        "rareItemId": 321,
        "uncommonItemId": 255,
        "rareItemDropRate": 2,
        "uncommonItemDropRate": 1,
        "hitboxWidth": 6,
        "hitboxHeight": 24,
        "flags": [
            "flag4",
            "flag10",
            "flag12",
            "flag13",
            "flag22",
            "flag27"
        ]
    },
```
The {something}_config.json files are more complicated now, and I wish it was easier to de-duplicate all the elements, but this should work for now. The json is definitely more useful now that the bit fields are in plaintext.

PR #672, opened 2023-10-09
Decompile DRA CheckSwordBrothersInput

bismurphy:
The last one of this set!

Before merging, we should work out how to deal with D_8006CBC4 since right now we don't have a good way to enum it.

Oddly this didn't use rodata for the switch even though the others did. It does use rodata for the FntPrint strings though.

Otherwise I think those are all the highlights.


bismurphy:
Made a solution with commenting to hopefully make it clear why this is needed. Happy to adjust naming if needed but I think this does the job.

PR #675, opened 2023-10-10
Add jumptable resolver to decompile.py

bismurphy:
This is some code I wrote a while ago and I think it might make sense to add it to `decompile.py`.

This function takes in an assembly file, and will scan through it for any jump tables. It will then find them in the asm folder, and move the jumptable information into the assembly file for the function. This means that users do not have to seek out rodata and paste it into the asm file manually.

As an example, the user can now run `dec EntityAlucard` and it will successfully decompile the function and insert it to the .c file, because now when it passes the assembly to m2c, the rodata is there at the bottom of the assembly function. Overall this has been a big convenience for me and I hope it will be helpful for others.

### DIFF
--- a/tools/decompile.py
+++ b/tools/decompile.py
@@ -6,6 +6,8 @@ import os
 import subprocess
 import tempfile
 from contextlib import redirect_stdout
+from pathlib import Path
+import re
 from enum import Enum
 import m2ctx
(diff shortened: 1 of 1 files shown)


sozud (approved):
Haven't tested this but the code seems reasonable to me

PR #688, opened 2023-10-16
DRA InitWeapon matching

Xeeynamo:


src/dra/5298C.c
```diff
@@ -2844,7 +2844,55 @@ bool LoadWeaponPrg(s32 equipIndex) {
...
+    // 0xE0-0xEF: weapon0 (left hand) entities
+    // 0xF0-0xFF: weapon1 (right hand) entities
+    entity = g_Entities;
+    for (i = 0; i < STAGE_ENTITY_START; i++) {
+        entityId = entity->entityId;
+        if (entityId >= itemSlot * 0x10 + 0xE0 &&
+            entityId <= itemSlot * 0x10 + 0xEF) {
+            DestroyEntity(entity);
+        }
+        if (entityId >= 0xE8 && entityId < 0xEE) {
```
sozud, line 2887: #defines for these ranges?

sozud (approved):

src/dra/5298C.c
sozud: If we don't really know or there's not a good solution the change isn't necessary. Was just noting that 


    // 0xE0-0xEF: weapon0 (left hand) entities
    // 0xF0-0xFF: weapon1 (right hand) entities

isn't encoded into our types anywhere as far as I know

PR #689, opened 2023-10-17
DRA DrawMenu

Xeeynamo:
Decompiles the function responsible of drawing all the menu. I am still not sure to understand how it works.

Thank you very much to @bismurphy's original scratch, which helped me getting this match.

EDIT: The `HD` version is matching, but to the missing rodata import I still cannot proceed to import it. I need `func_800FBC24` decompiled before importing the `HD` rodata.

### DIFF
--- a/src/dra/5298C.c
+++ b/src/dra/5298C.c
@@ -2597,7 +2597,7 @@ void DrawMenu(void) {
 
             cursorXMovement = cw / 2 * (128 - D_800A2D80[j]) / 128;
             cx += cursorXMovement;
-            cw -= cursorXMovement << 1;
+            cw -= cursorXMovement * 2;
         }
         if (menu->unk1C == 3) {
             menu->unk1D++;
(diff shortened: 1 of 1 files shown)


src/dra/5298C.c
```diff
@@ -2518,7 +2518,248 @@ void DrawConsumableCount(s32 itemId, s32 hand, MenuContext* ctx) {
...
+        if (menu->unk1C == 3) {
+            menu->unk1D++;
+            j = menu->unk1D; // FAKE?
+            if (menu->unk1D == 16) {
+                menu->unk1C = 0;
+            } else {
+                cursorXMovement =
+                    (cw / 2 * (128 - D_800A2D80[menu->unk1D])) / 128;
+                cx += cursorXMovement;
+                cw -= cursorXMovement << 1;
```
sozud, line 2611: I'm thinking `* 2` is more natural since it's coordinates

sozud (approved):
Great work

PR #690, opened 2023-10-17
Symbol clean-up

Xeeynamo:
I am not sure why I thought it was a good idea to populate the symbol table with so much useless stuff. It has been like this since the [first commit](https://github.com/Xeeynamo/sotn-decomp/commit/fcf1f077#diff-ad26526355e8bad3a3db379e8b039623e45c8620831de04ac21b76b9da6b7f4e), probably indicating some level of inexperience on how decompilation and linker works.

This frees some symbols from hardcoded addresses, allowing more relocation flexibility now that DRA.BIN is getting relatively close to completion.


sozud (approved):
I thought these were supposed to be getting automatically removed by the script?

PR #692, opened 2023-10-18
HD DRA aligns decompilation with US DRA

Xeeynamo:
Importing the rodata from HD took me a good amount of experimentation and trial&error. With the past PRs I was able to decompile all the functions required to import the `D_800A2D68` array and the jump table of #689, #688 and #687, which it had to be done all at the same time.

I also took the opportunity of decompiling the HD version of #445 , which actually appeared to be simpler than the US counterpart. I tried removing some fakes from the US function but without luck. But I managed to un-fake #447 by keep importing the `rodata` and by using `D_800DC6EC`.

There are now five functions left in 5298C, including a very big one. With HD aligned to US, finishing this source file will be a walk in the park.


sozud (approved):
D_800A2D68 looks like a custom string format but the number of bytes doesn't match the number of Japanese characters so I don't know what is going on. e.g if
```
"\x0B\x2F\x11\xFF", // c_strSystem
```
is supposed to be `システム` that's 3 bytes + the 0xff terminator rather than 4.

PR #693, opened 2023-10-18
DRA MenuEquipHandlePageScroll

Xeeynamo:
Decompiles what I believe is the function that handles the page scrolling via R1 and L1. It uses `MenuContext.unk16` but I am still not sure exactly what it does, therefore I left it undocumented for the time being.


src/dra/5298C.c
```diff
@@ -3145,7 +3145,58 @@ void func_800FA3C4(s32 arg0, s32 arg1, s32 arg2) {
...
+    const int Unk16 = 72;
+    s32 nItems;
+    s32 limit;
+    s32* cursorIndex;
+    MenuContext* menu = &g_MenuData.menus[3];
+
+    if (D_801375CC.equipTypeFilter == 0) {
+        cursorIndex = &g_MenuNavigation.cursorEquipHand;
+    } else {
+        cursorIndex = &(&g_MenuNavigation.cursorEquipHead)[D_801375D4];
```
sozud, line 3159: This gives the impression that cursorEquipHead is an array?

sozud (approved):

src/dra/5298C.c
```diff
@@ -3170,9 +3170,9 @@ void func_800FA3C4(s32 arg0, s32 arg1, s32 arg2) {
 
     // Here is where we use the menuitem, again, FAKE.
     if (D_801375CC.equipTypeFilter == 0) {
-        g_MenuNavigation.scrollEquipHand = menuitem[11];
+        g_MenuNavigation.scrollEquipType[HAND_TYPE] = menuitem[11];
     } else {
-        g_MenuNavigation.scrollEquipAccessories[D_801375D4] = menuitem[11];
```
sozud, line 3175: My guess is that this is split into hand and body like the equips https://github.com/Xeeynamo/sotn-decomp/blob/c443cdb1eee9053b6202cb8e60a4f3ff124a40b2/include/game.h#L775

That would explain the reason to use an if statement and the filter

PR #694, opened 2023-10-18
DRA MenuHandleCursorInput

Xeeynamo:
I had this decompiled for a while but I have been waiting for all the previous pile of functions to be decompiled before affording to propose this function here.

Very similar to #693 , this one seems to be more generic and used in multiple places, so I named it `MenuHandleCursorInput`.

I found a few fields from `MenuData` to be part of a third `MenuContext`. I cannot extract that `menus[3]` out of that struct or it will fail to align (@bismurphy  explained why this happens, but I do not remember where to pick the conversation).


bismurphy:
The conversation about the MenuData is here: https://discord.com/channels/1079389589950705684/1079395108501331990/1127396390927073451

The issue is that a MenuContext has a size of 0x1E, or 30 bytes. This does not fit in a "clean" number of bytes on a 32-bit system (you end up using 7.5 4-byte words), so when you try to make them an array, it makes the array use 8 bytes for each (and just ignore the last 2). By packing them into a struct, it will properly load each one to be directly continuous.

I am not certain about any of this, but this is my best understanding of what is happening and why it doesn't work if they are in a "loose" array rather than packaged into a struct.

src/dra/5298C.c
```diff
@@ -3132,40 +3133,159 @@ void func_800F9F40(void) {
...
+    prevCursor = nav->cursorMain;
+    switch (arg2) {
+    case 3:
+        if (g_pads[0].repeat & PAD_UP) {
+            if (nav->cursorMain) {
+                nav->cursorMain--;
+            }
+        }
+        if (g_pads[0].repeat & PAD_DOWN) {
+            if (nav->cursorMain != arg1 - 1) {
```
sozud, line 3151: rename arg1 to numItems or similar?

sozud (approved):

PR #696, opened 2023-10-22
Decompile DRA func_8011BBE0

bismurphy:
Not sure what this one is. The entity is ID 24. This comes right after EntityHolyWaterFlame in the entity updater functions list. I tried setting a breakpoint and playing around with subweapons for a bit but couldn't get the breakpoint to trigger.

I'll keep trying to play with this to figure out what it is, but for now we can at least check that my code is good.


src/dra/75F54.c
```diff
@@ -1848,7 +1848,38 @@ void func_8011B530(Entity* entity) {
 
 INCLUDE_ASM("dra/nonmatchings/75F54", func_8011B5A4);
 
-INCLUDE_ASM("dra/nonmatchings/75F54", func_8011BBE0);
+void func_8011BBE0(Entity* self) {
```
sozud, line 1851: Rename to EntityUnkId24?

sozud (approved):

PR #704, opened 2023-10-25
Decompile DRA func_8011E4BC

bismurphy:
No idea what this one is. It's very complicated though (5 switches!) and I'm amazed it matches.

It requires a few new unknown structs. There's a bizarre thing that gets loaded from g_PrimBuf which is not a primitive but acts like one in some ways so I called it FakePrim for now.

Overall I have a ton of confusion about this one and what the actual structures are so if anyone recognizes these patterns, I'd love to update this PR and make it make sense.

The only thing I've noticed from emulator testing is that this entity gets called when you are in a Wing Smash and you run into a sloped surface which redirects your flight.


include/unkstruct.h
```diff
@@ -227,4 +227,51 @@ typedef struct {
...
+    s16 unk10;
+    s16 unk12;
+    s16 unk14;
+    s16 unk16;
+    s16 unk18;
+    s16 unk1A;
+    s16 unk1C;
+} unkStr_8011E4BC;
+
+typedef struct FakePrim {
```
sozud, line 252: @Xeeynamo What's your take on this?

sozud (approved):

PR #708, opened 2023-10-28
Fix source code line numbers

Xeeynamo:
Fixes a very annoying issue it has been bothering me for a while. Since the introduction of the sotn-str tool in #663 the line output has been wrong and I failed to realise it immediately. This caused issues like `differ` reporting the wrong C line number for the correspondent assembly output, wrong debug info and so on.


sozud (approved):
The difference is not adding newlines?

PR #710, opened 2023-10-30
Split out data table for func_8011AC3C

bismurphy:
`func_8011AC3C` appears to be a very important function when it comes to spawning entities. From what I can tell, it seems to be responsible for creating all of the entities listed in `asm/us/dra/data/CF4C.data.s` from 800AD0C4 through  800AD1D0, which is the game's main entity updating function lookup table.

When func_8011AC3C is called, it uses its `self.params` to index into a table which contains spawning information for entities. For example, if func_8011AC3C is called with `self.params` of 0x21, it will create a child entity with entityID of 25, which is the entity for the Hellfire spell (which I recently decompiled and will submit as a PR soon).

I think func_8011AC3C is going to be useful for understanding entity spawning logic, so I have gone ahead and taken the data table that it uses at D_800AD1D4, and created it as an asset which will now be extracted by splat. This way, we can more directly view the data, rather than just seeing it as a giant array of u8 values. I am hoping that having this available as json will be helpful for understanding the entity creation logic. Overall, this PR is meant as a stepping stone to fully understanding func_8011AC3C and what it does to create the entities that it creates.

Doing this has helped me understand one struct member's purpose, which is to set the ID of the child entity, so I've added that to the struct, both the entity extension for func_8011AC3C and the struct which this is an array of, which for now is Unkstruct_800AD1D4.


tools/splat_ext/AC3CTable_config.json
```diff
@@ -0,0 +1,11 @@
+{
+    "struct": {
+        "childId":"u8",
```
sozud, line 3: I don't know if this system supports it but it seems to me that it should be extracted with the bitfields separated out.

e.g. something like

```
u8 childId;
u8 unk1;
u8 unk2_3f
u8 unk2_7;
u8 unk2_6;
etc.
```

bismurphy: Agreed, that is not currently supported; I tried doing it for a bit but the code got messy pretty fast, I'll think about how to implement that.

sozud: Can it be done by hand the old way for this particular instance? I'm thinking a system that's general is going to be hard to make. We could do it by hand for stuff with bitfields and use the general system for the easy cases.

bismurphy: I don't think the old way ever had support for bitfields, but I could be wrong about that. Do we have any existing assets which use bitfields, in the old way? 

The new system has support for entries that are entirely made of bitfields, but not for hybrid bytes which use one or two bits as fields and the rest as an int. But I don't think the old way ever had bitfields, it was always only ints (at least, as far as I know).

sozud: Our other splat extensions are able to do this as far as I understand? https://github.com/Xeeynamo/sotn-decomp/blob/9924fbf4e17ae5f482f0832ae4d0930f320351bd/tools/splat_ext/layers.py#L180

bismurphy: Got it, that example is super helpful, I think I can make something work.

bismurphy: Okay, I think I got something that does what we're looking for, here is an example of one of the entries in the generated .json file:
```
{
        "id": 49,
        "id_hex": "31",
        "ram_addr": "800AD2FA",
        "childId": 5,
        "unk1": 56,
        "unk2": {
            "bit7": "TRUE",
            "bit6": "TRUE",
            "bottom6": 1
        },
        "unk3": 3,
        "unk4": {
            "topHalf": 2,
            "bottomHalf": 8
        },
        "unk5": 0
    },
```

sozud: I was thinking a general system like this would work but I realized this only makes sense on big-endian platforms. In that case you could just have a sequence of `u` or `s` with the number of bits:
```
        "childId":"u8",
        "unk1":"u8",
        "unk2_7":"u1",
        "unk2_6":"u1",
        "unk2_3f":"u6"
```
And have a bitreader class to grab the number of bits specified and increment the current bit read position. In little-endian this is harder to conceptualize.

bismurphy: Haha, looks like we commented at about the same time :)

Indeed I made a general system, which describes bits in a big-endian-like manner (I believe endian usually only means the arrangement of bytes in multi-byte values, rather than arrangement of bits within a byte, but whatever). I figured that made the most sense since usually if bits are being stuffed into a number, those bits will be in the most-significant positions. I think the approach I have here should be flexible and expandable, and of course we can adjust in the future, but for now it works well in this case without being excessively tailored to this specific application.

sozud (approved):
The new version looks good to me, thinking @Xeeynamo should give this one a look before merging.

tools/splat_ext/AC3CTable_config.json
bismurphy: The issue with that is that in the "struct" field, we need to give a type for each of the struct members, which goes to the switch statements in assets.py (in get_serializer and get_parser_and_size). If we allow specifying bit portions at that top-level struct, then we would need the Python assets.py to have cases to handle every potential bit field structure. The way I'm doing it here, all the special cases are defined in the .json file, where the baseline struct just matches the C struct, and then the "fields" allows you to elaborate on anything which has deeper more complex structure of bits.

It also just becomes a bit of a mess when parsing the json - when building (going from the extracted assets to the compiled game), we iterate over the members of the assets file, and then generate bytes which are output to the serializer. Once a value is serialized to bytes, we can output it. But if we have a flat structure like you mention, then we can't just process "unk4_0" and then on the next iteration process "unk4_4": 8, because the bits for those entries need to be combined to share a byte. Under the way I have it now, "unk4" is a single byte, meaning it gets processed in a single iteration, because the serializer can run an internal loop to reassemble the byte from the specified parts. If we had a flat structure, then the serializer would need to have a lot more internal logic where every time it serializes a struct member, it has to check whether it needs to backtrack to the previous byte and keep modifying it, where right now once a byte is generated, we are done with it.

Overall while flattening the bit structure would have some aesthetic benefits, I think it would introduce a lot more complexity and backtracking in the code, rather than having the bit structure explicitly broken out in the assets json.

bismurphy: That's a good point, modders may want to modify the json in an automated fashion.
I think for now this system works as-is, and if it needs to have new capability in the future, we can figure out exactly how to implement that capability (which may include flattening or any other type of data restructuring) to suit those future needs once they arise.

PR #713, opened 2023-10-31
Fix offset for dra/704D0.c

bismurphy:
Not sure if this was a typo or what, but this file was at the wrong address. Because the first function was already decompiled, it didn't matter that the first 8 bytes of assembly were missing. Anyway, this fixes that.

Running `make force_extract` would previously generate a `asm/us/dra/nonmatchings/704D8/func_801104D8.s` which had 2 missing instructions which caused the start of the function to be really weird. With this change, the assembly file will be generated properly with all instructions present.

I imagine this could have been a headache for someone in the future but this will fix that possibility.


sozud:
So the problem was that some of CheckGravityBootsInput was truncated? Looks like this file was added here. https://github.com/Xeeynamo/sotn-decomp/commit/6806f5a0a5fe1066a4cd738f64227bc2fbbffd45#diff-db461468fcc07b8f1fe9642a0f0ae0af1953352fe6332aec15d7655495269806

bismurphy:
I still don't understand why github gives some PR comments a "Reply" box, and doesn't give it to others, but I don't have a direct way to reply to @Xeeynamo's request... Anyway, yes, this should be fine to merge any time

PR #735, opened 2023-11-06
DRA func_80134F50

sozud:
Co-authored-by: @sonicdcer


sozud:
I was planning to do another documentation followup so I'll merge as-is.

PR #736, opened 2023-11-06
Extract and document all functions from main.exe

Xeeynamo:
This took me a while. I used a [throw-away script](https://gist.github.com/Xeeynamo/58da1ff8f3831d0ba5d23da27cbca025) to help me with the quest, but I still had to manually check every single YAML subsegment.

I transported over what we were able to decompile from main. For a few functions I added the signatures and documented existing DRA code, especially on the sound department (got inspired by the recent @sozud PRs).

All the `.text` part from `main.exe` is now completely extracted as C/ASM where appropiate. Almost all the functions have their original PSY-Q names but 7 of them: `func_80012DBC`, `func_80012F84`, `func_80017008`, `func_80017078`, `func_8001929C`, `func_80021F0C` and `func_800286E0`. I did not feel confident enough to rename them, so I left them be. The rest of the functions I am 99% sure they are all accurate.

I am now excluding the [asm/](https://github.com/Xeeynamo/sotn-decomp/tree/master/asm/us/main/psxsdk) folder from the repo. It was useless.

This research confirms me the game uses the PSY-Q 3.5 libraries, with the exception of `LIBGPU.LIB`, which is from PSY-Q 3.0 for some unknown reason.

EDIT: `make format` was not taking care of duplicated symbols in our symbol list. To speed-up my work, all the duplicate symbols (duplicates =  name AND offset) are now removed.


sozud (approved):
Excellent work, I was thinking we needed libspu added to continue documenting sound stuff. It seems like the function calls script is getting stuck but other than that looks good.

```
 glabel func_82000_8017AAE0
Failed to return a value
```

PR #757, opened 2023-11-09
Decompile DRA func_8010E570

bismurphy:
This is one I first looked at many months ago, finally I got a match.

That being said, no idea what it does.

This also includes what I think is an innocent unrelated change to func_8010E168; if I'm being honest, I just forgot to remove that change from my local repo. I can push a new commit to remove it if desired but for now I'll leave it since it's an improvement and not worth its own whole PR.

### DIFF
--- a/src/dra/6D59C.c
+++ b/src/dra/6D59C.c
@@ -460,7 +460,7 @@ void func_8010E570(s32 arg0) {
     case 3:
     case 4:
         var_a0 = 6;
-        if (ABS(PLAYER.velocityX) > 0x28000) {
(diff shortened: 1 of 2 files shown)


bismurphy:
Will fix HD in a few hours, happy to take any other change requests in the meantime as well

sozud (approved):
I'd revert the func_8010E168 change unless we're ready to actually document the ID and flags but it's not a big deal.

bismurphy:
Maybe I will pick pl_vram_flag as my next target to deal with then, to enhance our understanding. It is certainly strange.

PR #759, opened 2023-11-10
Refactoring of things adjacent to entity factory

bismurphy:
With the decompiling of the entity factory framework being started, I went through and replaced some placeholders that have accumulated over time. Originally I was going to try to decompile the whole sequence of events that happens when you cast Hellfire, but I think the core of that is in EntityAlucard, since we set a player step.

Overall, this should have no functional changes and should all just be renaming things to improve based on new knowledge.

### DIFF
--- a/src/dra/6D59C.c
+++ b/src/dra/6D59C.c
@@ -294,7 +294,9 @@ void func_8010E0D0(s32 arg0) {
 }
 void func_8010E168(s32 arg0, s16 arg1) {
     if (arg0 == 0) {
-        CreateEntFactoryFromEntity(g_CurrentEntity, 0x15002C, 0);
+        // Create factory with unkA0 = 0x1500, blueprint #44.
+        // Blueprint 44 is to make child entity #11, or EntityPlayerBlinkWhite
+        CreateEntFactoryFromEntity(g_CurrentEntity, 0x1500 << 8 | 44, 0);
         if (arg1 >= g_Player.D_80072F1A) {
             g_Player.D_80072F1A = arg1;
         }
@@ -1147,7 +1149,7 @@ void func_8010FAF4(void) {
     g_Player.unk46 = 0;
 }
 
-void func_8010FB24(void) {
+void PerformHellfire(void) {
     PLAYER.velocityY = 0;
     PLAYER.velocityX = 0;
     SetPlayerStep(Player_SpellHellfire);
--- a/src/dra/7A4D0.c
+++ b/src/dra/7A4D0.c
@@ -160,48 +160,49 @@ void func_8011A9D8(void) {
     }
 }
 
-Entity* CreateEntFactoryFromEntity(Entity* source, u32 flags, s32 arg2) {
-    Entity* entity;
-    s16 temp = arg2;
+Entity* CreateEntFactoryFromEntity(Entity* source, u32 flags, s32 arg2_raw) {
+    Entity* newFactory;
+    // Weird thing needed for callers to match
+    s16 arg2 = arg2_raw;
 
-    entity = GetFreeDraEntity(8, 16);
-    if (entity == NULL) {
+    newFactory = GetFreeDraEntity(8, 16);
+    if (newFactory == NULL) {
         return NULL;
     }
 
-    DestroyEntity(entity);
-    entity->entityId = E_UNK_1;
-    entity->ext.generic.unk8C.entityPtr = source;
-    entity->posX.val = source->posX.val;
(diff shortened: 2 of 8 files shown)


src/dra/6D59C.c
```diff
@@ -294,7 +294,9 @@ void func_8010E0D0(s32 arg0) {
 }
 void func_8010E168(s32 arg0, s16 arg1) {
     if (arg0 == 0) {
-        CreateEntFactoryFromEntity(g_CurrentEntity, 0x15002C, 0);
+        // Create factory with unkA0 = 0x1500, blueprint #44.
+        // Blueprint 44 is to make child entity #11, or EntityPlayerBlinkWhite
+        CreateEntFactoryFromEntity(g_CurrentEntity, 0x1500 << 8 | 44, 0);
```
sozud, line 299: An enum for these? Not sure what the best term is. `ENTITY_FACTORY_PLAYER_BLINK_WHITE`? This applies to the other usages

src/dra/7A4D0.c
```diff
@@ -160,48 +160,49 @@ void func_8011A9D8(void) {
...
     }
-    entity->ext.generic.unkA0 = (flags >> 8) & 0xFF00;
-    entity->ext.generic.unk92 = temp;
+    newFactory->ext.entFactory.unkA0 = (flags >> 8) & 0xFF00;
+    newFactory->ext.entFactory.unk92 = arg2;
     if (source->flags & FLAG_UNK_10000) {
-        entity->flags |= FLAG_UNK_10000;
+        newFactory->flags |= FLAG_UNK_10000;
     }
     if (flags & 0x1000) {
```
sozud, line 192: `FACTORY_FLAG_UNK_10000` and document the other spots?

sozud (approved):

src/dra/6D59C.c
bismurphy: I'm inclined to hold off until we can determine more of how the factory structure works, particularly that second argument, before starting to make enums and such. I think I'd like to keep decompiling things and writing descriptive comments of where things happen, and start figuring out what the different things are, before making enums, especially since that seems like a pretty long enum name which will end up making things spill onto extra lines.

src/dra/7A4D0.c
bismurphy: It looks like the 10000 is used in EntityEntFactory to just pass it down to its child entity so probably isn't worth having its own separate enum since the 10000 doesn't affect the behavior of the factory itself.

Ultimately every time an entity factory is used, there are 3 relevant entities. The one here called the "source", the factory itself, and the entity (or maybe entities) created by the factory.

In creating the factory here, we're saying if the source has 0x10000, give the factory 0x10000 as well, which is just to make it so that the children will also have 0x10000. That makes the child inherit the source's 0x10000 status, and just uses the factory as a go-between. No reason to treat it as different for the factory than for other entities.

src/dra/75F54.c
```diff
@@ -174,7 +177,9 @@ void func_80116408(void) {
     case 4:
         if (PLAYER.animFrameIdx == 10 && PLAYER.animFrameDuration == 1) {
             g_Player.D_80072F18 = 4;
-            CreateEntFactoryFromEntity(g_CurrentEntity, 0x25U, 0);
+            // Make factory with blueprint 37. This creates entity with ID 28,
+            // which is EntityExpandingCircle.
+            CreateEntFactoryFromEntity(g_CurrentEntity, 37, 0);
```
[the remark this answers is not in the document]
bismurphy: The issue is that the blueprint number doesn't match the Entity ID. Blueprint 37 creates entity 28, so even if we populate 28 being EntityExpandingCircle, that wouldn't connect to the factory blueprint. We could consider making a new enum for the blueprint numbers and call it BlueprintIDs, but it would be distinct from EntityIDs.

src/dra/6D59C.c
bismurphy: This is a nice idea, I will put something together in the next few days and update this PR :)

bismurphy:
Yes, I went back and forth on whether to shift by 8 and have everything be 0x##00 or shift by 16 and have it be 0x##, the reason I decided to do it this way is that `entity->ext.generic.unkA0 = (flags >> 8) & 0xFF00;` does assign the 00 into unkA0, and so I think if those 00's are going to end up in unkA0 either way, they should also be explicitly written in the function call - calling with 0x33 and having unkA0 wind up being 0x3300 seems like a source of unnecessary confusion. That was my thought process but I'm not attached to it.

PR #761, opened 2023-11-11
DRA LoadFileSimToMem

Xeeynamo:
I managed to decompile a function I was stuck months ago. I left a comment above the function name to explain what the function is and why it exists. The rest of the code and changes should be self-explanatory enough.

I renamed `DEBUG_PTR` to `SIM_PTR` because `0x80280000` is only used on devkits as a memory area to load files from the simulator device. The name "DEBUG" was a bit too vague.


src/dra/47BB8.c
```diff
@@ -2,6 +2,8 @@
 #include "objects.h"
 #include "sfx.h"
 
+extern const char aPqes_1[]; // TODO: import rodata
```
sozud, line 5: Just FYI I'm pretty sure this is SEQ data. `0x53455170` is `SEQp`

sozud (approved):

PR #762, opened 2023-11-11
DRA VSyncHandler

Xeeynamo:
While the purpose of the function is clear, how it impacts the flow of the game is unclear to me. Since a few variables are coupled with the `LoadFileSim` function, I expect this callback effectively does nothing.

The name `VSyncHandler` comes from one of the PSY-Q SDK samples I found. Other names were `cbvsync`, which is a bit too similar to the callee `VSyncCallback` for my tastes. None of the samples from the SDK even remotely look like this function I decompiled.

This is the last function from the `47BB8.c` file that had to be matched, and the last blocker to keep importing data and rodata.


src/dra/47BB8.c
```diff
@@ -1,17 +1,57 @@
...
+        break;
+    }
+    D_800A04F0 = 0;
+}
 
 void func_800E7D08(void) {
     s32 i;
 
-    for (i = 0; i < 1024; i++) {
+    for (i = 0; i < VSYNC_UNK_LEN; i++) {
```
sozud, line 54: I'm not seeing a connection between D_800A04FC and this. How do we know that the 1024s were originally the same symbol?

sozud (approved):

PR #765, opened 2023-11-11
DRA func_80133290 WIP

sozud:
I can't seem to figure out how to get this one to match locally, it seems like something changes elsewhere in the binary.

I changed the signature of func_80132E38, which is used by these functions:

func_801341B4
func_80134388
func_801353A0

None of those seem to be the issue. I tried looking with vbindiff and wasn't able to figure out much. Putting this up in case anyone can see what's going on. Scratch: https://decomp.me/scratch/bgP55


sozud:
Thanks, forgot about the file splits. I updated the PR, but it's still not working. Is there a better way to figure this out than trying to eyeball it with vbindiff?

sozud:
Thanks for the help and explanation

PR #777, opened 2023-11-16
Decompile DRA EntityHellfireHandler

bismurphy:
Decompiled this one a little while ago, but spent some time on entity factory stuff before putting it in the repo, so here it is.

Not sure why the string in rodata isn't working, but I couldn't get it to work, maybe someone else can fix it.


src/dra/86ECC.c
```diff
@@ -0,0 +1,840 @@
...
+        LoadImage(&D_800B0790, &D_800B0728);
+        prim = &g_PrimBuf[self->primIndex];
+        prim->v0 = prim->v1 = prim->v2 = prim->v3 = 0xF8;
+        prim->u0 = prim->u2 = 1;
+        prim->u1 = prim->u3 = 0x30;
+        prim->b3 = 0x80;
+        prim->tpage = 0x11C;
+        prim->blendMode = 0x331;
+        prim->priority = self->zPriority = 0x1C0;
+        PlaySfx(0x636);
```
sozud, line 47: This should be the warp sound effect NA_SE_PL_WARP

sozud (approved):

PR #778, opened 2023-11-17
Extract XA music configs

sozud:
This is the data needed for the game to play XA music and voice tracks. The extracted data for the US version is here https://gist.github.com/sozud/09358ed163daaf87d1e28e897f07416d


sozud:
Going to merge since Xeeynamo is AFK, if there's a problem will do a follow-up.

PR #782, opened 2023-11-17
Decompile TT_000 func_80172C30

bismurphy:
After decompiling the bat fireball, I wanted to decompile all the places where it can be created, so I did this one, which contains the entity factory creation function call for the bat familiar to shoot its own fireball.

Kind of neat to learn that the player's own fireballs and the familiar's fireballs appear to be identical since they use the same function and everything.

This required a bit of shuffling of other code in other places; happy to elaborate on any of those changes but they seem fairly benign.


src/servant/tt_000/10E8.c
```diff
@@ -490,7 +491,128 @@ void func_80171ED4(s32 arg0) {
...
+
+    g_api.func_8011A3AC(self, 0, 0, &D_80174C30);
+    if (D_80174D3C != 0) {
+        self->zPriority = PLAYER.zPriority - 2;
+    }
+    switch (self->step) {
+    case 0:
+        func_801719E0(self);
+        if (self->ext.fam.unk82 == 0) {
+            func_8017160C(*(&D_80170660 + ((D_80174C30.unk0 / 10) * 5)), 0xD2);
```
sozud, line 516: I think this is `func_8017160C(D_80170660[(D_80174C30.unk0 / 10) * 5], 0xD2);` https://decomp.me/scratch/zxlKC 
The *5 may indicate a struct, also this overlaps with some other data.

sozud (approved):

src/servant/tt_000/10E8.c
bismurphy: Thanks! I was able to take the existing material and rearrange it a bit. We now have it as a 2D array of s32's. It might still be a struct, but for now the way we're indexing it seems to work. I brought this function up to match the others that touch that array, and moved the start of the array forward now that we see it should start at 0660.

sozud: So actually, the resulting array length is not a multiple of 5. I think it actually starts earlier than what you have. If you look at this: https://decomp.me/scratch/1n0Qq

```
                    if (*(&D_80170658 + ((D_80174C30.unk0 / 10) * 0x14)) < (s16) temp_a0_2) {

...
        temp_v0_11 = func_80173F74((s16) temp_v0_10, arg0->ext.unkAC3C.unk86, *(&D_8017065C + ((D_80174C30.unk0 / 10) * 0x14)));
```

We don't have to get this 100% right now, just pointing it out.

bismurphy: Yeah we will probably need to keep shifting stuff around. I think generally the data should match what has been decompiled - so once the function you linked is decompiled, the array can be adjusted in order to capture the full extent of the data. In general it seems that TT_000 is a bit messy, I think cleaning it up should be its own effort at some point. I don't want to have a PR for one function end up shifting too much to capture too much other stuff beyond the function's scope. Looks like there are only 3 functions left in TT_000, so might make sense for someone to finish those off and then make that whole overlay look as nice as possible.

bismurphy: Actually... I'm already changing stuff, it looks clear that the start of the array should be at 80170658, let me try shifting it to capture all the data and if it's quick and easy I'll add it to this PR.

bismurphy: I can't seem to get it to work. I make the table start at D_80170658, but the assembly for func_80172120 wants to reference D_8017065C (which is part of D_80170658). Then I get errors compiling with `undefined reference to D_8017065C` so I'm not sure how to address that. I think let's ignore that table for now and treat it as something to be resolved in parallel with func_80172120.

PR #791, opened 2023-11-19
TT_000 func_80172120

Xeeynamo:
Second last function of TT_000 decompiled.

* Add `printf '\x00' | dd of=$@ bs=1 seek=40959 count=1 conv=notrunc` to ensure the output binary is 40960 bytes long (useful once BSS is migrated)
* Add `$(BUILD_DIR)/tt_%.ld` so every time the splat YAML changes, doing `make tt_000` triggers the extract again
* All the variables from BSS are now in `10E8` and in order. Removing `extern` from will properly create the BSS section
* Fix 7A4D0.c to not use the familiar entity extension
* Delete the fake symbols `D_80170660` and  `D_8017065C`
* Replace all the instances of `ext.generic` as `ext.bat` when possible

NOTES:
* For `SquareRoot12((...) << 0xC) >> 0xC;` I was not able to use the macro `FLT`
*


include/entity.h
```diff
@@ -220,31 +220,31 @@ typedef struct {
 } ET_Merman_2;
 
 typedef struct {
-    /* 0x7C */ s32 unk7C;
+    /* 0x7C */ s16 unk7C;
+    /* 0x7E */ s16 unk7E;
     /* 0x80 */ s16 unk80;
     /* 0x82 */ s16 unk82;
     /* 0x84 */ s16 unk84;
-    /* 0x86 */ s16 unk86;
```
sozud, line 227: unk86 is located at 0x84?

include/entity.h
```diff
@@ -220,31 +220,31 @@ typedef struct {
...
-    /* 0xA2 */ s16 unkA2;
-    /* 0xA4 */ s16 unkA4;
-    /* 0xA6 */ s16 unkA6;
+    /* 0x9C */ s16 unk9C;
+    /* 0x9E */ s16 unk9E;
+    /* 0xA0 */ s32 unkA0;
+    /* 0xA4 */ struct Entity* target;
     /* 0xA8 */ s16 unkA8;
-    /* 0xAA */ s16 unkAA;
+    /* 0xA8 */ s16 unkAA;
```
sozud, line 241: unkAA is at 0xA8?

sozud (approved):

PR #794, opened 2023-11-20
Decompile DRA EntityGiantSpinningCross

bismurphy:
This was a weird one. This uses the Geometry Transformation Engine, which I don't think any previous function has done. It was confusing to get it working but it seems to work.

I think I put the macros in the right place, but let me know if not.

Overall this is the first PR of this type, so please do make sure everything looks good so we don't pollute the repo. If this goes in, I might try to find more of the gte functions so we can crack those all away. Making this work was difficult and required a lot of SDK research, so I'm hoping I can do more functions before the knowledge disappears from my brain :)


src/dra/8D3E8.c
```diff
@@ -871,8 +871,197 @@ void func_8013136C(Entity* entity) {
...
+    u8* primUVCoords;
+    SVECTOR* temp_a3;
+    SVECTOR** vectors_ptr;
+    u16 priority;
+
+    sp48 = D_800E2024;
+    sp50 = D_800E202C;
+
+    if (self->step == 0) {
+        self->primIndex = g_api_func_800EDB58(4, 0x2E);
```
sozud, line 895: g_api.func_800EDB58?

sozud (approved):
Great work. Seems good to me, will wait for @Xeeynamo's approval

bismurphy:
I don't actually understand what most of these operations are doing so I'll trust your naming choices. Thanks for going through these!

PR #806, opened 2023-11-23
MAD func_8019bed4

Xeeynamo:
Follow-up to #797 by decompiling the MAD counterpart and share all the functions within the same file. I renamed the function as `UnkPrimHelper` as I do not know what it does.


sozud:
Looks like this needs to be added back `EntityUnkId17 = 0x801A2E20;`

sozud (approved):

PR #807, opened 2023-11-24
Remove non-PSXSDK specific code

Xeeynamo:
As discussed on Discord. I did not touch `cdc.c` and `libcd.c` much as I did not know long-term how we plan to extract some platform-specific logic out of there.


sozud:
Is the difference between pc/pc.c and pc/sdl2.c supposed to abstract out sdl2 and allow other pc backends like directx etc?

sozud (approved):

PR #812, opened 2023-12-02
Decompile DRA HandleSaveMenu

bismurphy:
This appears to generate the menu that appears when you use a save room. This is obvious by the strings present in the many calls to `func_800F9D88`.

Appears to be closely related to some other functions which may make the logic a bit more obvious. Cool to have this working though, might be useful for some types of research into how saving works. I wonder if all these messages have actually showed up before for real players.

Many of the strings feature Shift-JIS characters (specifically, the periods, question marks, and the 0, 1, and 2 numerals). Big thank you to Sozud for helping me understand those weird bytes in the strings!

Parsing rodata for strings is pretty fun!

Happy to change the name of this function, but wanted to at least replace the numerical placeholder.


sozud (approved):
Great work

PR #814, opened 2023-12-03
Decompile DRA func_80103ED4

bismurphy:
Handles a bunch of the internal logic of saving the game. There is a small group of functions that deal with the process of saving; once we get them all decompiled it makes sense to rename them once we understand the individual steps involved.


src/dra/63ED4.c
```diff
@@ -0,0 +1,367 @@
...
+            i = 1;
+            if (GetMemcardFreeBlockCount(D_80097924) == 0) {
+                D_80137E4C = 10;
+                return;
+            }
+        }
+        // I believe the rand() call here selects the icon on the save
+        // in the save-select screen.
+        StoreSaveData(g_Pix, D_8006C378, rand() & 0xF);
+        if (MemcardWriteFile(D_80097924, 0, saveFile, g_Pix, 1, i) != 0) {
```
sozud, line 104: I'd suggest renaming i to shouldCreate or similar but I guess the variable gets reused. Does it still match if the i in case 4 is another var?
```
s32 MemcardWriteFile(
    s32 nPort, s32 nCard, const char* name, void* data, s32 flags, s32 create)
```

src/dra/63ED4.c
```diff
@@ -0,0 +1,367 @@
+#include "dra.h"
+#include "objects.h"
+#include "sfx.h"
+
+// Note: Including this as part of the previous file (62D70.c) caused
```
sozud, line 5: @Xeeynamo does this indicate a new file?

sozud (approved):

PR #819, opened 2023-12-07
Rearrange some files to remove padding

bismurphy:
As mentioned in the previous PR, this removes a few rodata_padding variables by simply un-splitting the files involved here.

There are two more paddings I found in other places; I will look into those next but don't want to do too much file reshuffling in one PR. This PR takes 3 files and merges them into one, removing the padding in the process.


sozud:
Is there an explanation for these cases?

https://github.com/Xeeynamo/sotn-decomp/blob/77987d7639517b8bc0380f1bf83fd083343c9789/src/dra/5087C.c#L3027
https://github.com/Xeeynamo/sotn-decomp/blob/77987d7639517b8bc0380f1bf83fd083343c9789/src/dra/5087C.c#L3922C6-L3922C6

bismurphy:
Okay, I have done some testing.

Any const rodata that inserts zeros in the middle of a .c file is meaningless. They can be removed entirely without changing the compiled code.

Any time that rodata is at the end of a .c file, it means that the .c file should be combined with the file that comes immediately after it.

Note, this ONLY applies in files where the relevant rodata (before and after the 00 00 00 00) has all been imported into the decompiled code. For example, in src/ric/26C84.c, there is a `const u32 rodataPadding_801569E4 = 0;`, but after that is additional INCLUDE_ASM functions, which will have rodata in the future. Once that future data is pulled into the code, the rodata padding will no longer be needed.

I am going to search all .c files for the string "rodata" and remove all of them that are currently possible, and update this PR with those changes. This will mean the only remaining ones in the repo are the ones that are still necessary (due to rodata that is not yet being parsed into the C).

bismurphy:
Okay, I think that's about all the ones I can solve for now, the remaining ones need to stay because they're coupled to functions that are not decompiled yet.

sozud:
```
make: *** No rule to make target 'build/pc/src/dra/gameover.o', needed by 'build/pc/sound'.  Stop.
```
Would you mind adjusting the makefile here? https://github.com/Xeeynamo/sotn-decomp/blob/80e9af288ea88617d44662cc9e9e9f66d43abae2/Makefile.pc.mk#L13

sozud:
make: *** No rule to make target 'build/pc/src/dra/46FD4.o', needed by 'build/pc/sound'.  Stop.

PR #824, opened 2023-12-09
Decompile RIC func_801590A0

bismurphy:
This looks like one connected to controlling Richter's movement. Not certain exactly which moves, but we'll work that out in the future.


src/ric/1AC60.c
```diff
@@ -342,7 +342,131 @@ void func_80158FA4(void) {
...
 
-INCLUDE_ASM("asm/us/ric/nonmatchings/1AC60", func_801590A0);
+void func_801590A0(void) {
+    if ((D_80154570 == 0) && (PLAYER.velocityY < FIX(-1)) &&
+        !(g_Player.unk44 & 0x40) && !(g_Player.padPressed & PAD_CROSS)) {
+        PLAYER.velocityY = FIX(-1);
+    }
+    if ((g_Player.pl_vram_flag & 2) && (PLAYER.velocityY < FIX(-1))) {
+        PLAYER.velocityY = FIX(-0.25);
+        g_Player.unk44 |= 0x20;
```
sozud, line 352: unk44 seems to be bit flags, could start towards documenting it, but not essential

src/ric/1AC60.c
```diff
@@ -342,7 +342,131 @@ void func_80158FA4(void) {
...
+            } else {
+                g_Player.unk44 &= ~0x10;
+            }
+        }
+        if (PLAYER.animFrameDuration < 0) {
+            g_Player.unk46 = 0;
+            PLAYER.step_s = 0;
+            func_8015C920(D_80155528);
+        }
+        return;
```
sozud, line 467: is this return required to match?

sozud (approved):

src/ric/1AC60.c
bismurphy: Agreed, I think that was actually why I started decompiling this function originally. I found `g_Player.unk44 & 1` seems to be the flag for being able to do a dive-kick and wanted to find any other uses to finish documenting it. So maybe I'll look for more uses of that variable to decompile.

src/ric/1AC60.c
bismurphy: No, good catch, thank you!

PR #825, opened 2023-12-10
Decompile DRA func_800EA720

cram0:
If there's a better place to put the macros let me know,

The linter doesn't like it when the following macros are formatted like this : 

`#define GET_RED(x) ((x) & RED_MASK)`

as it changes them to this : 

`#define GET_RED(x) ((x)&RED_MASK)`

(removes the spaces around the `&` operator)

It's probably not configured for it as it's not happening to the `|` operator.


sozud:
Nice work!

PR #826, opened 2023-12-10
Rename some entity functions

bismurphy:
After finding a copy of GetFreeDraEntity in RIC, I decided to rename the function (in both DRA and RIC) to GetFreeEntity. Similarly, the function right after it is GetFreeEntityReverse (since its logic runs in reverse).

I also found CreateEntFactoryFromEntity in RIC, so I renamed that to match the version in DRA. We will need to go through and use the FACTORY macro for its uses, but I'm not going to tackle that quite yet.

Seems there is a lot of work to be done making RIC catch up to DRA, we'll see if I keep working on that moving forward, it's neat to find the ways the two match.


bismurphy:
(working on the build issue now)

PR #836, opened 2023-12-12
Decompile DRA func_80133FCC

cram0:
Should be good to go ?


sozud:
I haven't checked this locally but I think it's more like this https://decomp.me/scratch/zc3Px

cram0:
> I haven't checked this locally but I think it's more like this https://decomp.me/scratch/zc3Px

It is matching with the mentionned variable, thank you !

sozud:
From what I remember D_800BD240 and D_800BD228 aren't real variables, they are offsets within g_XaMusicConfigs.

src/pc/stubs.c
```diff
@@ -119,6 +119,9 @@ s32 D_800987B4;
 u8 g_PadsRepeatTimer[BUTTON_COUNT * PAD_COUNT];
 s32 D_80136410;
 s32 D_80136414[];
+s32 D_8013B65C;
+s32 D_800BD228;
```
sozud, line 123:
```
g_XaMusicConfigs = 0x800BD224;
```
So use unk228 for D_800BD228 and  240 is probably the next element, by my count the struct is 16 bytes. So 0x800BD224+16+12 == g_XaMusicConfigs[pos + 1].unk22f?

cram0: What I needed to do was to add 1 to the position to finally get it to match with the field `unk230` from that same structure, thanks !

sozud (approved):
Thanks for fixing the g_XaMusicConfigs thing

PR #838, opened 2023-12-12
Decompile RIC func_8015F680

bismurphy:
As I've been saying on Discord, this is one of a set of functions that does a bunch of things with colliders.

This required some slight restructuring of existing variables. We can finally confirm that the second array in g_Player is a set of colliders, so that's nice.

Will add collider enums for all those different bitmasks momentarily.


src/dra/86ECC.c
```diff
@@ -1411,15 +1411,15 @@ void func_8012D28C(bool exitEarly) {
     // Start a routine where we look through this array for a value.
     bitNotFound = 0;
     for (i = 3; i < 7; i++) {
-        if ((g_Player.D_80072CF0[i][0] & 2)) {
+        if ((g_Player.colliders2[i].effects & 2)) {
```
sozud, line 1414: EFFECT_UNK_0002?

src/ric/22380.c
```diff
@@ -37,7 +37,81 @@ INCLUDE_ASM("asm/us/ric/nonmatchings/22380", func_8015EE28);
...
+
+    u16* yPosPtr = &PLAYER.posY.i.hi;
+    u16* xPosPtr = &PLAYER.posX.i.hi;
+    s32* vram_ptr = &g_Player.pl_vram_flag;
+
+    if (D_80097418 != 0) {
+        return;
+    }
+    temp_s0 =
+        g_Player.unk04 & (EFFECT_UNK_8000 | EFFECT_UNK_4000 | EFFECT_UNK_0800 |
```
sozud, line 55: How do we know that unk04 is these flags? I only see two places where it is set and there's no obvious connection to collider effects https://github.com/Xeeynamo/sotn-decomp/blob/b851d5e4d2a5db8b1c4a31dabe0bd3020b35c6c6/src/dra/692E8.c#L109
https://github.com/Xeeynamo/sotn-decomp/blob/b851d5e4d2a5db8b1c4a31dabe0bd3020b35c6c6/src/ric/1AC60.c#L122

bismurphy: I don't have any real evidence that it's collider flags, but it seemed like a very similar bitmask to what was being used on the collider (ie, if it was `g_Player.unk04 & 0x4021` I wouldn't have made this substitution) so I thought it was close enough to guess for now that it might be collider flags. Can revert to `0xCC03` if preferred.

sozud (approved):

PR #841, opened 2023-12-13
Menu refactor and portability

Xeeynamo:
Quite big PR that aims to document most of `menu.c` and make the code more portable.

I had to declare `u16* func_80106A28(u32 arg0, u16 kind);` on the top of `main.c`. Without it the compiler assumes it returns an `int`, which is not big enough for 64-bit pointers. It took me a while to realise that was the cause of the crash. We will have more of them in the future.

EDIT: Ignore the commit `******* FIX`. It was meant to be squashed for a clearer commit history but I forgot to do it.


src/dra/menu.c
```diff
@@ -3453,15 +3464,15 @@ void HandleMenu(void) {
...
                 PlaySfx(SE_UI_CONFIRM);
-                func_800FAB8C(9);
+                MenuHide(MENU_DG_CFG_BUTTONS);
                 g_MenuStep = MENU_STEP_SYSTEM;
             } else {
             block_117:
                 func_800F9808(2);
 #if defined(VERSION_US)
-                func_800F99B8("Assign the function to each button．", 2, 0);
+                ShowText("Assign the function to each button．", 2);
```
sozud, line 3473: If we are doing the "Menu" namespace thing should these be MenuShowText?

sozud (approved):

PR #848, opened 2023-12-15
Decompile dupe function rwrp func_801A2550

ser-pounce:
Hopefully straightfoward

### DIFF
--- a/src/st/rwrp/14590.c
+++ b/src/st/rwrp/14590.c
@@ -116,7 +116,16 @@ INCLUDE_ASM("asm/us/st/rwrp/nonmatchings/14590", BottomCornerText);
 
 INCLUDE_ASM("asm/us/st/rwrp/nonmatchings/14590", func_801955D8);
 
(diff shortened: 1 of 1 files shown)


sozud (approved):
Looks good, if you want to go further we have a process to work towards de-duplication where the function and data gets matching names across all the overlays and it gets pulled into an include. Here's an example https://github.com/Xeeynamo/sotn-decomp/pull/459/files

ser-pounce:
I can give it a shot, I suppoe the function needs a name? The `p3` member seems to be (ab)used to store state flags, so the function is something like `getNextXxxPrim` where `Xxx` is whatever is represented by `p3` being 0.

sozud:
I guess reading it what comes to mind is `FindPolyWithP3` or something like that. The name isn't super important and can be documented in more detail as we progress. Also I can merge this and if you want to do the deduplication as a follow up  (or not do it) that's fine.

ser-pounce:
Sounds good! I'll get to work on the refactor.

PR #850, opened 2023-12-15
Add sprite rendering

Xeeynamo:
As per title.

I created a new folder called `assetx/`, that is not cleaned with `make clean` and that contains assets that are not extracted from the game and that we want to keep in the repo. I am not sure if we agree on this, so I will be happy to remove it from the PR if it is an unwanted approach.

`MySetDrawMode` is implemented and it sets the texture to render. The problem is that currently there is not a batch system in place, meaning `MyRenderPrimitives` will operate on the latest texture set. This works as I am using just one texture, but it will not work long-term. The PS1 uses something called Ordering Table (our `g_CurrentOt`) that contains a list of ordered drawing calls that will be submitted to the GPU via `DrawOTag(g_CurrentOT)`. A later PR will come to properly handle this.

The PS1 has 32 texture pages (or `tpages`, or `tp`). I am thinking to store each tpage into 128x128 textures. I am not sure how it will work long-term or if it would be better to just create a single 1024x512 texture with the entire vram in it. When the tpage is 0x1E the font PNG is loaded.


assetx/font8.png
sozud, line 1: We might want to include a note of where this came from

src/pc/sdl2.c
```diff
@@ -46,9 +66,52 @@ void ResetPlatform() {
...
+               textureIndex);
+        if (g_SdlVramTextures[textureIndex]) {
+            SDL_DestroyTexture(g_SdlVramTextures[textureIndex]);
+        }
+        if (g_SdlVramSurfaces[textureIndex]) {
+            SDL_FreeSurface(g_SdlVramSurfaces[textureIndex]);
+        }
+    }
+
+    switch (textureIndex) {
```
sozud, line 108: I'm not sure how this approach is going to scale but I think it's fine for now

sozud (approved):
I think we need to install SDL_image in the CI to fix it

assetx/font8.png
sozud: If this enables further progress I don't see an issue with a placeholder solution. We can continue iterating on it

sozud (approved):
LGTM

PR #851, opened 2023-12-15
Load assets json

Xeeynamo:
Uses a small JSON parsing library called cJSON (MIT license) to parse the assets `equipment.json` and `accessory.json` and load them in memory rather than hard coding it or pre-building it.


sozud (approved):
I'm kind of wondering if we should use c++ to get access to libraries with more ergonomic APIs, but there's other tradeoffs with that like worse compile times.

PR #855, opened 2023-12-17
Decompile SEL func_801B3F94 + refactor

ser-pounce:
Started off as a dedupe, but DRA func_80102E04 and SEL func_801B3F94 ended up being slightly different. Also refactored some symbols.


src/st/sel/sel.h
```diff
@@ -168,9 +168,9 @@ extern s32 D_801BAFDC;
 extern s32 D_801BAFE0;
 extern u32 D_801BAFE4;
 extern s32 D_801BAFE8;
-extern s32 g_memCardRStep;
-extern s32 g_memCardRStepSub; // rstep sub
-extern s32 g_memCardRetryCount;
```
sozud, line 173: The capitalization of Card doesn't match the others here

src/st/sel/sel.h
```diff
@@ -168,9 +168,9 @@ extern s32 D_801BAFDC;
 extern s32 D_801BAFE0;
 extern u32 D_801BAFE4;
 extern s32 D_801BAFE8;
-extern s32 g_memCardRStep;
```
sozud, line 171: I'm thinking @Xeeynamo will want to examine these renames

sozud (commented):
The HD version symbol files need some adjustment to fix the build

ser-pounce:
> Thanks for decompiling and refactoring!
> 
> The only objection I have is regarding the renaming of `g_MemCardRStep`, `g_MemCardRStepSub` and `g_MemcardRetryCount`. Those names were taken from `FntPrint`, whcih are respectively `rstep` and `retry`. So my question is: do we prefer have names that are more accurate to `FntPrint` and therefore to possibly the original intended names? Or should we steer away from it to go for fully customised names in every case?
> 
> BTW I noticed we currently have `g_memCardRetryCount` and `g_MemcardRetryCount`. That is pretty bad :P one of the two needs to go.

I just (naively..) thought my names were more descriptive (`StepSub` especially doesn't make a lot of sense to me but I'm still very new here), but if the names are in the source then perhaps it makes sense to keep them

PR #856, opened 2023-12-17
spu_r

sozud:
This one is pretty odd, not sure if there's better ways to write this or not.


sozud:
> Looks good.
> 
> I am not sure where you got the code inspiration. But how about something like this? https://decomp.me/scratch/hmRhE

Nice find. I'm thinking the spucnt part was a macro originally and wrapped in do while https://decomp.me/scratch/h6PmL

PR #858, opened 2023-12-17
Decompile NP3 func_801CF7A0

ser-pounce:
Not too sure about the entity members

### DIFF
--- a/src/st/np3/4E69C.c
+++ b/src/st/np3/4E69C.c
@@ -171,7 +171,7 @@ int func_801CF7A0(s16* arg0) {
     Entity* var_a0;
     s32 var_s0;
     s32 var_s1;
-    
+
     if (g_CurrentEntity->ext.et_801CE4CC.unk8E != 0) {
         --g_CurrentEntity->ext.et_801CE4CC.unk8E;
     }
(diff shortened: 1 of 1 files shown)


src/st/np3/4E69C.c
```diff
@@ -167,7 +167,71 @@ void func_801CF778(void) {
     et->unk9C = et->next->ext.et_801CF254.unk9C + 0x300;
 }
 
-INCLUDE_ASM("asm/us/st/np3/nonmatchings/4E69C", func_801CF7A0);
+int func_801CF7A0(s16* arg0) {
```
sozud, line 170: Looking at this scratch, it seems like arg0 is an Entity* https://decomp.me/scratch/51iIJ

ser-pounce: Thanks, implemented along with some further deductions! Incidentally it seems like the bogus return value might be an original bug, because who needs compiler warnings.

sozud:
Why was this closed?

ser-pounce:
> Why was this closed?

Because I'm not paying attention, sorry! Hopefully can reopen it now.

PR #860, opened 2023-12-18
Decompile and dedupe func_801D2320 (+9 dupes)

ser-pounce:
Probably needs some variable name fine tuning but the logic is mostly clear.


ser-pounce:
> Sad we need `(s16*)&`. I wonder what the developers really wanted to do. Other than that it looks good. Congrats!

Might be a case for a `union` in `Primitive`, if it shows up elsewhere.

PR #872, opened 2023-12-19
Decompile RIC func_801572A8

bismurphy:
Not really sure what this does, it's called by UpdateEntityRichter 5 times. Has weird pointer caching but thankfully I got it to match.

Nice to see it using g_Player.colliders, colliders2, and colliders3, I think we probably have those all settled now.


src/ric/1AC60.c
```diff
@@ -173,7 +173,110 @@ void func_80156F40(s16 arg0) {
...
+        }
+    }
+    if ((PLAYER.velocityY < 0) && !(*unk04_ptr & 2)) {
+        PLAYER.posY.val += PLAYER.velocityY;
+    }
+    if ((PLAYER.velocityY > 0) && !(*unk04_ptr & 1)) {
+        PLAYER.posY.val += PLAYER.velocityY;
+    }
+    for (i = 0; i < 4; i++) {
+        argX = PLAYER.posX.i.hi + D_801545F4[i].unk0;
```
sozud, line 242: Given that there's 4 points and x and y points I think you can document D_801545F4 etc. as collision boxes

sozud (approved):

PR #874, opened 2023-12-19
Extract RIC SpriteParts

bismurphy:
This was done by comparing the src/servant/tt_000/spriteparts.c file to the data I found in RIC 170AC.data.s. I then wrote a Python script to take the data in the data file and output a C file which has all the same formatting as that existing spriteparts.c file. This is a big chunk of the .data section for RIC, so it's nice to have it pulled out to a C file now.


src/ric/spriteparts.c
```diff
@@ -0,0 +1,1140 @@
...
+    (SpriteParts*)D_80153B80, (SpriteParts*)D_80153B9C,
+    (SpriteParts*)D_80153BB8, (SpriteParts*)D_80153BD4,
+    (SpriteParts*)D_80153BF0, (SpriteParts*)D_80153C0C,
+    (SpriteParts*)D_80153C28, (SpriteParts*)D_80153C44,
+    (SpriteParts*)D_80153C60, (SpriteParts*)D_80153C7C,
+    (SpriteParts*)D_80153C98, (SpriteParts*)D_80153CB4,
+    (SpriteParts*)D_80153CD0, (SpriteParts*)D_80153CEC,
+    (SpriteParts*)D_80153D08};
+
+int D_80153AF4[] = {0x00300001, 0xFFF5FFF5, 0x00180018, 0x00000000,
```
sozud, line 783: I know you're following the other file but should we not fully extract these? https://github.com/Xeeynamo/sotn-decomp/blob/6898b8b60818dd222d770caf97704635d464164a/include/game.h#L1269

bismurphy: Sure, if that's of interest. I think the best option will be to merge this one, and then I will write a script to convert both spriteparts.c files to properly extract the structs, if that works.

PR #878, opened 2023-12-20
Decompile DRA func_80105078

bismurphy:
Not really sure what this one does, lots of math with vectors and matrices.


sozud:
Your issues are are a bunch of undefined references. To fix this the sotn-specific ones need to be defined in stubs.c https://github.com/Xeeynamo/sotn-decomp/blob/master/src/pc/stubs.c
There's also psy-q ones that need to be defined in the relevant psxsdk files https://github.com/Xeeynamo/sotn-decomp/tree/master/src/pc/psxsdk

For example D_801379D0 needs a non-extern definition in stubs.c. RotMatrix needs a definition in libgte.c

PR #886, opened 2023-12-25
Extract BIN/F_GAME.BIN

Xeeynamo:
Mostly a proof of concept to extract resources from `BIN/F_GAME.BIN`. There are tons of hacks to make this work:
* Tweaked png2s to decode binaries into PNGs
* Add a `config/gfx.game.json` that describes how to extract different sprites out of a binary file
* Coded the palette location by reading the `clut` parameter in the source code. A clut value of `0x196` corresponds to `"palette": "0x96"` or `"palette": 150`.
* Coded the bits per pixel for a couple of 8-bit images
* Coded the palette location of those graphics that expects the palette from DRA.BIN instead of F_GAME.BIN
* For all the undiscovered palette, I called the files `unk` and defaulted their palette to grey

The expectation is that sprites will slowly be documented and added into `config/gfx.game.json` by modders and enthusiasts. This does not yet pack back these PNGs into the binary file. The way these assets are extracted is probably not final or perfect, but we have to start from somewhere.

image

### DIFF
--- a/Makefile.psx.mk
+++ b/Makefile.psx.mk
@@ -11,6 +11,7 @@ PSX_HD_TARGETS	:= dra
 PSX_BASE_SYMS	:= $(CONFIG_DIR)/symbols.$(VERSION).txt
 
 extract_us: $(addprefix $(BUILD_DIR)/,$(addsuffix .ld,$(PSX_US_TARGETS)))
+	$(PNG2S) bdecode config/gfx.game.json disks/us assets/game
 extract_hd: $(addprefix $(BUILD_DIR)/,$(addsuffix .ld,$(PSX_HD_TARGETS)))
 
 extract_disk_us: extract_disk_psxus


tools/png2s.py
```diff
@@ -70,24 +94,196 @@ def encode_bitmap(data):
...
+        config = json.load(f)
+
+    source = os.path.join(source_dir, config["source"])
+    clut = get_num(config["clut"])
+    stride = get_num(config["stride"])
+    for item in config["items"]:
+        name = os.path.join(output, item["name"] + ".png")
+        ensure_dir(name)
+
+        if "palette" in item and "brute" not in item["palette"]:
```
sozud, line 160: Seems like brute is unused?

sozud (approved):
Nice work, seems like a good start to me. I'm thinking eventually we'd want this tool to be in a faster language but it's not important now

PR #890, opened 2023-12-26
Import menu strings for US

Xeeynamo:
I had the suspect menu strings were part of a string of arrays since a while. So I come up with a `g_MenuStr[]`, which removed the need of many fake symbols. I did not yet come with a bunch of `#define`s to resolve some magic numbers. Luckily you can guess which string is which from the context of the function. Resolving the HD strings might give me more clues on how to come up with good defines.

I decided to go for the route of having a `lang_us.c` and a `lang_jp.c` for the sake of maintaining things simpler. This is also due to the fact the HD build orders the japanese strings in a different way, making things a bit harder. Any modder can just come up with their own `lang_XX.c`, translate the menu in the language they want and just link it, without touching the original files.

For the test app things were tricky. I've been debating how to implement it as SOTN-encoded strings are not pre-encoded on compilation time like the PSX toolchain does. Instead they are burnt into the executable as UTF-8 characters. Instead of encoding them on-the-fly in `MenuDrawChar`, I decided to re-use `AnsiToSotnMenuString` to pre-encode all the strings and re-route each `g_MenuStr` pointer to the converted strings. This is both efficient at runtime and prevents to modify the existing matching code. Now the menu looks amazing.

I thought this was going to be much harder to do. I ended up doing incremental steps in each commit to ensure having healthy checkpoints in case things were going south. The tool ` ./tools/sotn-str.py parse disks/us/DRA.BIN 0xOFFSET` saved me sooo much time. But we need to tweak it to deal with Japanese characters.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ set(SOURCE_FILES_DRA
     src/dra/4A538.c
     src/dra/collider.c
     src/dra/demo.c
+    src/dra/lang_us.c
     src/dra/menu.c
     src/dra/5D6C4.c 
     src/dra/627C4.c


sozud (approved):
Looks good to me, if possible I'd like to merge the MSVC build before this one to avoid having to fix it for this PR

PR #893, opened 2023-12-26
Decompile RIC func_8015FEA8

ser-pounce:
99% copied from DRA func_80119F70, but cannot be deduped due to subtle differences.


bismurphy:
I would request that you add a comment on this function and func_80119F70, mentioning that they are the same function for anyone who might look at them in the future. For an example of this, see the comments above func_801603C4 and func_8011A4D0.

ser-pounce:
> I would request that you add a comment on this function and func_80119F70, mentioning that they are the same function for anyone who might look at them in the future. For an example of this, see the comments above func_801603C4 and func_8011A4D0.

Got it, just pushed, thanks!

sozud (approved):

PR #894, opened 2023-12-27
Decompile RIC func_8015C4AC

bismurphy:
Extremely similar to a DRA function, which is noted in the code.

### DIFF
--- a/src/ric/202A8.c
+++ b/src/ric/202A8.c
@@ -61,7 +61,61 @@ void func_8015C2A8(void) {
     }
 }
 
-INCLUDE_ASM("asm/us/ric/nonmatchings/202A8", func_8015C4AC);
+// Game Buttons unofficially refers to buttons used in playing the game.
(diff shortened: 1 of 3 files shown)


src/ric/202A8.c
```diff
@@ -61,7 +61,61 @@ void func_8015C2A8(void) {
     }
 }
 
-INCLUDE_ASM("asm/us/ric/nonmatchings/202A8", func_8015C4AC);
+// Game Buttons unofficially refers to buttons used in playing the game.
+// Direction, action and shoulder buttons. Any button except start or select.
+#define GAMEBUTTONS (~(PAD_START | PAD_SELECT))
```
sozud, line 66: Any reason not to share the macro in a header? https://github.com/Xeeynamo/sotn-decomp/blob/c5d0842edd7091fb1c154c1c9634ef35c7f0c0ac/src/dra/6D59C.c#L7

sozud (approved):

PR #898, opened 2023-12-27
Add File specific funcs

Xeeynamo:
I am not sure if we want to use this pattern. Here I am proposing to wrap functions to open files and read the entirety of a file.

The objective is to avoid performing sanity checks for `fopen`, `fread` and `malloc`, other than ensuring allocated resources are correctly de-allocated on both happy and unhappy paths.

How it works is the first parameter is always the callback, while starting from the second are all the parameters necessary to instantiate and allocate the resources to be passed to the callback.

I am not an expert in C, so I am happy to receive feedback on how to improve this. The idea comes from what I would do in functional programming. Unforutnately functions in C are not first-class citizen, so one-line helpers like `bool InitAccessoryDefs(FILE* f) { FileStringify(ParseAccessoryDefs, f); }` needs to be created instead of just creating lambda/anonymous functions.


sozud:
I'm going to go through and try to explain this to myself:
```
FileRead(InitEquipDefs, "assets/dra/equipment.json");

...

bool FileRead(bool (*cb)(FILE* file), const char* path) {
    INFOF("open '%s'", path);
    FILE* f = fopen(path, "rb");
    if (f == NULL) {
        ERRORF("unable to open '%s'", path);
        return false;
    }

    bool r = cb(f);
    fclose(f);
    return r;
}

...

bool InitEquipDefs(FILE* f) { FileStringify(ParseEquipDefs, f); }

...

bool FileStringify(bool (*cb)(const char* content), FILE* f) {
    fseek(f, 0, SEEK_END);
    size_t len = ftell(f);
    fseek(f, 0, SEEK_SET);

    void* content = malloc(len + 1);
    if (!content) {
        ERRORF("unable to allocate %d bytes", len);
        return false;
    }

    size_t bytesread = fread(content, 1, len, f);
    if (bytesread != len) {
        ERRORF("unable to read %d bytes", len);
        free(content);
        return false;
    }

    ((char*)content)[len] = '\0';

    bool r = cb(content);
    free(content);
    return r;
}

...

bool ParseEquipDefs(const char* content) {
...
}
```

So FileRead is called, the file is opened, and if it is successful, the callback is called. Now we call InitEquipDefs which doesn't seem to return a value. That calls FileStringify which reads the file then calls another callback. That callback is ParseEquipDefs, which parses the JSON.

I think InitEquipDefs not returning a value is problematic since I think if the JSON parsing fails we will crash or have odd behavior later? Also it seems like FileRead should be more like
```
if(!FileRead(InitEquipDefs, "assets/dra/equipment.json"))
{
  printf("Something went wrong parsing equipment.json\n");
  exit(1);
}
```

I guess one question is if a single error value should bubble up and be printed if FileRead fails, or these various printfs in each function? I suppose the information to print a useful error is only available at the place of failure.

I'm not voting no yet but my opinion at the moment is that this is kind of a lot of indirection given that C has weak abstractions. I don't think I've seen this pattern before in C? I'd be curious to hear other opinions.

PR #902, opened 2023-12-27
SsUtSetVagAtr

sozud:
https://decomp.me/scratch/X8he9


sozud:
> I was wondering about that `iVar2` to be `toneIndex` or something similar. I also see on the scratch you have `_svm_cur.field_7_fake_program` but in the PR you have `D_800978CF`? Other than that it looks good!

field_7_fake_program comes from here: https://github.com/AliveTeam/sound_rev/blob/d3f18849b08d0db5b51df854a49ba56a68c58e8b/psx_seq_player/lib_snd.cpp#L1605

I'm not sure how correct that is or not, I was thinking to decompile more of the memory area around `D_800978CF` and see what develops.

I'll make the iVar2 change

PR #915, opened 2023-12-30
Decompile RIC func_8016147C

bismurphy:
Similar to an existing DRA function, which is noted.


sozud:
Pointing out every example is going to take a while but there's a bunch of hex numbers that look good as FIX https://decomp.me/scratch/4DwsQ

PR #920, opened 2023-12-30
Dedupe DestroyEntity

ser-pounce:
Needs a little macro magic to work with DRA, unless there's a better way of doing this?


sozud:
I think the ric and stage dedupes are a clear win, but I'm not sure about the usage of G_API. Maybe just leaving the duplicate in DRA is cleaner? @Xeeynamo what is your opinion?

The big duplicate problem is mostly with the stage overlays

ser-pounce:
> The reason why we have those headers is because we currently cannot link those files as C functions due to some limitations of Splat that are likely to not going to be addressed any time soon.
> 
> Other than that, we are trying to find for a common function order between stage overlays to figure out how originally different C files were linked together. This is in the assumption that developers did not copy&paste code everywhere.
> 
> The overlay DRA does not fit into this plan, with the exception of `save_mgr`. Yours is a very smart idea. But I'd ditch `G_API` from `destroy_entity.h` and leave the DRA version as it was as it will most likely not going to be linked in the same way.
> 
> Let me know if you have any question!

Thanks, I understand now. I was of the impression that deduping was to reduce the codebase, but I see that the idea is to try and match the original source code 1:1, even if it was suboptimal.

PR #921, opened 2023-12-31
Dedupe SetTexturedPrimRect

ser-pounce:
config/symbols.us.ric.txt change unrelated


ser-pounce:
Not sure why `make format` is deleting the line in config/symbols.us.ric.txt, fixed for now.

ser-pounce:
> Here we kinda have a similar topic. We have `SetPrimRect` and then `SetTexturedPrimRect` right after in both DRA and SEL. But every other function that surrounds them are completely different. I bet they were just copy&pasted. As there are no hacks to make them working in harmony, I'd say this one is good to de-duplicate.
> 
> I'd even suggest to add `SetPrimRect` in the same header file. When we will be finally able to use C files instead of include headers, we can have a `set_texutred_prim_rect.c` and link it into both DRA and SEL.

Thanks for the suggestion, it's implemented.

PR #926, opened 2023-12-31
Decompile ST0 func_801B11E8

ser-pounce:


src/st/st0/30030.c
```diff
@@ -121,6 +121,28 @@ void func_801B1198(s16 arg0) {
...
-INCLUDE_ASM("asm/us/st/st0/nonmatchings/30030", func_801B11E8);
+u16* func_801B11E8(char ch) {
+    u16 var_a0;
+
+    if (arg0 >= 0x61) {
+        var_a0 = arg0 - 0x7DE0;
+    } else if (arg0 >= 0x41) {
+        var_a0 = arg0 - 0x7DE1;
+    } else if (arg0 == 0x2C) {
+        var_a0 = 0x8143;
```
sozud, line 132: I think you ought to be able to replace the shift-jis characters too.

Try this one here? `，`
Take a look at this table https://gist.github.com/stakira/7b87747e4341e09ff59eef38bba7b55f

ser-pounce: Does the old PS1 compiler support SJIS literals?

sozud: Yes https://github.com/Xeeynamo/sotn-decomp/blob/4b9db580763f69256c2dd7fdb25fc0874660cae4/Makefile#L72

The CI actually passed for that suggestion https://github.com/Xeeynamo/sotn-decomp/pull/926/commits/ca5910d0c98f9402cbc3540a5f803cce9a170d50

sozud: > Nope, but we injected `iconv` in the Makefile to always convert our utf-8 source code into shift-jis

I think Xeeynamo read the question as "does it support UTF-8", which I think is false. It definitely supports shift-jis and we have various examples https://github.com/Xeeynamo/sotn-decomp/blob/4b9db580763f69256c2dd7fdb25fc0874660cae4/src/dra/menu.c#L3469

PR #927, opened 2023-12-31
Import cd data

Xeeynamo:
and resolving some fake symbols


sozud (approved):
Significantly cleaner, nice work

PR #928, opened 2023-12-31
Decompile RIC func_8016F198

bismurphy:
This one is pretty weird, there are a whole bunch of temp variables that I couldn't figure out how to solve.

There are 8 blocks which set a variable, capped between 0 and an upper limit (0xF0 or 0xFF). For the upper limit, I have created a `MIN()` macro. I tried to make a `CLIP()` which would handle both limits in one macro, but couldn't create one that would match.

Scratch here if anyone can improve things:
https://decomp.me/scratch/fY8js


sozud:
Some temp removals and a CLIP that works https://decomp.me/scratch/A3eNa

bismurphy:
Neat - my goal with CLIP was to have syntax that would be `prim->y3 = CLIP(0, temp_a2_2, 0xF0);` , with the aim of removing var_v0_8 as a temp. Any chance that could work? If we're not removing a variable I'm not sure if this CLIP is worthwhile, especially since it's multi-line.

bismurphy:
Thank you! I think things look much better now.

PR #933, opened 2024-01-01
Decompile NP3 func_801CDF1C

ser-pounce:
Previously attempted by @sonicdcer https://decomp.me/scratch/CMSl2﻿.

I'm not sure the signature for func_801CDD80 is correct, which led me down this rabbit hole. It matches, but there's a lot of ugliness going on, I'm not sure if `arg1` has the correct type, or that `GH_Props` is the right struct to access..

### DIFF
--- a/src/st/np3/4B018.c
+++ b/src/st/np3/4B018.c
@@ -210,7 +210,25 @@ void func_801CDE88(s16* arg0) {
     }
 }
 
-INCLUDE_ASM("asm/us/st/np3/nonmatchings/4B018", func_801CDF1C);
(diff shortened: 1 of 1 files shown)


ser-pounce:
> Another way to be sure about the signature is to have a look at [one of the parent functions](https://raw.githubusercontent.com/Xeeynamo/sotn-decomp/gh-duplicates/function_calls/np3.func_801CDF1C.svg)

Haven't been able to deduce too much from those for now, but seeing as the function is shared by multiple entities then `GH_Props` might not be the right structure to reference. Probably best to revisit this function later when more of the calling functions have been decompiled.

PR #934, opened 2024-01-01
Texture render support

Xeeynamo:
Technically unimpressive, but I had to go through quite a good amount of changes. Here is the full changelog to make everything work:

* Ensure `ResetPlatform` is always called to avoid memory leaks
* Add `g_RawVram` to emulate the PS1 VRAM
* ~~The engine will load the optional file `disks/vram.bin`, a RAM dump from an emulator~~
* SDL2 will create 256x256 textures on-the-fly whenever a specific tpage is requested via `GetVramTexture`
* The function `GetVramTexture` caches the last called tpage to avoid tanking the performance
* `GetVramTexture` for only renders 4bpp and 8bpp textures with their specified palette
* Remove `SDL2_image` as the font is now loaded straight from the VRAM
* Calling `VSync` will call the set callback, which the game uses for DMA operations
* `MyLoadImage` is not yet implemented, but it is a placeholder to then interact with `g_RawVram`
* The menu font now uses the texture found in the VRAM
* Plugged a custom version of `LoadFileSim`
* The file `sim_pc.c` is similar to the original game's code but it is used here to load files from custom paths
* Using F5, F6 or F7 can dump the VRAM content on-screen, respectively in 16bpp, 8bpp and 4bpp

There are new graphical glitches on the font. In some occasions it appears black. It seems to be related to a flag in `P_TAG.code`. I plan to dig into it when I can render entities on screen to avoid potential mistakes. The same problem is present for the first half of Alucard's portrait. It seems to be related when a texture is transparent? 🤷 

I am not sure why the font is completely corrupted when entering in the Equip menu. It is hard to understand if I introduced any regression. Maybe the glitch was always there but it was hidden since I was always forcing the font texture to be rendered.

EDIT: Implemented `LoadImage`, `SaveImage` and `ClearImage`

### DIFF
--- a/src/pc/sdl2.c
+++ b/src/pc/sdl2.c
@@ -26,6 +26,7 @@ DebugSdl g_DebugSdl = DEBUG_SDL_NONE;
 unsigned int g_Tpage = 0;
 static SDL_Texture* g_VramTex = NULL;
 static int g_LastVramTexTpage = -1;
+static int g_LastVramTexClut = -1;
 
 void ResetPlatform(void);
 bool InitPlatform() {
@@ -68,6 +69,7 @@ bool InitPlatform() {
 void ResetPlatform(void) {
     g_Tpage = 0;
     g_LastVramTexTpage = -1;
+    g_LastVramTexClut = -1;
     if (g_VramTex) {
         SDL_DestroyTexture(g_VramTex);
         g_VramTex = NULL;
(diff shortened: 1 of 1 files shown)


src/pc/sdl2.c
```diff
@@ -236,19 +332,63 @@ DISPENV* MyPutDispEnv(DISPENV* env) {
...
+        g_LastVramTexTpage = tpage;
+
+        u8* src = g_RawVram;
+        src += (tpage & 0xF) * 128;
+        src += ((tpage >> 4) & 1) * VRAM_STRIDE * 256;
+
+        u16* pal = g_RawVram;
+        pal += clut * 0x10;
+        // Hack that rewrites the VRAM to adjust transparent palette colors
+        pal[0] &= ~0x8000; // first color always transparent?
```
sozud, line 349: According to this 0x0000 is transparent. What's the issue you're seeing?
image

sozud (approved):
Looks like a good step forward to me, we can continue iterating on the outstanding issues

PR #939, opened 2024-01-02
Add stage dummy

Xeeynamo:
Last test app specific PR for a few days. This adds a dummy stage that gets loaded by `InitStageDummy`. It does very similarly what the game does by copying the exported stage functions into `g_api.o`. The new entrypoint is now `InitRoomEntities`.

The dummy stage loads some assets from WRP. The stage layout and tileset are loaded in-memory. I created a very barebone `RenderTilemap` to draw the foreground without any scrolling implemented. I am not sure why the HUD is only showing when entering in the menu and it's black.

There is a hack job in the dummy stage to skip the SEL overlay and ask the game to load WRP. Choosing WRP is purely arbitrary as any stage but SEL would result to the same behaviour.

This is a bit of a memory hog. I introduced a memory pool instead of using `malloc` for the assets loaded dynamically. The dummy stage will not stay here for long, so I do not find useful to optimise it.

image

### DIFF
--- a/src/pc/sotn.c
+++ b/src/pc/sotn.c
@@ -46,8 +46,6 @@ GfxBank** g_GfxStageBank[0x40] = {
 };
 extern UnkStructClut* D_800A3BB8[];
 
-void StageOvlCb() {}
-
 void WeaponLoadPaletteStub(s32 arg0) { NOT_IMPLEMENTED; }
 
 // list of exposed API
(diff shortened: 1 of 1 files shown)


sozud (approved):
Looks great, MSVC isn't liking some of the 0-sized arrays though

```
D:\a\sotn-decomp\sotn-decomp\src\pc\stage_dummy.c(38,21): error C2466: cannot allocate an array of constant size 0 [D:\a\sotn-decomp\sotn-decomp\cmake_build\Sound.vcxproj]
D:\a\sotn-decomp\sotn-decomp\src\pc\stage_dummy.c(39,22): error C2466: cannot allocate an array of constant size 0 [D:\a\sotn-decomp\sotn-decomp\cmake_build\Sound.vcxproj]
```

PR #943, opened 2024-01-03
Decompile DRA func_80136010

bismurphy:
One of the few remaining DRA functions.

Some of the code here seems fake, but it matches so I'm going with it. I tried to change all the pointers to use the loop indexing but it didn't work. Might have been written by a coder who was more comfortable working with pointers than with loops.


sozud:
Thanks for helping out with this. So D_80138F64 is an array of size 24.

image

I have a couple of changes here to fix stuff that is overlappping with that:

https://decomp.me/scratch/jG2IN

I think the stuff like this is an optimization
```
        (((D_80138F64[14] + D_80138F64[15] + D_80138F64[16] + D_80138F64[17]) << 0x18) == 0)) 
```

and was actually like
```
if(D_80138F64[14] == SPU_OFF && ...
```
but I can't figure it out.

sozud:
func_80136010 needs to be removed from stubs.c since it is defined now
```
/home/runner/work/sotn-decomp/sotn-decomp/src/dra/953A0.c:133: multiple definition of `func_80136010'; build/pc/src/pc/stubs.o:/home/runner/work/sotn-decomp/sotn-decomp/src/pc/stubs.c:519: first defined here
```

D_80138F64 needs a stubs definition
```
/home/runner/work/sotn-decomp/sotn-decomp/src/dra/953A0.c:148: undefined reference to `D_80138F64'
```

SpuGetAllKeysStatus needs a stubs definition
```
/usr/bin/ld: /home/runner/work/sotn-decomp/sotn-decomp/src/dra/953A0.c:148: undefined reference to `SpuGetAllKeysStatus'
```

These too D_8013AE8A,  D_8013B64E, D_8013AEA6

PR #948, opened 2024-01-07
DRA assets config

Xeeynamo:
Creates the file `config_us.c` with all the configuration needed for the game to know where to locate files in the disc (previously `lba_stage.c`), equipments, accessories, menu strings (previously `lang_us.c`, , spells, relics and enemy structures. There is also an exp table for levelling up but I did not import it yet.

This is a major change in the codebase on how we handle assets. Previously they were exported as JSON in `assets/dra`. The main problem is that strings were not properly compiled, still relying on hardcoded offsets expected to be found in the `.rodata` section.

I deducted all those information belong to a single C file because the same strings in rodata were referenced from equipment to enemies. If those strings are found in the same C file, the compiler will optimise duplicated strings by de-duplicating them. This is how I reached the conclusion of having all those definitions in a single file.

Obviously the major pain point of this is that a JSON was much more declarative and easy to be consumed. Especially if tools were going to use those files. Counting the commas with the struct definition on the side is not the best thing for modding.

I had to comment out the HD `lba_stage` as it would require a major work to support the Japanese text.

### DIFF
--- a/config/splat.hd.dra.yaml
+++ b/config/splat.hd.dra.yaml
@@ -136,7 +136,7 @@ segments:
       - [0x3CCC0, .rodata, 627C4]
       - [0x3CCC0, .rodata, 627C4] # HandleSaveMenu, all the strings
       - [0x3CED4, .rodata, 63ED4] # jtbl_800DCED4
-      - [0x3CEF0, .rodata, lba_stage]
+      - [0x3CEF0, rodata, lba_stage]
       - [0x3D29C, rodata]
       - [0x40C74, .rodata, cd]
       - [0x40D64, rodata] # jtbl_800E0D64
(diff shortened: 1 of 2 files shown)


sozud (approved):
Nice work on unraveling this. Excepting the build issues seems good to me in general. Some thoughts are:
If scripts were written to generate this stuff, we may want those checked in? We can extract other versions etc. Also for the test program we may still want to config with json, so I think it may make sense to support either? Maybe like `#ifdef MODDING_MODE` etc. Having all this baked into C is probably better for ports to platforms without filesystems

sozud:
I'm fine with cutting the JSON for now, we aren't in a place where modding is really feasible at the moment

PR #949, opened 2024-01-07
assets.py exports C code

Xeeynamo:
Created due to some major changes introduced in #948


Xeeynamo:
I am closing this as it will be replaced by the same tool of #1343 (but in a different pull request later on)

PR #951, opened 2024-01-07
SEL import banks and fix headers

Xeeynamo:


include/types.h
```diff
@@ -12,7 +12,7 @@ typedef unsigned int uint32_t;
 typedef unsigned long long uint64_t;
 typedef unsigned char u_char;
 typedef unsigned short u_short;
-typedef unsigned int u_long;
+typedef unsigned long u_long;
```
sozud, line 15: Just noting that `long` is 8 bytes on linux and 4 bytes on psx

sozud (approved):

PR #952, opened 2024-01-07
Decompile SEL func_801B519C

ser-pounce:
One annoying `do { } while (0);` left but it matches.﻿


ser-pounce:
Poked around for a bit but couldn't refactor past your suggestions, thanks!

PR #954, opened 2024-01-09
Decompile DRA func_8013572C

bismurphy:
This is one of the few remaining sound functions.

It calls `ApplyQuadChannelSetting` and `func_80135624` in extremely similar ways, so I've adjusted the arguments to those functions so they match. `func_80135624` should probably get its own name which is similar to `ApplyQuadChannelSetting`, but I don't know enough about these functions to actually tell what it's doing.


sozud:
The func_8013572C stub needs to be deleted from stubs.c, other than that looks good!

PR #960, opened 2024-01-10
Remove the use of UnkStructClut

Xeeynamo:
Rewrites SEL's `banks.c` to use an array of `u_long` and make use of declarations to populate it.

This PR comes after giving up at believing the original developers would had the decency of using structs instead of crude array of data. They are really developers from another era.


sozud (approved):
Interesting how they have instructions to control the palettes

PR #963, opened 2024-01-10
Link stage SEL

Xeeynamo:
This depends from #962 . 

Without `func_800EA7CC` being decompiled, past the title screen you will be welcomed with a black screen. Using the [non matching scratch](https://decomp.me/scratch/dlQDD) will work. The input works regardless the black screen. Getting past the name selection screen there is a crash at `SetReleaseRate1` due to `D_80138FB4` being `NULL`. Getting past it you will get another crash at `HandleVideoPlayback` but I neither investigated on it or tried what's beyond that.

There is a divergence between the Make and CMake. For some reason the build from CMake gives a black screen. I am not able to detect where the issue is. I get no warnings or errors. I noticed `g_MegaMenuStrBuffer` gets full much faster in the CMake build. It might due to a silly size of any of the atomic C types. It needs further investigation.


src/pc/stage_sel.c
```diff
@@ -0,0 +1,141 @@
...
+#include "../st/sel/sel.h"
+#include "pc.h"
+#include "sfx.h"
+
+// stubs
+RECT D_80182584 = {0};
+RECT D_8018258C = {0};
+RECT D_801825A4 = {0};
+void* D_8018C404[100] = {NULL};
+u16 D_800737B0; // g_Entities[5].step
```
sozud, line 11: If these are actually g_Entities[5].step, isn't that going to be an issue since updates to one won't affect the other?

src/pc/stage_sel.c
```diff
@@ -0,0 +1,141 @@
...
+s32 D_801BAFC8;
+s32 D_801BAFCC;
+u8* D_801BAFD0; // Pointer to texture pattern
+s32 D_801BAFD4;
+s32 D_801BAFD8;
+s32 D_801BAFDC;
+s32 D_801BAFE0;
+u32 D_801BAFE4;
+s32 D_801BAFE8;
+// s32 g_MemCardRStep;
```
sozud, line 75: Do we want to keep the commented out stuff?

sozud (approved):

PR #964, opened 2024-01-10
Fix clang compilation errors

Xeeynamo:
It still complains on :apple: due to the missing SDL header. Doing `brew install SDL2` does not fix the issue. I think it would eventually be beneficial to add a macOS agent to the CI.


sozud (approved):

sozud:
If you change the header to `#include <SDL.h>` does that help?

PR #968, opened 2024-01-11
Add macOS agent to the CI

Xeeynamo:
`AppleClang` has `-Wimplicit-function-declaration` and `-Wreturn-type` defined by default. I had to disable them to have parity between non-Apple default compiler flags.


sozud:
For these

```
/Users/runner/work/sotn-decomp/sotn-decomp/src/st/sel/3642C.c:322:16: error: character too large for enclosing character literal type
        jCh = ('Ａ' - 'A') + ch;
```

I think we need to define our char types so that it's utf-8 width on non-psx platforms

sozud:
I think `ifdef`ing them out is fine, does that break any of the menu stuff at the moment?

PR #971, opened 2024-01-12
Initial sound_rev integration

sozud:
This integrates sound_rev and hardcodes playing the library song for now. There's some important API functions missing like `SsVabOpenHeadSticky` and `SsVabTransBodyPartly` so this is not deeply integrated into the engine yet. I fixed the extraction of the vb and vh files since the addresses and sizes were wrong. Right now sound_rev audio or XA audio has to be toggled with a define, the plan is to integrate proper mixing later.


sozud:
I changed sound_rev into a library so that it can have independent compile flags and fixed all the issues on mac and windows.

sozud:
Closing in favor of https://github.com/Xeeynamo/sotn-decomp/pull/1362

PR #972, opened 2024-01-12
Fix load of ST/SEL/F_SEL.BIN

Xeeynamo:
The file is 192KB long. Yet on PSX it seems the game still wants to load 256KB of it. Obviously that only works there where the memory is unprotected.

There are two other crashes happening at `func_801B1FD8` and `func_801B2108`. The first one is triggered by `func_801ACF7C` and it seems to be related to the termination character. The second I have no idea.

To skip `SEL` and boot straight to WRP, go to `sim_pc.c` and comment the following lines:
```c
        case STAGE_SEL:
            InitStageSel(&g_api.o);
            break;
```


src/pc/sim_pc.c
```diff
@@ -106,14 +106,16 @@ void func_800E7D08(void) {
...
     RECT rect;
     u8* pTilesetDataSrc;
-    s32 i;
+    u32 i;
+    u32 blockCount;
 
     rect.w = 0x20;
     rect.h = 0x80;
-    for (i = 0; i < 0x20; i++) {
+    blockCount = CLAMP_MAX(len >> 13, 0x20);
```
sozud, line 117: Not sure where the right spot is but maybe we should log a warning if we are clamping the length?

sozud (approved):

src/pc/sim_pc.c
```diff
@@ -107,14 +107,20 @@ void func_800E7D08(void) {
...
     u8* pTilesetDataSrc;
     u32 i;
     u32 blockCount;
 
     rect.w = 0x20;
     rect.h = 0x80;
-    blockCount = CLAMP_MAX(len >> 13, 0x20);
+    blockCount = len >> 13;
+    if (blockCount > MaxBlockCount) {
+        WARNF("tileset too large (%d > %d): it wil not be fully read",
```
sozud, line 120: I thought the issue was that the tileset is too small? What about the case where it's wanting to load more bytes than the file has?

sozud: I guess I'm not understanding what is going on. It seems to me there can be two cases:

The file is smaller than the number of bytes the game wants to load: we need to only load what actually exists to avoid a buffer overrun. We should warn about this since it's difference in behavior vs. psx, where I guess random junk will be written at the end?

The file is larger than the number of bytes the game wants to load: this shouldn't cause a crash but makes sense to warn about too.

It seems like we are warning about the latter case and not the former?

sozud: Ok whichever you think makes the most sense is fine with me

PR #976, opened 2024-01-13
Extract config_jp from DRA HD

Xeeynamo:
I need help on this pull request. While it matches as it is, I cannot decode the Japanese strings from bytes to UTF-8 convertible into Shift-JIS into a C file that is able to give me an `OK`. I think the problem lies on how Python decodes Shift-JIS byte arrays.

I [wrote a tool](https://gist.github.com/Xeeynamo/d75c73431eaf54b42f955ca8740bb754) to automate the process:
```bash
python3 sotn_config.py > src/dra/config_jp.c && make
```

To extract some of the strings as Shift-JIS, uncomment the first line at the function `get_sjis_str`. I am tempted to make a custom decoder. But first I want to gather your insight in case I am missing something obvious.


sozud:
I tried running this but I don't really understand which strings in the file are supposed to be sotn_str vs. shift-jis. Here's a couple of ideas:

If you convert the shift-jis string back to bytes, do you get the same result as the input?

Do something like this:

```
counter = 0

def get_sotn_str(f, ptr):
    if counter < 1:
        return get_sotn_str_new(f, ptr)
```
And increase the cutoff value until it stops matching, then you know which string is the issue

sozud:
Ok so just to check, the current state of the PR is: all shift-jis strings are converted except for the ソ, which is left as byte literals? The remaining strings are a custom encoding? If so seems good to me. My guess is that their GCC has a patch which is missing from our GCC

PR #980, opened 2024-01-14
Refactor DRA func_800FD5BC

bismurphy:
Came across this function today and found several improvements for it.

- Unkstruct_800FD5BC is a duplicate of DamageParam. Therefore, I have changed all instances of Unkstruct_800FD5BC to DamageParam. The fields make sense in context and everything.
- Because func_800FD5BC is in GameApi, which is defined in `game.h`, its argument (now a DamageParam) needs to be accessible in `game.h`. Therefore, DamageParam moves from dra.h to game.h.
- We had a fake division being done with using a `temp` and checking a comparison to zero; this has been changed to be a normal division. This is a common pattern I see in initial m2c output and I'm sure it exists in other places in the repo, would be nice if we could find an automated way to search for this type of structure to simplify the code.
- Changed a 0x14 to 20 since it makes way more sense that way
- Changed an `if(){return} else{morestuff} to just if(){return} morestuff, eliminating the `else`

I think this function should not be in `menu.c` since it has no close connection to any kind of menu. It would also be nice if we could figure out what damageKind actually is, and maybe make an enum for it.

Overall, no new decompiling here, but lots of improvement to the readability of the code. This might actually be a good PR to keep in mind as an example for people who are interested in the project but aren't comfortable with assembly, since this is a lot of improvement and change, without touching assembly at all.

### DIFF
--- a/include/game.h
+++ b/include/game.h
@@ -1174,6 +1174,13 @@ typedef struct {
 } Dialogue;                                  // size = 0x44
 
 typedef struct {
+    u32 unk0;
+    u32 damageKind;
+    s32 damageTaken;
+    s32 unkC;
+} DamageParam;
+
+typedef struct {
     /* 8003C774 */ Overlay o;
     /* 8003C7B4 */ void (*FreePrimitives)(s32);
     /* 8003C7B8 */ s16 (*AllocPrimitives)(PrimitiveType type, s32 count);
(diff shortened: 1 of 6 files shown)


bismurphy:
I will go ahead and move the function to the next file.

As far as damageKind, my point was just that we should work out what each value for it means. We know it's the kind, but we don't know what, for example, damageKind = 5 actually means.

PR #986, opened 2024-01-16
POC for JP sotn_strs

sozud:
This is a proof-of-concept for Japanese sotn_strs. The macro `_SJ` is used to designate them. I only converted one string in config_jp to serve as a test for the pipeline. I moved sotn_str to a folder and added some tests. I renamed it since python doesn't like modules with - in the name. The code could be cleaner and the conversion table could use some more work but I think this is a reasonable start.


sozud:
All the strings convert now except "\x8A\xFF\x9E\x99\x7B\xFF\x9E\x72\xC9\xB4\xDD\xFF\xFF\xED\xFF", not sure what's wrong with it.
Updated extraction script: https://gist.github.com/sozud/d61c0353d9e8c17add5985a5b87ea7d2

sozud:
I didn't realize they were the same. Did some messing around with it now and it's not matching, would need more investigation. Is combining them as a follow-up fine?

PR #995, opened 2024-01-19
Stage overlay dynamic linking

sozud:
I'm putting this up as a draft for feedback before attempting to figure this out on Windows. This compiles SEL as a dynamic library and loads the symbols at runtime. This allows us to have conflicting symbols in overlays. I changed the API to use function pointers because that solved the dynamic libraries not being able to find those symbols. The path to the libraries is currently hardcoded and different on Mac and Linux. This requires building in cmake_build.


sozud:
So I'm not sure if this is a reasonable approach due to limitations on Windows. We can share functions from the executable to the DLL by declaring things with `__declspec(dllexport)`. However my understanding is that you can't share data from the executable with a DLL unless you set a pointer to it with a function. For example `g_IsUsingCd` would need to become like `extern s32 *g_IsUsingCd;`, then we'd have `SetsUsingCdPointer(&g_IsUsingCd)` called in the exe before executing anything in the overlay. Maybe someone better versed in Windows programming can come up with something?

I'm thinking a better approach would be using macros to fix conflicting functions like:
```
#define PREPEND_PREFIX(prefix, func) prefix##func

PREPEND_PREFIX(SEL_, TestFunc)() {

}

PREPEND_PREFIX(SEL_, TestFunc)();
```

The macro would be off on PSX.

sozud:
There doesn't seem to be a way forward with this on Windows so closing in favor of a static linking approach.

PR #998, opened 2024-01-19
PC libgte rsin rcos

Xeeynamo:
A very simple implementation of `rsin` and `rcos`. For now without a pre-calculated sinus table.

I am not yet sure about `if (prim->blendMode & 1)`, so I will not create a dedicated define until further investigation. But this was required to render #997


sozud:
```
D:\a\sotn-decomp\sotn-decomp\src\pc\psxsdk\libgte.c(18,38): error C2065: 'M_PI': undeclared identifier [D:\a\sotn-decomp\sotn-decomp\pc\core.vcxproj]
```

sozud (approved):

PR #1004, opened 2024-01-21
Implement DrawOTag

Xeeynamo:
This was one of the main reasons why I did not want to create a custom SDK. But after reading online [how the ordering table works](https://psx.arthus.net/sdk/Psy-Q/DOCS/TECHNOTE/ordtbl.pdf), I just figured it would be a piece of cake to implement it.

Ohbboy I was wrong.

It took me half a day to reach this state. But this works on 32-bit and 64-bit builds without hiccups. Implementing `DrawOTag` was to me the biggest obstacle. Please enjoy a series of screenshots that gives the idea of how important this function is:

image

image

image

image


src/dra/42398.c
```diff
@@ -728,7 +728,7 @@ void func_800E385C(u_long* ot) {
             }
         } else {
             if (getaddr(ot) != ((u_long)ot & 0xffffff) + 4) {
-                *var_t1 = (u_long)ot & 0xffffff;
```
sozud, line 731: Are there any portability issues with casting from OT_TYPE to u_long and back?

include/psxsdk/libgpu.h
```diff
@@ -129,10 +129,17 @@ typedef struct {
     /* 0x6 */ short h; /* height */
 } RECT;                /* size = 0x8 */
 
+#ifdef VERSION_PC
+#define O_TAG                                                                  \
```
sozud, line 133: I think some documentation explaining the O_TAG / OT_TYPE problem and solution would be good, my understanding is pretty low

sozud (approved):
Looks amazing, well done!

PR #1005, opened 2024-01-22
Add support for Japanese characters

Xeeynamo:
This took me A LOT of time, mainly because I am not aware of any good dictionary/map implementation in C. One of the time consuming parts was adding the special characters due to how the game uses the dakuten and handakuten.

The function `AnsiToSotnMenuString` is still very fast as I am searching the characters via a binary search. If you have any suggestion for a O(1) implementation please let me know.

This fixes some buffer overflows when rendering Japanese strings.

### before

image

### after

image

### DIFF
--- a/src/pc/str.c
+++ b/src/pc/str.c
@@ -206,11 +206,14 @@ const char* AnsiToSotnMenuString(const char* str) {
         SotnMenuPair* pair = ReadCharacterInfo(&src);
         if (pair) {
             *dst++ = pair->value;
+            if (pair->value == 0xFF) {
+                *dst++ = 0xFF;
+            }
             if (pair->hasDakuten) {
                 *dst++ = 0xFF;
                 *dst++ = 0x9E;
             }
-            if (pair->hasDakuten) {
+            if (pair->hasHandakuten) {
                 *dst++ = 0xFF;
                 *dst++ = 0x9F;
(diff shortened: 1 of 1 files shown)


src/pc/str.c
```diff
@@ -0,0 +1,225 @@
...
+    char* start = str_buffer + str_buffer_index;
+    char* dst = start;
+    const unsigned char* src = str;
+    struct SotnMenuPair* pair;
+    while (*src != '\0') {
+        DEBUGF("%s", src);
+        SotnMenuPair* pair = ReadCharacterInfo(&src);
+        if (pair) {
+            *dst++ = pair->value;
+            if (pair->hasDakuten) {
```
sozud, line 209: It doesn't seem like the 月 case is being handled? See this item https://github.com/Xeeynamo/sotn-decomp/blob/4d2ccfbd3b29e312740addd87247d9633d9c080b/src/dra/config_jp.c#L153

And https://github.com/Xeeynamo/sotn-decomp/blob/4d2ccfbd3b29e312740addd87247d9633d9c080b/tools/sotn_str/jp.py#L267

The problem is that character is 255, but the terminator is also 255, so it has a special encoding.

sozud (approved):
Seems ok now with the fixes

PR #1008, opened 2024-01-24
DRA RenderTilemap

Xeeynamo:
I've been wanting to match this for a long time. I matched it by pure coincidence by moving things around. This is one of the few functions were the permuter was completely useless.

The function renders a room tilemap, with its background and foreground layers. This is also the first function that uses the PlayStation 1 scratchpad. It took me a few trials to understand how to use it. This match gives me the confidence to match the much bigger functions `RenderPrimitives` and `RenderEntities`. They will be my next targets.

I noticed there were a few wrongs in the rendering system. I used one of the PSY-Q SDK samples to help me understanding how to implement `PutDrawEnv` and how to render a `SPRT`. The stealth changes in this PR aims to easily have the tilemap rendering working. The map currently appears black. This is not a regression, but the lack of logic of loading a room.

### DIFF
--- a/.github/workflows/picci.yaml
+++ b/.github/workflows/picci.yaml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Install requirements
-        run: sudo apt-get install build-essential libsdl2-dev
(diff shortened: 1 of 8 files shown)


sozud (approved):
Nice work!

PR #1011, opened 2024-01-28
Decompile RIC EntityHydroStorm

bismurphy:
Several things with this one.

1) In order to determine what this actually was, I needed to extract Richter's subweapon definitions. I decided, at least for now, to add that extraction for RIC, using the `assets.py` script.

2) In order to match this function, we need a new Primitive, which is similar to the existing one (and FakePrim) but has a few totally incompatible fields. Most importantly, we have s16's at offsets 0x0C and 0x0E, which don't match either of those. Therefore, I introduce a new primitive that I am, at least for now, calling PrimLineG2, since this is the first time we're using LineG2, and is also the first use of this struct, so those may be connected. If a different primitive is found using these fields, we can rename this.

3) Recognizing that we now have 3 different varieties of Primitive, I have taken the definitions for all of them and moved them to a dedicated `primitive.h`, which I think is a good home for them. We should rename the variations eventually, but having them in one place makes sense, I think.

Otherwise, this one is fairly straightforward. It's interesting that we have the hardcoded values for the color of the hydro storm rain in there - using this knowledge I have made a fun PR for the randomizer which randomizes the colors of the rain. That PR is here: https://github.com/3snowp7im/SotN-Randomizer/pull/121

I think that's about it!

### DIFF
--- a/config/splat.us.ric.yaml
+++ b/config/splat.us.ric.yaml
@@ -33,6 +33,8 @@ segments:
       - [0x20, spritesheet, richter, disks/us/BIN/F_GAME2.BIN, 0x40400]
       - [0x170AC, .data, spriteparts]
       - [0x18568, data]
+      - [0x18688, assets, subweapondefs, D_80154688]
+      - [0x188F4, data]
       - [0x1895C, .data, 21250]
       - [0x18A6C, assets, factory_blueprint, g_RicFactoryBlueprints]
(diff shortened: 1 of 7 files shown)


sozud (approved):
Looks good to me but I'm thinking @Xeeynamo should weigh in on the multiple Prim structs thing

PR #1012, opened 2024-01-29
Start migrating psy-q

sozud:
This is the beginning of moving psy-q content to the other repo. I just did a couple of files to get the approach approved first.


sozud:
Alright, it doesn't have to be done right now.

sozud:
We have modifications to the libraries in-tree now so this doesn't make sense anymore

PR #1013, opened 2024-01-29
Decompile DRA DrawRichterHudSubweapon

bismurphy:
This one was tricky and required some restructuring of an area of memory, but it matches now, so that's cool!

No big comments from me but this one has a lot of random stuff so let me know if I missed some enum or something!


src/dra/5D5BC.c
```diff
@@ -1515,19 +1515,19 @@ void InitStatsAndGear(bool isDeathTakingItems) {
...
-    D_80137984 = 0;
-    D_80137990.unk0 = 0;
-    D_8013798C = 40000 / D_80137978;
-    D_80137988 = 40000 / D_8013797C;
-    D_80137970 = func_800EDD9C(PRIM_GT4, 9);
-    prim = &g_PrimBuf[D_80137970];
+    g_PlayerHud.unk14 = 48;
+    g_PlayerHud.unk18 = 0;
+    g_PlayerHud.unk24 = 0;
+    g_PlayerHud.unk20 = 40000 / (u32)g_PlayerHud.unk0C;
```
sozud, line 1527: unk0C is already u32 so I think this can be removed?
```
 /* 80072F2C */ u32 unk0C;
```

src/dra/5D5BC.c
```diff
@@ -1614,30 +1614,286 @@ void DrawRichterHud(void) {
...
-INCLUDE_ASM("dra/nonmatchings/5D5BC", DrawRichterHudSubweapon);
+    prim->u0 = (g_Status.hearts / 10) * 8;
+    prim->v0 = 0x60;
+    prim->u1 = ((g_Status.hearts / 10) * 8) + 8;
+    prim->v1 = 0x60;
+    prim->u2 = (g_Status.hearts / 10) * 8;
+    prim->v2 = 0x68;
+    prim->u3 = ((g_Status.hearts / 10) * 8) + 8;
+    prim->v3 = 0x68;
+    if ((g_Player.unk0C & 0x200000) && !(g_Timer & 2)) {
```
sozud, line 1802: My guess is that this does the flashing numbers when you have enough hearts to do an item crash

sozud (approved):
Excellent work

src/dra/5D5BC.c
```diff
@@ -1614,30 +1614,286 @@ void DrawRichterHud(void) {
...
-s32 D_800A3014[] = {
-    0x9,   0x00F, 0x018, 0x010, 0x0A8, 0x0C0, 0x01E, 0x17F, 0x009, 0x007, 0x018,
-    0x018, 0x080, 0x0C0, 0x01E, 0x17F, 0x00C, 0x00D, 0x010, 0x010, 0x028, 0x070,
-    0x01B, 0x102, 0x008, 0x00C, 0x018, 0x018, 0x038, 0x068, 0x01B, 0x102, 0x00C,
-    0x007, 0x010, 0x018, 0x098, 0x0D8, 0x01E, 0x17F, 0x00C, 0x007, 0x010, 0x018,
-    0x098, 0x0C0, 0x01E, 0x17F, 0x00F, 0x013, 0x008, 0x008, 0x0C0, 0x0D0, 0x01E,
-    0x163, 0x008, 0x00D, 0x018, 0x010, 0x0A8, 0x0D0, 0x01E, 0x17F, 0x008, 0x006,
-    0x018, 0x018, 0x080, 0x0D8, 0x01E, 0x17F,
-};
+TexturedRectParams D_800A3014[] = {
```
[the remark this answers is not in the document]
bismurphy: Changing to `RicSubwpnIconParams`. Thanks for pointing that out!

src/dra/5D5BC.c
bismurphy: Good point, that struct was previously all s32 so the cast was needed.

bismurphy: wait no, unk0C is in PlayerHud, your quoted line is in g_Player.unk0C.

In PlayerHud, it needs to be an s32 due to the `if (g_PlayerHud.unk0C < 0) {` at line 1673 of this file. So it must be s32, and must have a cast at this usage.

sozud: I misread, no problem then

PR #1014, opened 2024-01-30
Decompile RIC func_801682B4

bismurphy:
Exact copy (aside from using g_api) of an existing DRA function. It was not obvious that these would match, and took a lot of digging to work out that they were the same, but I'm glad I went through that process, because otherwise decompiling this would be annoying! Definitely will be important to keep doing that as we go forward. It's amazing how much of RIC and DRA are similar.

### DIFF
--- a/src/ric/2A060.c
+++ b/src/ric/2A060.c
@@ -510,7 +510,39 @@ void EntityHydroStorm(Entity* self) {
     g_Player.D_80072F00[3] = 16;
 }
 
-INCLUDE_ASM("ric/nonmatchings/2A060", func_801682B4);
+// Copy of DRA function func_80125A30
+s32 func_801682B4(s32 baseY, s32 baseX) {
+    s16 x;
+    s16 y;
+    Collider res1;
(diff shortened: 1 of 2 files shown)


src/ric/2A060.c
```diff
@@ -510,7 +510,39 @@ void EntityHydroStorm(Entity* self) {
...
+    Collider res1;
+    Collider res2;
+    s16 colRes1;
+    s16 colRes2;
+
+    x = baseX + g_CurrentEntity->posX.i.hi;
+    y = baseY + g_CurrentEntity->posY.i.hi;
+
+    g_api.CheckCollision(x, y, &res1, 0);
+    colRes1 = res1.effects & 0xF801;
```
sozud, line 526: The effects could be documented? https://github.com/Xeeynamo/sotn-decomp/blob/4c4c570ef08a57a6ef6bc42efa267f1a9897d141/include/game.h#L957

sozud (approved):

PR #1016, opened 2024-01-31
Decompile RIC EntityHolyWater

bismurphy:
There is already EntityHolyWater in DRA, but this one is quite different (which makes sense, because Richter's is very different in behavior from Alucard's). Still, there are some small overlaps here and there.

Due to the ongoing failure to match `func_80166784` (the previous rodata-using function), this is being split out to a new file, even though I'm pretty sure they're supposed to be in the same file. We will reunify the files once `func_80166784` can be solved.

There is a function which is identical in the overlays, and is only used by this function. I have gone ahead and named it identically in both overlays.


Xeeynamo (approved):
Looks good!

PR #1017, opened 2024-01-31
Decompile DRA and RIC EntitySubwpnCrashCross

bismurphy:
Similar thing to what I've had a lot of lately - I decompiled the RIC version, then went through the process of cross-referencing RIC and DRA to identify the duplicate function. There are a fair number of differences between these, so it wasn't super straightforward to go from the RIC match to the DRA match.

There's a lot of weird stuff in these, especially the "one" and "three" variables. Probably some weirdness that I didn't quite write the same as the original, but a match is a match.

Scratches here if you'd like to take a crack at any of the weird junk.
RIC: https://decomp.me/scratch/rRANW
DRA: https://decomp.me/scratch/jWF5h


src/dra/843B0.c
```diff
@@ -417,7 +417,124 @@ INCLUDE_ASM("dra/nonmatchings/843B0", EntityHolyWaterBreakGlass);
...
+        if (self->primIndex == -1) {
+            DestroyEntity(self);
+            return;
+        }
+        self->flags = FLAG_HAS_PRIMS;
+        self->posY.i.hi = 0x78;
+        self->ext.crashcross.unk80 = 1;
+        self->zPriority = 0xC2;
+        func_8011A290(self);
+        LoadImage(&D_800B0788, &D_800B06C8);
```
Xeeynamo, line 443: `D_800B06C8` is an array of data. I'd suggest to use it without `&` and change the type as `u16 D_800B06C8[0x30]`. I came up with that type as `D_800B0788` seems to suggest to load 0x30 half words in a row. It most likely contains three palettes.

Xeeynamo (approved):
left a comment about one of the `LoadImage` parameters that can be applied on both DRA and RIC versions. I couldn't understand the `one` and `three` issue and why it would match like that. I will be happy to accept any further de-faking in the future, but this looks good as it is already.

PR #1023, opened 2024-02-02
Decompile RIC EntitySubwpnCrashAgunea

bismurphy:
Nothing too special to share, but the array in the entity extension is weird. Especially the handling at the bottom, with `arr0`, `arr1`, and `arr2`. But another match! I think RIC is down to like 12 functions which is wild!


src/ric/2C4C4.c
```diff
@@ -829,7 +829,224 @@ void EntitySubwpnCrashCrossParticles(Entity* self) {
...
+        sp18 &= 3;
+        self->animFrameIdx = sp18;
+        self->animFrameDuration = 2;
+    } else {
+        self->animFrameDuration--;
+    }
+    sp10 = 0;
+    prim = &g_PrimBuf[self->primIndex];
+    prevPrim = prim;
+    sp18 = (((u32)g_GameTimer >> 1) & 1) + 0x1AB;
```
sozud, line 973: I think this cast can be removed

sozud (approved):

PR #1024, opened 2024-02-02
Decompile ST0 func_801AF6D0

ser-pounce:
Getting some easy stuff from my earlier decomp.me backlog out of the way.﻿


sozud (approved):
My opinion is that this is a little more clear https://decomp.me/scratch/8YfDq
This function is odd, @Xeeynamo will probably have ideas to improve it

ser-pounce:
Much better, thank you both for the suggestions.

PR #1027, opened 2024-02-03
Decompile RIC EntitySubwpnReboundStone

bismurphy:
I did the DRA version of the rebound stone a little while ago; this one is quite similar.

The rebound stone line is drawn using a series of PRIM_LINE_G2 primitives. It turns out that the new PrimLineG2 variant in `primitives.h` matches perfectly. The DRA version was using `LOH()` macros, but now it is clear that we don't need that, now that we understand the alternate primitive. It's very cool to find multiple places in the code using this struct as an alternative to the normal Primitive.

This also has two identical helper functions, so I have renamed those to match the DRA versions as well.

### DIFF
--- a/src/dra/86ECC.c
+++ b/src/dra/86ECC.c
@@ -484,7 +484,7 @@ void EntitySubwpnReboundStone(Entity* self) {
     Collider collider;
     u16 playerX;
     u16 playerY;
-    Primitive* prim;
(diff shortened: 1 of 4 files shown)


Xeeynamo (approved):
Very nice catch! RIC is almost done. Thank you :pray:

PR #1028, opened 2024-02-03
Decompile RIC EntitySubwpnThrownVibhuti

bismurphy:
Another one that is similar to, but subtly different from, the DRA version. Might be interesting to do a diff between these corresponding functions sometime to fully describe the differences between Richter and Alucard's behavior.


Xeeynamo (approved):
All good!

PR #1029, opened 2024-02-03
Decompile RIC EntitySubpwnAgunea

bismurphy:
Another one very similar to a DRA function.

Noticed the DRA one still had a placeholder `func_` name so I've also renamed that.


Xeeynamo (approved):
they don't stop coming

PR #1033, opened 2024-02-04
Decompile RIC func_8016DF74

bismurphy:
Big file shuffle to get this one working, but it works!

I have now made some degree of attempt at every function in RIC. Now we circle back to finish the hard ones!


Xeeynamo (approved):
LGTM

PR #1034, opened 2024-02-04
DRA RenderEntities (non matching)

Xeeynamo:
Another disappointing non-match, even closer than #1008 . There is only one registry swap in one single line. I've been permuting for a day without success. Putting this here if anyone wants to give it a try: https://decomp.me/scratch/AuSvv


bismurphy:
Looking into it

Xeeynamo:
> Looking into it
> 
> Edit: Behavior replicated locally.
> 
> Edit 2: This appears to be related to the fact that we are creating:
> 
> ```
> typedef struct {
>     void (*D_8013C000)(void);
>     void (*D_8013C004)(void);
>     void (*D_8013C008)(void);
>     void (*D_8013C00C)(void);
> } PlayerOvl;
> ```
> 
> But are also declaring `g_PlOvlAluBatSpritesheet = 0x8013C004;` so the assembly gets confused. `make force_extract` produces `func_800F298C` with
> 
> ```
> /* 52AB0 800F2AB0 1480023C */  lui        $v0, %hi(g_PlOvlAluBatSpritesheet)
> /* 52AB4 800F2AB4 04C0428C */  lw         $v0, %lo(g_PlOvlAluBatSpritesheet)($v0)
> /* 52AB8 800F2AB8 05006414 */  bne        $v1, $a0, .L800F2AD0
> /* 52ABC 800F2ABC 00000000 */   nop
> /* 52AC0 800F2AC0 09F84000 */  jalr       $v0
> /* 52AC4 800F2AC4 01000434 */   ori       $a0, $zero, 0x1
> /* 52AC8 800F2AC8 BACA0308 */  j          .L800F2AE8
> /* 52ACC 800F2ACC 00000000 */   nop
> ```
> 
> Which is not correct - it should be calling something like `%hi(g_PlOvl_D_8013C004)` - since it's a call to `g_PlOvl.D_8013C004`.
> 
> I think this issue relates to overlapping addresses getting the wrong symbols in the extraction. Not certain what the root is exactly, but the assembly is getting the wrong symbol for the call at 0x8013C004.

The problem is `8013C004` is used for both function calls (`BIN/RIC.BIN`) or as a pointer to data (`BIN/ARC_F.BIN`). Either we destroy the current `PlayerOvl` struct and use an array of pointers instead or we find another solution that I currently cannot see.

sozud:
Can we have the tool ignore usages of g_PlOvlAluBatSpritesheet?

bismurphy:
Sure, I can make that a special case for it to test for. Hopefully longer term we can figure out a way to make the extraction use the right symbols in the right places, but for now I'll change it.

PR #1035, opened 2024-02-04
Allow spritesheet to be extracted from the CLI

Xeeynamo:
A side-project I've been working on that helped me understanding some structures that are now a bit better documented in #1034 . Richter sprites are extracted as usual. Doing `python3 tools/splat_ext/spritesheet.py decode disks/us/BIN/ARC_F.BIN assets/arc alucard disks/us/BIN/F_GAME.BIN 1` can extract Alucard sprites too. There is no repacking yet as I want to avoid using splat for this one.


sozud (approved):
Some docs explaining what the spritesheet format is might be helpful, otherwise seems good

Xeeynamo:
> Some docs explaining what the spritesheet format is might be helpful, otherwise seems good

I am still in the process of understanding how things are interconnected. That is why I did not make any documentation nor a full packer.

PR #1038, opened 2024-02-06
bandaid fix for analyze_calls issue in renderentities PR

bismurphy:
Note that this is a PR into the renderentities branch, not into master.

This seems to be an acceptable fix for now.


Xeeynamo:
Thank you for taking care of it promptly :pray:

PR #1041, opened 2024-02-06
Decompile ST0 func_801ABBBC

bismurphy:
First function of my new effort to make progress in ST0.

There's a fun bug in case 5 where they forgot to increment their prim.

This required the creation of a new variety of Primitive, since there's strong evidence of an `f32` at offset 0x20. For now it's draculaPrimitive, but maybe we'll find more uses of it in the future.

I think those are the main highlights. No pressure to merge this one quickly, I'm expecting progress on ST0 to be slow.


sozud (approved):
I think @Xeeynamo should give an opinion on draculaPrimitive but otherwise looks good!

Xeeynamo (approved):
Another `PRIM_LINE_G2` shenanigan! LGTM

PR #1046, opened 2024-02-08
_SsSndPause, sync header name

sozud:
The old compiler needs 8.3 filenames, so libsnd_internal is renamed.


sozud:
I think this isn't too controversial so just going to merge so I can continue syncing the changes

PR #1049, opened 2024-02-08
Probably fix main funcs not appearing in function report

sozud:


Xeeynamo (approved):
Not sure if this is gonna work but let's try

PR #1055, opened 2024-02-11
Fix reversed dst and src for cd_memcpy

rossimo:


Xeeynamo (approved):
You can simplify the function by removing the temporary variables and by inlining the `+= 1` like the following:

```c
    for (i = size - 1; i != -1; i--) {
        *dst++ = *src++;
    }
```

PR #1057, opened 2024-02-12
_SsClose

sozud:


src/main/psxsdk/libsnd/ssclose.c
```diff
@@ -1,7 +1,33 @@
 #include "common.h"
 #include "libsnd_i.h"
 
-INCLUDE_ASM("main/nonmatchings/psxsdk/libsnd/ssclose", _SsClose);
```
Xeeynamo, line 4: You can just use a `for` loop: https://decomp.me/scratch/LEADb

also `var_a2` can be renamed `i`

src/main/psxsdk/libsnd/vmanager.c
```diff
@@ -189,7 +193,21 @@ short SsUtChangePitch(short voice, short vabId, short prog, short old_note,
     return -1;
 }
 
-INCLUDE_ASM("main/nonmatchings/psxsdk/libsnd/vmanager", SsUtChangeADSR);
+s16 SsUtChangeADSR(
```
Xeeynamo, line 196: Maybe it's a personal preference, but maybe using a guard clause makes it more readable? https://decomp.me/scratch/tCdUq

Also notice the use of `LEN(_svm_voice)` instead of the `24`

Xeeynamo (approved):
It looks good. I just left a couple of suggestions.

src/main/psxsdk/libsnd/vmanager.c
sozud: I added NUM_SPU_CHANNELS since there's a bunch of different arrays of that length. I think the nested way is more like how my brain thinks but if others prefer the early returns we can do that.

PR #1061, opened 2024-02-16
libspu get/set vol

sozud:


Xeeynamo (approved):
I left some minor comments, otherwise it looks good

PR #1062, opened 2024-02-17
Spu note2pitch

sozud:


src/main/psxsdk/libsnd/vmanager.c
```diff
@@ -67,7 +67,20 @@ void SpuVmNoiseOn(s32 arg0, s32 arg1) {
 
 INCLUDE_ASM("main/nonmatchings/psxsdk/libsnd/vmanager", SpuVmPBVoice);
 
-INCLUDE_ASM("main/nonmatchings/psxsdk/libsnd/vmanager", SpuVmPitchBend);
+s32 SpuVmPitchBend(s16 arg0, s16 arg1, s16 arg2, s16 arg3) {
+    s16 temp_v0;
+    s32 var_a0;
```
Xeeynamo, line 72: This is unused, it can be deleted

src/main/psxsdk/libspu/s_ini.c
```diff
@@ -1,7 +1,59 @@
...
+void _SpuInit(s32 arg0) {
+    u16* var_v0;
+    s32 var_v1;
+    u16 temp;
+
+    ResetCallback();
+    _spu_init(arg0);
+    if (arg0 == 0) {
+        temp = 0xC000;
+        for (var_v1 = 0x17, var_v0 = &_spu_voice_centerNote[0x17]; var_v1 >= 0;
```
Xeeynamo, line 38: `_spu_voice_centerNote` seems to be a fake symbol. I re-worked the for loop in a different way which I believe it is closer to how the original code would have looked like: https://decomp.me/scratch/F3iqN

src/main/psxsdk/libspu/s_n2p.c
```diff
@@ -1,5 +1,42 @@
...
+    s16 var_v1_2;
+    u16 sample_rate;
+    s32 pos;
+    var_v0 = offset_note_frac + center_note_frac;
+    if (var_v0 < 0) {
+        var_v0 += 7;
+    }
+    var_a1 = var_v0 >> 3;
+    var_v1_2 = 0;
+    if (var_a1 >= 16) {
```
Xeeynamo, line 20: You can convert the if/do/while into just a while loop to save lines of code and nesting: https://decomp.me/scratch/mr9Ya

Xeeynamo (approved):
This is very interesting stuff you decompiled in there. I left some suggestions to further uplift the code quality, but this is really getting in good shape! Great job

src/main/psxsdk/libspu/s_ini.c
sozud: This object is fully matched https://github.com/sozud/psy-q-decomp/blob/master/src/spu/s_ini.c
The symbol is at offset 40 (0x28) so I believe it's right. The loop change matches in the original so I'll fix that.
```
section .data symbol name _spu_voice_centerNote offset 40
```

PR #1063, opened 2024-02-17
DRA func_801309B4, func_80130E94, func_8013136C

Xeeynamo:
Completes `90264.c` with the exception of the BSS section.

After this PR I am thinking to either decompile and move `func_8012F178`, `func_8012F83C` and `func_8012F894` into `90264.c` as they all follow the same pattern, or move back `90264.c` before the split back into `8D3E8.c` as all the code seem to be related to the wolf form.


src/dra/90264.c
```diff
@@ -227,103 +227,392 @@ void func_80130618(Entity* self) {
...
-    entity->palette = PLAYER.palette;
+    self->palette = PLAYER.palette;
     func_8012C600();
 }
-#endif
 
 // When Alucard uses the cross subweapon for 100 hearts.
 // Entity ID 7, blueprint #7 (this is a coincidence)
 void EntityGiantSpinningCross(Entity* self) {
+    static const SVECTOR D_800E2024 = {0, 0, 0};
```
sozud, line 614: How did you determine these are static and local to the function?

sozud (approved):

PR #1065, opened 2024-02-18
Pass undefined_syms files to linker

mkst:
Creating a PR for this, will look at updating splat in a separate PR


Xeeynamo (approved):
This looks in a much better shape. Thanks for doing this!

PR #1067, opened 2024-02-18
weapon rodata migration

mkst:
this is a step along the way to getting splat updated, but im not there yet :)


Xeeynamo (approved):
Needs a `make format`. But otherwise it looks good!

PR #1071, opened 2024-02-19
_SsGetMetaEvent

sozud:
Thanks to an anonymous person who fixed this scratch https://decomp.me/scratch/Aw4jQ


Xeeynamo (approved):
Left minor comments for readability.

One important change to consider is for `SeqStruct.unk8` to be changed from `s32` to `u8*`, if this function confirms it to be a pointer, it will most likely going to crash if the code is compiled for 64-bit architectures.

PR #1081, opened 2024-02-24
Render the explored castle map

Xeeynamo:
image

This. It still misses the texture and I do not know what the green lines are about. All the stuff within the `#else` and `#endif` are placeholders that will hopefully be fixed later on the same way.

To get the same result from the screenshot I had to use the [func_800F1B08](https://decomp.me/scratch/nW7lV) scratch, comment  `g_CastleMap[i] = 0;` from `InitStatsAndGear` and use the following to read the map from a memory card's save:

```c
if (!FileRead(LoadMemcardSave, "memcard1.mcd")) {
    WARNF("unable to correctly load the save file");
}
```
```c
s32 LoadSaveData(SaveData* save);
bool LoadMemcardSave(FILE* f) {
    const int slot = 0;
    SaveData save;

    fseek(f, 0x2000 + slot * 0x2000, SEEK_SET);
    fread(&save, sizeof(SaveData), 1, f);
    return LoadSaveData(&save) == 0;
}
```

### DIFF
--- a/include/game.h
+++ b/include/game.h
@@ -161,6 +161,7 @@ typedef struct Prim {
 #define WEAPON_0_END (WEAPON_1_START - 1)
 #define WEAPON_1_START 0xF0
 
+#ifndef VERSION_PC
 #define RIC_PRG_PTR 0x8013C000
 #define FAMILIAR_PTR 0x80170000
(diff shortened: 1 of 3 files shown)


sozud:
Nice work!

PR #1084, opened 2024-02-25
Migrate rodata to functions for dra

mkst:


Xeeynamo:
The reason why we have comments right next to the `.data` and `.rodata` segments in the splat config is to easily allow merging and splitting C files. And it might be helpful whenever we decide to import the same overlay from different version of the game.
If we lose such information I think it would be very hard to do any of the above. Unless there's something I am missing.

mkst:
For the rodata splits, you can change from `.rodata` to `rodata` (and disable migrate_rodata_to_functions) to get the assembly version of the rodata out if you wanted to. I've added the .data sections back in (although the previous statement applies there too). 

Once a file is in .c, you'll struggling to identify the jump tables (without splatting the c file as asm instead), so having the name of the jumptable in the yaml isnt super helpful ( i think? ).

Xeeynamo (approved):
Thanks Mark for the incredible work!

PR #1085, opened 2024-02-25
Decomp w_001 func_ptr_80170008

rossimo:


Xeeynamo (approved):
I left a few comments to get the function in a slightly better shape, but overall it looks really good. Congrats! I realise most of the suggestions I left are not obvious as we lack of proper documentation and an in-depth tutorial. So do not feel discouraged by the comments :wink:

PR #1087, opened 2024-02-25
Decompile w_045's func_13F000_8017A718

rossimo:
If the control flow on this could be better, let me know.


bismurphy:
Found your scratch here (linking it can be convenient if you're asking for improvements to control flow or anything else): https://decomp.me/scratch/9hJCb

Control flow is definitely weird but none of the changes I tried to do to make sense were functional, so I think this PR is good as is.

bismurphy (approved):

PR #1090, opened 2024-03-03
Decompile few libgpu & libgte functions

dezgeg:


src/main/psxsdk/libgpu/prim.c
```diff
@@ -57,17 +57,24 @@ void SetTile1(TILE_1* p) { setTile1(p); }
...
 void SetLineG4(LINE_G4* p) { setLineG4(p); }
 
-INCLUDE_ASM("main/nonmatchings/psxsdk/libgpu/prim", MargePrim);
+int MargePrim(void* p1, void* p2) {
+    int length = getlen(p1) + getlen(p2) + 1;
+    if (length >= 0x21) {
+        return -1;
+    }
+    setlen(p1, length);
+    // Bug: this should also do *(u_long*)p2 = 0; so the link pointer of
```
sozud, line 74: Interesting info, where are you getting this from?

sozud (approved):
Looks good, thanks for contributing!

src/main/psxsdk/libgpu/prim.c
dezgeg: A game using PSY-Q 4.2 had this function otherwise equivalent except for that extra zero write. After reading bit more about how the GPU and its DMA works it seems clear that it's indeed a bugfix.

Xeeynamo: I wonder if it would be beneficial to start doing stuff such as:
```c
#if PSYQ_VER >= PSYQ(4, 2, 0)
*(u_long*)p2 = 0;
#endif
```
although it would be currently impossible to test it as we do not check the checksum with any libs other than 3.5.

sozud: I've been doing some work on other psy-q versions in the other repo although it's only 3.3 and 3.6 at the moment https://github.com/sozud/psy-q-decomp/commit/101a330ca9bde7d5efed5a31f721315f6b42284c

If 4.2 is being worked on somewhere I can add that to the repo. libgpu isn't integrated yet though

PR #1094, opened 2024-03-05
SpuVmSetSeqVol

sozud:
Not sure what the deal is with this line:
```
if ((u16)_svm_voice[voice].unke == (u16)temp) {
```
changing unke to unsigned causes a mismatch in SpuVmInit


Xeeynamo (approved):
I couldn't resolve that `(u16)temp`

PR #1095, opened 2024-03-07
DRA func_8010DA70

Xeeynamo:
Thanks to the anonymous person who managed to match the last part: https://decomp.me/scratch/3mKBR


src/dra/6D59C.c
```diff
@@ -140,8 +140,40 @@ void func_8010DA48(u32 arg0) {
...
-INCLUDE_ASM("dra/nonmatchings/6D59C", func_8010DA70);
+AnimationFrame* func_8010DA70(AnimationFrame** frames) {
+    Entity* entity;
+    AnimationFrame* frame;
+    s32 idx;
+    AnimationFrame* frameSub;
+    s32 idxSub;
+
+    entity = g_CurrentEntity;
+    frame = frames[entity->ext.generic.unkAC];
```
bismurphy, line 151: Given that using `ext.generic` is forbidden, I looked into the calls for this function.

`func_8010DA70` is called only by `func_8010DBFC`, which is itself called by 15 functions (3 in dra, and 12 in various weapons; all the dra ones are decompiled and none of the weapon ones are).

In all the existing decompiled functions, `g_CurrentEntity` is set to `&PLAYER` prior to the function call; to me this is enough to be willing to change this from `ext.generic` to `ext.player`, but if you want to wait to see all the calls (to see if there are any non-player uses) before locking this in, that's also reasonable. But I think it's worth it to avoid using `ext.generic`.

bismurphy (approved):

src/dra/6D59C.c
Xeeynamo: Very good point, thanks for noticing that! I replaced it with `.player`. We can change it later on if we find it is not the appropriate one.

sozud (approved):

PR #1098, opened 2024-03-13
Ss start

sozud:
This is an update of the previous PR. Not sure how to get rid of the get20 function.
https://decomp.me/scratch/cDYU4


Xeeynamo (approved):
no worries about `get20`, at this rate I think it is just real. Good one, thanks for doing this!

PR #1099, opened 2024-03-14
Correct the split of libapi/l10.c

dezgeg:
Correct the split of libapi/l10.c and decompile few trivial functions in it.

### DIFF
--- a/config/splat.us.main.yaml
+++ b/config/splat.us.main.yaml
@@ -35,7 +35,7 @@ segments:
       - [0x8F8, .rodata, psxsdk/libgpu/prim]
       - [0x9C8, .rodata, psxsdk/libgpu/sys]
       - [0xBC4, rodata, psxsdk/libapi/a22] # aVsyncTimeout
-      - [0xC08, .rodata, psxsdk/libapi/l10]
+      - [0xC08, .rodata, psxsdk/libetc/intr]
       - [0xC40, .rodata, psxsdk/libetc/intr_dma]
       - [0xC6C, .rodata, psxsdk/libc/sprintf]
       - [0xD48, .rodata, psxsdk/libcd/event]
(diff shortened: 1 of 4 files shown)


include/psxsdk/libetc.h
```diff
@@ -2,9 +2,9 @@
 #define LIBETC_H
 
 struct Callbacks {
-    void (*unk0)(void);
+    const char* rcsid; /* "$Id: intr.c,v 1.73 1995/11/10 05:29:40 suzu Exp $" */
```
sozud, line 5: I'm not sure about this one. Looking at startIntr https://decomp.me/scratch/MPlxQ
we have
```
        _96_remove(D_8002D340);
```
The id string is most likely `static const char*` at the top of the `.c` file based on it not having a symbol and being in the rodata section of the orignal object

```
~section name .rdata
rdata section len 105
24 49 64 3A 20 69 6E 74 72 2E 63 2C 76 20 31 2E 
37 33 20 31 39 39 35 2F 31 31 2F 31 30 20 30 35 
3A 32 39 3A 34 30 20 73 75 7A 75 20 45 78 70 20 
24 00 00 00 75 6E 65 78 70 65 63 74 65 64 20 69 
6E 74 65 72 72 75 70 74 28 25 30 34 78 29 0A 00 
69 6E 74 72 20 74 69 6D 65 6F 75 74 28 25 30 34 
78 3A 25 30 34 78 29 0A 00 
~section name .text
```

config/splat.us.main.yaml
```diff
@@ -35,7 +35,7 @@ segments:
       - [0x8F8, .rodata, psxsdk/libgpu/prim]
       - [0x9C8, .rodata, psxsdk/libgpu/sys]
       - [0xBC4, rodata, psxsdk/libapi/a22] # aVsyncTimeout
-      - [0xC08, .rodata, psxsdk/libapi/l10]
+      - [0xC08, .rodata, psxsdk/libetc/intr]
```
sozud, line 38: I believe these splits are correct. BTW I imported the current libetc work into the other repo, there's various fixes like importing all the data and fixing some static / non-static differences. https://github.com/sozud/psy-q-decomp/tree/master/src/etc

include/psxsdk/libetc.h
dezgeg: Yes, the string itself is stored in rodata. But first field of the struct is pointer to that string.

dezgeg: That `_96_remove(D_8002D340);` is weird though given its prototype is `extern void _96_remove(void);`...

dezgeg: If I looked it right, very first thing _96_remove() does is EnterCriticalSection() which trashes a0, so it doesn't end up used.

sozud: Ok, thanks for looking into it. Either way I think this is good for now, when the file is complete I can migrate it to the other repo and the obj comparison will make any issues clear.

sozud (approved):
Nice work, thanks!

PR #1100, opened 2024-03-15
Update maspsx

mkst:
Also tweaked the Dockerfile since this is what I use to build the project.


sozud (approved):
Looks good to me, thanks!

PR #1103, opened 2024-03-23
_SsSndDecrescendo

sozud:
Thanks to @dezgeg @mkst @Xeeynamo


Xeeynamo (approved):
Congrats everyone to get this one matching! I tried removing that `goto` but without success. It looks already very good as it is

PR #1106, opened 2024-04-02
TT_000 ProcessEvent matching

Xeeynamo:
Last function from tt_000. Given what the function does I am pretty sure it is also re-used by the other SERVANT overlays.


bismurphy (approved):
Wow! The code for this function looks really nice and extremely readable. Cool to see how much it improved beyond the simply-working scratch. Awesome to have tt_000 done!

PR #1107, opened 2024-04-02
Create and use entity flag FLAG_DEAD

bismurphy:
Due to the discovery of the 0x100 entity flag in EntityDraculaFinalForm, I have gone ahead and created the #define for 0x100 being FLAG_DEAD. I have also searched the codebase for the string `flags & 0x100` and `flags |= 0x100` and made suitable replacements. There are likely other places that use the 0x100 flag which use different syntax, and those are not changed, but hopefully we can clean those up over time as we discover them.

Nice to expand our knowledge of the different available flags!

### DIFF
--- a/include/game.h
+++ b/include/game.h
@@ -204,6 +204,8 @@ extern u8 g_BmpCastleMap[0x20000];
 #define FLAG_DRAW_UNK80 0x80
 
 // Flags for entity->flags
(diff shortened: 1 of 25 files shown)


Xeeynamo (approved):
This is awesome. Great finding!

PR #1111, opened 2024-04-04
Add TT_000 from the HD version

Xeeynamo:
100% decompiled.

`US` adds some padding to have the file exactly 40KB long. The HD version does not add any padding, hence why I ended up producing `tt_000_raw.bin`. I would have used just `tt_000.bin` as a name as found in `disks/pspeu/PSP_GAME/USRDIR/res/ps/hdbin/tt_000.bin`, but on Windows-based file system it would collide with `TT_000.BIN` due to the OS having case insensitive names.

I modified `make clean` as I found annoying that `VERSION=hd make clean` would wipe out `us` build stuff.

The only different function in HD is `ProcessEvent`, which has a weaker check. Another hint suggesting HD being older than US.

`s32 _unused[26];` added enough padding in the bss section to get an :ok: . I am pretty sure it is unused data because the final binary is not aligned by any power of 2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,7 @@ endef
 all: build check
 build: build_$(VERSION)
 build_us: main dra weapon ric cen dre mad no3 np3 nz0 sel st0 wrp rwrp tt_000
-build_hd: dra
+build_hd: dra tt_000
 clean:
 	git clean -fdx assets/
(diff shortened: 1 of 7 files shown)


bismurphy (approved):
Awesome! Glad they were so similar and easy to handle the decompiling!

PR #1113, opened 2024-04-20
Add TT_000 overlay from PSP

Xeeynamo:
TT_000 is the first overlay from PlayStation 1 that we are now able to compile from the source and produce a 1:1 binary. This lead me to start exploring the same overlay from the game Castlevania: Dracula X Chronicles, which contains a PSP re-build of Symphony of the Night.

This PR adds all the infrastructure to add the same flow for a PSP matching decomp. Use `export VERSION=pspeu` and then the usual `sotn` command to splat the overlay, build it and check if it matches. Running `make extract_disk` should not be necessary as the same ISO used from `VERSION=hd` is also used for `pspeu`, so you will probably have it extracted already.

Some important notes about the PSP build:
* The whole PSP build seems to be compiled with `-O0`, which makes it much easier to decompile
* Having ŧhe PSX, PSP and Saturn builds will allow to easily cross-reference the code and reduce fake matches
* `disks/pspeu/PSP_GAME/USRDIR/res/ps/hdbin/tt_000.bin` is the HD PSX build
* `disks/pspeu/PSP_GAME/USRDIR/res/ps/PSPBIN/tt_000.bin` has the same code from the HD build, but for PSP
* `disks/pspeu/PSP_GAME/USRDIR/res/ps/PACK_E/TP00.BIN` is the same as above, but it packs both overlay and graphics. This is the file the PSP game seems to actually use
* The PSP build uses the Metrowerk CodeWarrior's C compiler, which is very different from the GCC one used on PSX.
* Thanks to @mkst lead, we found a way to still use the GNU assembler and linker
* MWCC uses [wibo](https://github.com/decompals/WiBo/), a think compatibility layer to run Windows CLI tools on Linux. It is much more lightweight than Wine.
* MWCC does not support the `INCLUDE_ASM` dialect, so the custom pre-processor `tools/mwcpp` had to be created
* The exact MWCC compiler version is unknown, but I suspect it is `build 147`
* I am not yet sure of any implications for using GNU AS and GNU LD instead of the MW correspondent tools
* Not all the functions can be correctly disassembled, spimdisasm will just produce a bunch of `.word 0x________` due to the in-progress effort of adding the Allegrex-specific opcodes support

---

TO-DO list before marking the PR as ready:
- [X] Add PSP build to the CI
- [x] Add progress reporting to the PSP build
- [x] Integrate source file from `src/servant/tt_000_psp` to  `src/servant/tt_000` to promote the psp build as first-citizen

---

TO-DO in a follow-up PR:
 * Figure out what `header` is: can we extract it as assembly code? or maybe as custom re-compilable asset via splat? Is it a MW stuff or a Castlevania-specific file?
 * Get rid of the last line in `Makefile.psp.mk`


sozud (approved):
Looks good to me, awesome work!

PR #1114, opened 2024-04-21
Un-fake some previous matches for the psp build

Xeeynamo:
This improves the function readability from the already decompiled PSX functions. Since the PSP build is compiled with `-O0`, the statement order is much clearer. Even `if (entity->facingLeft != 0)` and `if (!entity->facingLeft)` produces different output on the PSP build. I was able to remove a couple of temp variables too.

---

There are a few functions I cannot match due to a trailing `nop` at the end of the function. It looks like the overlay have functions aligned by `8`. I do not know how to trigger that using MWCC or if it is a linker thing. Two functions that have this problem are `DestroyEntity` and `func_801746A0`.

image


sozud (approved):
Much nicer

sozud:
Does `subalign: 8` in the splat yaml help with the alignment?

Xeeynamo:
> Does `subalign: 8` in the splat yaml help with the alignment?

I tried and that unfortunately did not work :/

PR #1120, opened 2024-05-11
Weapon 051

Xeeynamo:
Helping Mazen with his weapon modding


bismurphy (approved):

src/weapon/w_051.c
```diff
@@ -4,11 +4,435 @@
...
+        break;
+    }
+    self->ext.generic.unkAC = PLAYER.ext.player.unkAC - props->unk0C;
+    if (g_api.UpdateUnarmedAnim(props->frameProps, props->frames) < 0) {
+        DestroyEntity(self);
+        return;
+    }
+    if (PLAYER.animFrameIdx == 1 && PLAYER.animFrameDuration == 1) {
+        g_api.CreateEntFactoryFromEntity(
+            self, ((g_HandId + 1) * 4096) | 0x3E, 0);
```
sozud, line 426: Does the factory macro work here?
```
#define FACTORY(A0, BLUEPRINTNUM) (A0 << 8 | BLUEPRINTNUM)
```

sozud (approved):

src/weapon/w_051.c
Xeeynamo: 0x3E is BLUEPRINTNUM. But that `g_HandId` wouldn't fit the current macro unfortunately. I was about to submit `func_ptr_80170008` too and it has something like `g_api.CreateEntFactoryFromEntity(
            self, ((g_HandId + 1) << 0xC) +  (((self->ext.factory.unk7E + 1) << 0x10) + 0x3E), 0);`, which is even more complicated and involves more bitfields. I'm staying away from documenting it from the time being.

bismurphy: Same is used a couple places in w_029, so there's precedent for this. Factories work well when they're simple but fail when we need to take in parameters that are dependent on external values like g_HandId.

PR #1121, opened 2024-05-12
Decompile weapon 15 func_ptr_80170014

bismurphy:
There are only a few weapons that make use of the func_ptr_80170014 function, and this is the first of them.

I don't really know how to pull in data for functions, so keeping these as externs for now. Particularly, I can't find where in the assembly the bytes of D_6D000_8017A2B0 exist. They don't appear to be in asm/us/weapon/data.

We identify a couple members of the `ET_Weapon` struct, and since they're a change to velocity, I named them acceleration. This breaks weapon 29's code (it uses different datatypes at offset 0x9C), but since there's already an ET_Weapon29 dedicated alternative entity, we can continue to use that instead.

Been a while since I've done a PR, nice to pick it back up, hope this is good :)


sozud (approved):

Xeeynamo:
>  can't find where in the assembly the bytes of D_6D000_8017A2B0 exist

It is exported as a file:
```yaml
- [0x6D2B0, animset, w_015_2]
```

PR #1124, opened 2024-05-13
Adjust func_8011A4D0 to improve readability and symbol

bismurphy:
While tracking the calling of the weapon functions, I made a neat discovery.

func_8011A4D0 is a crucial function at the heart of the game's engine. For every entity which is created by its ID, this function assigns the PfnEntityUpdate function which will run each frame for that entity.

If the ID is E0-EC, or F0-FC, it reads a function from a pointer to call. It turns out that the array we were indexing into before is not a real array, because of course it does not have elements below E0. When you take the minimum offsets into consideration, it turns out that these arrays are located precisely where the two weapons (in the player's two hands) are located in memory.

Because D_8017A000 and D_8017D000 are Weapon objects, we can't index into them as an array, so I had to do a bit of an ugly cast to a `PfnEntityUpdate*`, unfortunately I don't know a way to do this under our 80-character limit.

Neat to discover fake symbols and clean them up, as well as to make it obvious that these two lines are dealing directly with the weapon functions.


Xeeynamo:
Great catch!! It finally makes a bit more sense

PR #1125, opened 2024-05-13
Small logic cleanup on func_8010EDB8

bismurphy:
I'm working on tracing the logic involved in using weapons. `func_8010EDB8` is a core part of that process.

I noticed a small improvement that could be made to the control flow, removing one `goto` statement.

This function could use more work on its overall control flow (there are so many uses of `goto`), but even a small improvement is still an improvement.

### DIFF
--- a/src/dra/6D59C.c
+++ b/src/dra/6D59C.c
@@ -868,14 +868,10 @@ block_32:
         goto block_38c;
     }
     var_s2 = equipped_item->specialMove;
-    if (var_s2 == 0) {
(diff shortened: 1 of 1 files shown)


sozud (approved):
This seems like an improvement but I'm wondering if we should check the PSP version.

Xeeynamo:
DRA PSP symbols are still not cross-referenced, so we would need to hunt this function first. I think this is fine as it is, even knowing we would need to probably come back to it later.

PR #1127, opened 2024-05-14
Support Single BIN CUE Files

hohle:
In CUE files `FILE` is a stateful, global declaration that applies to all following `TRACK`s. `sotn-disk` was treating `FILE` declarations as `TRACK` delimiters which would result in incorrect parsing of CUE files with a single `FILE`, but multiple tracks.

Now when reading a `FILE` declaration, the path is stored and processing continues. A `TRACK` declaration will use the previously defined path, and if a previous track had been started, append that previous track to the list.

This also builds `sotn-disk` using the local repository instead of pulling the latest commit from GitHub. The target depends on `sotn-disk` sources and will rebuild as necessary (or with `make ~/go/bin/sotn-disk`).

As an aside, [pull/232](https://github.com/Xeeynamo/sotn-decomp/pull/232) ran into this same error, but fixes in a slightly different way that leaves side effects that may make supporting things like `INDEX` (for extracting the placeholder audio, for example) more error prone in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,6 +105,8 @@ define link
 		-T $(CONFIG_DIR)/undefined_funcs_auto.$(VERSION).$(1).txt
 endef
 
+SOTNDISK_SOURCES := $(shell find tools/sotn-disk -name '*.go')
+
 .PHONY: build
 
 all: build check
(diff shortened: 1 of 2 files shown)


Xeeynamo (approved):
The changes are very well welcomed. And it was silly of me to have the Makefile to always build the version from GitHub rather than locally.

Thanks for your contribution!

PR #1128, opened 2024-05-14
Decompile w16 func_ptr_80170014; change type

bismurphy:
This is the last of the weapons using func_ptr_80170014. It needs a non-void return type in order to match, so we also had to change the type for all the other instances so we don't get issues with declaration mismatches.


Xeeynamo (approved):
Looks good and the change on the return type is a safe bet!

PR #1130, opened 2024-05-15
Decompile weapon 21 helper function

bismurphy:
As noted in the comment, this is closely related to BottomCornerText, which exists in the overlays. It shows the text boxes in the lower left and lower right corners of the screen (an item you picked up, an enemy you encountered, etc).

I'm not sure how this is triggered in relation to food (I'll see when I decompile the caller), but there it is. The logic is weird since it parses characters two at a time.

I tried to make a symbol definition for this function, but it didn't work. Leaving my attempt in place to show what I did; I can remove it if desired, or do whatever is needed to rename functions in the weapon overlays. Anything works.

### DIFF
--- a/config/symbols.us.weapon.txt
+++ b/config/symbols.us.weapon.txt
@@ -1,1 +1,2 @@
 g_Weapon = 0x80170000;
+BottomCornerText = 0x8017AB54;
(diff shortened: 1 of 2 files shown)


Xeeynamo (approved):
I understand the point of using a `#define`. It is a bit confusing yet a smart solution when wanting to give context to variables that has to be reused to match.

The reason why two characters are processed when the first one is `0xFF` is due to the Japanese text parsing.

PR #1131, opened 2024-05-15
Add mwccgap

mkst:
OK. Here goes. Version 0.0.1 alpha of the MWCC Global Assembly Processor (mwccgap). It's currently very simple/limited, but it appears to work for `src/servant/tt_000/10E8.c`.

There is lot more that can be done to improve mwccgap - i.e. supporting .rodata migration would be a good addition, but let's see how far we can get with it in it's current state.

Note that the Makefile could do with some improvements - we don't nede to use mwccgap for any C file that *dont* have INCLUDE_ASM macros (it's a waste of time) so these could be ignored, i.e. for SSSV I do something like this to find the files that need fixing up:
```
GLOBAL_ASM_C_FILES := $(shell $(GREP) GLOBAL_ASM $(SRC_DIR) </dev/null 2>/dev/null)
```
.. although this is perhaps too simple given that SOTN has a mix of PSP and PSX functions (and therefore there may be INCLUDE_ASM for a PSX function but none for PSP functions...


Xeeynamo:
I think the build error is because the submodule is not getting cloned. We have `submodules: false` in the CI YAML to not pull unnecessary dependencies all the times, so we usually do the following in the `Makefile`:

```
MWCCGAP_DIR       := $(TOOLS_DIR)/mwccgap
MWCCGAP_APP       := $(MWCCGAP_DIR)/mwccgap.py
MWCCGAP           := $(PYTHON) $(MWCCGAP_APP)

[...]

$(MWCCGAP_APP):
	git submodule init $(MWCCGAP_DIR)
	git submodule update $(MWCCGAP_DIR)

[...]

$(PSP_BUILD_DIR)/%.c.o: %.c bin/mwccpsp.exe $(MWCCGAP_APP)
[...]
```

it's a bit verbose, but it works :'D

sozud:
I haven't looked at the source but I'm guessing the tool needs to make the output directory hierarchy if it doesn't exist?

```
Exception processing src/servant/tt_000/608.c: [Errno 2] No such file or directory: 'build/pspeu/src/servant/tt_000/608.c.o'
Traceback (most recent call last):
  File "/home/runner/work/sotn-decomp/sotn-decomp/tools/mwccgap/mwccgap.py", line 24, in main
    process_c_file(
  File "/home/runner/work/sotn-decomp/sotn-decomp/tools/mwccgap/mwccgap/mwccgap.py", line 229, in process_c_file
    with o_file.open("wb") as f:
  File "/usr/lib/python3.10/pathlib.py", line 1119, in open
    return self._accessor.open(self, mode, buffering, encoding, errors,
FileNotFoundError: [Errno 2] No such file or directory: 'build/pspeu/src/servant/tt_000/608.c.o'
```

mkst:
I guess I never ran from a completely clean build. Will fixup and push a new commit

mkst:
The "issue" was that the CI runs with `-j` and I was running locally single-threaded - and I was missing a mkdir instruction for any files that *didnt* have any INCLUDE_ASM macros present. The belt and braces fix was:
1. add the mkdir command for the scenario where there is no INCLUDE_ASM
2. add the mkdir command to the Makefile
3. dont use mwccgap if no INCLUDE_ASM is found when greping the source file

mkst:
Just made a small change, I realised the `asm/pspeu` path was hardcoded, so I've pulled it out as an argument.

sozud (approved):
Seems good to me but Xeeynamo probably needs to give final approval

mkst:
I fired up asm-differ and noticed that when MWCC fails to build it writes errors to stdout, but asm-differ needs them on stderr to be displayed - so fixed that. Also noticed that you guys were on an old version of asm-differ, so bumped that and set the arch accordingly (so `seh` instructions are parsed correctly).

Xeeynamo (approved):

PR #1132, opened 2024-05-15
Decompile `libgs` Wrappers

hohle:
`GsInitVcount`, `GsGetVcount`, `GsClearVcount` were all small wrappers around calls to respective root counter functions using the vsync counter spec. This provides implementations for all, replacing the previous ASM imports.

### DIFF
--- a/include/psxsdk/libapi.h
+++ b/include/psxsdk/libapi.h
@@ -82,5 +82,8 @@ extern long EnableEvent(long);
 extern void _96_remove(void);
 extern long SetRCnt(unsigned long, unsigned short, long);
 extern long StartRCnt(unsigned long);
+extern long GetRCnt(unsigned long);
+extern long StopRCnt(unsigned long);
+extern long ResetRCnt(unsigned long);
 
 #endif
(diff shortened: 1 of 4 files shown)


hohle:
I'm poking around trying to get familiar with the codebase, build system, and tooling.

bismurphy (approved):
Looks good! Especially nice that you made dedicated named files for these (renaming gs_00X to init and gs).

sozud:
You can run `make format` to fix the linter issue

src/main/psxsdk/libapi/counter.c
```diff
@@ -1,11 +1,12 @@
 #include "common.h"
+#include "psxsdk/libapi.h"
 
 typedef struct {
     u16 rootCounter;
-    s16 unk2;
+    u16 : 16;
```
sozud, line 6: Could you explain this and the other change to this struct?

hohle: I didn't see any references to `unk2` but understand the field alignment is important. The size of a struct field can be overridden with a `:` for bit packing but it also allows anonymous struct members that take up space but can't be addressed directly in code. It's a way to preserve the layout without accidentally using `unk2` for something if it is actually intended to remain unused.

sozud: Seems OK to me, this file is conflicted now BTW

hohle:
Coming back to this finally. Most of this change was made redundant, but still. few functions added.

PR #1133, opened 2024-05-15
Replace BIOS Trampolines

hohle:
This replaces several dozen BIOS trampolines from the extracted ASM to "decompiled" source. These are modeled on the `INCLUDE_ASM` macro, but generate the instructions necessary for each trampoline directly instead of importing an extracted source file.

Because these trampolines never return, and GCC 2.6 doesnt appear to have builtins for leaving off the return jump postamble, these will likely need to remain assembly.

This also changes the `main.elf` target to depend on `main.ld`, and undefined symbols files, allowing `make build` to regenerate those files if necessary.


hohle:
As with https://github.com/Xeeynamo/sotn-decomp/pull/1132 I'm trying to understand the repo and I like to do that by poking at areas and seeing what happens.

I know this doesn't necessarily move the needle on game engine decompilation, but replaces a bunch of extracted ASM with project code.

bismurphy (approved):
Looks nice. I like your defines for the bios functions.

Only suggestion might be: Do you need `DEFINE_` in your defines? Could you just have, for example, `BIOS_A_FUNCTION()`?

As far as your point about the fact that this doesn't directly involve decompiling the engine, that's completely fine. We are interested in all parts of the game and the SDK, APIs, and everything else are absolutely reasonable material for decompiling.

Thanks for your contributions and welcome to the project!

Xeeynamo:
The reason why we have weird names such as `c21` or `a50` is because those are the official names of the object files from the official PSY-Q SDK. I retrieved names and signatures to calculate the offset from here: https://github.com/lab313ru/psx_psyq_signatures/blob/main/350/LIBC.LIB.json

There's a tool to parse the custom `.LIB` files from the PSY-Q SDK, convert them into GNU-compatible objects and extract them to retrieve the original C file names and so on: https://github.com/grumpycoders/pcsx-redux/tree/main/tools/psyq-obj-parser

It is important to keep the YAML as it is. Just to give you more context, there's another project from one of our contributors @sozud that aims to do a matching decomp on PSY-Q SDK. Keeping the original names means we would be able to just copy & paste these SDK libraries from one repo to another and then eventually outsource it as a git submodule: https://github.com/sozud/psy-q-decomp

Are you in our Discord server by a chance? We are very active there. Contributors would be excited to engage with you into technical discussions and important changes to the project 🙂

Xeeynamo (approved):
It looks good!

PR #1137, opened 2024-05-18
More Descriptive func_801B9744

hohle:
This adds some additional context for func_801B9744 and renames globals to fit their assumed purpose.

Migrated various stream metadata to C definitions.

The makefile changes add a dependency on `sel.h` to all the objects that get built from the `st/sel` directory.


src/st/sel/stream.c
```diff
@@ -1,37 +1,28 @@
...
-} DECENV;
-
-s32 func_801B9744(void) {
+// Check if CD data is ready as long as its not related to reading the position of the laser. The status is cached in
+// the `g_StreamDiskIsReady` global.
+//
+// returns:
+//     0 - disc isn't ready for a request
+//     1 - disc is ready
+s32 StreamDiskIsReady() {
```
bismurphy, line 9: Very minor nitpick but you're using "Disk" in the function name, and "disc" in the comments with the return info. I believe "Disc" is preferred for optical media, but either way doesn't matter :)

bismurphy (approved):
Looks like a solid PR! Usually it's best to keep PRs limited in scope (in this case, decompiling the new function, and renaming stuff throughout other functions might be better as two separate PRs), but that's nitpicky. Looks nice from what I can tell, you're very quickly becoming a very solid contributor to the team :)

bismurphy:
If you do `make format` and then commit and push again, it should fix the linter error. You can always do `make format` before making your initial commit if you remember too, I usually end up forgetting so if you look at my past PRs you'll find many examples where I push a second commit where the message is just "format" or similar :)

src/st/sel/stream.c
hohle: Looking through the PSY-Q docs (which are new to me), Sony used "disk" in their API names and documentation, though I completely agree it should be "disc". I'll update the comment for consistency.

hohle:
> Usually it's best to keep PRs limited in scope (in this case, decompiling the new function, and renaming stuff throughout other functions might be better as two separate PRs)

I agree. I broke out the decompile to #1139 and will update this PR later with the renaming. It looks like there are some failures on other build tasks (PC, etc.) that I haven't looked at yet that need to be resolved as well.

hohle:
Resubmitting without the disassembly change.

sozud:
The PSX version doesn't have all the data imported, so we import some data just in the PC files. If it gets imported to the PSX version like this variable is, it needs to be removed from the PC version.

```
/home/runner/work/sotn-decomp/sotn-decomp/src/pc/stage_sel.c:102:5: error: conflicting types for ‘g_StreamRewindSwitch’; have ‘s32’ {aka ‘int’}
  102 | s32 g_StreamRewindSwitch;
      |     ^~~~~~~~~~~~~~~~~~~~
In file included from /home/runner/work/sotn-decomp/sotn-decomp/src/pc/stage_sel.c:2:
/home/runner/work/sotn-decomp/sotn-decomp/src/pc/../st/sel/sel.h:219:12: note: previous declaration of ‘g_StreamRewindSwitch’ with type ‘s32[1]’ {aka ‘int[1]’}
  219 | extern s32 g_StreamRewindSwitch[1];
```

Xeeynamo:
You also need to add `g_CurrentStream = 0x8003C100;` in `config/symbols.hd.txt`. The overlay `DRA.BIN` is compiled to target two builds, those we call `US` and `HD` (more info here: https://github.com/Xeeynamo/sotn-decomp/wiki/Game-versions). So whenever you add a symbol to DRA.BIN, you need to ensure it targets both builds.

Also remember to change the type at `src/pc/stage_sel.c:102` :wink:

Xeeynamo (commented):
Left a couple comments to fix the CI. But this is looking good so far.

PR #1138, opened 2024-05-18
Shared `CollectHeartVessel` Implementation

hohle:
This follows the model of  #[`collect_hearts.h`](../blob/master/src/st/collect_heart.h) to reuse the `CollectHeartVessel` implementation across stages that use the same implementation.

This adds implementations for CEN and WRP as well.

### DIFF
--- a/src/st/collect_heart_vessel.h
+++ b/src/st/collect_heart_vessel.h
@@ -1,0 +1,14 @@
+void CollectHeartVessel(void) {
+    if (g_PlayableCharacter != PLAYER_ALUCARD) {
+        g_api.PlaySfx(NA_SE_PL_COLLECT_HEART);
+        g_Status.hearts += HEART_VESSEL_RICHTER;
+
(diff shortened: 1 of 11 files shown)


sozud (approved):
Looks good, thanks!

src/st/collect_heart_vessel.h
```diff
@@ -0,0 +1,14 @@
+void CollectHeartVessel(void) {
+    if (g_PlayableCharacter != PLAYER_ALUCARD) {
+        g_api.PlaySfx(NA_SE_PL_COLLECT_HEART);
+        g_Status.hearts += HEART_VESSEL_RICHTER;
+
+        if (g_Status.heartsMax < g_Status.hearts) {
+            g_Status.hearts = g_Status.heartsMax;
+        }
```
joshlory, line 8: Tiny nit: I think this can be simplified to `CLAMP_MAX(g_Status.hearts, g_Status.heartsMax);`

https://decomp.me/scratch/0D98F

PR #1139, opened 2024-05-18
Decompile SEL func_801B9B7C

hohle:
Decompile SEL func_801B9B7C Mostly based on the work of @sonicdcer (who had the right code, but wrong struct), and @ser-pounce (who had the right struct, but unfinished code).


hohle:
Separated from #1137

sozud (approved):

Xeeynamo (approved):
Woah, nice one!! I didn't know that stuff is valid C syntax!

PR #1140, opened 2024-05-18
Decompile duplicates of weapon primitive helper function

bismurphy:
`func_107000_8017ADF8` is already in the repo and decompiled. I found that there are several duplicates of this function, so this PR decompiles all of them.

The only difference between the copies is 1. The address of the data array, obviously and 2. the assignment to the `size` variable. Some duplicates use 4 and 6, others use 2 and 3. Otherwise they're exactly the same.

This is the only heavily-duplicated weapon helper function remaining. I have decompiled all of the helper functions (the ones that are unique to the weapon they're part of, rather than being in the standard structure like `func_ptr_801700XX`) and will be doing PRs for each of them in sequence.

### DIFF
--- a/src/weapon/w_006.c
+++ b/src/weapon/w_006.c
@@ -7,7 +7,55 @@ INCLUDE_ASM("weapon/nonmatchings/w_006", EntityWeaponAttack);
 
 INCLUDE_ASM("weapon/nonmatchings/w_006", func_ptr_80170004);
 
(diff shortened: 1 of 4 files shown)


sozud:
It seems like this PR includes changes to asm-differ?

bismurphy:
No, but a different PR updated the asm-differ submodule, but I hadn't done that on my end, so I'm behind. Fixing.

bismurphy:
Cool, should be good now

PR #1142, opened 2024-05-18
Decompile Weapon 21 helper (food eating)

bismurphy:
Surprisingly, this uses an s16 for hitboxOffX, so it turns out we need to adjust the Entity struct in Weapon. Hopefully that's the last of the changes needed to Entity :)


Xeeynamo (approved):
Looks good (and tasty 🍖)

PR #1144, opened 2024-05-18
Decompile EntityWeaponAttack for Weapon 21 (food)

bismurphy:
A couple of the entity fields were different for this compared to the existing ET_Weapon, so I made a special one for food.

### DIFF
--- a/include/entity.h
+++ b/include/entity.h
@@ -190,6 +190,14 @@ typedef struct {
 } ET_Weapon29;
 
 typedef struct {
+    s16 timer;
+    s16 unk7E;
+    s32 unk80;
+    s16 pad[21];
+    s16 foodId;
+} ET_Food;
+
+typedef struct {
     /* 0x7C */ char pad_7C[0x4];
     /* 0x80 */ s16 timer;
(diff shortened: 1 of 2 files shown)


include/entity.h
```diff
@@ -189,6 +189,14 @@ typedef struct {
     /* 0x9C */ s16 unk9C;
 } ET_Weapon29;
 
+typedef struct {
+    s16 timer;
+    s16 unk7E;
+    s32 unk80;
+    s16 pad[21];
+    s16 foodId;
+} ET_Food;
```
Xeeynamo, line 198: I was wondering if you could re-use `ETWeapon`? `lifetime` and `timer` means the same thing. While `foodId` seem to match with `equipId`, which also mean the same thing.

bismurphy: The fatal mismatch is that food has `s32 unk80` while Weapon has `s16 unk80` and `s16 unk82`.

Xeeynamo: argh okay, that's fine then

Xeeynamo (approved):

PR #1147, opened 2024-05-18
Decompile CRN func_80193184

hohle:
This function appears to create all entities to the left of the provided horizontal position. It's implementation is shared across several stages but requires some refactoring to make it shareable.

### DIFF
--- a/src/st/cen/cen.h
+++ b/src/st/cen/cen.h
@@ -85,6 +85,7 @@ extern ObjInit2 D_8018125C[];
 
 extern LayoutEntity* D_8019C764;
 extern u16* D_8019C768;
+#define g_LayoutObjEnd ((LayoutEntity*)D_8019C768)
 extern u8 D_8019C76C;
 extern s16 D_8019D37E;
 extern u16 D_8019D380;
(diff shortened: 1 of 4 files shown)


src/st/cen/cen.h
```diff
@@ -85,6 +85,7 @@ extern ObjInit2 D_8018125C[];
 
 extern LayoutEntity* D_8019C764;
 extern u16* D_8019C768;
+#define g_LayoutObjEnd ((LayoutEntity*)D_8019C768)
```
Xeeynamo, line 88: I suggest you to just do `g_LayoutObjEnd = 0x8019C768` in `config/symbols.us.stcen.txt`. It reflects on `build/us/stcen.map` too. Using a `#define` will hide things a bit too much.

hohle: I was going to save that for a refactor but it turns out the refactor will probably capture this change anyway.

Nearly all of the layout scanning and creation functions are identical, they just need symbols mapped. That's probably a better change that this one off. I'm going to try to get that finished today and post a PR, but it's a much broader change across all of the stages.

Xeeynamo: It will be fine to just have the symbol added in stcen for this PR and have the refactor as a separate one if you want to avoid having too many changes in one go

hohle:
Opening a new request to modify these symbols across stages.

PR #1153, opened 2024-05-20
Implement SetRCnt

joshlory:
https://decomp.me/scratch/jawso

### DIFF
--- a/src/main/psxsdk/libapi/counter.c
+++ b/src/main/psxsdk/libapi/counter.c
@@ -3,9 +3,10 @@
 typedef struct {
     u16 rootCounter;
     s16 unk2;
-    s32 mode;
-    s32 target;
-    s32 padding;
+    s16 mode;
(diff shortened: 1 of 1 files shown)


sozud (approved):
Nice work, thanks for your contribution!

joshlory:
I'm not sure I have the function signature quite right ... args `u32 spec, u16 target` gives the same assembly.

sozud:
3.5 KERNEL.H says 
```
extern long SetRCnt(unsigned long, unsigned short, long);
```
https://github.com/sozud/psy-q/blob/9f541dbade4b25bcba449aea0e3dac9ab3cde14b/3.5/PSX/INCLUDE/KERNEL.H#L163

The public and internal headers may not be the same though. The manual gives a different signature:
image

sozud:
I think it's fine either way so going to merge.

PR #1155, opened 2024-05-21
Update func_8011A4D0 to match on PSP

bismurphy:
Based on Sozud's scratch (https://decomp.me/scratch/zVBXP), I got the function to match, and then confirmed that it also matches on PSX.

I also discovered a few improvements, especially in changing `D_8016FCC0[entityId];` to instead be `((PfnEntityUpdate*)FAMILIAR_PTR)[entity->entityId - 0xD0];`

While I was at it, I also updated the D_80170000 variable from a void* to be g_ServantDesc, just like we have in the servant overlay. The ServantDesc struct only exists in servant.h, so we update that as well.

Finally, we move g_Weapon to not be at 80170000, since it doesn't belong there.

### DIFF
--- a/src/dra/7A4D0.c
+++ b/src/dra/7A4D0.c
@@ -78,10 +78,7 @@ void func_8011A4D0(void) {
     s32 temp_s2;
     s32 i;
     s32 i2;
-    s32 i3;
-    u16 entityId;
     s32 enemy;
-    s32 enemy2;
 
     temp_s2 = *D_80097420;
     entity = g_CurrentEntity = &g_Entities[4];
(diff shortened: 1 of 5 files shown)


src/dra/7A4D0.c
```diff
@@ -93,36 +90,36 @@ void func_8011A4D0(void) {
...
                 if (!(entity->flags & FLAG_UNK_04000000) &&
-                    ((u16)(entity->posX.i.hi + 32) > 320 ||
-                     (u16)(entity->posY.i.hi + 16) > 272)) {
-                    DestroyEntity(entity);
+                    (entity->posX.i.hi > 288 || entity->posX.i.hi < -32 ||
+                     entity->posY.i.hi > 256 || entity->posY.i.hi < -16)) {
+                    DestroyEntity(g_CurrentEntity);
                 } else {
                     if (entity->flags & 0x100000) {
                         UpdateAnim(NULL, D_800ACFB4);
```
sozud, line 125: Not sure how it affects other things but I found that this variable needed to change to 
```
extern  AnimationFrame**  D_800ACFB4;
``` 
to compile with MWCC

sozud (approved):

src/dra/7A4D0.c
bismurphy: Yes, that's right, the problem is that `func_80118C28` references individual bytes from that variable and I don't know how to rewrite that function to be acceptable.

bismurphy: My scratch is here: https://decomp.me/scratch/UbCHl I couldn't get it to work with `AnimationFrame**`, I had to use `AnimationFrame* D_800ACFB4[]`

sozud:
This is a good improvement regardless of the D_800ACFB4 questions, merging

PR #1157, opened 2024-05-22
Decompile func_8019F3BC

joshlory:
https://decomp.me/scratch/EJ9uP

### DIFF
--- a/src/st/dre/1E1C8.c
+++ b/src/st/dre/1E1C8.c
@@ -365,8 +365,63 @@ void func_8019F304(void) {
     }
 }
 
-// DECOMP_ME_WIP func_8019F3BC https://decomp.me/scratch/Hfk9n
-INCLUDE_ASM("st/dre/nonmatchings/1E1C8", func_8019F3BC);
+void func_8019F3BC(Entity* arg0) {
+    s16 temp_v1;
+    s32 temp_v1_2;
+    s32 var_a0;
+    s32 var_v1;
+    s32 temp_s0;
+
+    if (arg0->step == 0) {
+        InitializeEntity(g_InitializeEntityData0);
+        arg0->animSet = 2;
+        arg0->palette = 0x81B6;
+        arg0->unk6C = 0x70;
+        arg0->zPriority = 0xC0;
+        arg0->unk19 |= 0xC;
+        arg0->drawMode |= 0x30;
+
+        switch (arg0->unk84.U8.unk0) {
+        case 1:
+            if (arg0->unk84.U8.unk1 >= 4u) {
+                arg0->unk84.U8.unk1 += 0xFD;
+                arg0->unk80.modeS16.unk0 -= 0x800;
+            }
+            break;
+
+        case 2:
+            arg0->unk80.modeS16.unk0 += (arg0->unk84.U8.unk1 * 0xC0);
+            break;
+        }
+
+        temp_v1_2 = arg0->unk80.modeS16.unk0 & 0xFFF;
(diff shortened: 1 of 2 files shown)


sozud:
I think you based this off an old scratch and Entity has changed since then https://github.com/Xeeynamo/sotn-decomp/blob/0917d1ee3b956260a7113095272c2236edfc2170/include/game.h#L643

joshlory:
Ah, looks like there are a few other fields to update as well. Where can I find the latest `ctx.c` to drop in?

Xeeynamo:
> Ah, looks like there are a few other fields to update as well. Where can I find the latest `ctx.c` to drop in?

First of all, thank you for taking your time to look into this function!

We generally create a context locally with `SOURCE=path/to/my/function.c make context`. Or we start decompiling locally and use the permuter to upload on decomp.me with `./import.py ~/repo/sotn-decomp/path/to/my/function.c ~/repo/sotn-decomp/asm/overlay/nonmatchings/blahblah/my_function_name.s --decompme`.

You might also want to check if it is compiling locally. Our CI is currently signaling a failing build at the line that has `arg0->unk19`. This is because `unk19` does no longer exist as it has been renamed. `arg0` is of type `Entity`, so I suggest to look where the `Entity` struct is declared and see what is the current field name for the offset `0x19`.

Once the build is fixed I will be very happy to welcome your changes and get them in!

joshlory:
@Xeeynamo thanks for the tips! I was able to regenerate the context and verify the changes. With the new `ctx.c` file I do see one warning:

```
src.c:57: warning: passing arg 1 of `AnimateEntity' from incompatible pointer type
```

I copied the type definition from another header: https://github.com/Xeeynamo/sotn-decomp/blob/867c7e12c896f81e04045b2fefee6f724ce39fb0/src/st/cen/cen.h#L82

Are these different "overlays"? Should I update `dre.h` (or both) to define it as `u8` instead? Sorry for the newbie questions!

sozud:
AnimateEntity is
```
s32 AnimateEntity(const u8 frames[], Entity* entity)
```

so u8 is most likely correct. You can also look at the data if you search for it in the asm folder, which might give a hint about the size/usage.

joshlory:
@sozud shouldni update it in both places or just the DRE header?

sozud:
The usage in CEN looks like this, so I'm not sure it's the same thing. https://github.com/Xeeynamo/sotn-decomp/blob/3b0943b25a91c9d32864cbdd549bd24206d42fd0/src/st/cen/1B274.c#L74
It may have the same name by accident.

Xeeynamo (approved):
Looks good as it is. Congrats for the first function match!

src/st/dre/1E1C8.c
```diff
@@ -365,8 +365,65 @@ void func_8019F304(void) {
...
+    if (arg0->step == 0) {
+        InitializeEntity(g_InitializeEntityData0);
+        arg0->animSet = 2;
+        arg0->palette = 0x81B6;
+        arg0->unk6C = 0x70;
+        arg0->zPriority = 0xC0;
+        arg0->drawFlags |= 0xC;
+        arg0->drawMode |= 0x30;
+
+        switch (arg0->ext.generic.unk84.U8.unk0) {
```
joshlory, line 384: I just saw https://github.com/Xeeynamo/sotn-decomp/issues/744 to not use `ext.generic.*` any more. @Xeeynamo is there a way to tell which entity type is most likely? The calls to this function haven't been mapped yet so may need to wait.

PR #1158, opened 2024-05-22
Rename Symbols For Scroll Position and Delta

hohle:
Based on their use in the game loop and during stage updates, it appears the renamed variables store the scroll position delta for the current loop and store off the current position for future delta calculation. These vars are then used by stage code to determine if any entities need to be created or destroyed based on the new position.


config/symbols.us.stcen.txt
```diff
@@ -1,3 +1,5 @@
+g_ScrollDeltaX = 0x80097908;
```
sozud, line 1: Stages start at 0x80180000, DRA starts at 0x800A0000, so this is actually a main symbol (starting from 0x80010000) it seems. Not sure but that might be your compile issue.

Xeeynamo: This will help understanding the memory layout and where to put the symbols: https://github.com/Xeeynamo/sotn-decomp/wiki/Internals

hohle: Thanks, I hadn't added the symbol to `symbols.hd.txt`.

Xeeynamo (approved):
Brilliant, it looks much better than what we currently have in our code-base. It is always great documenting the existing code-base. Much appreciated!

PR #1160, opened 2024-05-22
Decompile weapon 23 func_ptr_80170028

bismurphy:
Looks like just another weapon function.

Weird uses of the Weapon struct. Unfortunately Weapons don't seem to have any consistency among them. Would be nice if we could define the Entity extension in each weapon's `w_0XX.c` file, but I don't think that's a possibility in C.


sozud (approved):

Xeeynamo (approved):
LGTM. And yeah, unfortunately it looks like the `ext` section is just copy & pasted and then customised for each entity

PR #1161, opened 2024-05-22
EntityDamageDisplay matching

Xeeynamo:
```
         .----.  .----..----.    .-. .----.          
         | {}  }{ {__  | {}  }   | |{ {__            
         | .--' .-._} }| .--'    | |.-._} }          
         `-'    `----' `-'       `-'`----'           
  .--.  .-. . .-..----. .----. .----. .-.   .-..----.
 / {} \ | |/ \| || {_  { {__  /  {}  \|  `.'  || {_  
/  /\  \|  .'.  || {__ .-._} }\      /| |\ /| || {__ 
`-'  `-'`-'   `-'`----'`----'  `----' `-' ` `-'`----'
```

Thanks a lot to @SestrenExsis for doing the majority of the function matching! I only take the credit for cross-referencing it with the PSP counterpart and massaging the code until I got a match on PSX.


bismurphy:
Amazing!

I noticed there are a lot of `LOH` and `LOHU` on the Primitive members. Any chance they might match if you use FakePrim? Looks like they line up with the right offsets.

bismurphy:
Seems reasonable! We'll see if this primitive crops up somewhere else. For now this looks good and it's nice to be able to use descriptive struct members instead of abusing members like `r2` :)

PR #1162, opened 2024-05-22
Rename Stage Layout Functions and Vars Uniformly

hohle:
This change renames functions and global stage variables uniformly across the stages so that these functions can be pulled out and shared across all of the stages. Based on some other tests there are 12 or so functions that this will allow to be pulled out of each stage. Since these implementations are shared, an additional 12 asm functions can be eliminated in a subsequent pass.

**Vars**
* `g_pStObjLayoutHorizontal` - a horizontally sorted array of stage entities
* `g_pStObjLayoutVertical` - a vertically sorted array of stage entities
*  `g_LayoutObjHorizontal` - a pointer to a `LayoutEntity` in `g_pStObjLayoutHorizontal`
* `g_LayoutObjVertical` - a pointer to a `LayoutEntity` in `g_pStObjLayoutVertical`
* `g_LayoutObjPosHorizontal` - the direction last traversed in `g_LayoutObjHorizontal`
* `g_LayoutObjPosVertical` - the direction last traversed in `g_pStObjLayoutVertical `

**Functions**
* `FindFirstEntityToTheRight` - given an `x` position, update `g_LayoutObjHorizontal` with the first entity to the right of `x`
*  `FindFirstEntityToTheLeft` - given a `x` position, update `g_LayoutObjHorizontal` with the first entity to the left of `x` (backwards)
* `CreateEntitiesToTheRight` - given an `x` position, create all entities to the right (mutates `g_LayoutObjHorizontal`)
* `CreateEntitiesToTheLeft` - given an `x` position, create all entities to the left (mutates `g_LayoutObjHorizontal`)
* `FindFirstEntityAbove` - given an `y` position, update `g_LayoutObjVertical ` with the first entity to the above of `y`
* `FindFirstEntityBelow` - given an `y` position, update `g_LayoutObjVertical ` with the first entity to the below of `y`
* `CreateEntitiesAbove` - given an `y` position, create all entities above (mutates `g_LayoutObjVertical`)
* `CreateEntitiesBelow` - given an `y` position, create all entities beneath (mutates `g_LayoutObjVertical`)
* `UpdateRoomPosition` - look at the current game loop scroll delta and create any entities given the room layout

I believe all of these implementations are shared across all stages (including `InitRoomEntities`, and two more `CreateEntity` functions) (in my initial tests I had a small difference in `DER`, but I believe that had to do with an incorrect symbol table change).


Xeeynamo (approved):
Thank you very much for the valuable content this PR brings. The naming is on point. I cannot think of anything better.

This decomp also suffers of a few inconsistencies as it is a project that has been running since 2 years and a half. So it is good to see more stuff coming together.

bismurphy:
The comments you have here on each of these variables/functions are super beneficial - I would encourage you to add them in the code. You've clearly done work to identify what they are (for example, "g_LayoutObjPosHorizontal - the direction last traversed in g_LayoutObjHorizontal") and this text is unlikely to be remembered if it's in a closed PR; it would be nice if these were in the code so that future folks trying to understand things can have a basis for understanding the purpose of these variables, since I think their names do not make their purpose particularly obvious (which is okay).

Thank you very much for this highly beneficial contribution!

PR #1166, opened 2024-05-24
Share Stage Function Implementations

hohle:
Pulls out `CreateEntityFromLayout`, `CreateEntityWhenInVerticalRange`, `CreateEntityWhenInHorizontalRange`, `FindFirstEntityToTheRight`, `FindFirstEntityToTheLeft`, `CreateEntitiesToTheRight`, `CreateEntitiesToTheLeft`, `FindFirstEntityAbove`, `FindFirstEntityBelow`, `CreateEntitiesAbove`, `CreateEntitiesBelow`, `InitRoomEntities`, `UpdateRoomPosition`,
`CreateEntityFromCurrentEntity`, `CreateEntityFromEntity`, `EntityIsNearPlayer`, and `InitializeEntity` into headers to share the implementation between stages.

`libstage.h` is a new header intended to bring in a sequential block of functions shared by all stages.

`st_private.h` is a new header intended for declaring symbols used by all stages that aren't otherwise accessible to code outside of a stage.

The only discrepency is `MAD` which has a unique implementation of `CreateEntitiesToTheLeft`. Instead of being able to blanket include `libstage.h` it needs to bring in individual headers before and after its implementation.


src/st/cen/11280.c
```diff
@@ -3,274 +3,7 @@ INCLUDE_ASM("st/cen/nonmatchings/11280", TestCollisions);
...
-void CreateEntityFromEntity(u16 entityId, Entity* source, Entity* entity) {
-    DestroyEntity(entity);
-    entity->entityId = entityId;
-    entity->pfnUpdate = PfnEntityUpdates[entityId - 1];
-    entity->posX.i.hi = source->posX.i.hi;
-    entity->posY.i.hi = source->posY.i.hi;
-}
-
-#include "../entity_is_near_player.h"
+#include "../libstage.h"
```
hohle, line 6: I can image this being controversial. I chose a single header to make it easier to continue appending as things like `EntityRedDoor` or others are found to match. I don't have a good idea right now how many contiguous functions will be found to be shared, but based on the rest of this file, I'm assuming quite a bit more.

src/st/dre/dre.h
```diff
@@ -134,7 +134,7 @@ extern u8 D_801813FC;
 
 extern LayoutEntity* g_LayoutObjHorizontal;
 extern LayoutEntity* g_LayoutObjVertical;
-extern s8 g_LayoutObjPosHorizontal;
+extern u8 g_LayoutObjPosHorizontal;
```
hohle, line 137: This was a pain and I wonder if it thwarted previous attempts to decompile the horizontal entity creation methods here and in `mad`.

sozud (approved):
This all looks good to me, we can continue adjusting as we find out more

PR #1119, opened 2024-05-06
WIP: Disassemble PSP dra.bin

Xeeynamo:

PR #1172, opened 2024-05-25
Add ST/WRP from PSP

Xeeynamo:
Extract ST/WRP out of #1119 . All the function symbols should have been cross-referenced. There as some PSX functions missing from PSP and some new functions from PSP that are not present on PSX (e.g. `st_init_wrp.c`). 

The files `st_debug.c` and `e_breakable.c` are shared between WRP PSX and WRP PSP. Everything else from PSP is isolated into its own folder. I had to do some tricks on the YAML config to allow shared code.

`ST_WRP_MERGE = st_debug e_breakable` in the `Makefile` is a bit annoying as MWCC complains about every single minute detail from the C source that has been already decompiled for the PSX US build.

`EntityWarpSmallRocks` is matching on PSP but I couldn't extract the rodata without having a bunch of linker errors. This might be a Splat issue. I need to investigate further.

`func_psp_09244760` is soooo interesting. The values from `0x11` to `0x17` matches the Entity IDs that are unique to the WRP overlay. This aligns to what we have in `typedef enum EntityIDs`.

Overall I am very excited to the recent discoveries from the PSP build!


sozud (approved):
Nice work!

PR #1175, opened 2024-05-26
Rework g_unkGraphicsStruct

bismurphy:
I noticed that this struct had some overlap with other values in memory, so I have pulled all those values into this struct.

The boundaries of this struct are uncertain and are a matter of ongoing research.


config/symbols.us.weapon.txt
```diff
@@ -1,2 +1,2 @@
-g_Weapon = 0x8017A000;
+g_unkGraphicsStruct = 0x80097400;g_Weapon = 0x8017A000;
```
Xeeynamo, line 1: `\n` xD

bismurphy: The interesting thing is that I think this problem was created by `make format`, since I had the graphics struct as the last line of this file, and `make format` sorted it. I think what happened is that when I had it as the last line, I did not have a line break at the end of the file, so when the line got sorted to the top, it got sorted without a \n, since as the final line, there was no \n. Neat little edge case.

bismurphy:
I believe the current pspeu build error is inherited from master.

bismurphy:
Rebased on new master

Xeeynamo (approved):
Looks better, especially that nasty loop at `func_800F2404`

PR #1176, opened 2024-05-26
Prevent entity from respawning refactor

Xeeynamo:
Thanks to @sozud for the hint and @bismurphy for the refactoring idea. The function `PreventEntityFromRespawning` on PSP hints that the struct starts 8 bytes earlier. Also there's a missing `nop` at `PreventEntityFromRespawning`, suggesting the function had to be moved to the previous function.


src/st/wrp_psp/BDA0.c
```diff
@@ -1,7 +1,13 @@
...
 
 INCLUDE_ASM("st/wrp_psp/psp/wrp_psp/BDA0", EntityRoomForeground);
 
 INCLUDE_ASM("st/wrp_psp/psp/wrp_psp/BDA0", DestroyEntity);
 
-INCLUDE_ASM("st/wrp_psp/psp/wrp_psp/BDA0", PreventEntityFromRespawning);
+void PreventEntityFromRespawning(Entity* entity) {
+    if (entity->entityRoomIndex) {
+        u16 index = entity->entityRoomIndex - 1 >> 5;
+        g_unkGraphicsStruct.D_80097428[index] |=
```
sozud, line 10: I'm wondering if we are going to get out of bounds accesses from shortening this array, but based on the scratch it seems like this must be closer to the original. We can continue iterating as more of wrp_psp gets done.

Originally this was called g_entityDestroyed, are we thinking it has another meaning or multiple uses now?

sozud (approved):

src/st/wrp_psp/BDA0.c
Xeeynamo: I was wondering the same... only time will tell I guess. I have a few other functions lined up like `CreateEntitiesToTheRight`. I might re-discover the meaning behind it? Not sure. I will keep iterating over it as I find new things

PR #1180, opened 2024-05-26
UpdateRoomPosition

sozud:


Xeeynamo (approved):
Works well!

PS. Same warning as the previous PR

PR #1181, opened 2024-05-26
ST WRP re-organise C files

Xeeynamo:
A massive PR, I know. It's the smallest change I could do to make this happen.

After comparing the function order from both ST/WRP PSX and ST/WRP PSP, I was able to come up with a file split much different from the one suggested from Splat. It still returns an :ok: and it follows the same group functions order from the two game builds.

As soon as we are on board with the C names, next PR will be to do the same with the PSX overlay. Then in another PR we will be able to start merging the C files between the two builds. Ideally the same approach can be used for sharing C files across the different overlays instead of relying to floating header files in `src/st`


sozud (approved):
Seems reasonable to me

PR #1186, opened 2024-05-26
Decompile NP3 TestCollisions

bismurphy:
Well here we are, TestCollisions matching!

I'm going to start with this one, and once it's in, I will follow up by de-duplicating it.

Please review this closely, especially the changes that are happening in other files than the main C file.

I suspect we will learn a lot from studying this function; I've already made some changes (including to the Entity struct) but there will surely be more.

Nice that this was possible with only two `goto` statements.

I expect a fair bit of revision here, especially related to any enums/defines I might not be aware of. Please be liberal with comments :)


sozud (approved):
Great work!

src/st/np3/394F0.c
```diff
@@ -1,5 +1,514 @@
...
+        }
+        if (miscVar2) {
+            if (iterEnt1->unk5C != NULL) {
+                entFrom5C = iterEnt1->unk5C;
+                entFrom5C->unk44 = (u16)iterEnt1->unk44;
+                entFrom5C->hitFlags = (u8)iterEnt1->hitFlags;
+            } else {
+                entFrom5C = iterEnt1;
+            }
+            if (!(entFrom5C->flags & 0x100) && miscVar2) {
```
Xeeynamo, line 191:
```suggestion
            if (!(entFrom5C->flags & FLAG_DEAD) && miscVar2) {
```

src/st/np3/np3.h
```diff
@@ -121,7 +121,15 @@ extern Unkstruct_80180FE0 D_8018129C[];
 
 // *** Merman properties START ***
 
-extern u16 D_80181230[];     // pos TBL
+extern u16 D_80181230[]; // pos TBL
+extern u16 D_80181678[];
+extern u8 D_80181998[];
+extern u8 D_8018199C[];
+extern u8 D_801819A4[];
+extern u16 D_801819AC[];
```
Xeeynamo, line 129:
```suggestion
extern u16 D_801819AC[10];
```

src/st/np3/np3.h
```diff
@@ -121,7 +121,15 @@ extern Unkstruct_80180FE0 D_8018129C[];
...
+extern u8 D_801819C0[];
```
Xeeynamo, line 130:
```suggestion
extern u8 D_801819C0[12];
```
I know it should be `10`, but I am not sure how happy it is going to be alignment-wise

src/st/np3/394F0.c
```diff
@@ -1,5 +1,514 @@
...
+                        miscVar1 = 0;
+                    }
+                    if (miscVar1 != 0xC000) {
+                        if (miscVar1 & 0x8000) {
+                            miscVar2 = 9;
+                        } else {
+                            miscVar3 = (u16)iterEnt2->attackElement;
+                            // includes all elements 0x40 and up
+                            if (miscVar3 & 0xFFC0) {
+                                for (i = 0; i < 10; i++) {
```
Xeeynamo, line 285:
```suggestion
                                for (i = 0; i < LEN(D_801819AC); i++) {
```
but you would need to set the array size for `D_801819AC` first

src/st/np3/394F0.c
```diff
@@ -1,5 +1,514 @@
...
+                        }
+                    }
+                    otherEntity = otherEntity->unk60;
+                } while ((otherEntity != NULL) && (otherEntity != entFrom5C));
+            }
+        }
+    }
+    for (prim = &g_PrimBuf[g_unkGraphicsStruct.D_800973F8]; prim != NULL;
+         prim = prim->next) {
+        if ((prim->drawMode) != 8) {
```
Xeeynamo, line 497:
```suggestion
        if (prim->drawMode != DRAW_HIDE) {
```

Xeeynamo (approved):
I put some nitpicks here and there. I couldn't comment more on some flags such as `hitState` or `hitFlag` as we yet know very little

Very cursed variables such as `miscVar3` that seems to be everywhere. I think when the types are vague or opaque it might be better to keep them as `var_REGISTERNAME` so we can move them around or even declare them in the middle of the function by reducing the scope and giving them a more meaningful name (as long as it keeps matching).

Out of curiosity, do we have a PSP scratch? It would be good to resume it as some point, even if it is still a bit soon.

As I said on Discord, this was truly the final boss. You got the two biggest beasts of this decomp between this one and `EntityAlucard`. Congratulations @bismurphy!

bismurphy:
> Very cursed variables such as `miscVar3` that seems to be everywhere. I think when the types are vague or opaque it might be better to keep them as `var_REGISTERNAME` so we can move them around or even declare them in the middle of the function by reducing the scope and giving them a more meaningful name (as long as it keeps matching).

Yes, I thought about this - the trouble is that the registers are different on PSP vs PSX. I started this function with a PSP scratch, then once that was nearly-matching, copied the C over to the PSX scratch. Interestingly, the PSX scratch was much closer to matching than PSP, even though I had been writing my C based on only the PSP version. But anyway, because my variables were based on the PSP version, I would have to rename them to make the names meaningful on PSX. Because the registers/stack locations used for a given variable will change depending on small differences, I figured it was better to give them real names, even if they aren't super helpful ones. If you would like me to rename all the local variables to go back to their register/stack names, let me know and I'll make those replacements.

Here is the PSP scratch I made.
https://decomp.me/scratch/rZYV2

Thanks for the kind words!

src/st/np3/np3.h
bismurphy: I think with 10 it should be okay since it will get padded out to 32-bits. Seems to work at least for now.

PR #1189, opened 2024-05-27
WRP PSP check-in more function

Xeeynamo:
The highlight is that `src/st/wrp/st_update.c` is shared across the two overlays. `EntityMessageBox` and `EntityStageNamePopup` are vastly different.


src/st/prevent_entity_from_respawning.h
```diff
@@ -0,0 +1,8 @@
+void PreventEntityFromRespawning(Entity* entity) {
```
sozud, line 1: It's probably intentional but just pointing out this is the old version https://github.com/Xeeynamo/sotn-decomp/blob/4d4c039ac49e0cc6563508528636a6aa34501dba/src/st/wrp_psp/st_common.c#L5

sozud (approved):

src/st/prevent_entity_from_respawning.h
Xeeynamo: Yep I am aware! Merging is currently not possible and I stopped because of a current blocker. I am preparing another PR to unblock this

PR #1190, opened 2024-05-27
Decompile ST0 TestCollisions

bismurphy:
Got this one done too! ST0 is slightly simpler than the others.

I tried to adapt the existing matching TestCollisions directly, but I found that it was easier to work on PSP to get the logic right and then move to PSX instead.


Xeeynamo (approved):
> Attempting to de-duplicate this would involve a lot of #ifdef.

I was about to ask. Yeah this one has quite the good amount of stuff missing or different. There are also a couple of functions different in ST0, like `ReplaceBreakableWithItemDrop` and another one I do not remember the name of. This makes me wonder if this function comes from an earlier version of the game code.

On that regard, do you have any plan on decompiling the MAD counterpart as well? I would be fascinated to see the various differences.

bismurphy:
> do you have any plan on decompiling the MAD counterpart as well?

Didn't know there was one! Looks like according to duplicates.txt it's func_8018EDB8, so I'll start working on that :)

PR #1191, opened 2024-05-27
Decompile MAD TestCollisions

bismurphy:
This should finish out the set :)

Can't say I understand the differences here (it's modifying g_CastleFlags during the item drop randomization routine for crying out loud) but hey, a match is a match.


Xeeynamo (approved):
So it's more similar to a normal overlay than ST0. Trurly `flaggy_flags` are the biggest mistery of all 😆.
GGWP!

PR #1192, opened 2024-05-27
Decompile MAD BottomCornerText

bismurphy:
When decompiling MAD TestCollisions in my last PR, I noticed that the call to BottomCornerText was going to func_80198BC8 in MAD, which was not yet decompiled.

The duplicates report indicated that func_80198BC8 was a duplicate of func_97000_8017AB54, a weapon function that I decompiled a few weeks ago.

I copied the code for that function into here, and renamed it to BottomCornerText to match the other overlays, even though it's not the exact same code as is used in other overlays.


Xeeynamo (approved):
Very curious it is equal to w_021. Thanks for looking into it

PR #1195, opened 2024-05-27
PrimDecreaseBrightness

sozud:
I created Primitive2 to avoid doing a big refactor. Not sure how correct the struct definition is or not but putting it up for comments.
https://decomp.me/scratch/DzATf


Xeeynamo (approved):
Thank you, I saw that on Discord. It's not the first instance of a customised `Primitive` ad-hoc for a specific function. I wouldn't read too much into it until we get more proof.

PR #1200, opened 2024-05-27
Dedupe CheckColliderOffsets

bismurphy:
Found this function which is in every overlay, and is a dependency of another function I'd like to de-duplicate.

See the comment on the .h file for a description of what this function appears to do. I think the fact that it's taking the entity's position, offsetting it based on the provided array, and running CheckCollision makes CheckColliderOffsets a reasonable name without being too verbose. Happy to use a different name if anyone has opinions.

This function had been decompiled in a few overlays, but was still INCLUDE_ASM in others, so this will also unify them to all be decompiled.

### DIFF
--- a/config/symbols.stmad.txt
+++ b/config/symbols.stmad.txt
@@ -65,6 +65,7 @@ EntityExplosion = 0x80194218;
 EntityEquipItemDrop = 0x801943EC;
 EntityRelicOrb = 0x801949C0;
 EntityHeartDrop = 0x80195520;
+CheckColliderOffsets = 0x8019563C;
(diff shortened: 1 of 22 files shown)


bismurphy:
Need to get better at using git; not sure why the commits from my previous PR got attached to this one (maybe I messed up when rebasing), but looks like those don't show up in the changes for the PR as a whole, so should be ok.

Xeeynamo (approved):
> not sure why the commits from my previous PR got attached to this one

s'all good man, PRs gets squash and merged anyway

PR #1204, opened 2024-05-28
Deduplicate EntityUnkId14Spawner

bismurphy:
This spawns several copies of EntityUnkId14, very similarly to the existing EntityUnkId15Spawner.

Most of the work for this PR (renaming all the functions, cross-referencing symbols across overlays) was done by an automated script I made. I am still testing this script, but it is nice to see that it appears to be starting to work. I will test it on a few more de-duplication PRs, and then consider adding it to the `tools` directory.


src/st/dre/1E1C8.c
```diff
@@ -29,8 +29,8 @@ void EntityUnkId13(Entity* entity) {
     }
 }
 
-// DECOMP_ME_WIP func_8019E3C8 https://decomp.me/scratch/lcx4I
-INCLUDE_ASM("st/dre/nonmatchings/1E1C8", func_8019E3C8);
+// DECOMP_ME_WIP EntityUnkId14Spawner https://decomp.me/scratch/lcx4I
```
sozud, line 32: If this function is decompiled we can remove the WIP comment

sozud (approved):

src/st/dre/1E1C8.c
bismurphy: Thanks, good catch!

src/st/entity_unkId14_spawner.h
```diff
@@ -0,0 +1,24 @@
...
+    for (i = 0; i < count; i++) {
+        newEntity = AllocEntity(&g_Entities[160], &g_Entities[192]);
+        if (newEntity != NULL) {
+            newEntity->posX.i.hi = newX + xGap * i;
+            newEntity->posY.i.hi = newY;
+            newEntity->entityId = E_ID_14;
+            newEntity->pfnUpdate = EntityUnkId14;
+            newEntity->params = params;
+            newEntity->ext.generic.unk94 = arg5 + i;
+            newEntity->rotY = newEntity->rotX = eUnk14SpawnRots[arg5 + i];
```
Xeeynamo, line 19: Can you prepend the `g_` to `eUnk14SpawnRots`? Unless it's marked as `static` and it is self-contained in a C file, it is better to be explicit the variable being globally accessible

bismurphy: Ah yes of course, fixed

PR #1208, opened 2024-05-29
EntityEnemyBlood

sozud:
Thanks to @dezgeg


Xeeynamo (approved):
Looks good! Can you put the changes into `src/st/entity_enemy_blood.h` and include it? That way is much easier to keep all the changes between PSP and PSX in sync

sozud:
> Looks good! Can you put the changes into `src/st/entity_enemy_blood.h` and include it? That way is much easier to keep all the changes between PSP and PSX in sync

Done

PR #1209, opened 2024-05-29
Fix g_LayoutObjHorizontal and g_LayoutObjVertical types

Xeeynamo:
As discovered in [one of the recent PSP functions](https://decomp.me/scratch/thtl9), the types for `g_LayoutObjHorizontal` and `g_LayoutObjVertical` are not of `LayoutObject*` but `u16*`. This saves some awkward casts that caused Clang and modern GCC to fail to compile some recent code (specifically `((u16*)g_LayoutObjVertical) = ` being not valid C).

This is a bit sad as it reverts part of the changes from #1166 that aimed to improve code clarity.


src/st/create_entities_horizontal.h
```diff
@@ -21,17 +21,17 @@ void CreateEntitiesToTheLeft(s16 posX) {
     }
 
     while (true) {
-        if (g_LayoutObjHorizontal->posX == LAYOUT_OBJ_START ||
-            posX > g_LayoutObjHorizontal->posX) {
+        if (g_LayoutObjHorizontal[0] == LAYOUT_OBJ_START ||
```
sozud, line 24: `g_LayoutObjHorizontal[LAYOUT_OBJ_POS_X]` etc?

src/st/find_entity_horizontal.h
```diff
@@ -11,13 +11,13 @@
...
-        if (g_LayoutObjHorizontal->posX != LAYOUT_OBJ_START &&
-            g_LayoutObjHorizontal->posX >= posX) {
+    while (1) {
+        u16* layoutEntity = g_LayoutObjHorizontal;
+        if (layoutEntity[0] != LAYOUT_OBJ_START && layoutEntity[0] >= posX) {
             break;
         }
 
-        g_LayoutObjHorizontal++;
+        g_LayoutObjHorizontal += sizeof(LayoutEntity) / sizeof(u16);
```
sozud, line 20: Why keep LayoutEntity here when the rest of the PR removes the structs?

sozud (approved):

src/st/create_entities_horizontal.h
Xeeynamo: It wouldn't make sense in this context because the first element wouldn't indicate Position X but just the start of the data.

Same for `InitRoomEntities` ([scratch](https://decomp.me/scratch/thtl9), where the array elements `0` and `1` from `g_LayoutObjHorizontal = (u16*)((g_LayoutObjHorizontal[1] << 0x10) + g_LayoutObjHorizontal[0])` do not indicate PosX or PosY but just two halves or a pointer array... it is unfortunately horrible, hacky and very not-well structured.

src/st/find_entity_horizontal.h
Xeeynamo: It is to avoid using the magical number `+= 5` or `-= 5`. The array is still aligned by the struct `LayoutEntity`, which is still getting propagated. I also use it in 6FD0 to declare the data. I am still undecided if to remove `LayoutEntity` completely... I was forced to use a `u16*` for these functions just due to the PSP match.

Let's continue with the project and make any further adjustments accordingly!

PR #1211, opened 2024-05-30
De-duplicate CollectGold

bismurphy:
This one had some small differences in MAD and ST0, so I worked them in as ifdef.

Also, I cleaned up the function a bit, by removing the unneeded pointers.

The GoldSizeIndex local variable is unneeded. It is possible to just use the goldSize argument, and do `-=2`, but I think using the local variable makes it more readable, so I left it in place.


bismurphy:
I encourage you to examine the 3 commits here individually, because they are interesting.

Commit 1 changes `g_Status.gold` from `u32` to `s32`. Surprisingly, this breaks the PSP version.

Commit 2 moves the cast to s32 from g_Status.gold to c_goldPrizes.

Commit 3 was me realizing that the need for a cast indicates that g_Status.gold and c_goldPrizes are different types. So I changed c_GoldPrizes to be u32 instead, which made everything match on both versions, with no casts needed at all.

Could be a useful pattern to be aware of.

src/st/wrp_psp/e_collect.c
```diff
@@ -72,13 +72,13 @@ void CollectHeart(u16 arg0) {
 }
 
 extern u8* D_80180E08[];
-extern s32 c_GoldPrizes[];
+extern u32 c_GoldPrizes[];
 
 void CollectGold(u16 goldSize) {
```
sozud, line 77: I'm thinking Xeeynamo might want the PSP version de-duplicated too? https://github.com/Xeeynamo/sotn-decomp/pull/1208#pullrequestreview-2086627080

sozud (approved):

src/st/wrp_psp/e_collect.c
bismurphy: Sounds good, done

PR #1212, opened 2024-05-30
Nicer logs

Xeeynamo:
image

Add colours to the various logs to make sense out of all the produced noise. Also gets rid of the square brackets and use a nicer column separator. Don't mind the error log being mis-labeled. My hunger for copy&pasting code consistently betrays me.

image

but most importantly groups the logs to avoid spamming the same one for every frame. The log count keeps getting updated on the left-hand side of the log, replacing the beginning of the path (almost never relevant). If the log is old, the logging function will go back, update the line and go back down.

It is a tad bit slower than the original implementation, but logs are never meant to be seen anyway, especially debug logs. I used a relatively fast hashing function paired with a dictionary to avoid unnecessary loops. There are two hashes. `logHash` simply works as an index to the massive 65K entry dictionary to pick an index. If an entry is found then check the other hash `altHash` to check if the dictionary collided with an actually different log, effectively replacing it. Since `entry->hash` is `0` at the beginning, that also works for new elements.

I believe the slowest part of the code are the two `while (nLinesBack-- > 0)`. I am thinking to maybe unroll the loop in a similar way `memcpy` works internally to save `fputs` calls. But I have little clue on how to measure improvements. If you know better on how to take advantage of moving the cursor through the TTY, please let me know.

You can disable this entire functionality with `#define FORCE_LINEAR_LOGS` and it gets disabled if `stderr` is not printing on a terminal.

### DIFF
--- a/include/log.h
+++ b/include/log.h
@@ -11,7 +11,7 @@ typedef enum {
 } LOG_LEVEL;
 
 #define NAMEOF(var) #var
-#define NOT_IMPLEMENTED DEBUGF("not implemented")
+#define NOT_IMPLEMENTED WARNF("not implemented")
 
 #ifndef NO_LOGS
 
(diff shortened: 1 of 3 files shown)


sozud (approved):
Seems reasonable to me, if speed ends up being an issue it can be optimized

PR #1213, opened 2024-05-30
patchgte to asm

sozud:
I think this file is all handwritten asm

patch_gte
https://gist.github.com/sozud/f25a15fe190fc8fa77b59b2278de4d85

func_8001929C
https://gist.github.com/sozud/27df133a0207fa732991b0786ee2f34c

func_8001929C is the code used to patch the bios. The sequence
```
/* 9A3C 8001923C B0000A24 */  addiu      $t2, $zero, 0xB0
/* 9A40 80019240 09F84001 */  jalr       $t2
/* 9A44 80019244 56000924 */   addiu     $t1, $zero, 0x56
```
looks like a bios call to me and I'm not sure this can be generated by a compiler? Scratch here
https://decomp.me/scratch/TlFnw

### DIFF
--- a/config/splat.us.main.yaml
+++ b/config/splat.us.main.yaml
@@ -154,7 +154,7 @@ segments:
       - [0x956C, c, psxsdk/libgte/fgo_05]
       - [0x970C, c, psxsdk/libgte/fgo_06]
       - [0x98AC, c, psxsdk/libgte/ratan]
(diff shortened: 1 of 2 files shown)


Xeeynamo (approved):
Makes sense. We can always revert it back if we find a way to write that stuff in C

PR #1214, opened 2024-05-30
 Factor out UpdateRoomPosition 

sozud:


Xeeynamo (approved):
Looks good. We should keep in mind to put all the variable declarations on top for function matching stuff, since that's how probably the source code looked like

PR #1216, opened 2024-05-30
Factor out CollectHeartVessel

sozud:


Xeeynamo (approved):
Simple but an effective change

PR #1218, opened 2024-05-30
 Factor out EntityIntenseExplosion 

sozud:


Xeeynamo (approved):
Great to see `src/st/wrp_psp` shrinking

PR #1219, opened 2024-05-30
Merge g_CdVolumeTable with g_VolumeTable

Xeeynamo:
I found the variable `distance` in `func_80132A04` can be negative. The array index will underflow unless the two global variables `g_CdVolumeTable` and `g_VolumeTable` gets merged


sozud (approved):
Nice find

PR #1220, opened 2024-05-30
EntityMessageBox

sozud:
https://decomp.me/scratch/nMwCS


Xeeynamo (approved):
Very interesting. I started looking into this function but I gave up. Thanks for looking into it.

For that `char* str` to `u8* str` I think we should add the `-char unsigned` parameter to MWCCPSP so we can avoid casts here and there. Can you make this change either in this PR or in a separate one before we forget?
JFYI, on PSX we also have `char`s as `unsigned`.

```
$ bin/wibo bin/mwccpsp.exe --help | grep char

  -char keyword              # set sign of 'char'                              
     signed                  #    chars are signed; default                    
     unsigned                #    chars are unsigned    
```

sozud:
> Very interesting. I started looking into this function but I gave up. Thanks for looking into it.
> 
> For that `char* str` to `u8* str` I think we should add the `-char unsigned` parameter to MWCCPSP so we can avoid casts here and there. Can you make this change either in this PR or in a separate one before we forget? JFYI, on PSX we also have `char`s as `unsigned`.
> 
> ```
> $ bin/wibo bin/mwccpsp.exe --help | grep char
> 
>   -char keyword              # set sign of 'char'                              
>      signed                  #    chars are signed; default                    
>      unsigned                #    chars are unsigned    
> ```

Done

PR #1223, opened 2024-05-30
Allow PSP stage diffing

sozud:
I needed to diff locally to figure out issues with EntitySoulStealOrb, seems like diff.py can't do psp so added some support.

Example usage
```
./tools/asm-differ/diff.py --version pspeu --overlay st/wrp EntitySoulStealOrb
```

### DIFF
--- a/diff_settings.py
+++ b/diff_settings.py
@@ -70,6 +70,17 @@ def apply_psx_stage(config, version, name):
     ]
     config["objdump_executable"] = "mipsel-linux-gnu-objdump"
 
+def apply_psp_stage(config, version, name):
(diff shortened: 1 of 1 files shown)


bismurphy:
Maybe I'm missing something, but I can already run the differ on pspeu without these changes. As long as `VERSION` is set to `pspeu`, simply using `differ st/wrp CollectSubweapon` (I use the recommended `differ` alias) successfully shows me the interface for the psp version.

Xeeynamo:
differ works on my side locally as well. What error were you getting locally before this PR?

sozud:
Yeah this seems to be working now, not sure what the issue was.

PR #1225, opened 2024-06-01
Decompile Weapon 28 shield spell

bismurphy:
This one was a beast, but nice to get it working.

This seemed to match the same entity extension as weapon 29. Weapons 28 and 29 are both shields, so I decided to rename the Weapon29 struct to be Shield instead. We will see if it continues to match for the future shields.

This had a ton of FIX(), enums, #defines, etc, so hopefully I got them all :)


Xeeynamo (approved):
Looks good!

PR #1228, opened 2024-06-01
Decompile weapon 28 EntityWeaponAttack

bismurphy:
This finishes out the Skull Shield.

Nothing too terribly special. I added a new #define for the player status flag at 0x10000, since that's used here. Hopefully we can find the function that actually sets that flag; I couldn't find it.


src/weapon/w_028.c
```diff
@@ -4,9 +4,142 @@
...
+        }
+        break;
+    }
+    if (self->step != 4) {
+        g_api.func_8010DBFC(D_C8000_8017AA98, D_C8000_8017AAE0);
+    }
+    self->drawFlags = PLAYER.drawFlags;
+    self->rotY = PLAYER.rotY;
+    self->rotPivotY = PLAYER.rotPivotY;
+    return;
```
Xeeynamo, line 141: `return` at the end unnecessary?

Xeeynamo (approved):

src/weapon/w_028.c
bismurphy: thanks, leftover from the classic "default case in a random spot in the middle of the switch needs to be moved to the end"

PR #1234, opened 2024-06-01
Decompile Weapon 27 func_ptr_80170024

bismurphy:
I'm pretty sure this has to be the laser beams for the Medusa Shield. They get spawned by the EntityWeaponShieldSpell.

This function is a pretty big mess. It could probably benefit from a cleanup pass, potentially using PSP to assist.

But for now, the goal is matching, so hey, this matches :)


Xeeynamo (approved):
I think the function is fine. I can see some potential things that might be a bit more real than this current match, but as you said it matches lol. Nice match!

bismurphy:
I'm confused on why the CI is failing. I've checked it many times now on my computer and it matches. Is there any way to obtain the non-matching binary that was built in the CI, to determine what might be going wrong? Or, could someone else check this branch and see if it matches on their end?

Xeeynamo:
> I'm confused on why the CI is failing. I've checked it many times now on my computer and it matches. Is there any way to obtain the non-matching binary that was built in the CI, to determine what might be going wrong? Or, could someone else check this branch and see if it matches on their end?

It matches locally on my end too... I am re-running the CI

bismurphy:
Build error is related to the arguments for func_80118B18. I am going to try to find the PSP version of that function to see what the problem is.

PR #1237, opened 2024-06-02
BlitChar

sozud:
The function itself is matching for me locally but wrp.bin doesn't match. I think it's possibly a wrong file split?
https://decomp.me/scratch/tsOag

### DIFF
--- a/include/game.h
+++ b/include/game.h
@@ -1389,8 +1389,14 @@ typedef struct {
     /* 8003C89C */ void (*DebugInputWait)(const char* str);
     /* 8003C8A0 */ void* unused12C;
     /* 8003C8A4 */ void* unused130;
+    // this matches on both versions but doing this to show the difference
+#if defined(VERSION_PSP)
+    /* 8003C8A8 */ u16* (*func_ptr_91CF86C)(u32 arg0, u16 kind);
+    /* 8003C8AC */ u16 (*func_ptr_91CF870)(char*, u8* ch);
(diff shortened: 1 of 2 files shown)


Xeeynamo:
There are a few things to fix here. The mis-alignment comes from these two lines:

```c
u16* (*func_ptr_91DF86B)(u32 arg0, u16 kind);
u16 (*func_ptr_91DF86F)(char*, u8* ch);
```

As they are global variables and not function prototypes, you need an `extern`. Otherwise the compiler will add some gap for the `.data` section.

`func_ptr_91DF86F` is the wrong offset. I read this in the code:
```asm
/* 4E4C 0923C4CC 1D09023C */  lui        $v0, (0x91D0000 >> 16)
/* 4E50 0923C4D0 70F8428C */  lw         $v0, -0x790($v0)
```
`0x91D0000 - 0x790` is `0x91CF870`, not `91DF86F`.

The offset looks very close to where `g_api` is. And actually `func_ptr_91DF86B` can be put to `unused134` and `func_ptr_91DF86F` to `unused138` to then just call `g_api.func_ptr_91DF86B` and still get an :ok: .
Very interesting as those g_api fields are unused on PSX and set to `NULL`.

sozud:
Thanks, it's matching now

Xeeynamo (approved):

Xeeynamo:
Congrats! It was not an easy one

PR #1238, opened 2024-06-02
Decompile Weapon 27 EntityWeaponAttack

bismurphy:
Started working on this one and then I realized it's identical to the one on the previous shield, so it was a direct copy-paste, aside from the names of the D_C1000 variables.

Maybe we'll de-duplicate in the future but for now I think weapons should just be written individually.


Xeeynamo (approved):
I agree on not de-duplicating functions between weapons that repeat less than a few times. They were probably copy&pasted anyway.

PR #1240, opened 2024-06-02
DRA import data at 2F324

Xeeynamo:
I think those are a bunch of animation indices? I am not yet sure how to represent them as structured data, so there you go having them imported as C files.


sozud (approved):
We can continue making it more structured as we go. I'm thinking this was script-generated? Maybe post a gist of the script here or in the file so we can refer to it and regenerate if needed

Xeeynamo:
Unfortunately it was done by hand with the help of a few regex. The MSVC compiler is throwing a C2133, which is very annoying. I am not sure how to fix it :/

PR #1242, opened 2024-06-02
Start importing wrp data

sozud:
The yaml changes were generated with a script. https://gist.github.com/sozud/ae6837a24bc8bb4e8bcc89d3e14d83f4

The alignment of the yaml seems to be wrong so I had to check if the address % 16 == 0 and only split at those spots. I put the data in create_entity since it's the first file in the psp yaml.


Xeeynamo (approved):
This is a great initiative! Feel free to put this in whenever you want a check-point

PR #1245, opened 2024-06-03
Decompile DRA func_8010D010

bismurphy:
The next one! And yet another renaming struct to Point16.

Almost done!


sozud:
D_800ACEE0 needs to be stubbed or the data imported it seems

sozud (approved):

Xeeynamo:
> D_800ACEE0 needs to be stubbed or the data imported it seems

That part is done in my PR #1246

PR #1248, opened 2024-06-03
Deprecate the s16_pair struct

bismurphy:
A while ago, I created the s16_pair struct, and used it in various functions. As the name implies, it is a pair of s16 variables.

I later found that there already exists a Point16 struct which has the same data, and is a better description anyway, since its members are `x` and `y`, and this structure is almost always used to represent x-y values.

I've gone through the repo for every use of this structure, and changed it over to be Point16. Therefore, s16_pair will no longer exist in this project.


Xeeynamo (approved):
I love some good code haircut :haircut:

PR #1254, opened 2024-06-04
Decompile DRA func_800EA7CC

joshlory:
https://decomp.me/scratch/lSGji

Credit to @Xeeynamo for 99% of the initial match, I just fixed up one of the registers and did some light refactoring. The repetition of `temp_s0 = ptr->desc` makes me think it's actually an array reference but I couldn't get anything else to fit.

### DIFF
--- a/src/dra/4A538.c
+++ b/src/dra/4A538.c
@@ -119,7 +119,182 @@ u16 func_800EA720(u32 colorDst, u32 colorSrc) {
     return colorRes;
 }
 
-INCLUDE_ASM("dra/nonmatchings/4A538", func_800EA7CC);
+void func_800EA7CC(void) {
+    // stores which palettes have been invalidated and need to be re-uploaded
+    // to the VRAM
+    u8 palettes[0x30];
+    Unkstruct_8006C3C4* ptr;
+    u_long* temp_s0;
+    u_long* temp_v0_5;
+    u_long* temp_s0_4;
+    s32 count;
+    s32 var_a3;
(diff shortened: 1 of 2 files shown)


sozud:
I grabbed the PSP version because I wanted to check
```
                new_var = temp_v0_5[-2];
                clut += new_var;
                count = temp_v0_5[-1];
```
but it has the same negative offsets 
image

https://decomp.me/scratch/mlr4O

Note this function is -O4 so it's not as easy

src/dra/4A538.c
```diff
@@ -119,7 +119,182 @@ u16 func_800EA720(u32 colorDst, u32 colorSrc) {
...
+        palettes[i] = 0;
+    }
+
+    // cycles through a list of palette operations stored in D_8006C3C4
+    for (i = 0, ptr = D_8006C3C4, var_a3 = 0xFFFF; i < 0x20; i++, ptr++) {
+        if (ptr->unk8 == 0) {
+            continue;
+        }
+        if (ptr->index != 0) {
+            ptr->index = ptr->index + var_a3;
```
joshlory, line 159: Given `var_a3 = 0xFFFF` this pattern is likely `ptr->index--` or similar.

Xeeynamo (approved):
This is huge. It is one of the functions I've been struggling the most to get it in. Thank you very much for dealing with it!

Let me know if you would like to polish it a bit more or you want to get this merged in now. To me it is good enough.

PR #1255, opened 2024-06-04
Decompile DRA EntityPlayerPinkEffect

joshlory:
PSX: https://decomp.me/scratch/2YXLH
PSP: https://decomp.me/scratch/yJ6yP

image

After correcting the `PLAYER` and `g_Player` references I have a small remaining diff, do I need to create symbols for these so they resolve without dot notation? Also the location of the last `.rodata+0xd8` is off.

Thanks @sozud for help with the PSP version!


sozud:
Could you post a scratch with the fixed symbols so we can see what the difference could be? e.g. not this kind `g_Player_D_80072F00_0`

joshlory:
Ah I see, the `.rodata` diff means there's no default case on the inner switch statement. I will take another look.

src/dra/dra.h
```diff
@@ -411,9 +411,11 @@ extern AnimationFrame D_800ADC10[];
 extern AnimationFrame D_800ADC44[];
 extern u16 D_800ADC7C[2];
 extern Point16* D_800ADCC8[];
+extern u8 D_800AE120[];
 extern RECT D_800AE130;
 extern s32 D_800AE270[9];
 extern AnimationFrame D_800AE294[];
+extern Unkstruct_800ADEF0 D_800ADEF0[];
```
Xeeynamo, line 418: Storing this in a shared header is breaking the build. For simplicity I suggest to just move it into `7E4BC.c`

src/dra/7E4BC.c
```diff
@@ -893,7 +893,94 @@ void EntityHitByIce(Entity* self) {
...
+            }
+        }
+
+        self->ext.factory.unk7E++;
+        self->ext.timer.t = data_idx->unk0[self->ext.factory.unk7E];
+        if (self->ext.timer.t == 0xFF) {
+            temp2 = data_idx->unk18;
+            switch (temp2) {
+            case 0x83:
+                if (PLAYER.step == 0xB) {
```
Xeeynamo, line 938:
```suggestion
                if (PLAYER.step == Player_StatusStone) {
```
Most of the `PLAYER.step` are documented by an enum. `0x0B` should be `Player_StatusStone`!

Xeeynamo (commented):
Left a couple of comments. You are almost there with this one!!

joshlory:
I think we need to replace the definition in `src/dra/d_DBD4.c` with the new struct.

Xeeynamo:
> I think we might need to replace the definition in `src/dra/d_DBD4.c` with the new struct.

100%. If you want, let's do it as part of a separate pull request. I do not want to overwhelm you with more changes on this one. I think it is good enough to be merged. Thank you!

Xeeynamo (approved):
> I think we might need to replace the definition in `src/dra/d_DBD4.c` with the new struct.

100%. If you want, let's do it as part of a separate pull request. I do not want to overwhelm you with more changes on this one. I think it is good enough to be merged as it is. Thank you!

PR #1256, opened 2024-06-04
Decompile weapon 26 func_ptr_80170024

bismurphy:
Kind of a crazy one with the way the vectors are dealt with, but eventually got it to work. Hopefully this can be a nice reference for some of those SDK functions.

### DIFF
--- a/src/weapon/w_026.c
+++ b/src/weapon/w_026.c
@@ -50,24 +50,209 @@ void func_BA000_8017B37C(void) {
 
 INCLUDE_ASM("weapon/nonmatchings/w_026", EntityWeaponShieldSpell);
 
-const s32 D_BA000_8017AB7C[] = {
-    0xFFE40008,
-    0x00000000,
-};
-const s32 D_BA000_8017AB84[] = {
-    0xFFE40038,
-    0x00000000,
-};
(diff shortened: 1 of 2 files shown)


src/weapon/w_026.c
```diff
@@ -50,24 +50,209 @@ void func_BA000_8017B37C(void) {
...
+    SetGeomOffset(self->posX.i.hi, self->posY.i.hi);
+    RotMatrix(&rotation_axis, &m);
+    TransMatrix(&m, &transfer_vector);
+    SetRotMatrix(&m);
+    SetTransMatrix(&m);
+    prim = &g_PrimBuf[self->primIndex];
+
+    rotTransResult = RotTransPers4(
+        vecs[0][0], vecs[0][1], vecs[0][2], vecs[0][3], (long*)&prim->x0,
+        (long*)&prim->x1, (long*)&prim->x3, (long*)&prim->x2, &spA0, &spA4);
```
sozud, line 244: spA0/A4 could be renamed but they are unused anyway
image

sozud (approved):

src/weapon/w_026.c
bismurphy: Yeah, I figured if they are unused there's not much point in naming them

PR #1261, opened 2024-06-05
DRA func_8010DBFC refactor 

Xeeynamo:
`func_8010DBFC` now matches on PSP, so it's most likely to be real: https://decomp.me/scratch/xuNg5

`func_8010DA70` has a reg-swap but it's matching better than the original: https://decomp.me/scratch/mGIed
I am a bit concerned about this function as it seems to suggst `AnimationFrame` is a structure that does not exist and it might just be a `u16` array.


sozud (approved):
To me it seems likely that AnimationFrame is just an array, we can continue collecting data on it as we do more PSP functions

Xeeynamo:
> To me it seems likely that AnimationFrame is just an array, we can continue collecting data on it as we do more PSP functions

I feel the same... which is a bit of a shame. I will merge this as it is, despite it looking a bit ugly. I want more proof that this is an array before removing that struct

PR #1262, opened 2024-06-05
DRA fix D_800CF324 data type

Xeeynamo:
Slightly better representation of some data I previously imported


src/dra/d_2F324.c
```diff
@@ -1,1656 +1,556 @@
...
-u32 D_800CFE40[] = {
-    0xFFF480D9,
-    0x0000FFEC,
-};
+typedef struct {
+    u16 frame;
+    s16 pivotX;
+    s16 pivotY;
+    u16 clut;
+} AluFrame;
```
sozud, line 8: Have we done any PSP functions that use AluFrame? Wondering if it's going to turn out to be an array again

sozud (approved):

src/dra/d_2F324.c
Xeeynamo: Not really, it's just for convenience I am doing this here. We have no function that uses `AluFrame`, which the struct definition is even self-contained within the C file.
I am still figuring out many things about the data, but I feel this is at least another stepping stone?

PR #1263, opened 2024-06-05
Decompile DRE func_801A2848

bismurphy:
Kind of a messy function to match, but it matches. Looks like there are a few functions in this file with similar structure.

### DIFF
--- a/src/st/dre/22848.c
+++ b/src/st/dre/22848.c
@@ -9,8 +9,61 @@ long RotTransPers3(SVECTOR* v0, SVECTOR* v1, SVECTOR* v2, long* sxy0,
                    long* sxy1, long* sxy2, long* p, long* flag);
 long NormalClip(long sxy0, long sxy1, long sxy2);
 
-Primitive* func_801A2848(Primitive* prim, u8* model, u16* arg2);
-INCLUDE_ASM("st/dre/nonmatchings/22848", func_801A2848);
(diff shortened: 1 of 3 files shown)


src/st/dre/22848.c
```diff
@@ -9,8 +9,61 @@ long RotTransPers3(SVECTOR* v0, SVECTOR* v1, SVECTOR* v2, long* sxy0,
...
-Primitive* func_801A2848(Primitive* prim, u8* model, u16* arg2);
-INCLUDE_ASM("st/dre/nonmatchings/22848", func_801A2848);
+Primitive* func_801A2848(Primitive* prim, u8* arg1, u16* arg2) {
+    s32 pad0;
+    s32 pad1;
+    s32 pad2;
+    s32 pad3;
+    s32 pad4;
+
+    pad0 = *SPAD(arg1[0]);
```
Xeeynamo, line 19: Wanna use `SP` instead?

Xeeynamo (approved):
LGTM, just left a minor comment

src/st/dre/22848.c
Xeeynamo: What if SPAD internally uses SP? Reason is the scratch pad offset on PSP is different and on PC I got far too many unexpected crashes not knowing a hardcoded address was used for the scratchpad. Using SP, even if SPAD is wrapping it, will at least a compilation error

bismurphy: Sure, that sounds like a nice idea. Will implement.

PR #1265, opened 2024-06-05
Refactor Unkown Entity 15

hohle:
Pulls out `EntityUnkId15` and maps related symbols.


src/st/entity_unkId15.h
```diff
@@ -0,0 +1,26 @@
...
+        arg0->rotY = arg0->rotX = unk15_rot[arg0->params];
+        arg0->velocityY = unk15_yVel[arg0->params];
+        arg0->step++;
+    } else {
+        arg0->animFrameDuration++;
+        arg0->posY.val -= arg0->velocityY;
+        if (!(arg0->animFrameDuration & 1)) {
+            arg0->animCurFrame++;
+        }
+        if (arg0->animFrameDuration >= 37) {
```
sozud, line 22: You're going to need to use the PSP version, this is different for example
```
if (self->animFrameDuration > 36) {
```

config/symbols.pspeu.stwrp.txt
```diff
@@ -107,3 +107,5 @@ g_eBreakableExplosionTypes = 0x0924BCC8;
 D_80180608 = 0x0924CCD0;
 D_80180530 = 0x0924CCD8;
 D_80180648 = 0x0924CCE0;
+unk15_rot = 0x80180FF8;
```
sozud, line 110: I"m thinking these addresses are off.
```
mipsel-linux-gnu-ld: build/pspeu/src/st/wrp_psp/e_misc.c.o: in function `EntityUnkId15':
(.text+0x7c): undefined reference to `unk15_rot'
mipsel-linux-gnu-ld: (.text+0x80): undefined reference to `unk15_rot'
mipsel-linux-gnu-ld: (.text+0xac): undefined reference to `unk15_yVel'
mipsel-linux-gnu-ld: (.text+0xb0): undefined reference to `unk15_yVel'
```
You can look at the original asm here for the addresses https://decomp.me/scratch/ez7sR

src/st/entity_unkId15.h
hohle: Ah, I need to set up PSP builds.

hohle:
I have a local PSP EU build going now and found that the order of some lines appears different. Rather than trying to get a shared implementation working, I've reverted everything but the symbol changes from the PSP version.

sozud:
The PSP and PSX versions are the same though, here's the PSP version pasted into the PSX scratch https://decomp.me/scratch/Ai7Op
The PSP version is just closer to the original code

hohle:
I pulled out the PSP version which, as you said, compiled perfectly with the PSX and PSP versions.

src/st/wrp/e_misc.c
```diff
@@ -83,7 +83,7 @@ u16 g_eUnk14SpawnRots[] = {
 
 #include "../entity_unkId15_spawner.h"
 
-u16 D_80180FF8[] = {
+s16 unk15_rot[] = {
```
Xeeynamo, line 86: One small change here. We tend to call global variables with the prefix `g_` (so in this case `g_Unk15_rot`). That way it's clearer if a variable is either local or global when reading functions.

The `g_` thing does not apply whenever we can mark the variable as `static`. This is a great opportunity to mark the array as `static. as it looks like it is not used anywhere else! No need to do anything it in other overlays as it can give us a clear indicator in the future that `unk15_rot` can be self-contained in any of the C files the data will be imported to.

Xeeynamo (approved):
Thank you for doing this!

PR #1268, opened 2024-06-06
Decompile weapon 21 (food) func_ptr_80170004

bismurphy:
Unfortunately, this required an evil change of the return type, even though it doesn't return. And since it's one of many weapons that share the functions, they all had to change.

Otherwise, nice to have the food figured out (this is the last food function), since food is a somewhat major part of this game.

### DIFF
--- a/src/weapon/w_021.c
+++ b/src/weapon/w_021.c
@@ -8,6 +8,8 @@
 #include "weapon_private.h"
 #include "shared.h"
 
+extern SpriteParts D_97000_8017A040[];
+
 // Similar to BottomCornerText which exists in many overlays
 // This one seems to process 2 characters at a time, weird.
 void func_97000_8017AB54(u8* str, u8 lowerLeft) {
@@ -206,8 +208,6 @@ s32 func_97000_8017AF2C(Entity* self, s32 arg1) {
     }
     return 1;
 }
-
-extern SpriteParts D_97000_8017A040[];
 
 void EntityWeaponAttack(Entity* self) {
     Collider sp10;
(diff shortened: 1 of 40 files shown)


sozud (approved):
Looks good to me, thinking Xeeynamo should look at this one too due to the API changes

bismurphy:
Sounds good. For context, this is the same kind of change as was applied in https://github.com/Xeeynamo/sotn-decomp/pull/1128. (Wow, there have been 140 PRs since then! Crazy!)

src/weapon/w_021.c
```diff
@@ -415,7 +415,114 @@ void EntityWeaponAttack(Entity* self) {
...
+        self->animCurFrame = upperParams - 0x1C;
+        self->posY.i.hi -= 4;
+        SetSpeedX((rand() & 0x3FFF) + FIX(0.875));
+        self->velocityY = FIX(-2);
+        DestroyEntityWeapon(1);
+        self->facingLeft = 0;
+        self->hitboxWidth = 8;
+        self->hitboxHeight = 8;
+        self->step++;
+        return;
```
Xeeynamo, line 452:
```suggestion
        break;
```

src/weapon/w_021.c
```diff
@@ -415,7 +415,114 @@ void EntityWeaponAttack(Entity* self) {
...
+                g_api.FreePrimitives(g_unkGraphicsStruct.BottomCornerTextPrims);
+                g_unkGraphicsStruct.BottomCornerTextTimer = 0;
+            }
+            g_api.PlaySfx(0x67C);
+            temp_s1 = g_api.equipDefs[upperParams].name;
+            g_api.AddToInventory(upperParams, EQUIP_HAND);
+            func_97000_8017AB54(temp_s1, 1);
+            DestroyEntity(self);
+        }
+    }
```
Xeeynamo, line 524:
```suggestion
        break;
    }
```
let's add a `break` there, it makes easier to read the end of a `switch` statement

src/weapon/w_021.c
```diff
@@ -415,7 +415,114 @@ void EntityWeaponAttack(Entity* self) {
...
+    const char* temp_s1;
+    s16 xVar;
+    s16 yVar;
+
+    upperParams = (self->params >> 8) & 0x7F;
+
+    switch (self->step) {
+    case 0:
+        SetSpriteBank1(D_97000_8017A040);
+        self->animSet = -0x7FF0;
```
Xeeynamo, line 432: This should be `ANIM_OVL(0x10)` I think

Xeeynamo (approved):
Cursed match but it's okay if it is required. I left a few minor :nail_care: , other than that it looks good

PR #1270, opened 2024-06-06
De-duplicate EntityUnkId13

bismurphy:
In the interest of reducing uses of `generic`, and especially the `entityPtr` member of it, I found this function which had several duplicates already, all of which used `generic`. Rather than cleaning it up in every instance, I decided instead to de-duplicate it, and then clean it up in the .h file.

I needed to make a new entity extension for this, and noticed that we had `ET_Entity16`, but then I realized that this was being used for `g_Entities[16]`, so I renamed that to be `ET_EntitySlot16`, leaving `ET_EntityXX` available, for XX being the ID of an unknown entity.


sozud:
Would you mind adding the PSP version? https://github.com/Xeeynamo/sotn-decomp/blob/1df0706a0473103c91a93c012d1a7898897b8112/src/st/wrp_psp/e_misc.c#L45

bismurphy:
> Would you mind adding the PSP version?
> 
> https://github.com/Xeeynamo/sotn-decomp/blob/1df0706a0473103c91a93c012d1a7898897b8112/src/st/wrp_psp/e_misc.c#L45

Ah yes, will do. My de-duplicator script pulls from duplicates.txt so it doesn't catch PSP. PSP deduplication incoming!

bismurphy:
Done, PSP had a slight mismatch due to the formulation of an inequality test.

PR #1271, opened 2024-06-06
Add blueprint support

Xeeynamo:
I had to cherry-pick some existing changes from another branch, but that part is tested and working.

For the blueprint stuff, I changed the fields from the strings `"TRUE"` and `"FALSE"` to native JSON boolean values `true` and `false`. That made things much easier to parse.

The parsing mechanic is straightforward. I'd like to have @bismurphy review on the blueprint parser and asset changes specifically.

One of the blueprints in action:

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,8 @@ set(SOURCE_FILES_PC
     src/pc/str.c
     src/pc/sim_pc.c
     src/pc/pl_ric.c
-    src/pc/stage_dummy.c
+    src/pc/stages/stage_loader.c
+    src/pc/stages/stage_dummy.c
 )
 
 if(WIN32)
@@ -112,7 +113,7 @@ set(SOURCE_FILES_DRA
 )
 
 set(SOURCE_FILES_STAGE_SEL
-    src/pc/stage_sel.c
+    src/pc/stages/stage_sel.c
     src/st/sel/banks.c
     src/st/sel/CD54.c
     src/st/sel/2C048.c
--- a/src/pc/pc.h
+++ b/src/pc/pc.h
@@ -24,8 +24,16 @@ typedef struct {
     size_t length;
 } FileLoad;
 
+typedef struct {
+    const char* path;
+    const char* content;
+    size_t length;
+    void* param;
+} FileStringified;
+
 bool FileRead(bool (*cb)(FILE* file), const char* path);
-bool FileStringify(bool (*cb)(const char* content), const char* path);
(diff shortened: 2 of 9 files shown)


bismurphy:
This looks reasonable to me! If I'm understanding it correctly:

The current process extracts the blueprints to JSON, and then when compiling, re-creates the byte sequences in the game binary.

In your implementation here, you're loading the JSON file at runtime, which would mean being able to change blueprints on the fly?

Not sure why the coloration is wrong on the gravity boot jump, but you're definitely spawning the entity, which is super cool!

Xeeynamo:
> In your implementation here, you're loading the JSON file at runtime, which would mean being able to change blueprints on the fly?

Yeah that's what I am currently doing, which might help to quickly test changes in-game and to quickly create mods. Obviously this might change if we want to target weaker devices that can't afford to parse JSON files. 

> Not sure why the coloration is wrong on the gravity boot jump, but you're definitely spawning the entity, which is super cool!

The PSX (and also PS2) graphics pipeline is a bit weird on something I keep failing to understand. There must be somewhere a flag that tells the GPU to render `0x80` as a full white colour. Or maybe I just have to multiply the RGBs by two and clamp at `0xFF`, but that breaks other things. The colors look a bit dark because RGBs are at half rate. Also the keyhole background is much darker than what is supposed to be

bismurphy:
> There must be somewhere a flag that tells the GPU to render 0x80 as a full white colour.

Hmmm, almost sounds like it's seeing it as a signed byte, in that case. I don't see why that would be, but it lines up.

dezgeg:
> The PSX (and also PS2) graphics pipeline is a bit weird on something I keep failing to understand. There must be somewhere a flag that tells the GPU to render 0x80 as a full white colour. Or maybe I just have to multiply the RGBs by two and clamp at 0xFF, but that breaks other things. The colors look a bit dark because RGBs are at half rate. Also the keyhole background is much darker than what is supposed to be

From nocash' docs:
```
Caution: For untextured graphics, 8bit RGB values of FFh are brightest. However, for texture blending, 8bit values of 80h are brightest (values 81h..FFh are "brighter than bright" allowing to make textures about twice as bright as than they were originially stored in memory; of course the results can't exceed the maximum brightness, ie. the 5bit values written to the framebuffer are saturated to max 1Fh).
```

Xeeynamo:
> dezgeg

:eyes: very interesting to know, thanks for letting us know. I will make a patch around the week-end!

PR #1272, opened 2024-06-06
Add Alucard sprite support

Xeeynamo:
And it does not crash any more.

trust me bro

### DIFF
--- a/src/pc/pl_arc.c
+++ b/src/pc/pl_arc.c
@@ -1,0 +1,40 @@
+#include <game.h>
+#include <log.h>
+#include "pc.h"
+
+u8* g_PlOvlSpritesheet[256];
+static u8* sprite_data = NULL;
+
+void InitPlayerArc(FileLoad* file) {
+    int i;
+
+    g_PlayableCharacter = PLAYER_ALUCARD;
+
+    if (sprite_data) {
+        free(sprite_data);
+    }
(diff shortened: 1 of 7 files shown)


src/pc/pl_arc.c
```diff
@@ -0,0 +1,40 @@
...
+
+void InitPlayerArc(FileLoad* file) {
+    int i;
+
+    g_PlayableCharacter = PLAYER_ALUCARD;
+
+    if (sprite_data) {
+        free(sprite_data);
+    }
+    // this is not freed anywhere else, technically a memory leak but it's not
```
sozud, line 16: You mean because it doesn't continue to grow at runtime?

Xeeynamo: It is because it is not freed before the application exits. I've been meaning to implement a wrapper around `atexit` to handle these kind of scenarios without coupling the code too much.

It never grows though. If the init function is called twice, it just frees the previous one

sozud: I think we should turn on leaksanitizer eventually but it's not a big deal right now https://clang.llvm.org/docs/LeakSanitizer.html

sozud (approved):

src/pc/pl_arc.c
Xeeynamo: Let's keep that in mind, it definitively sounds like a good option!

PR #1273, opened 2024-06-06
Refactor Entity Explosion

hohle:
Pulls out `EntityExplosion` into a common file.

`cen` and `rwrp` appear to have not imported a declaration for `AnimateEntity` and call it as an undeclared function with a int-sized return value. Since `AnimateEntity` is declared in the same compilation unit a workaround is done to fix how the return value is interpreted.


Xeeynamo (approved):
It looks very good! A bit of a shame for relying on `MISSING_ANIMATE_ENTITY_PROTOTYPE` but I am sure it is a temporary issue while we figure out how to split C files on other overlays. This is good to be merged, thanks!

Issue #739, opened 2023-11-07
Make 4.4 does not work when building in parallel


Xeeynamo:
# What

The latest minor version bump of GNU Make does not work well with this project due to some [documented breaking changes](https://lists.gnu.org/archive/html/help-make/2022-10/msg00020.html). The problem only occurs when building happens in parallel (e.g. `make -j build`. The problem is GCC wants to compile C files and output them to directories that do not yet exist.

This good first issue will help you to familiarise with our build chain.

## Technical details

In the Makefile there are targets that are responsible for creating the output directories:

```
tt_%_dirs:
	$(foreach dir,$(ASM_DIR)/servant/tt_$* $(ASM_DIR)/servant/tt_$*/data $(SRC_DIR)/servant/tt_$* $(ASSETS_DIR)/servant/tt_$*,$(shell mkdir -p $(BUILD_DIR)/$(dir)))
st%_dirs:
	$(foreach dir,$(ASM_DIR)/st/$* $(ASM_DIR)/st/$*/data $(SRC_DIR)/st/$* $(ASSETS_DIR)/st/$*,$(shell mkdir -p $(BUILD_DIR)/$(dir)))
%_dirs:
	$(foreach dir,$(ASM_DIR)/$* $(ASM_DIR)/$*/data $(SRC_DIR)/$* $(ASSETS_DIR)/$*,$(shell mkdir -p $(BUILD_DIR)/$(dir)))
```

When an overlay is built, these targets are called. For example doing `make cen` calls the target `cen: stcen_dirs $(BUILD_DIR)/CEN.BIN` to first create the directories and then the output binary. This behaviour was working on Make 4.3 and below as dependent targets were always processed in order.

## Potential solution

Before GCC is called, `mkdir -p $(dir $@)` can be called to ensure the directory exists. This is already done in `Makefile.saturn.mk`. Once this part is done, the various targets `%_dirs` can be deleted.

On Arch Linux I personally made a back-up of make [4.3-5.1](https://discord.com/channels/1079389589950705684/1079395108501331990/1105436043399929856) to easily downgrade and test things around.

## What is affected

* [Arch Linux](https://archlinux.org/packages/core/x86_64/make/) since October 2022
* [Gentoo](https://packages.gentoo.org/packages/sys-devel/make) since October 2022
* [Fedora](https://packages.fedoraproject.org/pkgs/make/make/) since version 38
* [openSUSE](https://software.opensuse.org/package/make) since February 2023

[...]

PR #1275, opened 2024-06-07
Makefile Quality of Life Changes

hohle:
The `check` recipe now has dependencies on all of the files it checks allowing them to be rebuilt if necessary.

Consistent use of `dd` and suppressing `dd`'s output which noise for the replacement and padding it's doing.

Run 7z in non-interactive mode to avoid confirming overwrite of files during extraction (equivalent behavior to `VERSION=us` `extract_disk`).

`patch` is a phony target.


Makefile
```diff
@@ -147,8 +149,8 @@ format: bin/clang-format
 	./tools/symbols.py remove-orphans config/splat.us.stmad.yaml
 patch:
 	$(DIRT_PATCHER) config/dirt.$(VERSION).json
-check: patch
-	sha1sum --check config/check.$(VERSION).sha
+check: config/check.$(VERSION).sha patch $(CHECK_FILES)
```
Xeeynamo, line 152: The `$(CHECK_FILES)` is clever, but I do not want to couple `check` with `build` (this is already done by `all`). This is because there are times we do quick checks like `make clean && make build/us/dra.ld && make dra && make check` to check a single overlay instead of having to build everything.

Also I do not want to couple the Makefile task name with the config check name, which already proven to be quite difficult to move things in the past.

Xeeynamo (commented):
I left one comment for something I am not fond of the proposed changes. Everything else looks good.

Makefile
hohle: It seems odd from a build perspective that at the end of that command sequence the shell result would be failure but the human interpretation would be success. Would you be opposed to a separate recipe to ignore do a check while ignoring missing objects (`shasum --check --ignore-missing`) so a workflow like that would still result in success?

> (this is already done by all)

`all` has dependencies on `check` and `build` but make doesn't guarantee an order that dependencies are run (also seen in #739 where the tasks that actually depend on `*_dirs` don't have dependencies but rely on a dependent, which isn't how make is designed). Make only guarantees that dependencies will be completed prior to running the dependent target. Since `check` doesn't explicitly depend on any built artifacts there are often times were `check` will be run (e.g. during a `make -j`) prior to build finishing. I just ran this after a rebase:

```
$> make -j
⋮
./tools/splat_ext/animset.py gen-asm assets/st/wrp/D_80186D88.animset.json build/us/assets/st/wrp/D_80186D88.s
./tools/splat_ext/roomdef.py assets/st/wrp/g_Rooms.roomdef.json build/us/assets/st/wrp/g_Rooms.bin
./tools/splat_ext/layers.py assets/st/wrp/rooms.layers.json build/us/assets/st/wrp/rooms.s
sha1sum --check config/check.us.sha
build/us/main.exe: OK
build/us/DRA.BIN: OK
build/us/WEAPON0.BIN: OK
build/us/RIC.BIN: OK
build/us/CEN.BIN: OK
build/us/F_CEN.BIN: OK
build/us/DRE.BIN: OK
build/us/F_DRE.BIN: OK
build/us/MAD.BIN: OK
build/us/F_MAD.BIN: OK
build/us/NO3.BIN: OK
build/us/F_NO3.BIN: OK
build/us/NP3.BIN: OK
build/us/F_NP3.BIN: OK
build/us/NZ0.BIN: OK
build/us/F_NZ0.BIN: OK
build/us/SEL.BIN: OK
build/us/ST0.BIN: OK
build/us/F_ST0.BIN: OK
build/us/WRP.BIN: OK
build/us/F_WRP.BIN: OK
build/us/RWRP.BIN: OK
build/us/F_RWRP.BIN: OK
build/us/TT_000.BIN: OK
mipsel-linux-gnu-objcopy -O binary build/us/stnz0.elf build/us/NZ0.BIN
mipsel-linux-gnu-ld -r -b binary -o build/us/assets/st/wrp/D_80181250.o build/us/assets/st/wrp/D_80181250.bin
mipsel-linux-gnu-ld -r -b binary -o build/us/assets/st/cen/g_Rooms.o build/us/assets/st/cen/g_Rooms.bin
mipsel-linux-gnu-ld -r -b binary -o build/us/assets/st/wrp/D_80181228.o build/us/assets/st/wrp/D_80181228.bin
mipsel-linux-gnu-ld -r -b binary -o build/us/assets/st/wrp/D_80181278.o build/us/assets/st/wrp/D_80181278.bin
⋮
```

Adding the `CHECK_FILES` dependency ensures the correct ordering and that all of those have completed before `check` runs. 

I'm trying to find a tractable way to correctly model target dependencies so thing something like `make -j dra` would work without the explicit need to clean or require separate make invocations. If those things worked, there would be less cases where a clean is required to begin with and parallel builds have reliable order.

Would you be opposed to an alternate target for `all` that adds the dependency (e.g. `all-check` or `build-check`)?

> Also I do not want to couple the Makefile task name with the config check name, which already proven to be quite difficult to move things in the past.

I'm not sure I follow this. Do you mean you don't want something like a `check-dra` target?

Xeeynamo: My brain must have been drunk when reviewing this. I thought $(CHECK_FILES) would extract target names out of the checksums and expand as `check: dra cen dre ...`, but it actually expands as `check: build/{VERSION}/DRA.BIN ...` . 

I was not aware of `--ignore-missing`.

I am pretty much inexpert on writing `Makefile`s . The sole truth of the current build system working the way it is, is down to some kind of miracle of me trying random things until it worked.

Thank you for the explanation, it makes much more sense where you're heading to with these changes you're proposing in the past few PRs

Xeeynamo (approved):
I will be happy to get this in. I think it makes things a bit easier!

PR #1278, opened 2024-06-07
Decompile DRE Entity3DBackgroundHouse

bismurphy:
This one was fun! Lots of use of the coordinate transformation functions, and lots of weirdly packed data.

Had to shift the splat a little bit. This is the last function in the file (location-wise, not the last to be decompiled) and the end of the C segment was slightly off.

### DIFF
--- a/src/st/dre/22848.c
+++ b/src/st/dre/22848.c
@@ -248,7 +248,7 @@ void Entity3DBackgroundHouse(Entity* self) {
         TransMatrix(&m, &trans);
         SetRotMatrix(&m);
         SetTransMatrix(&m);
-        vertices = D_80181544;
+        vertices = &D_80181544[0][0];
         scratchpad = SPAD(0);
         for (i = 0; i < 6; i++) {
             RotTransPers3(vertices[0], vertices[1], vertices[2], &scratchpad[0],
--- a/src/st/dre/22848.c
+++ b/src/st/dre/22848.c
@@ -221,13 +221,13 @@ void Entity3DBackgroundHouse(Entity* self) {
         self->ext.bghouse.unk80 = 0x80;
         self->ext.bghouse.unk82 = 0x80;
         self->ext.bghouse.unk84 = 0x80;
-        return;
+        break;
     case 1:
         // surprised this isn't just if abs(posX) > 0x200
         if (self->posX.i.hi > 0x200 || self->posX.i.hi < -0x200) {
             for (prim = self->ext.bghouse.prim; prim != NULL;
                  prim = prim->next) {
-                prim->drawMode = 8;
+                prim->drawMode = DRAW_HIDE;
             }
             return;
         }
@@ -235,7 +235,7 @@ void Entity3DBackgroundHouse(Entity* self) {
         SetGeomScreen(0x400);
         SetGeomOffset(128, 192);
         // this is a 7-by-X array, so params picks a set of 7 values
-        houseData = &D_801815CC;
+        houseData = D_801815CC;
         // Params is 0 or 1. 0 if house is straight, 1 if rotated 90 degree.
         houseData += 7 * self->params;
         rot.vx = 0;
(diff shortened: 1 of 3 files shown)


src/st/dre/22848.c
```diff
@@ -176,11 +173,113 @@ void Entity3DHouseSpawner(Entity* self) {
...
+
+        } else {
+            DestroyEntity(self);
+            return;
+        }
+        // These seem to be unused?
+        self->ext.bghouse.unk80 = 0x80;
+        self->ext.bghouse.unk82 = 0x80;
+        self->ext.bghouse.unk84 = 0x80;
+        return;
```
Xeeynamo, line 224:
```suggestion
        break;
```

src/st/dre/22848.c
```diff
@@ -176,11 +173,113 @@ void Entity3DHouseSpawner(Entity* self) {
...
+    case 1:
+        // surprised this isn't just if abs(posX) > 0x200
+        if (self->posX.i.hi > 0x200 || self->posX.i.hi < -0x200) {
+            for (prim = self->ext.bghouse.prim; prim != NULL;
+                 prim = prim->next) {
+                prim->drawMode = 8;
```
Xeeynamo, line 230:
```suggestion
                prim->drawMode = DRAW_HIDE;
```

src/st/dre/22848.c
```diff
@@ -176,11 +173,113 @@ void Entity3DHouseSpawner(Entity* self) {
...
+            }
+            return;
+        }
+        prim = self->ext.bghouse.prim; // unused
+        SetGeomScreen(0x400);
+        SetGeomOffset(128, 192);
+        // this is a 7-by-X array, so params picks a set of 7 values
+        houseData = &D_801815CC;
```
Xeeynamo, line 238:
```suggestion
        houseData = D_801815CC;
```

src/st/dre/22848.c
```diff
@@ -176,11 +173,113 @@ void Entity3DHouseSpawner(Entity* self) {
...
+        rot.vy = *houseData++;
+        rot.vz = 0;
+        RotMatrix(&rot, &m);
+        trans.vx = self->posX.i.hi - 0x80;
+        trans.vy = self->posY.i.hi - 0xC0;
+        trans.vz = *houseData++ + 0x400;
+        TransMatrix(&m, &trans);
+        SetRotMatrix(&m);
+        SetTransMatrix(&m);
+        vertices = &D_80181544[0][0];
```
Xeeynamo, line 251:
```suggestion
        vertices = D_80181544;
```
can we remove the [0][0]?

src/st/dre/22848.c
```diff
@@ -176,11 +173,113 @@ void Entity3DHouseSpawner(Entity* self) {
...
+            model += 4;
+        }
+        houseData += 3;
+        model = D_801815A4;
+        for (i = 0; i < 2; i++) {
+            prim = func_801A2C9C(prim, model, houseData);
+            model += 20;
+        }
+        while (prim != NULL) {
+            prim->drawMode = 8;
```
Xeeynamo, line 280:
```suggestion
            prim->drawMode = DRAW_HIDE;
```

src/st/dre/22848.c
```diff
@@ -176,11 +173,113 @@ void Entity3DHouseSpawner(Entity* self) {
...
+            prim = prim->next;
+        }
+        return;
```
Xeeynamo, line 283:
```suggestion
        break;
```

Xeeynamo (approved):
Left some minor readability comments, but the rest looks great!

src/st/dre/22848.c
bismurphy: That came from PSP stuff, the PSP compiler is more aggressive about types matching, and it doesn't like doing `myPointer = myArray;` because the pointer is supposed to be a pointer to an int, not a pointer to an array of ints. I'll clean these up.

sozud: It seems like we should keep the PSP match? We're going to combine PSP+PSX eventually

Xeeynamo: Agree. If only the match `vertices = &D_80181544[0][0];` is valid on PSP, then we should keep it. Sorry for making you going back and forth on this.

bismurphy: Sounds good.

It's also possible that there might be a compiler flag that tells it to not make errors on this kind of pointer type mismatch. In that case we could use the simpler expression for PSP, without needing &[0][0].

PR #1279, opened 2024-06-07
EntityWarpSmallRocks

sozud:


Xeeynamo (approved):
It looks good! I was about to suggest to copy `EntityWarpSmallRocks` into PSX WRP but I guess we're gonna merge the two files soon anyway

PR #1281, opened 2024-06-08
Share more code and data between stage overlays

Xeeynamo:
Given the fact that sharing compiled C objects is not exactly possible (code heavily copy&pasted maybe? this initiative is now abandoned 👉 https://github.com/Xeeynamo/sotn-decomp/commit/53a566f075a395e064f42db28737d5ef351f0ea1) I decided to keep pressing forward with shared headers. Thanks a lot to @hohle for making our life much easier by cross-referencing symbols.

The file split on WRP seems to be the closest file split we might have compared to the original source code (still speculating here). I think it would be a good idea to start splitting other overlays too with the same approach.

My idea is to have a file split like the following:

```
st/
   cen/
       e_particles.c
       e_misc.c
       st_common.c
   nz0/
       e_particles.c
       e_misc.c
       st_common.c
   wrp/
       e_particles.c
       e_misc.c
       st_common.c
   e_particles.h
   e_misc.h
   st_common.h
```

each of those C files will just be a one-line `#include "../the_shared_code.h"` as usual. Right now we create individual headers for single functions, sometimes for more than one function when we think grouping makes sense. But I think we can start merging some of those headers and consolidate the code. This can be done gradually. For example `src/st/e_particles.h` is still importing function headers under the hood. That is okay for now, but later on I wish to import those headers functions into their respective parent headers.

Another important aspect to consider that will validate a correct file split is to start importing the data inside these new C files. Right now we have floating data such as `src/st/wrp_psp/wrp_data_EA00.c` or monstrosities such as `src/st/wrp/6FD0.c`. An example of a (possibly) correct migrated data is what this PR does with WRP PSP and NZ0 PSX, with data pointing in `src/st/*/e_particles.c`. 

~⚠️ I am marking this PR as a Draft. I want to work a bit more on this to further validate this is the correct way going forward. I am making this public to be transparent about this initiative, to gather feedback and to not step on each other's toes.~ it's ready


sozud:
I'm not really sure we can combine the data like you want since it's not in the same order that it was declared in the files on PSX. See here https://github.com/HighwayFrogs/frogger-psx/tree/main/vlo#step-3-the-assembler

The above issue is especially clear with `main` for example where `main` variables and libsnd variables are interleaved

```
glabel _snd_openflag
/* 880B0 800978B0 */ .word 0x00000000
/* 880B4 800978B4 */ .word 0x00000000
.size _snd_openflag, . - _snd_openflag

glabel g_randomNext
/* 880B8 800978B8 */ .word 0x00000000
/* 880BC 800978BC */ .word 0x00000000
/* 880C0 800978C0 */ .word 0x00000000
/* 880C4 800978C4 */ .word 0x00000000
.size g_randomNext, . - g_randomNext

glabel _svm_cur
/* 880C8 800978C8 */ .byte 0x00
.size _svm_cur, . - _svm_cur
```

From the small amount of data I've imported on PSP WRP I can see it's in a different order, which may actually be the correct one since I don't think mwcc has the same issue?

Xeeynamo:
`g_randomNext` and others are in the BSS section though, where the _"randomness"_ happens. Data and rodata seem to be well ordered and I actually see everything aligning pretty well.

I see very minor differences in the PSP version. Like `src/st/wrp_psp/wrp_data_EE50.c` is almost identical if not for `D_80180E58` being probably in a different place. The other stuff close to it seem to align. Even `g_ESoulStealOrbAngles`, `g_ESoulStealOrbSprt`,  `g_ESoulStealOrbAnim` is another example where data is in the same order between the two builds. 

I think where data is ordered differently between PSX and PSP is where a potential file split is as the PSP seem to just have files compiled in a different order. The same is also applied between the various PSX overlays.

Can you provide an example of `.data` that is completely out of order between PSX and PSP? Maybe I can spot a new pattern.

sozud:
I missed that it was bss only that has the issue. An example of out-of-order would be

https://github.com/Xeeynamo/sotn-decomp/blob/4dff51303356884358a612f9944ac8ca14beac39/src/st/wrp_psp/wrp_data_EA00.c#L3

On PSP you have:

g_eRedDoorUV
UNK_Invincibility0

on PSX you have:

https://github.com/Xeeynamo/sotn-decomp/blob/93cadd80778d8459569f17efbc3c6b875a1d9fa2/src/st/wrp/6FD0.c#L652

UNK_Invincibility0
... (other stuff)
g_eRedDoorUV

sozud (approved):
This seems reasonable to me and aligns with my current "copy-paste a folder of C files to make a new overlay" theory. We don't have to solve all the data order questions at once

src/st/no3/collision.c
```diff
@@ -0,0 +1,3 @@
+#include <common.h>
```
hohle, line 1: Is this necessary (can this be included in collision.h instead)? It would be nice if the files were literally just a single include line.

The one place I can think of where this might not work is what I found with `AnimateEntity` in `cen` and `rwrp` where they don't expect a particular prototype to be included. I think those should be considered exceptions though and make those more complicated instead of the ones that follow a similar pattern.

src/st/np3/collision.c
```diff
@@ -0,0 +1,12 @@
+#include <common.h>
+
+extern u16 g_testCollEnemyLookup[];
+extern u8 g_testCollLuckCutoff[];
+extern u8 g_testColluCoords[];
+extern u8 g_testCollvCoords[];
+extern u16 g_testCollElementLookup[10];
+extern u8 g_testColliFrames[10];
+extern u16 g_testCollPrizeTable[];
+extern u16 g_testCollRandTable[];
```
hohle, line 10: Same idea here.

hohle (commented):
This is awesome and will make working on future levels much easier! Can the declarations in these other headers be easily moved into the shared header or is there some other reason why they need to be separate?

src/st/np3/collision.c
Xeeynamo: Thank you, good point. I forgot to clean that up while I was moving stuff. Great catch!

PR #1282, opened 2024-06-08
Decompile NZ0 EntityGaibon

bismurphy:
This was great to get working!

There was a Decompme WIP, but I decided to ignore that and do it from the beginning, and I think that worked out well. PSP was of course very helpful.

This ended up being pretty readable with all the steps that were already documented (I think that was Sonic's doing, so thanks!). Pretty happy to have this working!

Slogra and Gaibon each have a separate `FntPrint("charal %x\n", self->animCurFrame);` call in their debug step. Importantly, this string is in rodata in two places. If Slogra and Gaibon are in the same file, they share the same string address, and the rodata only gets the string in one place. Splitting the files restores the duplicated string in rodata, meaning that these files need to be split.

I have already decompiled EntityCloseBossRoom and found the same issue there, so my next PR will end up pulling out Slogra into his own file as well.

Lots of code here, review carefully!


src/st/nz0/gaibon.c
```diff
@@ -0,0 +1,708 @@
...
+    s32 speedLimit;
+    s8* hitboxPtr;
+
+    s32 yVar;
+    s32 xVar;
+
+    s32 stupid;
+
+    if ((self->step) && (!self->ext.GS_Props.nearDeath) &&
+        self->hitPoints < g_api.enemyDefs[0xFE].hitPoints / 2) {
```
Xeeynamo, line 84: > g_api.enemyDefs[0xFE]

I wonder if there is a better way to reference that `0xFE` to the actual enemy entry. I do not think an enum would suffice here. Let's keep it as it is but explore what we can do later on

Xeeynamo (approved):
WOW. This is great to see matched!! Great, great work you've done here :clap: congratulations

PR #1283, opened 2024-06-08
Decompile NZ0 EntityBossFightManager

bismurphy:
A few changes here.

First, Slogra needs to be split out to his own file, as mentioned in my previous PR.

I decompiled EntityCloseBossRoom, but I decided it needed a better name, since, while it does close the door to the boss room, it actually does a lot more and in general manages the fight (most importantly, it starts the fight, spawns Slogra and Gaibon, and when they die, spawns the life max up). Therefore I named it BossFightManager. Since this function, and the door blocks it spawns, are the only ones in this file, I decided to call it bossfight. I think we should try to be liberal with naming files, once we know all the functions contained in them.

Otherwise I think that should do it! Very cool to have the first boss fight in the game all figured out.


Xeeynamo (approved):
>  Detect whether player is in the room. If so, close the door.

thanks for putting some comments here!

PR #1285, opened 2024-06-08
Wrp data

sozud:
Going to do a few files


src/st/wrp_psp/wrp_data_F7B0.c
```diff
@@ -0,0 +1,40 @@
+#include "../wrp/wrp.h"
+
+u16 g_eRedDoorTiles[2][8] = {
```
Xeeynamo, line 3: This seems to be consistently close to the top in all overlays, together with that array passed to `InitializeEntity`. Sooo weird.

Xeeynamo (approved):

PR #1288, opened 2024-06-08
Decompile NZ0 EntityLifeUpSpawn

bismurphy:
Boy, this one was a beast!

It created yet another evil awful horrible variant of Primitive, where it's referencing several s32 values in the struct, so we had to add another to Primitive.h.

The splat was slightly off, it was missing the nop at the end of this function.

It appears that this function manages 385 primitives to make the life max up spawning animation, and puts them all through a bunch of GTE functions, so I'm wondering if that's why the animation is so laggy.

I'm confused on why this function doesn't show up in duplicates.txt, since I would have thought it would be universal across all overlays which contain a boss (which is most of them). Although... now that I think of it, Slogra and Gaibon are the only bosses in the overlays that are currently in the decomp, so that's probably why. CEN, DRE, ST0, NP3, etc don't have bosses so they don't need to have this entity. Hopefully this ends up duplicating in the other overlays once those get pulled in.

An awful function, but at least it matches!


sozud (approved):

Xeeynamo (approved):
Wow what a beast! This is super great to see complete. I bet a lot of curious people will look into this function to understand the slowdowns.

I left a few minor touch-ups. The biggest touch-up would probably be to use `LifeUpPrim` instead of `Primitive` to get rid of all the casts. So we could read stuff like `((LifeUpPrim*)prim->next)->x0` as simply `prim->next->x0`

bismurphy:
> I left a few minor touch-ups. The biggest touch-up would probably be to use `LifeUpPrim` instead of `Primitive` to get rid of all the casts. So we could read stuff like `((LifeUpPrim*)prim->next)->x0` as simply `prim->next->x0`

The problem is that the primitives aren't all LifeUpPrim. Look at the i-j-k loop. We set u,v, r, g, b values for `prim`, and then we set weird values (including unkC and unk10) for `prim->next`. But `prim->u0` and `prim->v0` are located at 0xC, and `prim->r1, g1, b1` are at 0x10.

It appears that the primitives alternate. There is a normal prim, and then a LifeUpPrim. And at the end of the i,j,k loop, we increment the prim twice. So it seems like even-numbered prims are normal, and odd-numbered are LifeUpPrim. We can not change them to all use one type of Primitive, it's definitely a mix.

bismurphy:
I guess the other option is if this is only used in one function, we could just `LOW()` it.

bismurphy:
Alright this is probably better

Xeeynamo:
> I guess the other option is if this is only used in one function, we could just `LOW()` it.

Up to you, feel free to change it to LOW or merge as it is with the new primitive type

bismurphy:
I ended up changing to LOW since there were only two members different from a normal Primitive. Should be good to merge.

PR #1291, opened 2024-06-09
func_psp_09244020

sozud:
This uses a bunch of functions from PS.ELF so not sure about any of the structs or stack variables


sozud:
I checked in these symbol changes from the formatter since they've been bothering me locally but the script seems to be wrong in removing them.

Xeeynamo (approved):
It's nice to have this exclusive PSP function. Great job!

PR #1295, opened 2024-06-10
EntityNumericDamage

sozud:


Xeeynamo (approved):
It looks good! Before merging, shall we align the function name as the rest of the overlays? I think this might have slipped. The most up-to-date function name is `EntityDamageDisplay`. 

Also did you check if this also matches `src/st/entity_damage_display.h`? If this match also matches the PSX build we could merge the two.

sozud:
These are unified now except for
```
#if defined(VERSION_PSP)
            xOffset = -(*nDigits) * 2;
#else
            xOffset = *nDigits * -2;
#endif
```
if anyone wants to try that

PR #1297, opened 2024-06-10
Two small ST0 functions

joshlory:
One duplicate and one near-duplicate.

`func_801B101C` ([decomp](https://decomp.me/scratch/K87Ql)) is listed as a 96% match of `func_801B76F0` but it seems to be using a slightly different Dialogue struct.


sozud (approved):
Looks good!

PR #1299, opened 2024-06-11
Decompile DRA func_80121F58

joshlory:
PSX: https://decomp.me/scratch/w2wDx
PSP: https://decomp.me/scratch/s95Fn


Xeeynamo:
Similarly to #1255, the symbols has to be cross-referenced between the US and the HD build. For example, the CI is currently complaining about the symbol `D_800AE230` being unreferenced, yes?

image

I looked at the data in US for `D_800AE230` and found that the equivalent on HD is located at `800AE2F4`. So adding `D_800AE230 = 0x800AE2F4` in `config/symbols.hd.dra.txt` will fix it. 

image

Another way to do it is to run `VERSION=us make -j extract` and `VERSION=hd make -j extract` and then compare both the disassembled functions with VS Code and it will highlight the symbols. For instance in the screenshot you can see `D_801381F4` in the US version, but on HD is `D_80137DC0`. To cross-reference you will add the symbol `D_801381F4 = 0x80137DC0`. That way GCC will translate `D_801381F4` to point at `0x80137DC0` at compilation time, despite the name suggesting something a tiny bit different.

Let me know if any of this does not make sense. I will be happy to explain this further!

joshlory:
Thanks! When do symbols need to be added to the `us` build, if ever? Are the offsets from `us` to `hd` always the same for a given overlay or can they vary?

sozud:
> When do symbols need to be added to the us build
When you rename a variable for function and there's usages of it in asm form still

> Are the offsets from us to hd always the same for a given overlay 
It depends on whether functions or data were altered in a way that changes their length

To fix the build you need to add a stub definition of D_801381F4 to stubs.c or import the data
```
/usr/bin/ld: libcore.a(7E4BC.c.o): warning: relocation against `D_801381F4' in read-only section `.text'
/usr/bin/ld: libcore.a(7E4BC.c.o): in function `func_80121F58':
/home/runner/work/sotn-decomp/sotn-decomp/src/dra/7E4BC.c:1017: undefined reference to `D_801381F4'
/usr/bin/ld: /home/runner/work/sotn-decomp/sotn-decomp/src/dra/7E4BC.c:1023: undefined reference to `D_801381F4'
```

sozud:
```
core.lib(7E4BC.obj) : error LNK2019: unresolved external symbol D_801381F4 referenced in function func_80121F58 [D:\a\sotn-decomp\sotn-decomp\pc\Sound.vcxproj]
```
You can't have 0 sized arrays on msvc

src/pc/stubs.c
```diff
@@ -209,6 +209,7 @@ unk_800B08CC D_800B08CC[6] = {0};
 s32 D_800B07C8 = 0x12345678;
 Unkstruct_80138094 D_80138094[100]; // unknown size
 s32 D_8013808C;
+Primitive D_801381F4[8]; // unknown size
```
joshlory, line 212: Assumed size, since it's indexed by:

https://github.com/Xeeynamo/sotn-decomp/blob/b2c1fb3151c742fbd1c32fa889b8172a115226bc/src/dra/d_DBD4.c#L143-L149

joshlory:
Thanks for all your help with this one! Now that I've broken the pipeline in (almost) every possible way I should be all set for the next non-trivial change 😅

PR #1300, opened 2024-06-11
Decompile Several Similar `EntityWeaponAttack`

hohle:
Decompiles several `EntityWeasponAttack` functions that were grouped by the deduper. Many of the variations are subtle - change to how the subtype is calculated, draw mode, setting sprite bank or not, setting vars or not.


config/symbols.stmad.txt
```diff
@@ -67,12 +67,10 @@ EntityIntenseExplosion = 0x801964E4;
 EntitySoulStealOrb = 0x801982BC;
 EntityEnemyBlood = 0x80198650;
 EntityRoomForeground = 0x80198B00;
-GetSideToPlayer = 0x801920AC;
 UNK_Invincibility0 = 0x801806B4;
 g_ItemIconSlots = 0x80199E54;
 g_InitializeEntityData0 = 0x80180508;
 g_InitializeData0 = 0x801804FC;
-g_UnkPrimHelperRot = 0x8018D8C0;
```
hohle, line 75: I'm not sure why the linter grabbed these as they're unrelated to this change.

Xeeynamo: Somehow the linter in one of the merged PRs did not pick up there were symbols to be cleaned-up. It is an issue in the CI linter step.

Xeeynamo (commented):
@bismurphy are you happy to review this? I feel weapons are more your expertise

src/weapon/w_003.c
```diff
@@ -3,7 +3,59 @@
 #include "weapon_private.h"
 #include "shared.h"
 
-INCLUDE_ASM("weapon/nonmatchings/w_003", EntityWeaponAttack);
+extern WeaponAnimation D_19000_8017AC28[];
+
+void EntityWeaponAttack(Entity* self) {
+    WeaponAnimation* anim;
+    s8 subType;
```
bismurphy, line 10: How did you pick the name `subType` for this? I'm not sure what the `params` actually controls. For some entities, it is a subtype, but for others it might control other things like whether an entity is in an "angry" state, or HP values, or color, or whatever else. In this case, it looks like ti selects the animation? I might call it animIndex or something. Would be good if we could determine where the `params` are coming from to really be able to tell how the animations are selected.

bismurphy (commented):
Looking good! Left a few comments that apply across all these weapons. Hopefully at some point we can find a way to de-duplicate these since they're incredibly similar. Would also be nice if we could cross-reference this set of files to determine how these weapons might have something in common. Nice work!

src/weapon/w_003.c
hohle: `subType` was just taken from other similar functions. I haven't gone through and triggered any of these to see what they were doing.

hohle: I've renamed this `animIndex` to make it more generic.

src/weapon/w_002.c
```diff
@@ -3,7 +3,57 @@
 #include "weapon_private.h"
 #include "shared.h"
 
-INCLUDE_ASM("weapon/nonmatchings/w_002", EntityWeaponAttack);
+extern WeaponAnimation D_12000_8017ABC4[];
```
hohle, line 6: I considered naming these symbols (something like `WeaponAttackAnimation`). They are private to each weapon and only used by this function afaict, so while a common name doesn't help with reuse, it's nicer to read than an address label. I'm open to comments and opinions.

Xeeynamo: I think all those data symbols used to be `static`

hohle: Likely

sozud (approved):

PR #1302, opened 2024-06-11
Decompile CEN EntityHolyGlassesCutscene

bismurphy:
This is very similar to the Maria Cutscene that I recently finished in NZ0, so I'm doing this one too while the knowledge is still fresh.


Xeeynamo (approved):
I have not much to say here, it's more of the same as the function in NZ0. It's :heavy_check_mark: for me

PR #1304, opened 2024-06-11
Decompile DRE EntitySuccubusCutscene

bismurphy:
Continuing to run the cycle on all these cutscenes.

This one is interesting because it switches on self->params to do different things - this might signify when the Succubus shows up as "Lisa", versus after the reveal when it shows up as Succubus. Unsure, that can be a project for the future.


sozud (approved):
LGTM going to merge to enable the other PR

PR #1305, opened 2024-06-11
Link WRP

Xeeynamo:
I think this is pretty much ready to be merged it. I do not see much of a point of having the branch floating around.

I am using a `#ifndef HARD_LINK` to avoid compiling duplicate functions with the same symbol name.

For data with the same symbol name but different content I am doing a `#define OVL_EXPORT(x) WRP_##x`. So for example `OVL_EXPORT(g_EntityGfx)` will be expanded as `WRP_g_EntityGfx`. It is an unfortunate hack. On the other hand all the data migration allowed me to mark most of the data and functions as `static`. If we use the same approach on other overlays we could easily statically link them as well.

SEL is enabled by default. It will take you straight to the real WRP, not the dummy one.

Most of the changes are loading assets from their respective JSON files. Linking the stage was pretty straight forward.


sozud (approved):
Looks like stubs for D_80181420 etc need to be added to placate msvc but other than that looks good

PR #1306, opened 2024-06-11
Batch draws

Xeeynamo:
Instead of sending one single draw call, this batches as much calls as possible. I am supporting the vertex buffer with an index buffer to minimse how much data is sent to the driver. 

The `StoreImage` got some minor boost too.


sozud:
> Instead of sending one single draw call

You mean instead of sending multiple draw calls you're batching I think?

sozud (approved):

Xeeynamo:
> > Instead of sending one single draw call
> 
> You mean instead of sending multiple draw calls you're batching I think?

that is correct!

PR #556, opened 2023-09-07
Sonicdcer EntityCSMoveAlucard

sozud:
I'm putting these back up so the work isn't lost. @Xeeynamo are you fine with merging this since we haven't come up with a solution for `D_80097400`?

PR #1307, opened 2024-06-12
Decompile DRE EntityCSMoveAlucard

bismurphy:
I was asked to help with #556 , but that is an old PR and I'm not sure I want to be doing a PR into a PR, so here's this one.

I did not touch the code at all, just shifted around the files and such in order to get rodata to match.


sozud (approved):
Thanks for the help, I guess I didn't notice the string and that was throwing me off

PR #1309, opened 2024-06-12
Decompile NO3 EntityDeathCutscene

bismurphy:
As far as the duplicate-finder can tell, this is the last of the cutscenes, but who knows - we may find more. Luckily, the pattern is pretty recognizable.

Named it due to what we see at the very end in `case 7`, with the TimeAttackController getting triggered.


sozud (approved):
Great work on these

PR #1310, opened 2024-06-12
msvc fixes

sozud:
I looked into the msvc issues, after these are fixed it crashes for me in the memset here, not sure what that is about:

https://github.com/Xeeynamo/sotn-decomp/blob/77edb0c33e71c70459e9ff732b673ac9b2bd7c75/src/pc/pl_arc.c#L22

### DIFF
--- a/src/dra/953A0.c
+++ b/src/dra/953A0.c
@@ -1345,7 +1345,10 @@ void func_80136010(void) {
                     D_8013AEA0[i] = 0;
                 }
             }
+#ifndef VERSION_PC
+            // this can be uninitialized
             fakeptr++;
+#endif
             var_t4 += 1;
             var_t3 += 4;
             var_t2 += 4;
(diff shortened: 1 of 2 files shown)


Xeeynamo:
I have a fix pending for these I am about to push. But instead of correcting from `[]` to `[1]` I am just marking the global variable as `extern`. If you could kindly rebase over it that would be good :pray:

sozud:
This PR is pretty small so maybe just incorporate the fixes and then I close this one?

Xeeynamo:
> This PR is pretty small so maybe just incorporate the fixes and then I close this one?

The other one does not have the `fakeptr++;` fix though. I am not sure how critical that is

Xeeynamo:
I thought this was targeting `master`, not `pc-wrp` :facepalm:

sozud:
> The other one does not have the `fakeptr++;` fix though. I am not sure how critical that is

MSVC stops execution and enters the debugger when that line is hit, so I think that change should be merged. It looks like it didn't get into pc-wrp?

PR #1312, opened 2024-06-13
Decompile ST0 EntityDraculaCutscene

bismurphy:
The duplicate finder missed this one, but I found it indirectly by looking for the preceding functions, which are all duplicates. This should actually be the last of the cutscenes (at least, until we get new overlays).

All these cutscenes have little tweaks to their logic, but nothing very interesting, so not much more to share here. But cool to have them all done!


sozud (approved):
Congrats on completing all these

PR #1313, opened 2024-06-13
Decompile DRA func_8012C600

joshlory:
PSX: https://decomp.me/scratch/74j2Y
PSP: https://decomp.me/scratch/ons5L

Let me know if you'd like to inline `CLAMP_MIN` and `CLAMP_MAX`, or move them to `macros.h` or `common.h`. I see a reference to `CLAMP` in STYLE.md but couldn't find it in the repo.


sozud:
The HD version seems to be the same so I'm guessing there's an issue with your addresses https://decomp.me/scratch/RJqvI

PR #1314, opened 2024-06-13
Decompile NZ0 EntityMaria

bismurphy:
Relatively simple function.

I'm trying to get better at things like using good variable names, organizing files, etc, and I think it pays off. This function is more readable than it would have been in the initial scratch. Excited to really get the game into a state where all the code is directly readable.


Xeeynamo (approved):
It is very interesting to see how an entity is controlled by a cutscene!

PR #1315, opened 2024-06-13
Decompile NZ0 func_801C4198

bismurphy:
This is a weird function, we don't really know what it does. But matching is always the first step. Just glad it works.

The behavior on the `dataPtr` makes me sad. Pointers are treated in such evil ways in this game.


Xeeynamo (approved):
Another confirmation that structs were not really used in this game. Meh.

PR #1317, opened 2024-06-13
Decompile NZ0 EntityMagicallySealedDoor

bismurphy:
I was able to start with the red door code and rely on that a fair bit. I modified the code to recreate the parts that are needed for the blue door.

Their entity extensions matched up (with the blue door needing a couple extra members) so I changed ET_RedDoor to just be ET_Door, and both doors use it. I'll be curious to see if the weird golden door to the Olrox fight is similar to these ones.

NZ0 is really getting empty now!


Xeeynamo (approved):
At some point we should start to mark those helper functions such as `EntityIsNearPlayer` as `static`. I doubt they were exported in any way.

Congrats for taking another one down. You're crushing it!

PR #1318, opened 2024-06-13
Decompile NZ0 801B0AA4

bismurphy:
3 NZ0 functions remaining? I think you mean 2 NZ0 functions remaining :)

I don't really understand this one, it's a bit of a mess. Lots of tilemap stuff, so maybe this does something with the weird shifting backgrounds in the Alchemy lab? Who knows.


Xeeynamo (approved):
LET'S GO!!! You've owning this game hahaha

PR #1319, opened 2024-06-14
Defake NP3 func_801CD78C

bismurphy:
This function is part of a small network of functions that pass around Entities and Point32s, so it ends up being a little weird. I guess the first two elements of an Entity are equivalent to a Point32, but still weird.

Within this function though, there is nothing to indicate that the args are Entity, and a PSP function seems to allocate the right stack size to call this with Point32, so I think we should do that, and have the callers cast their entities to Point32 if needed.

### DIFF
--- a/src/st/np3/4B018.c
+++ b/src/st/np3/4B018.c
@@ -62,16 +62,14 @@ void func_801CD734() {
         func_801CD658();
 }
 
-void func_801CD78C(Entity* src, s32 speed, s16 angle, Entity* dst) {
-    if (g_CurrentEntity->facingLeft != 0) {
+void func_801CD78C(Point32* src, s32 speed, s16 angle, Point32* dst) {
+    if (g_CurrentEntity->facingLeft) {
         angle = -angle;
     }
+    *dst = *src;
 
-    //! FAKE:
-    (*(Point32*)dst) = (*(Point32*)src);
-
-    (*(Point32*)dst).x -= speed * rsin(angle) * 16;
-    (*(Point32*)dst).y += speed * rcos(angle) * 16;
+    dst->x -= speed * rsin(angle) * 16;
+    dst->y += speed * rcos(angle) * 16;
 }
 
 void func_801CD83C(Entity* self) {


Xeeynamo (approved):
If you still want to use `Entity*` as a parameter, you can rewrite the function as the following:

```c
void func_801CD78C(Entity* src, s32 speed, s16 angle, Entity* dst) {
    if (g_CurrentEntity->facingLeft != 0) {
        angle = -angle;
    }

    __builtin_memcpy(dst, src, sizeof(f32) * 2);
    dst->posX.val += speed * rsin(angle) * 16;
    dst->posY.val += speed * rcos(angle) * 16;
}
```

if this is a good idea or not it is entirely up to debate. As @sozud mentioned in the past, `__builtin_memcpy` is none other than a struct copy like you did with `*dst = *src;`. I couldn't match this in any other way.

Either approaches are valid.

bismurphy:
> If you still want to use `Entity*` as a parameter

I don't think we do want to. Here's the scratch I used to decide this: https://decomp.me/scratch/7EZNl

sp10 is on the stack, and if we make it an Entity, it ends up being way too big. So sp10 is definitely a Point32. We could of course call this function with `(Entity*)sp10`, but that doesn't seem right to me. I think it makes more sense to call the function with Point32, and sometimes that Point32 might be a `(Point32*)myEntity`, but I think it makes more sense to cast an entity to a point (thus grabbing the position from the entity) than to cast a point to an entity.

Xeeynamo:
> > If you still want to use `Entity*` as a parameter
> 
> I don't think we do want to. Here's the scratch I used to decide this: https://decomp.me/scratch/7EZNl
> 
> sp10 is on the stack, and if we make it an Entity, it ends up being way too big. So sp10 is definitely a Point32. We could of course call this function with `(Entity*)sp10`, but that doesn't seem right to me. I think it makes more sense to call the function with Point32, and sometimes that Point32 might be a `(Point32*)myEntity`, but I think it makes more sense to cast an entity to a point (thus grabbing the position from the entity) than to cast a point to an entity.

I based my assumption on `func_801CDA14`, where we do `ent2->ext.GH_Props.unk9C` indicating it is a real entity, but when we pass `ent2` itself. I did not know about `func_801CDAC8`, but now I more understand why you say that `Point32` might have been used at the beginning on `Entity`. Thanks for the info!

PR #1320, opened 2024-06-14
Enhance `sotn_str` & Encode Menu Strings

hohle:
`sotn_str` can now handle escaped quotes, right parenthesis, and more in in `_S` strings. This allows all of the strings in `dra/menu.c` to be encoded as static C-strings.

This also converts the spell button sequences to UTF-8.


src/config_us.h
```diff
@@ -444,34 +444,34 @@ const char* g_MenuStr[] = {
...
-    {"", "\xFF", "", 0, 20, 8, 2, 129, 0, 16448, 30, 0},
-    {"", "\xFF", "", 0, 20, 8, 2, 129, 0, 16448, 45, 0},
+    {"Dark Metamorphosis", _S("←↖↑↗→"), "Heal HP by shedding blood", 10, 0, 0, 0, 1, 0, 0, 0, 0},
+    {"Summon Spirit", _S("←→↑↓"), "Summons a fierce spirit", 5, 20, 8, 2, 1, 0, 0, 15, 0},
+    {"Hellfire", _S("↑↓↘→"), "Transport and fireball attack", 15, 20, 8, 2, 1, 0, 32768, 30, 0},
+    {"Tetra Spirit", _S("↑ため↗→↘↓"), "Summons four spirits", 20, 20, 8, 2, 1, 0, 0, 15, 0},
+    {"Wolf Charge", _S("↓↘→"), "Powerful attack by Wolf", 10, 40, 10, 2, 129, 64, 64, 30, 0},
+    {"Soul Steal", _S("←→↘↓↙←→"), "Steals HP from nearby enemies", 50, 10, 8, 130, 2, 0, 0, 15, 0},
+    {"Wing Smash", _S("↑↖←↙↓↘→"), "Powerful attack by Bat", 8, 10, 10, 2, 1, 16, 0, 50, 0},
+    {"Sword Brothers", _S("↓↘→↗↑ため↓"), "Summons Sword Brothers", 30, 20, 8, 130, 129, 0, 64, 20, 0},
```
hohle, line 454: The `ため` sequence appears to be used to color the proceeding character red (indicating that it should be held for two seconds).

tools/sotn_str/jp.py
```diff
@@ -119,6 +119,11 @@ def dakuten(chr, prev):
 for index, value in enumerate(table):
     utf8_to_index[value] = index
 
+# ASCII characters not mapped above, but useful for encoding strings in C files.
+ascii_supplement = "*;<=>@"
+for c in ascii_supplement:
+    utf8_to_index[c] = ord(c) - 0x20
+
```
hohle, line 126: Entries in `g_MenuStr` have several adjacent ASCII characters, but in the mapping table above some of those have non-ascii characters in their positions. This allows these ASCII characters to be used as well when it makes sense.

tools/sotn_str/jp.py
```diff
@@ -259,7 +264,7 @@ def dakuten_to_bytes(input_chr):
 
 
 def utf8_to_byte_literals(input_str):
-    clean_str = input_str.replace("_SJ(", "").replace(")", "")
+    clean_str = input_str.replace("_SJ(", "(")
```
hohle, line 267: I dont' see any use (or definition) of the `_SJ` macro, but the second replacement on this line made it impossible to have a right parenthesis in an `_S` encoded string. The first replacement now replaces `_SJ(` with a left parens, so something like `_SJ("foo")` would become `("foo")`.

tools/sotn_str/sotn_str.py
```diff
@@ -30,17 +30,15 @@ def process_string(match: re.Match[str]):
...
     return f'"{escaped}"'
 
 
 def do_sub(line):
-    pattern = r'_S\("([^"]*)"\)'
-    # english_str_processed = re.sub(pattern, process_string, line)
-    # pattern_jp = r'_SJ\("([^"]+)"\)'
+    pattern = r'_S\("((?:\\["]|[^"])*)"\)'
     jp_str_processed = re.sub(pattern, process_string_jp, line)
-    jp_str_processed = jp_str_processed.replace("_S(\"\")", "\"\\xFF\"")
```
hohle, line 43: Long term this might be better written as a state machine to better handle escape sequences. This now allows `\"` (escaped double quotes) in strings (on line 40). These need to be dealt with after the match (on line 33).

tools/sotn_str/jp.py
sozud: Anything related to SJ can be removed, it's left over from when I thought there was a US encoding and a Japanese one

src/dra/menu.c
```diff
@@ -66,42 +66,42 @@ const char* D_800A2D18[] = {
 
 #if defined(VERSION_US)
 const char* D_800A2D68[] = {
-    "\x00\x01\x01\xFF",
```
sozud, line 69: It seems like something is off with the ones in this file, there's not a single entry that's coherent. This function doesn't seem to be doing the same encoding? https://github.com/Xeeynamo/sotn-decomp/blob/a4a556239ecab52a7c309fb70fdafe4dde30c4d1/src/dra/menu.c#L930

hohle: That's certainly possible. Let me see if I can get more context on this.

src/config_us.h
Xeeynamo: I think this is fine! There are only two places, as far I know, this is used.

Xeeynamo (commented):
This looks an improvement as it is. I know we could do some clean-up on `_SJ` as it's no longer used.

Could you add some unit tests in `tools/sotn_str/test.py` to cover the changes you made? That would be good in case we decide to refactor stuff or to document parts of the code

tools/sotn_str/jp.py
sozud: I'm not sure I'm understanding why this is needed, the mapping is like this
image
These characters are not in the table?

src/dra/menu.c
sozud: You can see in that function it's the 12x16 font
```
    const int ChWidth = 12;
    const int ChHeight = 16;
```
However these strings aren't shift-jis. I'm guessing they resize some bios font characters and make a table at runtime?

tools/sotn_str/jp.py
hohle: It's not relevant after removing the `menu.c` change.

sozud:
If you want to revisit this the table for the D_800A2D68 strings starts at D_800A2D10
"\x0C\x0D\x0E\xFF"  should decode to 攻撃力 for example

hohle:
Yeah, I can take a look at that table as well and add a separate encoding macro if necessary.

tools/sotn_str/test.py
```diff
@@ -22,47 +22,42 @@ def test_dakuten_to_bytes(self):
...
-
     def check_sei():
         assert(utf8_to_index['聖'] == 222)
 
-    def test_glasses(self):
-        input = "_SJ(聖なるめがね)"
-        out = utf8_to_byte_literals_wrapped(input)
-        assert out == "_SJ(\\xEE\\xC5\\xD9\\xD2\\xB6\\xFF\\x9E\\xC8\\xFF)"
-
-    def test_moon(self):
```
sozud, line 43: Why are we removing test cases? 月 is is 0xFF, which is the same as the end byte, so it's useful to have a test. I don't remember the details of the others but they're probably there for a reason

tools/sotn_str/jp.py
```diff
@@ -124,6 +124,21 @@ def get_chr(chr):
...
+    "装備技システム短剣必殺使攻撃力防",
+    "御魔導器拳こ一覧棒両手食物爆弾盾",
+    "投射薬ん右左武兜鎧マントその他い",
+])
+
+alt_utf8_to_index = {}
+for index, value in enumerate(alt_table):
+    alt_utf8_to_index[value] = index
+
+for index, value in enumerate("ＡＴＤＥＦ"):
```
sozud, line 138: This is a separate table at D_800A2D64. I think the right approach is having two different macros because this table is only in the US version, and D_800A2D10 etc. are only in the JP version. It's not the same encoding

sozud:
BTW we just squash everything when merging so you don't need to rebase every time if it's extra work for you

tools/sotn_str/jp.py
hohle: OK, there's are `_S2` and `_S2_HD` macros for the separate lookup tables.

sozud:
I think this is ready if the test cases here are restored 
```
        input = "_SJ(あかつきの剣)"
        input = "_SJ(聖なるめがね)"
        input = "_SJ(バルザイのえん月刀)"
```
https://github.com/Xeeynamo/sotn-decomp/blob/d45701638fe050e3e88d3b7a2d75f541bd8fcfb1/tools/sotn_str/test.py#L30

hohle:
I had gotten rid of them along with `utf8_to_byte_literals_wrapped`, but I see their value. I've modified them to use `utf8_to_byte_literals_escaped`.

I also added an additional test (which I should have added in the first place), with escaped quotes (i.e. `"\"(\")"`)

sozud (approved):

PR #1322, opened 2024-06-14
Decompile WRP func_801907EC (!)

bismurphy:
One of the last two functions!

Since this is duplicated in every overlay (but never used in any of them), it will make sense to de-duplicate this. I will use my automatic de-duplication script to put it into the appropriate .h file. That can either be in this PR (after getting the seal of approval), or can be a separate PR (after merging this initial WRP version). I'm good with any option.

Very cool though, glad I figured this one out finally! Hopefully the last one will come along soon...

### DIFF
--- a/include/types.h
+++ b/include/types.h
@@ -53,4 +53,9 @@ typedef struct {
     /* 0x2 */ s16 y;
 } Point16; // size = 0x4
 
+typedef struct {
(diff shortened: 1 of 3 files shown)


sozud (approved):
Nice work

Xeeynamo (approved):

PR #1327, opened 2024-06-15
Decompile NP3 EntityHammer

bismurphy:
Kind of a complicated process to get this one working, beyond the normal function decompilation.

Files are split for now, because there are weird things with rodata. I have already done the two remaining Gurkha functions, so my next PRs will pull those in and will end up bringing the files back together.

Messy things here, tried to make it looks nice but as always, feedback is appreciated.


Xeeynamo (approved):
Looks good!

PR #1330, opened 2024-06-16
Decompile NP3 EntityOwl

bismurphy:
Another big one! This required mangling the files a little bit; once the other functions are decompiled the files should come back together nicely. The function itself is in 4B018.c, and the function is currently alone in that file.


sozud (approved):
Nice work

PR #1331, opened 2024-06-16
Decompile NP3 func_801CBF18

bismurphy:
Helper function for the Owl Knight, does a bunch of stuff with prims. Might be related to his death animation.


Xeeynamo (approved):
Looks good! Let's start marking those helper functions as `static` too

PR #1333, opened 2024-06-17
Decompile NP3 EntityOwlKnight

bismurphy:
This is a huge one! Lots of file shuffling needed to happen, but now that things are in place I feel relatively confident that we have our file splits in the right places, which is great. It's really cool when we can make file splits that are motivated by game structure, and not by rodata limitations.

Anyway, here it is, let me know which flags and stuff I forgot about :)


sozud (approved):
Good work

PR #1334, opened 2024-06-17
Decompile func_80123B40

hohle:
Finishes off work originally started by @joshlory.


hohle:
Thanks to @sozud for pointing out `memcpy`s could be simple assignments.

Xeeynamo:
team work <3

src/dra/7E4BC.c
```diff
@@ -1202,7 +1202,97 @@ void func_80123A60(Entity* entity) {
...
+
+        if (!(self->flags & 0x800000)) {
+            DestroyEntity(self);
+            return;
+        }
+
+        self->flags = 0x0C830000;
+        self->velocityY = FIX(-3.0);
+        SetSpeedX(-0x1AAAA);
+        self->ext.stub[0x28] = 0x80;
```
sozud, line 1233: @Xeeynamo Do we need to create an ET_XXX here?

Xeeynamo: Yep 100%. The use of `self->ext.stub` should be disallowed and it's there just for padding

hohle: Sounds good. I'll see if any of the others are relevant otherwise create a new one.

Xeeynamo: We usually create new ones based on the context of the function. For example this one is calling `EntityPlayerBlinkWhite` with the same parameter. So maybe we can create a new struct called `ET_BlinkWhite blinkWhite` and use `self.ext.blinkWhite.unkA4`, or something like that.

sozud:
Conflicts need to be resolved

PR #1335, opened 2024-06-17
Conditionally compile libsnd

sozud:
This makes changes to allow libsnd/libspu to compile when WANT_LIBSND_LLE is set. If turned on it will immediately crash due to trying to write to the SPU. A build is added to make sure it's compiling.


Xeeynamo (approved):
It looks good, very nice piece of work! Let's see in the future if we would be forced to stub libsnd or we can fix to not need `WANT_LIBSND_LLE`

PR #1336, opened 2024-06-17
Decompile DRA menu func_800FB23C

bismurphy:
This one got stalled a long time ago, so I came back to it. It wasn't easy, and PSP was compiled with -O4, but I got it, so that's nice :)


src/dra/menu.c
```diff
@@ -3020,8 +3020,214 @@ bool func_800FB1EC(s32 arg0) {
...
+                block_36:
+                    func_800FB160(
+                        g_EquipmentCursor, nav->cursorMain, D_801375CC);
+                    ret = 2;
+                    g_IsSelectingEquipment = 0;
+                    PlaySfx(0x633);
+                } else {
+                    PlaySfx(0x686);
+                }
+            } while (0);
```
hohle, line 3146: Does the `do/while(0)` generate different code than without it?

bismurphy: Yes, unfortunately this is a common hack that ends up making code match, which otherwise does not match. In this case, see the scratch: https://decomp.me/scratch/BghLP

If you remove the do-while, you get the `s5` and `s6` registers swapped throughout the function.

If you can find an alternative fix, that would be very appreciated.

src/dra/menu.c
```diff
@@ -3020,8 +3020,214 @@ bool func_800FB1EC(s32 arg0) {
     return false;
 }
 
-s32 func_800FB23C(MenuNavigation* nav, u8* order, u8* count, u32* selected);
-INCLUDE_ASM("dra/nonmatchings/menu", func_800FB23C);
+typedef enum ItemTypes {
```
sozud, line 3023: Is there a difference vs.
```
typedef enum EquipKind {
    EQUIP_HAND,
    EQUIP_HEAD,
    EQUIP_ARMOR,
    EQUIP_CAPE,
    EQUIP_ACCESSORY,
    NUM_EQUIP_KINDS,
} EquipKind;
```
?

bismurphy: Nope, this was just the enum that was used in Xeeynamo's old scratch, didn't know there was a different one! I'll update.

sozud (approved):
The do/while can probably be eliminated by studying the psp version but this is already good, merging

PR #1341, opened 2024-06-18
Decompile NP3 EntityAlucardWaterEffect

bismurphy:
This function is kind of crazy and huge.

I also made its helper static.

The name for the function comes from the fact that the duplicate finder matched this up with a function in NO3, which had been given that name, so I named this one to make them match up.


src/st/np3/player_water_effect.c
```diff
@@ -0,0 +1,343 @@
...
+    u16 sp30;
+    u16 var_s6;
+    u32 var_s7;
+    u16 sp4A;
+    s32 gPlayerUnk0C;
+    Tilemap* tilemap = &g_Tilemap;
+    Entity* player = &PLAYER;
+
+    sp18 = player->posX.i.hi + tilemap->scrollX.i.hi;
+    gPlayerUnk0C = g_Player.unk0C;
```
Xeeynamo, line 53: you can rename `gPlayerUnk0C` as `status` and use the various `PLAYER_STATUS_` to document the bitfields

src/st/np3/player_water_effect.c
```diff
@@ -0,0 +1,343 @@
...
+                D_80097448[1] = 0;
+            }
+            if (var_s3 > 0x30) {
+                D_80097450 = var_s3 - 0x30;
+            } else {
+                D_80097450 = 0;
+            }
+        }
+    } else {
+        InitializeEntity(&D_80180A60);
```
Xeeynamo, line 335:
```suggestion
        InitializeEntity(D_80180A60);
```
`D_80180A60` is an array

Xeeynamo (approved):
Congrats, this function is huge! I left some suggestions to document that `unk0c`

src/st/np3/player_water_effect.c
bismurphy: Awesome idea!

src/st/np3/player_water_effect.c
bismurphy: Good catch, fixed

PR #1344, opened 2024-06-19
Enable spu init

sozud:
This reworks enough writes to libsnd/libspu to make it through the call to SsInitHot. Instead of raw pointers we have read/write_16 functions. The writes are saved to an array so we can test against output from mednafen. See `src/pc/psxsdk/expected/_spu_init.txt`. The gold file is generated by running a real psx psy-q program that calls the same function. A modified mednafen prints the writes. There are reads to spustat and other regs that require particular values to be returned, so I've also imported and hooked up the SPU part of mednafen.


Xeeynamo (approved):
Exciting times. Thanks also for making a E2E test. I actually have a few ideas on how we could leverage that in the future

PR #1345, opened 2024-06-19
Decompile NP3 func_801B32A8

bismurphy:
Just an entity function, not clear what it is or how it is created.


Xeeynamo (approved):
Looks gud

PR #1347, opened 2024-06-19
Decompile NP3 EntityGaibon

bismurphy:
This finishes NP3.

Gaibon's logic is extremely similar here as it is in NZ0. There are a few differences, especially in the initialization code (which runs before the primary switch), but as a whole, it's a copy-paste.

Slogra and Gaibon are in separate files. The only way we know this is because they both have debugging `charal` strings which exist separately in rodata (if they were the same file, it would reuse the same rodata location for both strings).


sozud:
How hard is it to share these with NZ0?

sozud (approved):

src/st/np3/gaibon.c
```diff
@@ -0,0 +1,756 @@
...
+    GAIBON_FLY_SHOOT_FIREBALLS,
+    GAIBON_LANDING_AFTER_SHOOTING,
+    GAIBON_SHOOT_FROM_GROUND,
+    GAIBON_FLY_SHOOT_BIG_FIREBALL = 8,
+    GAIBON_PICKUP_SLOGRA,
+    GAIBON_NEAR_DEATH = 15,
+    SLOGRA_GAIBON_RETREAT,
+    GAIBON_DYING,
+    GAIBON_DEBUG = 255,
+} GaibonSteps;
```
hohle, line 18: This is incredible!

PR #1348, opened 2024-06-19
Rewrite DRA and ST0 gpubuffer function to remove NON_MATCHING

bismurphy:
Found this function which had a NON_MATCHING.

I'm getting pretty good at finding PSP functions now, so finding the PSP told me that this was an if-block, not a ternary operator.

PSP also indicated that some of the other lines were using multiple-assignment, rather than a sequence of individual assignments, so I made those changes too.

As usual, I don't really know what these do, but it's nice to have pure matches now.

Now the only functions left with NON_MATCHING are RenderEntities and RenderPrimitives.


Xeeynamo (approved):
Great finding!

PR #1352, opened 2024-06-24
Duplicate DRA EntitySubwpnAxeCrash

joshlory:
Near-duplicate of `EntitySubwpnCrashAgunea` in RIC.


config/symbols.hd.dra.txt
```diff
@@ -190,9 +191,7 @@ EntityPlayerDissolves = 0x80120A7C;
 EntityLevelUpAnimation = 0x8012162C;
 func_80121F58 = 0x80121C04;
 EntityMist = 0x80121FC8;
-func_80123B40 = 0x801237CC;
```
joshlory, line 193: Cleanup, symbol no longer needed after https://github.com/Xeeynamo/sotn-decomp/pull/1334.

src/dra/843B0.c
```diff
@@ -265,7 +265,222 @@ void func_80124A8C(Entity* entity) {
 INCLUDE_ASM("dra/nonmatchings/843B0", EntitySubwpnThrownDagger);
 
 // axe thrown when using subweapon
-INCLUDE_ASM("dra/nonmatchings/843B0", EntitySubwpnThrownAxe);
+// near-duplicate of EntitySubwpnCrashAgunea
+void EntitySubwpnAxeCrash(Entity* self) {
```
sozud, line 269: I just checked and this is also called when Alucard uses the axe subweapon. Naming it Crash seems to imply it's only used there?

PR #1353, opened 2024-06-24
Fix Directory Creation Dependencies in Makefile

### LINKED ISSUE #739: Make 4.4 does not work when building in parallel

hohle:
This updates the Makefile so that targets which depend on phony targets
(like `*_dirs`) have direct dependencies on those targets rather than
relying on transient dependencies from dependent targets. Other targets
explicitly create parent directories for their files.

This resolves #739, making it possible to using make-4.4 to build in parallel.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,9 +107,6 @@ SOTNDISK_SOURCES := $(shell find tools/sotn-disk -name '*.go')
 
 CHECK_FILES := $(shell cut -d' ' -f3 config/check.$(VERSION).sha)
 
-print_target = \
-    @printf "\033[0;32m$@\033[0m\n";
-
 .PHONY: build
 
 all: build check
@@ -117,7 +114,6 @@ build: build_$(VERSION)
 build_us: main dra weapon ric cen dre mad no3 np3 nz0 sel st0 wrp rwrp tt_000
 build_hd: dra tt_000
 clean:
-	$(call print_target)
 	git clean -fdx assets/
 	git clean -fdx asm/$(VERSION)/
 	git clean -fdx build/$(VERSION)/
@@ -125,7 +121,6 @@ clean:
 	git clean -fdx function_calls/
 	git clean -fdx sotn_calltree.txt
 format: bin/clang-format
-	$(call print_target)
 	bin/clang-format -i $$(find $(SRC_DIR)/ -type f -name "*.c" | grep -v 'src/pc/3rd')
 	bin/clang-format -i $$(find $(SRC_DIR)/ -type f -name "*.h" | grep -v 'src/pc/3rd')
 	bin/clang-format -i $$(find $(INCLUDE_DIR)/ -type f -name "*.h")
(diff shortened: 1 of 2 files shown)


sozud:
Deferring to @Xeeynamo on this one

Xeeynamo (approved):
This is really good. You did a much better job than what I had in mind to fix this. I tried make 4.4.1 on Arch and I can build it just fine.

PR #1354, opened 2024-06-25
Decompile DRA EntityLevelUpAnimation

joshlory:
PSX: https://decomp.me/scratch/tedz6
PSP: https://decomp.me/scratch/TWe0Y

Not an exact match on PSP yet. Happy to take another pass if you want that to also match!


sozud:
I think D_800AE180 is a struct, see here: https://decomp.me/scratch/DqdmE
There's some other fixes too

joshlory:
Thanks @sozud! Do you have a guess at which entity ext to use or should I create a new one for now?

sozud:
> Do you have a guess at which entity ext to use or should I create a new one for now?

I think just make one for this entity, we generally don't reuse them unless it's a duplicate or near duplicate

Xeeynamo:
structs and arrays are interchangeable when it comes to matching functions. I agree with sozud that an ad-hoc struct will make this a bit more readable. Just remember to use the same struct in `src/dra/d_DBD4.c` too.

I think we could even inline  `src/dra/d_DBD4.c` in `src/dra/7E4BC.c` and mark all the data as `static`, so you do not need to write the struct in a shared header. You would need to ensure the DRA HD builds correctly though.

PR #1355, opened 2024-06-25
Libsnd malloc and reverb funcs

sozud:
This makes changes to get SpuSetReverb, SpuMallocWithStartAddr and SpuClearReverbWorkArea running. We don't have matches for SpuMallocWithStartAddr and _SpuMallocSeparateTo3 so the current WIPs are used. A unit test is added to check that these are creating the correct _spu_memList. Spu write tests are added for the other functions. To avoid pulling in the DMA controller _spu_t just does the writes itself.


src/pc/psxsdk/libspu.c
```diff
@@ -30,9 +30,276 @@ void SpuSetKey(long on_off, unsigned long voice_bit) {
...
+                return var_v0;
+            }
+            return -1;
+        }
+    } else {
+        return -1;
+    }
+
+    return var_v0;
+    return 0;
```
Xeeynamo, line 302: Is the double-return intentional here?

Xeeynamo (approved):
I cannot hit those two functions within the game, but I think the unit tests are enough for now

sozud:
> I cannot hit those two functions within the game,
Are you building with `-DWANT_LIBSND_LLE=1`? These functions are being called as part of SoundInit https://github.com/Xeeynamo/sotn-decomp/blob/1006c9f04480859681380b9069135b07f634e6ba/src/dra/91EBC.c#L212

src/pc/psxsdk/libspu.c
sozud: This is just copy-pasted right off my decomp.me scratch, removed

Xeeynamo:
I was not building with that flag, thanks!

I noticed it gets stuck at `SsVabTransBodyPartly` due to `_svm_vab_used[vabid]` being `0` and not `-2`, which makes the game init stuck into `LoadVabData`. But it's a separate problem that has nothing to do with this PR.

Good one, can't wait to see more of this!

PR #1343, opened 2024-06-19
New asset manager

Xeeynamo:
This aims to deprecate all the Splat tools in `tools/splat_ext` in favour of a more centralised asset manager. This brings the following advantages:

* Much faster extraction
* Faster build
* Automatically define `static` symbols or unique names whenever `static` is not possible
* Allow to embed assets into the output binary
* Drastically simplify `Makefile` by removing all the asset build rules
* Avoid situations where it is not possible to extract and build assets that is not 4-byte aligned

This is achieved by having the splat YAML targeting a normal C file as data and have an external tool to take care of the following:

1. Extract asset files straight from the overlay binary file into human readable file in `assets/st/STAGE_NAME`
2. Build assets as header files that go into `src/st/STAGE_NAME` to just include them from any C file

This requires each stage header to have the following new format: please see `src/st/nz0/header.c`

Built assets in `src/st` are ignored by Git.

As for now, for simplicity sake, the steps `make extract_assets` and `make build_assets` are just executed within `make extract` exclusively for the US version.

I plan to auto-generate files such as `src/st/nz0/tile_data.c`.

For a first iteration I am aiming to handle the following:

* [X] Extract rooms: `assets/st/*/rooms.json`
* [X] Extract room layers: `assets/st/*/entity_layouts.json`
* [X] Extract tilemap data: `assets/st/*/tilemap_*.bin`
* [X] Extract tilemap definitions: `assets/st/*/tiledef_*.json`
* [X] Extract sprites: `assets/st/*/sprites.json`
* [x] Extract entity layouts
* [X] Build rooms: `src/st/*/rooms.h`
* [X] Build room layers: `src/st/*/layers.h`
* [X] Build tilemap data: `src/st/*/tilemap_*.h`
* [X] Build tilemap definitions: `src/st/*/tiledef_*.h`
* [x] Build sprites (aka `g_SpriteBanks`)
* [x] Build entity layouts
* [x] Allow the tool to suggest how to adjust the Splat config for each overlay

I want the tool to cover the following stages:
* [x] CEN
* [x] DRE
* ~MAD~ I do not think this can be done, it is way too different from the other overlays
* [x] NO3
* [x] NP3
* [X] NZ0
* [x] ST0
* [X] WRP
* [x] RWRP
* ~WRP (PSP)~ Maybe in a follow-up PR

For a later iteration I plan on extracting and build:

* Entity GFX thingie
* The CLUT thingie in the header
* Uncompressed GFX data
* Cutscene data
* Blueprints
[...]

PR #1356, opened 2024-06-25
CEN, DRE, MAD, NO3, NP3, RWRP, & ST0 Stage Headers

hohle:
Decompiles stage header data for all in progress stages.

Of note is that NO3 and NP3 (and likely other future stages) use an `AbbreviatedHeader` because their `SpriteBanks` start immediately after `UpdateStageEntities`. There are several areas which use `sizeof(Overlay)` to copy this data over, in those cases sprite bank data is copied into the tail fields of the overlay but are never used.


sozud:
There's overlap here with this work and the work going on with the asset manager https://github.com/Xeeynamo/sotn-decomp/pull/1343 so I think this is going to need @Xeeynamo 's input

Xeeynamo (approved):
How did you determine `StageEndCutScene`?

I think I have an idea on how to not use `AbbreviatedOverlay` and make it work anyway, but I will wait until the asset manager is complete to confirm my theory.

The PR is good. I am fine that it conflicts with my PR that is still work in progress. Your work is still a step forward and I will take inspiration from it.

I am also thinking to rename the various `header.c` as `nz0_header.c` or `header_nz0.c` and similar, in case in the future we want to have all the C stage files in the same folder. It's just an heads-up, not a change for this PR.

PR #1359, opened 2024-06-28
Speed Up `make format`

hohle:
After building for all supported architectures `make format` was taking around 30s. This commit includes various changes which get that down to around 3½s on the same machine.

This changes the `format` target in a few ways:

  * filters out files generated by CMake
  * runs `clang-format` in parallel
  * breaks `format` into 3 separate targets which can be run in parallel

When building the PC version cmake creates a file to test compiler features. This file is picked up by the pattern which matches files for `clang-format`. This file took around 20s, filtering it out brought total format time to around 10s.

Instead of passing a list created by a `find` command directly as the args to `clang-format` the list is now passed to `xargs` which runs several processes in parallel, each looking at 10 files (determined by experimentation). This has the added advantage of ensuring as the source list grows `ARG_MAX` won't be a concern. This brought total format time to around 6s.

`format` was broken up into three targets `-src`, `-tools`, and `-symbols`. These targets can be run in parallel, bringing the total time down to 3½s.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,6 +122,7 @@ clean:
 	git clean -fdx sotn_calltree.txt
 
 format: format-src format-tools format-symbols
+
 format-src: bin/clang-format
 	@# find explainer:
 	@#    find $(SRC_DIR) $(INCLUDE_DIR)                      : look in src and include
@@ -159,6 +160,11 @@ format-symbols:
 	./tools/symbols.py remove-orphans config/splat.us.strwrp.yaml
 	./tools/symbols.py remove-orphans config/splat.us.tt_000.yaml
 	./tools/symbols.py remove-orphans config/splat.us.stmad.yaml
+
+# fast-format
+ff: MAKEFLAGS += --jobs
+ff:
+	$(MAKE) format
 
 patch:
 	$(DIRT_PATCHER) config/dirt.$(VERSION).json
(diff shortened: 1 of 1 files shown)


Xeeynamo (approved):
Great catch. This PR of yours is slashing the formatting time from 9.0s down to just 2.2s on my machine!

I noticed we can only benefit the speed-up if we do `make -j format`. What do you think about modifying the `format` target like the following, if it makes sense?

```
format:
    make -j format-src format-tools format-symbols
```

hohle:
Having a target override jobs seems unexpected, especially with no straightforward way to override it.

What do you think about a separate target (I called it `ff` - fast format) as an alternative for those who know, but won't affect those who don't. Alternatively this can be done through `MAKEFLAGS` env vars (if you want all targets parallel) or local aliases.

Xeeynamo:
> What do you think about a separate target (I called it `ff` - fast format) as an alternative for those who know, but won't affect those who don't. Alternatively this can be done through `MAKEFLAGS` env vars (if you want all targets parallel) or local aliases.

I think it is fine doing `make ff` too. I just want to make it obvious to the existing contributors that there's now a faster way to format, or your enhancement will risk to go unseen. Let's merge it as it is and see how it goes. Please make noise in our Discord so other people will know!

PR #1361, opened 2024-06-28
Implement SpuSetCommonAttr

sozud:


Xeeynamo (approved):
For the future we need to remember to do the following:
* Add `.cpp` to `make format`
* Remember to replace these custom implementations whenever we have matching functions on `main`

This is good as it is tho

PR #1362, opened 2024-06-29
SsVabOpenHeadWithMode Attempt

sozud:
This PR is trying to get SsVabOpenHeadWithMode running, but I'm getting the wrong result vs. my test app. I'm putting this up to see if others have ideas. https://github.com/sozud/psy-q-tests/blob/c7484b6800d3dfa5337c1c1e2d6e39299b3f3ef8/test.c#L154

I check a bunch of preconditions on the test app and the implementation so I think they are starting from the same state. The test app is able to play the library song so I think it's OK.

The problem is that SpuMalloc returns the wrong value. See
```
_svm_vab_start[vabId] = spuAllocMem;
```

_svm_vab_start[0] is supposed to be 0x00001010 but instead we get 0x11010.

https://github.com/sozud/psy-q-tests/blob/c7484b6800d3dfa5337c1c1e2d6e39299b3f3ef8/test.c#L201

```
69648 != 4112 in check_ss_vab_open_head_with_mode /home/d/sotn-decomp-2/src/pc/psxsdk/emu.cpp:306
```

The scratch for SpuMalloc is https://decomp.me/scratch/UjIPd

The version here is based off this one since I think that scratch isn't usable yet. https://github.com/AliveTeam/sound_rev/blob/4ff48d466068cc40b2a0b27d75f642bcdb553aa6/psx_seq_player/lib_spu.cpp#L2462

The scratch for func_800286E0 is https://decomp.me/scratch/msP8t


sozud:
I wrote a better test that checks SpuMalloc by itself and it seems like those results match. So this may actually be a bug with SsVabOpenHeadWithMode

sozud:
I think I got an equivalent https://decomp.me/scratch/MFD6z
This PR ought to be merged first https://github.com/Xeeynamo/sotn-decomp/pull/1365 then I'll update this one

sozud:
This is rebased with the matching version of SsVabOpenHeadWithMode now

Xeeynamo (approved):

PR #1366, opened 2024-07-01
Decompile heaven sword func_ptr_8017000C

bismurphy:
Someone in the Long Library discord was asking about Heaven Sword (especially its special attack) so I'll go ahead and decompile that next.

This initial function appears to reveal that Heaven Sword doesn't follow the main Weapon struct. I'm going to keep using it until the weapon is done, so we can get a better picture of what the mismatching fields are. All the struct members are the same size, they're just different names.


Xeeynamo (approved):
Looks good. And welcome back!

PR #1367, opened 2024-07-01
Decompile heaven sword func_ptr_80170008

bismurphy:
This one is kind of ugly in places.

It uses offset 0x84 as an s32, while Weapon uses it to hold an entity, so this forced the creation of a new ext for the heaven sword.

There is only one more function left in this file (EntityWeaponAttack), so once that one is done, I will move the whole file over to use the Heaven Sword extension.

### DIFF
--- a/src/weapon/w_013.c
+++ b/src/weapon/w_013.c
@@ -3,7 +3,9 @@
 #include "weapon_private.h"
 #include "shared.h"
 
+extern SpriteParts D_5F000_8017A040[];
 extern u8 D_5F000_8017A5B0[6][8];
+extern s16 D_5F000_8017A5E0[];
 
 s32 func_5F000_8017A9CC(Primitive* prim, s32 x, s32 y) {
     s32 size;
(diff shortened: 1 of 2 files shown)


src/weapon/w_013.c
```diff
@@ -80,7 +82,182 @@ s32 func_ptr_80170004(Entity* self) {
...
+                prim->x0 = prim->x1 = self->posX.i.hi;
+                prim->y0 = prim->y1 = self->posY.i.hi;
+                prim->priority = self->zPriority;
+                prim->drawMode = DRAW_TPAGE2 | DRAW_TPAGE | DRAW_TRANSP;
+            } else {
+                DestroyEntity(self);
+                return;
+            }
+            self->step++;
+            return;
```
Xeeynamo, line 214:
```suggestion
```
`return` not needed?

Xeeynamo (approved):
Left a single comment.

It makes sense to have an ad-hoc ext struct for this

PR #1368, opened 2024-07-01
Fix loadfilesim, connect cd audio

sozud:
The loading code comes from https://github.com/Xeeynamo/sotn-decomp/blob/8e78845e8b110cb8a465086ba51d8e1e0225c491/src/dra/47BB8.c#L472 which I think is correct

This will eventually hit an assert in _SpuSetVoiceAttr.


Xeeynamo (approved):
This is great!

PR #1370, opened 2024-07-02
Decompile heaven sword EntityWeaponAttack

bismurphy:
Finishes the heaven sword.

More weirdness with the ext - it appears that unk84 is used differently in different functions within this value.

Overall, this file could use some improvements, but to start out, having it decompiled is better than not. Hopefully someone can figure out how this actually all works.

### DIFF
--- a/src/weapon/w_013.c
+++ b/src/weapon/w_013.c
@@ -121,9 +121,9 @@ void EntityWeaponAttack(Entity* self) {
         }
         return;
     case 2:
-        self->hitboxWidth = 0xC;
-        self->hitboxHeight = 0xC;
-        self->drawFlags |= 4;
+        self->hitboxWidth = 12;
+        self->hitboxHeight = 12;
+        self->drawFlags |= DRAW_COLORS;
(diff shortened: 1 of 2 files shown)


src/weapon/w_013.c
```diff
@@ -55,7 +61,148 @@ s32 func_5F000_8017A9CC(Primitive* prim, s32 x, s32 y) {
...
+            self->velocityY = FIX(-1.5);
+            self->ext.timer.t = 0x18;
+            SetSpeedX(FIX(-3));
+            self->step++;
+        }
+        return;
+    case 2:
+        self->hitboxWidth = 0xC;
+        self->hitboxHeight = 0xC;
+        self->drawFlags |= 4;
```
Xeeynamo, line 126:
```suggestion
        self->drawFlags |= DRAW_COLORS;
```

Xeeynamo (approved):
A single line comment. But overall it looks good.

I think the reason why the ext must be different is because those two entry points work on different entities. One is the actual weapon and the other one is an effect that is spawned with it

PR #1371, opened 2024-07-02
Decompile DRA RenderPrimitives

bismurphy:
Got this working with the aid of PSP!

Scratches here:

PS1: https://decomp.me/scratch/FDWxW
PSP: https://decomp.me/scratch/SObUR

These are identical aside from the versions defined at the top of the context, and everything else is done with `ifdef`.

Several of the `ifdef` differences are in structs that are defined in `include/psxsdk/libgpu.h`, and since they're in the PSX SDK, I don't want to add those PSP differences into that file. I imagine at some point we'll have a PSP libgpu.h or something, but until then, this function will not actually match on PSP. But since PSP is not the focus of this decomp, I think that's okay.

This is 99% similar to Xeeynamo's scratch, the main difference is the treatment of the lines that look like `setSemiTrans(&primbuf->g2, var_s0->drawMode & DRAW_TRANSP);`.

### DIFF
--- a/src/dra/4A538.c
+++ b/src/dra/4A538.c
@@ -1806,7 +1806,7 @@ void RenderPrimitives(void) {
     #define RECT_LOC 0x128
     #endif
     RECT* rect = (RECT*)SP(RECT_LOC);
-    PrimitivesRenderer* r = (PrimitivesRenderer*)SP(0x000); //a0
+    PrimitivesRenderer* r = (PrimitivesRenderer*)SP(0x000);
     PrimBuf* primbuf = (PrimBuf*)SP(0x024);
     
     s32 var_s1;
--- a/src/dra/4A538.c
+++ b/src/dra/4A538.c
@@ -1768,22 +1768,8 @@ void FreePrimitives(s32 primitiveIndex) {
     }
 }
 
-#ifndef NON_MATCHING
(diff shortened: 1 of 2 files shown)


src/dra/4A538.c
```diff
@@ -1847,69 +1849,85 @@ void RenderPrimitives(void) {
...
     } else {
-        var_a1 = 0;
-        i = 0;
-        prim = g_PrimBuf + i;
-        for (; i < MAX_PRIM_COUNT; prim++, i++) {
-            primType = prim->type;
-            if (!primType) {
+        var_a2 = 0;
+        var_s1 = 0;
+        for (var_s0 = &g_PrimBuf[0]; var_s1 < MAX_PRIM_COUNT; var_s0++,
```
Xeeynamo, line 1854: Shall we rename some vars back to the previous names? I think `var_s0` as `prim` or `var_s1` as `i` and so on are a bit clearer.

Xeeynamo (approved):
I've been wanting to get this done for so long. I am glad you persisted on it and found a way to get a match. Congrats!

src/dra/4A538.c
bismurphy: Good call. Done!

PR #1374, opened 2024-07-04
Fixes to load wrp/alk

sozud:
I don't understand the file loading code that well but this gets the ALK vab running. Maybe SIM_PTR should just be D_80280000 so we have that buffer to use?


Xeeynamo (approved):
This is fine for now. I plan to largely refactor the `sim_pc` at some point too and document things properly. Let's get this in

PR #1375, opened 2024-07-04
Decompile Karma Coin EntityWeaponAttack

bismurphy:
Decided to target this one since it's the largest remaining function in `weapon` and now it's done. Kind of surprising that such a simple weapon has such a huge function, but I guess it has a lot of graphics and therefore a lot of primitive manipulation.

### DIFF
--- a/src/weapon/w_020.c
+++ b/src/weapon/w_020.c
@@ -3,12 +3,422 @@
 #include "weapon_private.h"
 #include "shared.h"
 
+extern SpriteParts D_90000_8017A040[];
+extern AnimationFrame D_90000_8017A850[];
+extern AnimationFrame D_90000_8017A864[];
+extern AnimationFrame D_90000_8017A8C8[];
+extern AnimationFrame D_90000_8017A8D8[];
+extern AnimationFrame D_90000_8017A8E8[];
 extern AnimationFrame D_90000_8017A954[];
 extern AnimationFrame D_90000_8017A9D8[];
+extern FrameProperty D_90000_8017A9E0;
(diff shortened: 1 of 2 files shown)


src/weapon/w_020.c
```diff
@@ -3,12 +3,422 @@
...
+    s16 sp6a;
+    s16 sp68;
+    s16 sp66;
+    s16 sp64;
+    s16 sp62;
+    u16 sp60;
+    s16 var_s8;
+    s16 var_s7;
+
+    sp60 = D_90000_8017AA10[g_GameTimer % 3] & 0x7FFF;
```
sonicdcer, line 45: i feel like modulus makes more sense in these 0x7FFF i've been seeing around, idk what you think.
```suggestion
    sp60 = D_90000_8017AA10[g_GameTimer % 3] % 0x8000;

Comment (bismurphy) [reply src/weapon/w_020.c:45]: Sure, why not, I'll make the change.

Comment (sonicdcer) [reply src/weapon/w_020.c]: Although it seems it doesn't work in cases like the rand() below, that's a pitty.

**Review by Xeeynamo — APPROVED** [review]

File: src/weapon/w_020.c (hunk 1 above)
Comment (sozud) [reply src/weapon/w_020.c]: What's the rationale for using modulus? This gets used like:
```
prim->clut = sp60;
```
The top bit of the clut is supposed to be 0 so it seems like it's getting masked off

https://psx-spx.consoledev.net/graphicsprocessingunitgpu/#clut-attribute-color-lookup-table-aka-palette

PR #1376, opened 2024-07-04
Refactor/clean-up of some of Alucard's jumping logic

bismurphy:
There was a question in the decomp discord about the bounce when Alucard dive-kicks and hits an enemy. I tracked this down, and in the process, found some moderately old functions that could use some improvements to their code style. I also cross-referenced these with PSP to make the code make a bit more sense.

The PSP version makes it clear that D_800ACF7C is not a struct and is just an array of s16, so I made that change. The struct version matched on PS1, but now that PSP reveals that it was fake, we turn it back to just an array.

Let me know if I missed anything! Nice to have real names for some of these functions.

### DIFF
--- a/src/dra/6D59C.c
+++ b/src/dra/6D59C.c
@@ -406,7 +406,13 @@ void DecelerateY(s32 amount) {
     }
 }
 
-s32 func_8010E27C(void) {
+// Checks hte player's left/right inputs and compares to the facing direction.
+// If the player is pressing the opposite of facing, we change the facing value
+// to turn the player around, and return -1.
+// If the player is pressing the same direction they are facing, return 1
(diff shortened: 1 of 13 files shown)


src/dra/6D59C.c
```diff
@@ -406,7 +406,13 @@ void DecelerateY(s32 amount) {
     }
 }
 
-s32 func_8010E27C(void) {
+// Checks hte player's left/right inputs and compares to the facing direction.
```
Xeeynamo, line 409:
```suggestion
// Checks the player's left/right inputs and compares to the facing direction.
```

Xeeynamo (approved):
An improvement, thanks for looking into it!

PR #1380, opened 2024-07-06
Cleanup DRA func_8012C600

bismurphy:
Another adventure in examining a function alongside PSP.

The CLAMP_MAX and CLAMP_MIN seem like nice ideas, but I wasn't getting them to match on PSP. And since they are very little-used in the repo, keeping them around in here just didn't seem worth it.

As far as the `D_8013AE##` variables, I noticed they were consecutive in addresses (they're all s32, and are all different by 4 bytes). Rather than being a weird mix of individual values and arrays, I thought it made more sense to treat them all as an array, and indeed it still matches this way, so I think this is a more reasonable way to handle this data.

I think those are all the changes that matter.


src/dra/843B0.c
```diff
@@ -3126,15 +3124,15 @@ void func_8012C600(void) {
             x = g_Entities[i + 21].posX.i.hi + 4;
         }
     }
-    D_8013AEBC = x - t0;
```
sozud, line 3129: Are these consecutive on psp? I think this is in the bss section so on psx they're in a different order than they were originally declared

bismurphy: D_8013AEBC[0] is D_psp_09236714
D_8013AEBC[1] is D_psp_09236718
D_8013AEBC[2] is D_psp_0923671C
D_8013AEBC[3] is D_psp_09236720

So yes, looks like they are consecutive and are in the same order on both systems.

bismurphy:
I was doing a different thing and found a function that was needlessly taking an `f32` input with a fake usage, so I fixed that. It's super tiny and inconsequential so I tossed it into this open PR, didn't seem like a major enough thing to make a whole dedicated PR for it.

PR #1384, opened 2024-07-07
Progress towards linking/loading weapons

sozud:
This links w_000 and allows the texture to be loaded to vram, as well as the functions executed.

g_Cluts was renamed since the stages also have g_Cluts. A macro and preprocessor defines are used to allow unique weapon function names. This doesn't load the animsets since it seems like that stuff is getting redone with the asset manager.


src/pc/weapon_pc.c
```diff
@@ -0,0 +1,863 @@
...
+    // seek to target position
+    fseek(file, start, SEEK_SET);
+
+    // Read the file into the buffer
+    size_t read_size = fread(dest, 1, size, file);
+    if (read_size != size) {
+        printf("Failed to read file");
+        fclose(file);
+        return 1;
+    }
```
hohle, line 850: Failure here could be due to `EOF` or other error. Instead of `printf` do you think `perror` and/or an `feof` check to specify the failure would be appropriate?

src/pc/weapon_pc.c
```diff
@@ -0,0 +1,863 @@
...
+    return 0;
+}
```
hohle, line 852: In the case that the read is successful, this function does not call `fclose` on the file and leaks the pointer.

hohle (changes requested):

src/pc/weapon_pc.c
sozud: OK I added checks for ferror and feof

src/pc/weapon_pc.c
sozud: I call fclose at the end now

hohle (approved):

Xeeynamo (approved):
Just one question, but everything else looks great. The `WEAPON_ID=w_000` idea is a smart solution and I am pretty sure it will work without problems!

I plan on replacing `readSubsetToBuf` with a proper I/O library as the current stuff we have is a bit undocumented and fragmented.

I will think what to do with the animset. It would be easier with the asset manager than parsing a JSON.

include/weapon.h
```diff
@@ -24,6 +24,10 @@ typedef struct {
 } Weapon /* 0x40 */;
 
 #define N_WEAPON_PAL 12
+#ifdef VERSION_PC
```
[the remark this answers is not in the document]
sozud: Seems to work, changed

PR #1385, opened 2024-07-07
Decompile DRA EntityMist

bismurphy:
This is a big messy function, but it's decompiled!

Did what I could to rename variables in a useful way, but I could only get so far. Hopefully one day we can work out exactly what all the different angles, XY variables, etc are. But this is one of the biggest remaining functions and it's great to have matching.


src/dra/7E4BC.c
```diff
@@ -1235,7 +1235,435 @@ Primitive* func_80121F58(bool arg0, s32 arg1, Primitive* arg2, s16 facingLeft) {
...
+        D_801381D4 = rand() & 0x7FFF;
+        D_801381D8 = rand() & 0x7FFF;
+        D_801381DC = rand() & 0x7FFF;
+        D_801381E0 = rand() & 0x7FFF;
+#else
+        D_801381D4 = rand();
+        D_801381D8 = rand();
+        D_801381DC = rand();
+        D_801381E0 = rand();
+#endif
```
joshlory, line 1321: Nit: might be cleaner to extract a `rand_mask` const as in https://decomp.me/scratch/G3ECA

src/dra/7E4BC.c
```diff
@@ -1235,7 +1235,435 @@ Primitive* func_80121F58(bool arg0, s32 arg1, Primitive* arg2, s16 facingLeft) {
...
+        self->params = FLAG_UNK_4000 | FLAG_UNK_1000 | FLAG_UNK_800 |
+                       FLAG_UNK_400 | FLAG_DEAD;
+        EntityPlayerBlinkWhite(self);
+        if (self->primIndex == -1) {
+            goto block_147;
+        }
+        prim = &g_PrimBuf[self->primIndex];
+        mistPrim = D_801381F4;
+        for (i = 0; i < 8; i++, mistPrim++) {
+            *mistPrim = *prim;
```
joshlory, line 1299: Nice find!

Xeeynamo: woah that's smooth!

src/dra/7E4BC.c
Xeeynamo: I think it would be better not having a constant here? At most a `const int RandMax = 0x7FFF` but it is so self-contained and context-explicit that I would keep it as it is. IMHO.

src/dra/7E4BC.c
```diff
@@ -1235,7 +1235,435 @@ Primitive* func_80121F58(bool arg0, s32 arg1, Primitive* arg2, s16 facingLeft) {
...
+                    prim->b1 = 0;
+                prim->r2 = prim->g2 = prim->b2 = prim->r3 = prim->g3 =
+                    prim->b3 = (D_80138394 * primColor4 / 256) & 0xFF;
+                break;
+            }
+            prim = prim->next;
+        }
+    }
+
+block_147:
```
Xeeynamo, line 1664: Tricky business... Only the first two `goto` can be replaced with:
```c
return;
```
I am not sure if you want to remove those two `goto` or be consistent and keep all of them falling to the last function statement

Xeeynamo (approved):
Congrats for the match!

src/dra/7E4BC.c
bismurphy: I think I will keep it this way at least for now. PSP seems to want to use the goto, so I think unfortunately the goto might be real.

PR #1389, opened 2024-07-07
Cleanup on func_800F24F4

bismurphy:
Lots of small changes here, but ultimately this ends up rewriting the whole function.

I opened this one in PSP (it's `-O4`!) and found several differences which I corrected. I also rewrote some of the if-statements to be a bit more readable.

Kind of a weird function the way it's testing for hard-coded X and Y values. Not sure what this function actually does.

We also end up only calling `func_801042C4` once, which is a nice improvement for readability.

`func_801042C4` is not decompiled, so I'll move on to that next.


Xeeynamo (approved):
It looks much better, thanks for investigating it!

PR #1392, opened 2024-07-07
Decompile DRA EntityPlayerBlinkWhite

bismurphy:
This is the largest function remaining in the decomp, and now it's dealt with.

It's a giant beast and I can't claim to be anywhere near understanding a single thing that it does. Researching this and documenting it will be a second giant PR. Maybe I'll do that someday, but for now I'll leave it to anyone else who wants to give it a try. None of the variables have useful names. We could have "give the variables in this function useful names relating to how they are used" be a "good first PR" issue, since we set up several of those a while ago.

Data types on some of the externs it uses have been changed to match how the function uses them.

This also uses g_PlOvlSpritesheet, which we have found seems to need to be accessed as a numerical pointer, so I added a #define for that. It seems to be working, so big thank you to everyone who was involved in the process of setting that up. I believe this is the first function to actually make use of that.

Besides that, I'll leave it to others to make their comments. There is really too much in this function for me to talk about it, but hopefully this is a reasonable start and we can make whatever tweaks are needed to get this merged.


src/dra/7A4D0.c
```diff
@@ -721,7 +721,641 @@ bool func_8011BD48(Entity* entity) {
...
+                }
+#endif
+            } else {
+                sp5c = D_800CF324[D_801396EC & 0x7FFF];
+#ifdef VER_PSP
+                if (sp5c) {
+#endif
+                    sp72 = *sp5c++;
+                    sp72 &= 0x7FFF;
+                    sp58 = ((u8**)SPRITESHEET_PTR)[sp72];
```
sozud, line 816: Is this still necessary after the recent maspsx fix?

bismurphy: My understanding was that the maspsx fix was in order to allow this to work. Previously even this would not work. Simply doing g_PlOvlSpritesheet[sp72] maintains the old issue. with $at inversion.

bismurphy: I haven't been able to get it to work without using this setup, so I think we go with this, and leave it open for cleaning up later, if we want to. I can add a comment if you like.

Xeeynamo (approved):
This is great! I am happy to have this merged now and have the `SPRITESHEET_PTR` issue being handled later on

PR #1393, opened 2024-07-08
Decompile DRA RenderEntities

bismurphy:
Now with the maspsx changes, this can also match by using SPRITESHEET_PTR.

Tried to document all the variables, use flags when possible, etc, so hopefully this is mostly complete. Should be exciting to have RenderEntities and RenderPrimitives both working now!


src/dra/4A538.c
```diff
@@ -719,84 +719,81 @@ void ResetEntityArray(void) {
...
+    u8* spriteData;
+    PlayerDraw* plDraw;
+    u16 palette;
+    u16 animFrameIdx;
+    u8 uvLeft;
+    u8 uvBottom;
+
+    POLY_GT4* poly;
+    EntitiesRenderer* r;
+    poly = (POLY_GT4*)SPAD(0);
```
sozud, line 764: This part is still needed to prevent crashes
```
#ifdef VERSION_PC
    POLY_GT4 _poly;
    POLY_GT4* poly = &_poly;
    EntitiesRenderer _r;
    EntitiesRenderer* r = &_r;
#else
```

bismurphy: Okay, I think I fixed it properly then. I also moved the declaration of the `sp` array down, so that we only need one #ifdef for this. Let me know if it's still wrong :)

sozud: This is untested but it's probably more like:

#ifdef VERSION_PC
    POLY_GT4 _poly;
    POLY_GT4* poly = &_poly;
    EntitiesRenderer _r;
    EntitiesRenderer* r = &_r;
#else
    POLY_GT4* poly;
    EntitiesRenderer* r;
    poly = (POLY_GT4*)SPAD(0);
#ifdef VERSION_PSP
    r = (EntitiesRenderer*)SPAD(14);
#else
    r = (EntitiesRenderer*)SPAD(13);
#endif
#endif

The issue is there's no scratchpad so you have to create some storage and use that instead.

bismurphy: But we create the local emulated scratchpad with 
```#ifdef VERSION_PC
    u8 sp[SP_LEN];
```
, right?

sozud: Ok if we're going that route then you don't need this part
```
POLY_GT4 _poly;
POLY_GT4* poly = &_poly;
EntitiesRenderer _r;
EntitiesRenderer* r = &_r;
```

sozud: I think generally this way is cleaner since the size of these objects may not be the same, so hardcoded scratchpad offsets aren't going to work right. 
```
POLY_GT4 _poly;
POLY_GT4* poly = &_poly;
EntitiesRenderer _r;
EntitiesRenderer* r = &_r;
```

bismurphy: Okay, changed PC to use this syntax. I also changed the two other versions to use `sizeof` in order to remove the ifdef, which is nice.

Xeeynamo: > But we create the local emulated scratchpad with `u8 sp[SP_LEN];`

Unfortunately that is not very reliable when working with pointers. For example if you have `a = (void*)SP(0);` and `b = (void*)SP(4)`, on 64-bit builds a pointer is 8 bytes long, so now `b = SP(4)` will fall into an unexpected portion of memory.

This is why I came up with `void *_b; b = &_b` as it is more reliable than anything else on any kind of configuration.

src/dra/4A538.c
```diff
@@ -719,84 +719,81 @@ void ResetEntityArray(void) {
...
+    s16 y;
+    s16 xPivot;
+    s16 yPivot;
+    u16 unused12;
+    s32 index;
+    s32 index2;
+    u32 spriteSheetIdx;
+    s32 facingLeft2; // set to 2 when facing left, 0 when not
+    s32 eDrawFlags;
+    u_long* ot;
```
Xeeynamo, line 738:
```suggestion
    OT_TYPE* ot;
```
This needs to get back to `OT_TYPE`, otherwise the graphics will look messed up

src/dra/4A538.c
```diff
@@ -719,84 +719,81 @@ void ResetEntityArray(void) {
...
-    entity = g_Entities;
-    setPolyGT4(poly);
-    r->index = 0;
     r->poly = &g_CurrentBuffer->polyGT4[g_GpuUsage.gt4];
     r->ot = g_CurrentBuffer->ot;
-    for (; r->index < 0x100; r->index++, entity++) {
-        animSet = entity->animSet;
-        r->animSet = animSet;
+
+    for (r->index = 0, entity = &g_Entities[0]; r->index < 0x100; r->index++,
```
Xeeynamo, line 775:
```suggestion
    for (r->index = 0, entity = &g_Entities[0]; r->index < LEN(g_Entities); r->index++,
```

src/dra/4A538.c
bismurphy: If we use `sizeof` with all `SP()` uses, would that fix things? That way even if a pointer is 8 bytes, the scratchpad will always align in a way that keeps everything in the right place.

src/dra/4A538.c
```diff
@@ -942,238 +940,247 @@ void RenderEntities(void) {
...
-                    if (!(temp_s2_2 & 1)) {
-                        poly->u0 = texSX;
-                        poly->v0 = texSY;
-                        poly->u1 = texEX;
-                        poly->v1 = texSY;
-                        poly->u2 = texSX;
-                        poly->v2 = texEY;
-                        poly->u3 = texEX;
-                        poly->v3 = texEY;
+                tempFlagCheck = (animFrameFlags & 2) ^ r->facingLeft2;
```
Xeeynamo, line 1074: I think you could rename `r->facingLeft2` as `r->flipYX`?

bismurphy: Sure. Why `flipYX` instead of just `flipX`?

src/dra/4A538.c
sozud: I think sizeof() fixes the issue, but what happens if the size of the structs exceeds SP_LEN? The `&_poly` method automatically allocates the right amount of space and you're not casting to u8 and back with is cleaner in my opinion. I think address sanitizer won't catch buffer overruns if you cast to `sp`. e.g. if the code did something like `poly->lastitem[too_big]` it would overwrite the next item in `sp` but not get caught

src/dra/4A538.c
Xeeynamo: I believe the first bit is to flip by Y? That's what I am understanding from line `1075`

src/dra/4A538.c
Xeeynamo: Agree with Sozud. I was trying to use `sizeof()` on `RenderPrimitives` but it didn't work that well either unfortunately

src/dra/4A538.c
bismurphy: I think in `animFrameFlags`, bit 1 is to flip Y (as in 1075), and bit 2 is to flip X, but `tempFlagCheck` only samples bit 2, so `r->flipX/flipYX` only controls the X flipping?

src/dra/4A538.c
bismurphy: That makes sense. We'll keep this as-is then. PSP/PS1 use `sizeof()`, while PC will use `&_poly`.

src/dra/4A538.c
```diff
@@ -805,125 +802,125 @@ void RenderEntities(void) {
...
             if (r->animSet > 0) {
-                if (r->animSet == 13) { // BAT FORM
-                    stuff = r->unk1C + g_PlOvlAluBatSpritesheet[0];
-                    spriteData = *stuff;
+                if (r->animSet == 13) {
+                    spriteData = (*g_PlOvlAluBatSpritesheet)[r->spriteSheetIdx];
                 } else {
-                    stuff = r->unk1C + g_PlOvlSpritesheet;
-                    spriteData = *stuff;
+                    spriteData = ((u8**)SPRITESHEET_PTR)[r->spriteSheetIdx];
```
Xeeynamo, line 826: I know it is not part of your PR, but `SPRITESHEET_PTR` in `game.h` at line `214` needs to be changed to `#define SPRITESHEET_PTR g_PlOvlSpritesheet`, otherwise the game will crash. I am still investigating why this PR does not render the Alucard sprite.

src/dra/4A538.c
```diff
@@ -719,84 +719,81 @@ void ResetEntityArray(void) {
...
     POLY_GT4* poly = &_poly;
     EntitiesRenderer _r;
     EntitiesRenderer* r = &_r;
 #else
-    POLY_GT4* poly = (POLY_GT4*)0x1F800000;
-    EntitiesRenderer* r = (EntitiesRenderer*)0x1F800034;
+    POLY_GT4* poly;
+    EntitiesRenderer* r;
+    poly = (POLY_GT4*)SPAD(0);
+    setPolyGT4(poly);
```
Xeeynamo, line 767: This needs to go outside the `#ifdef / #else / #endif` block, otherwise there will be no entity rendering

Xeeynamo (commented):
A few changes, some nice to have and other required. This is almost ready to go

src/dra/4A538.c
bismurphy: Done!

src/dra/4A538.c
```diff
@@ -805,125 +802,125 @@ void RenderEntities(void) {
...
+            r->yPivot = animFrame[1] + spriteData[3];
             r->w = spriteData[0];
             r->h = spriteData[1];
             poly->tpage = plDraw->tpage;
-            texSX = (plDraw->rect0.x & 0x3F) * 4;
-            texSY = plDraw->rect0.y;
-            texEX = texSX + r->w;
-            texEY = texSY + r->h;
+            uvRight = (plDraw->rect0.x & 0x3F) * 4;
+            uvTop = plDraw->rect0.y & 0xFF;
```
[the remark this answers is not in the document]
bismurphy: This is needed for PSP. See line 117 here: https://decomp.me/scratch/kdYz5

If we would rather have the code be clean than have it match on PSP, that's understandable, I just thought multi-system matching was desirable when possible. For now I'll leave it this way unless you prefer removing the `& 0xFF`

src/dra/4A538.c
```diff
@@ -942,238 +939,247 @@ void RenderEntities(void) {
...
-                    var_a0 = (entity->zPriority + 0x1A0) - PLAYER.zPriority;
-                    if (temp_s2_2 & 0x100) {
-                        var_a0 = var_a0 + new_var;
+                    *r->poly = *poly;
+                    polyX = (entity->zPriority + 0x1A0) - PLAYER.zPriority;
+                    if (animFrameFlags & 0x100) {
+                        polyX += 4;
                     }
+// Only exists in US, absent in PSP and HD
+#if !defined(VERSION_PSP) && !defined(VERSION_HD)
```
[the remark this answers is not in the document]
bismurphy: Great point, fixed

src/dra/4A538.c
bismurphy: Unfortunately the setPolyGT4 line is sandwiched between `poly` and `r` lines, so I decided to just duplicate it in both blocks, rather than having two separate sets of #ifdef/#else. Hopefully that's a suitable solution.

Xeeynamo (approved):
I am blessing this PR

Xeeynamo:
Can you do a rebase upon master or do a `git rebase -i 3b24fb2~1` and remove `3b24fb2` and `abaf7e8`? I think because of that it is showing a lot of files changed

bismurphy:
oh jeez i broke my git

bismurphy:
working on it :)

bismurphy:
Okay I think things are resolved now? CI is passing and I think all the extra file changes are gone.

PR #1396, opened 2024-07-08
Implement PadInit

Xeeynamo:
Tested with a dual sense

### DIFF
--- a/src/pc/sdl2.c
+++ b/src/pc/sdl2.c
@@ -27,13 +27,13 @@ DebugSdl g_DebugSdl = DEBUG_SDL_NONE;
 static SDL_Texture* g_VramTex = NULL;
 static int g_LastVramTexTpage = -1;
 static int g_LastVramTexClut = -1;
-static SDL_Joystick* joystick = NULL;
 
 void ResetPlatform(void);
 bool InitPlatform() {
     atexit(ResetPlatform);
 
-    if (SDL_Init(SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_JOYSTICK) < 0) {
+    if (SDL_Init(SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_GAMECONTROLLER) <
+        0) {
         ERRORF("SDL_Init: %s", SDL_GetError());
         return false;
     }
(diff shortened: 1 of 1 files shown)


src/pc/sdl2.c
```diff
@@ -164,6 +169,35 @@ u_long MyPadRead(int id) {
             pressed |= PAD_R1;
         }
 
+        if (joystick) {
+            SDL_JoystickUpdate();
+            pressed |= SDL_JoystickGetButton(joystick, 0) ? PAD_CROSS : 0;
```
sozud, line 174: I think you're supposed to use these?
https://github.com/davidsiaw/SDL2/blob/6ecaa6b61372e5b2f9bd01201814d07e34bb4186/include/SDL_gamecontroller.h#L311

src/pc/sdl2.c
```diff
@@ -164,6 +169,35 @@ u_long MyPadRead(int id) {
...
+            pressed |= SDL_JoystickGetButton(joystick, 2) ? PAD_TRIANGLE : 0;
+            pressed |= SDL_JoystickGetButton(joystick, 3) ? PAD_SQUARE : 0;
+            pressed |= SDL_JoystickGetButton(joystick, 4) ? PAD_L1 : 0;
+            pressed |= SDL_JoystickGetButton(joystick, 5) ? PAD_R1 : 0;
+            pressed |= SDL_JoystickGetButton(joystick, 6) ? PAD_L2 : 0;
+            pressed |= SDL_JoystickGetButton(joystick, 7) ? PAD_R2 : 0;
+            pressed |= SDL_JoystickGetButton(joystick, 8) ? PAD_SELECT : 0;
+            pressed |= SDL_JoystickGetButton(joystick, 9) ? PAD_START : 0;
+
+            short x_axis = SDL_JoystickGetAxis(joystick, 0);
```
sozud, line 185: https://github.com/davidsiaw/SDL2/blob/6ecaa6b61372e5b2f9bd01201814d07e34bb4186/include/SDL_gamecontroller.h#L265

sozud (approved):

src/pc/sdl2.c
Xeeynamo: Weird, they do not match? Only the first four fields A, B, X and Y matches. Everything else is off. Maybe it's my configuration only. Do you have any joystick you can test this with?

sozud: I didn't test this, I just meant it seems like we should use the enum rather than magic numbers?

sozud: Reading this it seems like the better API is SDL_GameControllerGetButton?  The enums are for that and not Joystick. https://discourse.libsdl.org/t/difference-between-joysticks-and-game-controllers/24028

sozud (approved):

PR #1397, opened 2024-07-09
Cleanup NP3 EntityBloodSplatter

bismurphy:
Another one of these PRs that takes an old function that was decompiled a while ago, and updates it to today's standards.

We move out of using ext.generic, and we use LOW() and LOH() macros where relevant.

We also cleanup the logic by using PSP as a guide.

This looks nicer, and is also a great example of lots of LOW and LOH on primitive members, which may be useful in studying what Primitive might really be, and whether any of its members (such as the RGBP sets) might actually be structs.


bismurphy:
I think there is enough LOH and LOW happening here that it makes sense to make a new alternative Primitive to represent these. I don't think it's reasonable to keep acting like these fields are normal Primitive fields when there is so much alternative typing needed.

sozud:
> I think there is enough LOH and LOW happening here that it makes sense to make a new alternative Primitive to represent these. I don't think it's reasonable to keep acting like these fields are normal Primitive fields when there is so much alternative typing needed.

I'm fine with creating a union, @Xeeynamo what do you think?

sozud (approved):

bismurphy:
> > I think there is enough LOH and LOW happening here that it makes sense to make a new alternative Primitive to represent these. I don't think it's reasonable to keep acting like these fields are normal Primitive fields when there is so much alternative typing needed.
> 
> I'm fine with creating a union, @Xeeynamo what do you think?

I think as of right now, we should just add another entry to Primitive.h along with the 6ish others that we have. Once we know the patterns that exist, and we have found how everything is used, then maybe exploring a union makes sense. But since 99% of the primitives in the game are just the normal Primitive, seems a bit clunky to have to specify which union member is used, every time a primitive is invoked.

sozud:
> I think as of right now, we should just add another entry to Primitive.h 

I think that's fine, Xeeynamo may disagree

Xeeynamo:
It is a nasty one. I have always used `RenderPrimitives` as a reference on how what we call `Primitive` is structured. But as Bismurphy observed there are too many patterns to be a coincidence. We also have a bunch of functions in `prim_helpers.h` that re-use certain fields.

I really think developers just had different structs that they carefully crafted to have certain fields aligned. We might need to do the same. Putting an `union` in the current `Primitive` might be a bit too much considering how extended the change is. For `prim->r0` alone I am counting more than 550 references.

Let's create a new struct in a separate Pull Request.

Xeeynamo (approved):

PR #1398, opened 2024-07-09
`rwrp` Decomp

hohle:
Replace 14 ASM functions with common files, duplicates, near duplicates, or new decomps.


hohle:
This leaves only `EntityRWarpRoom` which is [almost done](https://decomp.me/scratch/HanHi).

src/st/rwrp/8DF0.c
```diff
@@ -7,9 +7,94 @@
...
+                g_Tilemap.width = 1280;
+                entity->step++;
+                return;
+            }
+            break;
+        }
+
+        if (entity->unk44 != 0) {
+            ret = GetSideToPlayer();
+            phi_v1 = entity->ext.generic.unk7C.s;
```
sozud, line 72: There's a bunch of entity->ext.generic, we're trying to get rid of these https://github.com/Xeeynamo/sotn-decomp/issues/750

src/st/rwrp/D580.c
```diff
@@ -63,7 +63,61 @@ void FallEntity(void) {
         g_CurrentEntity->velocityY += FALL_GRAVITY;
     }
 }
-INCLUDE_ASM("st/rwrp/nonmatchings/D580", func_8018D990);
+
+#if !defined(VERSION_PSP)
```
sozud, line 67: We extracted rwrp on psp and confirmed this function isn't present?

PR #1399, opened 2024-07-10
_otc, register docs

sozud:


src/main/psxsdk/libgpu/sys.c
```diff
@@ -319,9 +326,31 @@ u_long get_tw(RECT* arg0) {
...
+    set_alarm();
+    if (*DMA6_CHCR & 0x01000000) {
+        while (1) {
+            if (get_alarm()) {
+                return -1;
+            } else {
+                if (!(*DMA6_CHCR & (1 << 24))) {
+                    break;
+                }
+            }
```
Xeeynamo, line 349:
```suggestion
            } else if (!(*DMA6_CHCR & (1 << 24))) {
                break;
            }
```
maybe like this is a bit more compact

Xeeynamo (approved):
left an optional suggestion. Exciting stuff!

src/main/psxsdk/libgpu/sys.c
sozud: Done

PR #1403, opened 2024-07-11
Use SFX constants

joshlory:
Open to suggestions for how to break this change up to make it easier to review. The data extraction necessary for naming values in `D_800ACF60` and `D_800ACF84` could be a separate commit... but I'm not sure it's worth it.


sozud:
Looks good!

PR #1405, opened 2024-07-11
Decompile DRA EntityPlayerOutline

bismurphy:
This was EntityMPReplenished, but then I looked into how it is used and found that it is used for many other things besides replenishing MP, so I named it more generically.

I think I hunted down every situation where it is used and documented them fairly well. In general, I think this is a relatively good-looking function and is easy to read.

Looking forward to seeing any feedback!


src/dra/7A4D0.c
```diff
@@ -1357,8 +1357,277 @@ void EntityPlayerBlinkWhite(Entity* self) {
...
+                      FLAG_UNK_20000 | FLAG_UNK_10000;
+        prim = &g_PrimBuf[self->primIndex];
+        // This is just not a for-loop, that's weird
+        for (i = 0; i < 1; i++) {
+            prim->tpage = 0x18;
+            prim->clut = primData[3]; // Always 259
+            prim->priority = PLAYER.zPriority + 2;
+            prim->drawMode =
+                primData[4] +
+                0x304; // Always 49, or DRAW_UNK_40 | DRAW_HIDE | DRAW_TRANSP
```
joshlory, line 1439: Nit: `(DRAW_UNK_200 | DRAW_UNK_100 | DRAW_COLORS)`

Also I think the comment is meant for the previous line?

bismurphy: This was all one line until the formatter broke it into 3, I'll clean this up, thanks!

include/game.h
```diff
@@ -721,7 +722,7 @@ typedef struct Entity {
     /* 0x1C */ s16 rotY;
     /* 0x1E */ s16 rotZ;
     /* 0x20 */ s16 rotPivotX;
-    /* 0x22 */ s16 rotPivotY;
+    /* 0x22 */ u16 rotPivotY;
```
joshlory, line 725: Why only change the type for `rotPivotY` and not `rotPivotX`?

bismurphy: Probably because Y only goes up from zero while X can be positive or negative for left/right.

Either way, this only matches if it's done this way. Most importantly, line 544 of src/dra/90264.c is `self->rotPivotX = -8;` so that one is definitely signed. But rotPivotY has to be unsigned or this function (the one in this PR) explodes.

joshlory (approved):

include/game.h
sozud: Did we do the psp version too? Does it need to be unsigned here?

sozud (approved):

include/game.h
bismurphy: PSP is here: https://decomp.me/scratch/MOeKz

I never got a good match on it so I don't think we can count on it to say one way or another.

sozud: Is this the ps1 scratch? https://decomp.me/scratch/TdHwY Unless I'm missing something it seems like s16 works? Changing it to u16 on the psp scratch definitely matches less

bismurphy: I guess I got things mixed up, you're both right - this doesn't need to change. Reverted back, now both rotPivot values are s16.

PR #1407, opened 2024-07-12
Add software rendering

sozud:
This takes the GPU from Mednafen and hooks it up by using parts of libgpu and writing to the fifo. I avoid using DMA and just write directly and it seems to work fine. This means that the OTAG part isn't required since that's part of the DMA system and not the GPU apparently. 

The main bug I'm aware of is the TILE that fades to white in the warp room doesn't show up. The GL renderer is still available. I made the software renderer the default since it's faster and more accurate on average. The two renderers can be toggled with F8. I turned on linear logs by default since it's significantly faster.

The biggest gap I'm aware of is PutDispEnv needs to be properly decompiled and hooked up.


src/pc/log.c
```diff
@@ -4,7 +4,7 @@
 #include <string.h>
 
 // uncomment to disable the logs grouping feature
-// #define FORCE_LINEAR_LOGS
+#define FORCE_LINEAR_LOGS
```
Xeeynamo, line 7: Let's comment this back?

src/pc/sdl2.c
```diff
@@ -24,9 +18,13 @@ SDL_Renderer* g_Renderer = NULL;
 SDL_AudioSpec g_SdlAudioSpecs = {0};
 SDL_AudioDeviceID g_SdlAudioDevice = {0};
 DebugSdl g_DebugSdl = DEBUG_SDL_NONE;
-static SDL_Texture* g_VramTex = NULL;
-static int g_LastVramTexTpage = -1;
-static int g_LastVramTexClut = -1;
+SDL_Texture* g_VramTex = NULL;
+int g_LastVramTexTpage = -1;
```
Xeeynamo, line 22: Can we try to move some of this stuff inside `render_gl.c` and mark it as `static` again?

src/pc/render_shared.c
```diff
@@ -0,0 +1,98 @@
...
+        vram += rect->x + rect->y * VRAM_W;
+
+        for (int i = 0; i < rect->h; i++) {
+            memcpy(vram, mem, rect->w * 2);
+            mem += rect->w;
+            vram += VRAM_W;
+        }
+    }
+
+    // copy to software vram
```
Xeeynamo, line 85: Can we do a `if (render_mode == RENDER_SOFT) {` here too? Or do you plan to switch between hw and sw rendering on the fly?

src/pc/render_shared.c
```diff
@@ -0,0 +1,98 @@
...
+void SoftDrawOTag(OT_TYPE* p);
+
+int GlDrawSync(int mode);
+DISPENV* GlPutDispEnv(DISPENV* env);
+void GlSetDrawEnv(DR_ENV* dr_env, DRAWENV* env);
+DRAWENV* GlPutDrawEnv(DRAWENV* env);
+void GlDrawOTag(OT_TYPE* p);
+
+int MyDrawSync(int mode) {
+    if (render_mode == RENDER_SOFT) {
```
Xeeynamo, line 24: Not for this PR, but I think we could just do something like the following later on:

```c
typedef struct {
    int (*DrawSync)(int mode);
    ...
} Renderer;

static Renderer renderer_hw = {
    .DrawSync = GlDrawSync,
    ...
};
static Renderer renderer_sw = {
    .DrawSync = SoftDrawSync,
    ...
};

static Renderer* renderer = &renderer_sw;
int MyDrawSync(int mode) {
    return renderer->DrawSync(mode);
}

```

Xeeynamo (commented):
I think this is a brilliant piece of work and you thought very well to use something already established that would aim to accuracy. I am not surprised it is also much faster.

I have a single concern, but I will move the conversation to Discord as I'd like to hear different opinions.

src/pc/log.c
sozud: Done, this is really slow on my system though. I think we should reduce the amount of logging

src/pc/sdl2.c
sozud: g_WndWidth/Height are used by both renderers.
g_LastVramTexTpage/Clut are used by ResetPlatform as well as the gl renderer
g_VramTex/g_DebugSdl is used by sdl.c and gl.c
 Is there something in particular you want moved?

src/pc/render_shared.c
sozud: Right now it's able to switch with F8, I don't know how useful it is or not. I prefer software rendering for this case so I don't need to be able to switch. What do you prefer?

src/pc/sdl2.c
Xeeynamo: mmh then I think it's fine for now as it is. I will think of a more structured solution as a separate PR

src/pc/render_shared.c
Xeeynamo: I have yet to try this PR (hence why it's taking so long for me to review). I missed the F8 thingie. That's very cool actually!

Xeeynamo (approved):

PR #1408, opened 2024-07-13
Decompile DRA func_8012F178

bismurphy:
The way this one uses scratchpad is super weird, but hey, it matches.

It's always fun to come across super-unique functions and I'm pretty confident to say this qualifies.


src/dra/8D3E8.c
```diff
@@ -764,7 +764,194 @@ void func_8012EF2C(void) {
...
+#define TPAGE 0x118
+#define V_OFFSET 0xC0
+#endif
+#ifdef VERSION_PC
+    scratchpad_8012F178 _s;
+    scratchpad = &_s;
+#else
+    scratchpad = (scratchpad_8012F178*)SPAD(0);
+#endif
+    scratchpad->prim = prim;
```
sozud, line 802: scratchpad is used so much in this function I kind of wonder if it should have a minimal name like
```
            s->prim->u0 = s->prim->u2 = 0;
            s->prim->u1 = s->prim->u3 = 0x3F;
            s->prim->v0 = s->prim->v1 = V_OFFSET;
            s->prim->v2 = s->prim->v3 = V_OFFSET + 0x3F;
            s->prim->tpage = TPAGE;
            s->prim->priority = PLAYER.zPriority - 6;
            s->prim->drawMode = 0x13F;
            s->helper->state++;
```
Should get approval from @Xeeynamo before changing though probably

sozud (approved):

src/dra/8D3E8.c
Xeeynamo: Yeah I think `s` or `sp` is enough. I _personally_ prefer shorter names for local variables as they are very targeted and self-contained with the function.
Also a good blog post that describes what to do in these kind of situations: https://www.lysator.liu.se/c/pikestyle.html

bismurphy: I'll go with `s` since `sp` is used as the actual scratchpad for PC in `common.h`

PR #1409, opened 2024-07-13
Decompile DRA func_8012F894

bismurphy:
This function was previously the last function in its file.

Upon decompiling, the rodata (specifically the `00 00 00 00` pattern) indicated that the file does not end here, and that in fact the entire following file should be part of this file, so I had to change the splits to accomplish this.

It's an ugly function, and was hard to get a match, but here it is. Did what I could to document variables but it doesn't seem to make much sense, if you ask me.


src/dra/8D3E8.c
```diff
@@ -951,4 +951,1098 @@ s32 func_8012F83C(s32 x0, s32 y0, s32 x1, s32 y1, s32 distance) {
...
+        // make it look more reasonable.
+        while (1) {
+            if (PLAYER.facingLeft) {
+                hitboxX = playerX + var_s1;
+            } else {
+                hitboxX = playerX - var_s1;
+            }
+            hitboxY += 2;
+            CheckCollision(hitboxX, hitboxY, &collider1, 0);
+            if (collider1.effects & 1) {
```
sozud, line 1094: enum, same for other effects checks https://github.com/Xeeynamo/sotn-decomp/blob/d73363b83b55b148a350d8c979f4806447a50443/include/game.h#L1168

PR #1413, opened 2024-07-16
Tetra spirit fixes, macro system

sozud:
This adds some fixes to get tetra spirit mostly working. It copies part of the framebuffer with MoveImage to warp it, so that's added. Right now there's some bugs with DrawEnv or DispEnv stuff so that the copy is from the wrong place every other frame and shows as black. There's a couple of out of bound issues that have speculative fixes I'm not sure about. 

The spells are hard for me to do so I added a macro/hotkey system to be able to test them quickly. I tried calling the spell functions directly but they didn't work right. I think there's probably code scattered around to support them so doing the inputs is the only way to activate it all correctly


src/dra/d_10798.c
```diff
@@ -17,13 +17,26 @@ s32 D_800B0830[] = {FIX(-2.5), FIX(-2.5), FIX(-2.5)};
...
+#else
 u8 D_800B0848[] = {0x80, 0x80, 0x80, 0x90, 0x80, 0xA0, 0x80, 0xB0,
                    0xA0, 0x80, 0xA0, 0x90, 0xA0, 0xA0, 0xA0, 0xB0};
+#endif
 
 u16 D_800B0858[] = {0x03C0, 0x0440, 0x0340, 0x04C0};
 
+// tetra spirit out of bounds problem, add another value at end
+#ifdef VERSION_PC
+s16 D_800B0860[] = {31, 0,   31, -32, 0,  -32, -32, -32, -32,
```
Xeeynamo, line 33: No need of this here. You can add the `0` at the end on the original array and it will still match

src/pc/render_shared.c
```diff
@@ -68,7 +68,8 @@ extern u8* GetFb();
 extern u16 g_RawVram[VRAM_W * VRAM_H];
 
 int MyLoadImage(RECT* rect, u_long* p) {
-    DEBUGF("(%X, %X, %X, %X): %p", rect->x, rect->y, rect->w, rect->h, p);
+    // extremely spammy
+    // DEBUGF("(%X, %X, %X, %X): %p", rect->x, rect->y, rect->w, rect->h, p);
```
Xeeynamo, line 72: Let's just remove this

src/pc/macros.c
```diff
@@ -0,0 +1,98 @@
...
+    if (keyb[scanCode] && !previousKeyStates[scanCode]) {
+        press_spell = 1;
+        press_pos = 0;
+        next_time = 0;
+        spell_num = which;
+        spell_dir = PLAYER.facingLeft;
+    }
+}
+
+void handle_macros(const u8* keyb, u_long* pressed) {
```
Xeeynamo, line 68: I wonder if we can leverage `DemoUpdate(void)` and `g_DemoMode` to pass macros? nothing that needs to be done now, this is good as it is

Xeeynamo (commented):
Left a few comments before I want to try it out

src/dra/d_10798.c
sozud: Done

src/pc/render_shared.c
sozud: Done

src/pc/macros.c
sozud: Looking at the usages of `g_DemoMode`, there's a bunch of entanglements that change the state of the game so I'm thinking it's cleaner to not use it.

src/pc/sdl2.c
```diff
@@ -10,6 +10,7 @@
 #endif
 #include "pc.h"
 #include "sdl_defs.h"
+#include "macros.h"
```
Xeeynamo, line 13: Could you rename `macros.c` and `macros.h` into `sdl2_macros.c` and `sdl2_macros.h`? so we remember to keep them isolated

Xeeynamo (approved):
Unfortunately I couldn't find the secret behind `D_800B0848`. As you mentioned it might simply be an oversight from the developers.

I left a file rename suggestion in the review, but everything else looks great!

PR #1414, opened 2024-07-17
Decompile RIC EntityPlayerBlinkWhite

bismurphy:
Wow! With this one, Richter is now down to a single non-matching function (it's a duplicate of a DRA function which is missing - the one with the funny pointer math which is connected with EntityPlayerDissolves).

Anyway, we're matching. This is different from DRA's EntityPlayerBlinkWhite, but in weird ways that made it a huge pain to get matching. Definitely was not a matter of replacing a few lines like some duplicated functions are.

Anyway, looking forward to see if there's anything to fix!


src/ric/26C84.c
```diff
@@ -79,7 +79,332 @@ bool func_80162E9C(Entity* entity) {
...
+    temp_v0 = ((u8**)SPRITESHEET_PTR)[temp_v1->count & 0x7FFF];
+    temp_v1++;
+    four = 4;
+    one = 1;
+    sp10 = four + temp_v0[0];
+    temp_a0 = one + temp_v0[1];
+    xSize = sp10 - four;
+    ySize = temp_a0 - one;
+    // Hmm, this seems to me like temp_v1 isn't actually SpriteParts
+    xOffset = temp_v1->count + temp_v0[2];
```
sozud, line 129: Where's the scratch for this? It definitely seems like D_801530AC is not SpriteParts. 

```

SpriteParts* D_801530AC[] = {
    (SpriteParts*)NULL,       (SpriteParts*)D_80153400,

...

int D_80153400[] = {0xFFF08000, 0x0000FFE8};

```

Seems like it's more like this but the second number seems too big

```
s16 D_80153400[] = {-16, -32768, 0,-24};
```

bismurphy: Scratch is here: https://decomp.me/scratch/Ul7qI

Notice that here, I have temp_v1 as `s16*` instead. When I moved the scratch into the repo, I found that `D_801530AC` was already SpriteParts so I changed this function to also use SpriteParts. I'm not sure how this could index into D_80153400, but it's possible. I guess it could be that     `temp_v1 = D_801530AC[PLAYER.animCurFrame & 0x7FFF];` is supposed to be `    temp_v1 = D_80153400[PLAYER.animCurFrame & 0x7FFF - 0xWhatever];`

sozud: I'm not sure I'm following the second part of you comment. Here's the relevant part:
```
    s16* temp_v1_2;
    temp_v1_2 = D_801530AC[PLAYER.animCurFrame & 0x7FFF];
    temp_v0 = ((u8**)SPRITESHEET_PTR)[temp_v1_2[0] & 0x7FFF];
    temp_v1_2++;
    var_s2 = temp_v1_2[0] + temp_v0[2];
    var_s3 = temp_v1_2[1] + temp_v0[3];
```
My reading is we get an entry of D_801530AC here
```
    temp_v1_2 = D_801530AC[PLAYER.animCurFrame & 0x7FFF];
```
For example that could be `D_80153400` if `PLAYER.animCurFrame == 1` since

```

SpriteParts* D_801530AC[] = {
    (SpriteParts*)NULL,       (SpriteParts*)D_80153400,
```

the first entry of `D_80153400` etc. is an index into `SPRITESHEET_PTR`. Then the next two are x and y offsets. It seems like the 4th value is unused?

bismurphy: Ah, understood. So the argument is that D_801530AC is not a SpriteParts*, and *also* that its members (such as D_80153400) are not SpriteParts.

I'll start converting the data over to just be s16 integers then, will update this PR once that's done.

sozud: We might want to get Xeeynamo's input before converting it. The first value is like a u16 since the top bit gets masked off like it has an alternate meaning
```
temp_v1_2[0] & 0x7FFF
```
Then the next two are like they are signed since they are added to the offsets
```
    var_s2 = temp_v1_2[0] + temp_v0[2];
    var_s3 = temp_v1_2[1] + temp_v0[3];
```
unless I'm miscounting though then you'd get the following for D_80153400:

```
index into sprite sheet: 0x7FF0 (pretty big)
xOffset: -32768, (unreasonably big for a sprite offset)
yOffset: 0,
```

bismurphy: Okay, I'll hold off on any changes for now.

bismurphy: Oh, looks like xeeynamo is doing it himself in the D_801530AC-refactor branch.

Xeeynamo: how about this? https://github.com/Xeeynamo/sotn-decomp/commit/596e4231955eb1616a15d0eebc3870755c9f0c71

bismurphy: Seems good to me. Will you push that to master and then I will rebase, or should I copy your changes into this PR?

Xeeynamo (commented):
A bit more context:
* `temp_v1[0]` is somewhat related to some kind of X transposition
* `temp_v1[1]` same but for the Y coord
* `temp_v0[0]` is the richter sprite width (check `tools/splat_ext/spritesheet.py`)
* `temp_v0[1]` is the richter sprite height
* `temp_v0[2]` is the richter sprite X pivot
* `temp_v0[3]` is the richter sprite Y pivot

A few variable name suggestions:
* `var_s2` as `xPivot`
* `var_s3` as `yPivot`
* `temp_v0` as `plSprite`
* `four` as `xMargin`
* `one` as `yMargin`
* `sp10` as `wSprite`
* `shadow_var_a0` as `hSprite`
* `sp18` as `width`
* `sp20` as `height`

In general I'd prefer give `four` and `one` a name based on the context they're used on, not for what they are. Because if I check the value of that variable and I find `4`, it still provides me no information on what it is supposed to do.

src/ric/26C84.c
Xeeynamo: You can just cherry-pick that commit, maybe it's cleaner than pulling that from my PR

bismurphy:
> In general I'd prefer give `four` and `one` a name based on the context they're used on, not for what they are. Because if I check the value of that variable and I find `4`, it still provides me no information on what it is supposed to do.

Strongly agree, I just completely failed in my attempts to understand what those were doing. Thank you for deciphering this function, I will rename all variables as specified!

src/ric/26C84.c
Xeeynamo: BTW for `0xC4 | 0x8000`, note that `0xC4` is `196`, the exact number of frames that Richter has in `assets/ric`

PR #1418, opened 2024-07-18
 Rework D_8003C908 

bismurphy:
The next function of SEL, func_801B99E4, makes it clear that D_8003C908 is not a struct member. I do not know why it was placed in this struct. I have that function now matching, but since we're changing around variables that are used elsewhere, I decided to split the PR so that we can handle one type of change at a time.

The D_8003C908 variable has not yet been used in any C code, so this change does not affect anything.

Further, since both struct members were s32, I decided to make it a 2-member array for the sake of conciseness in the code.


sozud:
Can you post the scratch for func_801B99E4?

bismurphy:
> Can you post the scratch for func_801B99E4?

Sure. Here's the existing scratch I found on decompme: https://decomp.me/scratch/7Ram0

I noticed that the mismatch was due to storing the address to s0 and then loading/storing to that cached pointer, while the target was writing directly to the address. I've seen this pattern before, where struct members are accessed with pointer caching while "loose" variables are done with a full two-instruction load from an absolute address.

Therefore, here is my matching scratch: https://decomp.me/scratch/quG1P

Also, note that in this PR, we have `pfnWeapon(g_EquipDefs[D_8003C90C[0]].palette);` and `pfnWeapon(g_EquipDefs[D_8003C90C[1]].palette);`, which makes it look like these are storing the left-hand and right-hand weapons, so it makes sense for this to only be a pair of values.

sozud:
Ok, the weapon 0/1 makes sense as an array and it's unlikely that an movie decoding variable was in the same struct

sozud:
Should we rename D_8003C90C to g_WeaponIds or similar?

bismurphy:
> Should we rename D_8003C90C to g_WeaponIds or similar?

How about g_EquippedWeaponIds, to make it clear that it's what is currently equipped, and not a full set of all IDs in the game?

sozud:
Sounds good to me

PR #1421, opened 2024-07-18
HD WRP matching

Xeeynamo:
Mostly identical to MAD and PSP. There are four very different functions from US that needs to be decompiled to be 100% complete:
* [x] `e_misc.c` `EntityMessageBox`
* [x] `e_misc.c` `EntityRelicOrb`
* [x] ~`e_stage_name.c` `StageNamePopupHelper`~ it does not exist
* [x] `e_stage_name.c` `EntityStageNamePopup`

These four very different functions are the only ones that use the JP text instead of the US one. This overlay also lacks of `BlitChar`.

`EntityStageNamePopup` is very similar to the PSP counterpart. I used @joshlory scratch to match the HD part.

`EntityRelicOrb` had some #ifdef between US and BETA (aka the MAD overlay). I found some code that crossed between BETA and HD, so I just put an `#else` after `VERSION_US`. We are starting to reconstruct how the source code originally evolved across different game builds.


config/splat.us.stwrp.yaml
```diff
@@ -54,7 +54,7 @@ segments:
       - [0x1228, .data, e_layout] # layout entries data
       - [0x1420, cmp, D_80181420]
       - [0x1764, cmp, D_80181764]
-      - [0x1D08, cmp, D_80181D08]
+      - [0x1D08, raw, D_80181D08]
```
sozud, line 57: It turns out this isn't cmp?

sozud (approved):

config/splat.us.stwrp.yaml
Xeeynamo: Yeah, the game performs a raw copy of that potion of data. It represents CLUT data and it's not compressed

PR #1422, opened 2024-07-18
Import gte code, candle test button

sozud:
This imports the psycross gte code and hooks it up. I added a hotkey to create a candle with a subweapon. As far as I can tell you can't give yourself a subweapon directly without breaking stuff since EntityBreakable and the entity it spawns set up various things. I used this to test the gte. This also fixes a couple of function pointers like CheckCollision to fix the cross animation. This removes a few gte functions we previously implemented in favor of their versions, I'm not sure about the tradeoffs there.


Xeeynamo:
I think we need to put the LICENSE file inside the PsyCross folder? https://github.com/OpenDriver2/PsyCross/blob/master/LICENSE

sozud:
> I think we need to put the LICENSE file inside the PsyCross folder? https://github.com/OpenDriver2/PsyCross/blob/master/LICENSE

Done

Xeeynamo (approved):
Sorry for the late review. I was waiting for @SoapyMan feedback (which I did not yet have).

It looks good. I am happy that all the testing logic remains isolated in the macros file.

PR #1424, opened 2024-07-20
Use allegrex assembler for PSP

mkst:
I moved *some* of the PSX-specific stuff out of the common `Makefile` and into the `Makefile.psx.mk` -- there is still more that could be done here, but this should be enough to have the correct rules for PSX and PSP with the desired assembler.

Also threw in a bonus feature where I've removed the `move` macro from `macro.inc` because `maspsx` does this automagically for you now.


sozud (approved):
Thanks for helping with this. Looks reasonable to me but needs a review from Xeeynamo

Xeeynamo (approved):
Looks good on my side too! Thanks for doing this, it will be helpful later on.

I renamed the compiler file from `allegrex.tar.gz` to `allegrex-as.tar.gz` as well if you want to make the PR a bit cleaner

PR #1425, opened 2024-07-21
Merge RWRP with WRP

Xeeynamo:
This will decompile all the functions but the new `EntityRWarpRoom` from the RWRP overlay.

Note that as `e_misc.c` has a function with a different castle flag, I had to use a shared header. Everything else was identical.


sozud (approved):
Seems good to me once analyze_calls is fixed

PR #1428, opened 2024-07-22
Attempt fixing analyze_calls build issue

bismurphy:
Iterating on this problem in master.

Now that wrp/rwrp are sharing so much code, analyze_calls is having trouble telling what functions call each other in which overlay.

I think this will resolve all the issues we have for now, but I'm sure more will appear over time.


sozud (approved):
Going to merge and see what happens

PR #1430, opened 2024-07-22
NP3 first steps to align C extraction

Xeeynamo:
Start to perform a series of code refactoring to align the various overlays on how NZ0, WRP and RWRP are currently organised, aiming to share as much code as possible. This is a good stop gap that does not change too much. There are virtually no major changes other than the removal of the fake symbol `D_8003BEEC`.

I found it is very hard to share every single line of code due to some minor differences in the castle flags. I couldn't yet spot a reliable pattern on how castle flags are manager across stages.

My vision is to share as much code as possible, possibly not as header files but as C files in a similar fashion of #1425 . I am aiming to model NP3 so we can merge it with NO3 given the fact they share a good chunk of the code base.

I am waiting to split the rest of the entities into their own individual files. My idea is to have a file per entity or per entity group with their own set of data and helper functions, possibly all marked as `static`. We agreed it is probably a good idea to wait exploring the PSP build of NZ0, NO3 or NP3 before performing this step.


src/st/nz0/entity_heart_drop.c
```diff
@@ -1,35 +1,36 @@
 #include "nz0.h"
 
 void EntityHeartDrop(Entity* self) {
```
sozud, line 3: Why not create a .h for EntityHeartDrop since we're updating it in so many places?

src/st/np3/e_misc.c
```diff
@@ -285,7 +324,7 @@ void func_801C1848(void) {
 
 #include "../clut_lerp.h"
 
-void func_801C2598(s16 sfxId) {
+void func_801916C4(s16 sfxId) {
```
sozud, line 327: I'm thinking these renames are being done to align the names among multiple overlays? We might want to pick actual names if the address doesn't match the function name anymore

sozud (approved):

src/st/nz0/entity_heart_drop.c
Xeeynamo: Because each one has its own `g_CastleFlag` index. Each overlay has a slightly different variant of the function. And I couldn't spot a reliable pattern unfortunately.

src/st/np3/e_misc.c
Xeeynamo: I was thinking `PlaySfx3D` specifically for this one. But I have no idea what the others are doing. I was going to replace this one with `#include "../e_misc.h` if it was not for `EntityHeartDrop`

src/st/nz0/entity_heart_drop.c
sozud: Each one seems to be + 0x100? Am I missing something?
```
g_CastleFlags[(index >> 3) + 0x100] |= 1 << (index & 7);
```

Xeeynamo: It's not that line but two lines above, where some overlays do ` + 0x118` and then `-= 0x118` a bit later. I could realistically add an ad-hoc `#define` and still use a shared header, which is what I was doing with `CASTLE_FLAG_BANK`. But that plan failed. I need to make more research with overlays that actually use candles and item drops.

The other issue is `index = D_80180***[index]`, which is different in each overlay. It wouldn't be a problem if it was not for WRP and RWRP where the array type is different from other overlays for some reason. Again, it needs more exploration.

This is why `e_misc` is the only C file I did not put it yet as a shared header. I do not think it is a blocker for this PR?

PR #1438, opened 2024-07-26
Decompile RIC func_8015A9B0

bismurphy:
This finishes RIC.

Very similar to the most recent DRA function 80115394, so noted that here.


Xeeynamo:
Congratulations :clap:

PR #1439, opened 2024-07-26
Decompile w58 (Alucart Shield) Shield Spell

bismurphy:
This modifies a variable at 0x80097420, I noticed that was in the g_UnkGraphicsStruct, so I went through all the uses of that variable and made them all use the struct.


src/weapon/w_058.c
```diff
@@ -17,7 +19,196 @@ s32 func_ptr_80170014(Entity* self) {}
...
+        prim->v0 = prim->v1 = 8;
+        prim->v2 = prim->v3 = 8 + 7;
+        prim->priority = self->zPriority + 4;
+        prim->drawMode = DRAW_HIDE;
+        SetSpeedX(FIX(-2));
+        self->velocityY = 0;
+        DestroyEntityWeapon(1);
+        g_api.PlaySfx(SFX_UNK_641);
+        g_unkGraphicsStruct.unk20 = 1;
+        // Would be nice if this could be removed
```
Xeeynamo, line 103: Now that you have removed the fake symbol in favour of `g_unkGraphicsStruct.unk20`, you should be able to remove the do while(0)

Xeeynamo (approved):

src/weapon/w_058.c
bismurphy: Oh holy crap, you're totally right, awesome catch!

PR #1440, opened 2024-07-26
Decompile w58 (Alucart Shield) EntityWeaponAttack

bismurphy:
Finishing the Alucart Shield.

Noticed we are doing stuff with offset 0xAC which sounds a lot like the one in the Player entity, so I called it anim for now especially since it seems connected with animations.

This finishes this shield. I'm going to try to go through all the shields in the weapon overlays and get them done, they seem like some of the most complicated since a lot of the shield spells do some kind of crazy stuff with graphics.

### DIFF
--- a/src/weapon/w_058.c
+++ b/src/weapon/w_058.c
@@ -43,7 +43,7 @@ void EntityWeaponAttack(Entity* self) {
         self->flags = FLAG_UNK_08000000;
         self->animCurFrame = 0x3E;
     }
-    if ((PLAYER.step == 2) && (PLAYER.step_s != PLAYER.step)) {
+    if ((PLAYER.step == Player_Crouch) && (PLAYER.step_s != 2)) {
         anim2++;
     }
     switch (self->step) {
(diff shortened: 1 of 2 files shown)


src/weapon/w_058.c
```diff
@@ -4,8 +4,145 @@
...
+            self->velocityX = PLAYER.velocityX;
+        } else {
+            self->velocityX = PLAYER.velocityX * 2;
+        }
+        self->velocityY = FIX(-3.5);
+        self->ext.timer.t = 0x80;
+        self->flags = FLAG_UNK_08000000;
+        self->animCurFrame = 0x3E;
+    }
+    if ((PLAYER.step == 2) && (PLAYER.step_s != PLAYER.step)) {
```
sozud, line 46: Player_Crouch?
https://github.com/Xeeynamo/sotn-decomp/blob/12f8f7149d0976df5f2b740710cfbc6e7b5ec143/include/game.h#L660

sozud (approved):

PR #1441, opened 2024-07-26
SFX defines for RIC, Teleport, CrossCrash

synamaxmusic:

PR #1442, opened 2024-07-26
Various weapon pointer funcs

joshlory:


bismurphy:
Can we try to keep PRs limited to single functions, or highly related functions, moving forward? This seems like 4 totally unrelated functions, would be nice to make sure a given PR has a fixed scope rather than being "various" things. This one isn't too bad, but keeping PRs tight and tidy is appreciated.

src/weapon/w_014.c
```diff
@@ -5,7 +5,55 @@
...
+        baseX = (u32)((rcos(angle) >> 4) * 5) >> 6;
+        baseY = (u32)((rsin(angle) >> 4) * 5) >> 6;
+        break;
+    case 2:
+        baseX = (u32)((rcos(angle) >> 4) * 0xF) >> 7;
+        baseY = (u32)((rsin(angle) >> 4) * 0xF) >> 7;
+        break;
+    case 3:
+        baseX = (u32)((rcos(angle) >> 4) * 5) >> 5;
+        baseY = (u32)((rsin(angle) >> 4) * 5) >> 5;
```
bismurphy, line 51: Do these all need the u32 cast?

src/weapon/w_045.c
```diff
@@ -41,7 +41,63 @@ INCLUDE_ASM("weapon/nonmatchings/w_045", EntityWeaponAttack);
...
+        PLAYER.step_s = 0;
+        PLAYER.ext.player.anim = 0xCF;
+        PLAYER.animFrameIdx = 0;
+        PLAYER.animFrameDuration = 0;
+        PLAYER.velocityX >>= 1;
+        return;
+    }
+    if (g_Player.padTapped & PAD_SQUARE) {
+        PLAYER.ext.player.anim = 0xCE;
+        g_api_PlaySfx(0x6F0);
```
bismurphy, line 76: I think we're trying to use enums for all SFX moving forward

src/weapon/w_047.c
```diff
@@ -5,7 +5,47 @@
...
+            }
+            self->unk4C = D_14D000_8017A528;
+            self->flags = FLAG_UNK_08000000 | FLAG_UNK_100000;
+            self->zPriority = PLAYER.zPriority + 2;
+            speed = (rand() % 1535) + 0x100;
+            self->velocityX = rcos((s32)speed) << 7;
+            self->velocityY = -(rsin((s32)speed) << 7);
+            self->posX.val += self->velocityX * 2;
+            self->posY.val += self->velocityY * 2;
+            self->drawMode = 0x30;
```
bismurphy, line 35: Should use enums for drawmode here

src/weapon/w_047.c
```diff
@@ -5,7 +5,47 @@
...
+
+s32 func_ptr_80170004(Entity* self) {
+    s16 speed;
+
+    if (self->step == 0) {
+        if (self->ext.factory.parent->entityId == 0) {
+            DestroyEntity(self);
+        } else {
+            SetSpriteBank2(D_14D000_8017A0D4);
+            self->animSet = 0x8011;
```
bismurphy, line 19: ANIMSET here

src/weapon/w_014.c
joshlory: Yeah, I couldn't get a match without: https://decomp.me/scratch/Y89cu

bismurphy: Change to `u16 baseX, baseY;` and then you can ditch the (u32) casts.

src/weapon/w_047.c
joshlory: Whoops missed this one!

src/weapon/w_045.c
joshlory: Should also be `g_api.PlaySfx`, will fix

joshlory: I'm tempted to hold off creating enums for these until https://github.com/Xeeynamo/sotn-decomp/pull/1441#issuecomment-2253470492 is resolved but I'll add these two

bismurphy: Fair, I think in general I'm in favor of over-documenting, if we end up duplicating entries or whatever it's easier to consolidate them later than to hunt through the repo digging for undocumented uses

sozud (approved):

PR #1452, opened 2024-07-29
Decompile Goddess Shield EntityWeaponShieldSpell

bismurphy:
Next shield, these seem to be coming together nicely, I'm looking forward to being able to study and compare all the shields when they're done.

### DIFF
--- a/src/weapon/w_025.c
+++ b/src/weapon/w_025.c
@@ -224,7 +224,7 @@ void EntityWeaponShieldSpell(Entity* self) {
                 0x80;
         prim->priority = self->zPriority - 4;
         prim->drawMode =
-            0x40 | DRAW_TPAGE2 | DRAW_TPAGE | DRAW_COLORS | DRAW_TRANSP;
+            DRAW_UNK_40 | DRAW_TPAGE2 | DRAW_TPAGE | DRAW_COLORS | DRAW_TRANSP;
         prim = prim->next;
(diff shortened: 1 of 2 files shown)


src/weapon/w_025.c
```diff
@@ -162,9 +163,315 @@ s32 func_ptr_80170014(Entity* self) {}
...
+        prim->u1 = prim->u3 = 0x80 + 0x4F;
+        prim->v0 = prim->v1 = self->ext.shield.unk7C + 0;
+        prim->v2 = prim->v3 = self->ext.shield.unk7C + 0x4F;
+        self->ext.shield.unk82 = 0;
+        prim->r0 = prim->g0 = prim->b0 = prim->r1 = prim->g1 = prim->b1 =
+            prim->r2 = prim->g2 = prim->b2 = prim->r3 = prim->g3 = prim->b3 =
+                0x80;
+        prim->priority = self->zPriority - 4;
+        prim->drawMode =
+            0x40 | DRAW_TPAGE2 | DRAW_TPAGE | DRAW_COLORS | DRAW_TRANSP;
```
sozud, line 227: We don't have an enum for 0x40?

sozud (approved):

src/weapon/w_025.c
bismurphy: Good catch, this code was copied and pasted from a different weapon which was made when there was no 0x40, fixed in both places now.

PR #1453, opened 2024-07-29
Decompile Dark Shield func_ptr_80170024

bismurphy:
Yet another one, seems many of these shields are padding out the struct as we go since we now have an unkA0.

### DIFF
--- a/src/weapon/w_024.c
+++ b/src/weapon/w_024.c
@@ -263,7 +263,7 @@ void func_ptr_80170024(Entity* self) {
         self->velocityY += self->ext.shield.unkA0;
         if (self->posY.i.hi < 0x80) {
             self->ext.shield.unk92 += 2;
-            self->ext.shield.unkA0 -= FIX(1.0/8);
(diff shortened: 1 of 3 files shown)


src/weapon/w_024.c
```diff
@@ -181,7 +181,111 @@ int GetWeaponId(void) { return 24; }
...
+            self->step += 1;
+        }
+        break;
+    case 3:
+        self->posX.i.hi = self->ext.shield.unk94 + (self->ext.shield.unk80 & 1);
+        self->posY.val += self->velocityY;
+        self->velocityY += self->ext.shield.unkA0;
+        if (self->posY.i.hi < 0x80) {
+            self->ext.shield.unk92 += 2;
+            self->ext.shield.unkA0 -= 0x2000;
```
sozud, line 266: This is added to velocityY so I think it's FIX?

sozud (approved):

src/weapon/w_024.c
bismurphy: Good call, thanks

PR #1455, opened 2024-07-30
Symbol Matching

hohle:
Matching several symbols for consistent names across duplicates.


hohle:
While working on other dups, I was running into symbols that were known in other objects and did a quick pass through `duplicates.txt` to find known names.

hohle:
Looking into build failures.

joshlory:
Nice cleanup!

src/st/no3/44C64.c
```diff
@@ -255,16 +255,20 @@ u16 GetNormalizedAngle(u16 arg0, u16 arg1, u16 arg2) {
 }
 
 void SetStep(u8 step) {
```
sozud, line 257: Do the psp versions of these not match everywhere?

hohle: Yes, the PSP version should be the default. Fixed.

sozud: It's still not the psp version?

Xeeynamo (approved):
Looks good!

For `SetStep` it is fine as it is. We will split that portion of code into `st_common.c` and normalize it as we're doing with the rest of the code anyway

PR #1456, opened 2024-07-31
Early Stage Bootstrap Tool

hohle:
PSP overlay objects have enough metadata to automate creating a basic Splat config. This looks at a stage object and creates the boundaries as specified in the MetroWorks overlay header.

### DIFF
--- a/tools/sotn-disk/bootsrap.go
+++ b/tools/sotn-disk/bootsrap.go
@@ -1,0 +1,118 @@
+package main
+
+import (
+    "bytes"
+    "crypto/sha1"
+    "encoding/binary"
+	"errors"
+	"fmt"
+	"io/ioutil"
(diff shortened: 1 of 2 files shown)


hohle:
I've been using this to quickly get to ASM files and thought it might be useful to others. `sotn-disk` might not be the right place for this.

Xeeynamo:
I might have missed to make it clearer we already have a script that does this, plus a bit more: https://gist.github.com/Xeeynamo/4541578da3c6aa07f46ec0fa2b8baff2

I am not sure if we should close this and continue to iterate to the existing script and maybe put it in the repo.

hohle:
Yeah, yours does quite a bit more! I don't mind closing this in favor of something else that integrates that script.

PR #1457, opened 2024-07-31
Decompile w_038, w_039 `EntityWeaponAttack`

hohle:
Decompiles `EntityWeaponAttack` for `w_038` and `w_039`. These match several other `EntityWeaponAttack` methods with the addition of a conditional call to `CreateEntFactoryFromEntity` at the end.


Xeeynamo (approved):
Nice one!

PR #1458, opened 2024-08-01
Added SFX enum and SFX_RIC_WHIP_RATTLE_A

synamaxmusic:
Starting off with something simple.  Per [sozud](https://github.com/sozud)'s suggestion in #1441 , I put in a new enum for sounds.  There's only one reference to Richter's whip sound in the codebase so far.  According to my research, BO6 and RBO1 reference this ID as well.  

There's several other whip rattle sounds (0x603-0x606) immediately following the commonly heard 0x602 cue.  Unless BO6 or RBO1 uses random sound calls (like Alucard's randomized grunts), I don't believe these are referenced in the game and are unused.

### DIFF
--- a/include/sfx.h
+++ b/include/sfx.h
@@ -386,13 +386,18 @@ typedef enum { MONO_SOUND, STEREO_SOUND } soundMode;
 #define SE_UNK_TE3_6B8 0x6B8
 #define NA_VO_MAR_AIM_HEAD 0x87F
 
+// The VAB IDs appear in large chunks so all sounds proceeding
+// a vabid label comment will belong in that VAB group unless noted.
+
 enum Sfx {
-    SFX_HARPY_WING_FLAP = 0x601,   // vabid 0
-    SFX_RIC_WHIP_RATTLE_A = 0x602, // vabid 0 (also used in BO6 & RBO1)
-    SFX_RIC_WHIP_RATTLE_B = 0x603, // vabid 0 (603-606 are unused)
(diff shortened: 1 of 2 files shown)


include/sfx.h
```diff
@@ -389,4 +386,13 @@ typedef enum { MONO_SOUND, STEREO_SOUND } soundMode;
 #define SE_UNK_TE3_6B8 0x6B8
 #define NA_VO_MAR_AIM_HEAD 0x87F
 
+enum Sfx {
+    SFX_HARPY_WING_FLAP = 0x601,   // vabid 0
```
sozud, line 390: The compiler will automatically number each entry so we only need the 0x601.

sozud: I'm also thinking instead of labeling the vabid of each sfx, to just comment when it changes. The vabids are generally in large blocks See https://gist.github.com/sozud/e7dbf3f3192d66c33b003e52d8a54355

sozud:
```
import json

with open('assets/dra/sfx.json', 'r') as f:
    data = json.load(f)

def format_output(data):
    prev_id = None
    for asset in data['asset_data']:
        id = format(asset['id'] + 0x601, 'X')
        if asset['vabid'] != prev_id:
            print(f"// vabid {asset['vabid']}")
        print(f"SFX_UNK_{id},")
        prev_id = asset['vabid']

format_output(data)
```

sozud (approved):

PR #1465, opened 2024-08-01
More SFX added to enum

synamaxmusic:
Slowly adding more sounds and updating all the references in the codebase.


sozud (commented):
Other than the saturn breakage seems good to me

PR #1468, opened 2024-08-02
More SFX additions to enum

synamaxmusic:
I decided to rename `SFX_SKULL_HIT` to `SFX_METAL_CLANG`, as well as `SFX_SWORD_SWISH` to `SFX_WEAPON_SWISH`.

These are more generic names that better match up with the actual sound.  While `SFX_SWORD` is mainly used for sword weapon cues, it is used for other weapon entities so I renamed it.

Same idea also applied to `METAL_CLANG`; it's mainly used for the CGI spinning Stone Skull enemies but it also shows up in CEN as an elevator sound cue.

I also used `SFX_RIC_RSTONE_TINK` as an abbreviation for the Rebound Stone "tink" sfx, to help keep these names short.


include/sfx.h
```diff
@@ -393,15 +392,16 @@ enum Sfx {
...
-    SFX_SKULL_HIT_E,
-    SFX_SKULL_HIT_F,
+    SFX_WEAPON_SWISH_A,
+    SFX_WEAPON_SWISH_B,
+    SFX_WEAPON_SWISH_C,
+    SFX_METAL_CLANG_A,
+    SFX_METAL_CLANG_B,
+    SFX_METAL_CLANG_C,
+    SFX_METAL_CLANG_D,
+    // Stone Skull Hit
```
sozud, line 402: Could we put comments about individual entries to the right? e.g.
```
SFX_METAL_CLANG_E, // Stone Skull Hit
```
This makes it so the vabid changes and info about the sounds are easier to read in my opinion

synamaxmusic: Sure thing, one sec.

sozud (approved):

PR #1469, opened 2024-08-03
Added and updated SFX references 

synamaxmusic:
Stopping at 0x633 (`SFX_UI_SELECT`) for now, but will continue to add and update more sfx references soon.  

So far we have 51 out of a potential 736 sfx IDs (`0x8E0` - `0x600`).  

For what it's worth, the last valid sound ID I have listed in my spreadsheet is `0x8D2` found in `DRA`, `RIC`, `BO4`, `BO6`, and `RBO5` so the actual number might be 722 enum IDs, but we'll see what happens since the last sfx group in the list belongs to vabid 3.

### DIFF
--- a/src/dra/627C4.c
+++ b/src/dra/627C4.c
@@ -592,7 +592,7 @@ s32 HandleSaveMenu(s32 arg0) {
                 return 2;
             }
             if (g_pads[0].tapped & CONFIRM) {
-                PlaySfx(SE_UI_CONFIRM);
+                PlaySfx(SFX_UI_SELECT);
                 FreePrimitives(D_80137E58);
                 FreePrimitives(D_80137E5C);
                 FreePrimitives(D_80137E60);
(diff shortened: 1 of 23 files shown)


src/dra/627C4.c
```diff
@@ -592,7 +592,7 @@ s32 HandleSaveMenu(s32 arg0) {
                 return 2;
             }
             if (g_pads[0].tapped & CONFIRM) {
-                PlaySfx(SE_UI_CONFIRM);
+                PlaySfx(SFX_UI_SELECT);
```
sozud, line 595: In my mind confirm is pressing X and select is pressing up/down/left/right. If this is select what are you going to call up/down/left/right?

synamaxmusic: I was planning on having the `0x67B` sound cues labeled as `SFX_UI_MOVE` since it involves the D-pad.

I might just revert back to confirm because it does make a little more sense.  One of the reasons why I changed it to select is that I'm trying to find a good name for the Start Button confirm bang at the title screen, but that cue is used for other cues besides UI like door sounds and Dracula monster stomps.

Xeeynamo (approved):
Happy to see very good progress on documenting the SFXs. Thank you a lot!

src/dra/627C4.c
synamaxmusic: Reverting now...

synamaxmusic:
> Happy to see very good progress on documenting the SFXs. Thank you a lot!

Absolutely, glad to finally help out!

sozud (approved):

PR #1470, opened 2024-08-04
Weapon assets and statically linked

Xeeynamo:
Allow to compile weapon overlays as a single C files. The benefit is to disallow symbols to be exported by statically compile them all but the header. We would no longer need to use `OVL_EXPORT` as long as the symbols are marked as `static`. The frames data is also generated as an array of data that can be just included into their respective C files, generated by JSON files on the fly.

I changed `animset` to `frameset` as the data does not represent a set of animations but rather a set of frames, where every frame has multiple sprites. The information on how the frames are played (e.g. animation) is found elsewhere. 

Following the full list of changes to achieve this project:
* The `animset` from splat is now dummied out
* The `frameset` JSON is now generated by the asset manager
* The `frameset` JSON is now compiled into `src/weapon/w_0xx_y.h`
* The `header.c` has been moved into `shared.h`
* For simplicity I added a `#define g_Animset w_000_1`. I did not want to hack the asset manager too much to force the name to be `g_Animset` and I did not want to change all the symbols yet.
* Simplified `weapon_pc.c` as we now only need the exported header
* Always try to build the most up-to-date asset manager: this should avoid weird errors whenever people pull new changes

As the asset manager now can accept those `config` files, now the CLI supports the following commands:
* `stage extract`
* `stage build`
* `config extract` (NEW)
* `config build` (NEW)

The config structure is heavily inspired to Splat to maintain consistency and continuity.


sozud:
The msvc line that fails is the same as in https://github.com/nodejs/build/issues/3739
```
  (compiler file 'D:\a\_work\1\s\src\vctools\Compiler\Utc\src\p2\main.c', line 242)
```
That fix still seems to be in preview. We could try this:
```
The only "solution" with the current MSVC release is to #pragma optimize off/on around the problem function. 
```

sozud:
Could also try disabling optimizations at the cmake level like https://stackoverflow.com/questions/33540578/cmake-how-to-disable-optimization-of-a-single-cpp-file

sozud:
It seems like if we get rid of this syntax the older version would work
```
D:\a\sotn-decomp\sotn-decomp\src\pc\stages\../../st/wrp/layers.h(11,6): error C2059: syntax error: '}' [D:\a\sotn-decomp\sotn-decomp\pc\core.vcxproj]
```
```
static MyLayer layers[] = {
    {},
    { tilemap_01D68, tiledef_06D78, 0x01328328, 0x60, 3, 0 },
```

sozud:
How about just do `#if 0` on w_000.c to get this compiling and we monitor the situation with msvc? If only one file is busted on msvc we are still getting 99% coverage. This keeps getting conflicted endlessly

Xeeynamo:
> How about just do `#if 0` on w_000.c to get this compiling and we monitor the situation with msvc? If only one file is busted on msvc we are still getting 99% coverage. This keeps getting conflicted endlessly

I think I finally got it!

sozud (approved):

PR #1472, opened 2024-08-04
Decompile Axelord Shield func_ptr_80170024

bismurphy:
Next one! Lots of collisions here which is interesting.

### DIFF
--- a/src/weapon/w_011.c
+++ b/src/weapon/w_011.c
@@ -301,11 +301,9 @@ void func_ptr_80170024(Entity* self) {
                 goto make_other_entity;
             }
         }
-        if ((((collider.effects & (EFFECT_UNK_8000 | EFFECT_UNK_4000)) ==
-              (EFFECT_UNK_8000 | EFFECT_UNK_4000)) &&
+        if ((((collider.effects & EFFECT_UNK_C000) == EFFECT_UNK_C000) &&
              (self->facingLeft)) ||
-            (((collider.effects & (EFFECT_UNK_8000 | EFFECT_UNK_4000)) ==
-              EFFECT_UNK_8000) &&
+            (((collider.effects & EFFECT_UNK_C000) == EFFECT_UNK_8000) &&
              (self->facingLeft == 0))) {
             if (!(collider.effects &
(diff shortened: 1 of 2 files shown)


src/weapon/w_011.c
```diff
@@ -155,8 +158,218 @@ int GetWeaponId(void) { return 11; }
...
+            yOffset = 8;
+            xVar = self->posX.i.hi + xOffset;
+            yVar = self->posY.i.hi + yOffset;
+            g_api.CheckCollision(xVar, yVar, &collider, 0);
+            if ((collider.effects & EFFECT_SOLID) &&
+                !(collider.effects & EFFECT_UNK_8000)) {
+                goto make_other_entity;
+            }
+        }
+        if ((((collider.effects & (EFFECT_UNK_8000 | EFFECT_UNK_4000)) ==
```
Xeeynamo, line 304:
```suggestion
        if ((((collider.effects & (EFFECT_UNK_8000 | EFFECT_UNK_4000)) ==
```
I wonder if extracting `(EFFECT_UNK_8000 | EFFECT_UNK_4000)` into a `const` or `#define` would make that if statement less of a nightmare to read

src/weapon/w_011.c
```diff
@@ -155,8 +158,218 @@ int GetWeaponId(void) { return 11; }
...
+        } else {
+            goto make_other_entity;
+        }
+        break;
+    case 3:
+        if (self->animFrameDuration >= 0) {
+            break;
+        }
+        DestroyEntity(self);
+        return;
```
Xeeynamo, line 353:
```suggestion
        if (self->animFrameDuration < 0) {
            DestroyEntity(self);
            return;
        }
            break;
```
flipping this for consistency. I did not test if this matches

Xeeynamo (approved):

sozud (approved):

src/weapon/w_011.c
bismurphy: Ah yes good point of course, will swap

src/weapon/w_011.c
bismurphy: Maybe I'll just make an UNK_C000 and then we can name it properly once we know what each of the individual 8000 and 4000 are

PR #1473, opened 2024-08-04
Import BSS section across all overlays

Xeeynamo:
I did not do it for `weapon`. This is the script I used:

`python3 a.py asm/us/st/dre/data/23264.sbss.s > src/st/dre/bss.c`

```python
import sys

with open(sys.argv[1], "r") as f:
    lines = f.readlines()

print('#include "common.h"')
print("")
for line in lines:
    if line == "\n":
        continue
    elif line.startswith(".include"):
        continue
    elif line.startswith(".section"):
        continue
    elif line.startswith("glabel"):
        label = line[7:].replace("\n", "")
        len = 0
    elif ".word" in line:
        if len > 0 and n != 4:
            print(f"WARN: {label}", file=sys.stderr)
        n = 4
        len += 1
    elif ".short" in line:
        if len > 0 and n != 2:
            print(f"WARN: {label}", file=sys.stderr)
        n = 2
        len += 1
    elif ".byte" in line:
        if len > 0 and n != 1:
            print(f"WARN: {label}", file=sys.stderr)
        n = 1
        len += 1
    elif line.startswith(".size"):
        if len == 1:
            if n == 1:
                print(f"u8 {label};")
            elif n == 2:
                print(f"u16 {label};")
            elif n == 4:
                print(f"u32 {label};")
        else:
            if n == 1:
                print(f"u8 {label}[{len}];")
            elif n == 2:
                print(f"u16 {label}[{len}];")
            elif n == 4:
                print(f"u32 {label}[{len}];")

```

the script is a bit dumb. It does not account of the header. Some types are wrong compared to their prototype. The memory layout matches though, so we can keep iterating on top of this.


config/splat.hd.dra.yaml
```diff
@@ -198,9 +198,14 @@ segments:
...
       - [0x94B2C, c, 94F50]
       - [0x94F7C, c, 953A0]
-      - [0x95E88, sbss]
+      - [0x95E88, .bss, bss]
   - name: dra_data
     type: data
     start: 0x9B270
     vram:  0x8013B370
+    subsegments:
+      - [0x9B270, raw, vb_0] # D_8013B6A0
```
sozud, line 207: So there's no gap like in `us`? Haven't done the math to see if these line up with the sizes declared in the code or not
```
    subsegments:
      - [0x9B6A0, raw, vb_0] # D_8013B6A0
      - [0xDD350, raw, vb_1] # D_8017D350
      - [0xEB4E0, raw, vb_2] # D_8018B4E0 (len 108048)
      - [0x105AF0]
      - [0x109C80, raw, vb_3] # D_801A9C80 (len 64496)
```

sozud (approved):

config/splat.hd.dra.yaml
Xeeynamo: TBH I am not sure? That part gave me quite the headache, so I kept bruteforcing until I got an :ok: lmao

sozud: The lengths ought to match
```
    LOAD_VAB(0, aPbav, g_VabAddrs[0], D_8013B6A0, vab0Len);
    LOAD_VAB(1, aPbav_0, g_VabAddrs[1], D_8017D350, vab1Len);
    LOAD_VAB(2, aPbav_1, g_VabAddrs[2], D_801A9C80, vab2Len);
    LOAD_VAB(3, aPbav_2, g_VabAddrs[3], D_8018B4E0, vab3Len);
```
which differ across US and HD

sozud: 0: 0xDCF1C-0x9B270=269484
1: 0xEAFEC-0xDCF1C=57552
2: 0x10978C-0xEAFEC=1087357
3: 0x115EDC-0x10978C=51024

HD
0: 269488
1: 57808
2: 65200
3: 107792

I think I did the math wrong but some of these don't seem right

Xeeynamo: I will double-check

Xeeynamo: The problem is that the VABs  data between HD and US do not match. The only portion of data I could find matching is `vb_0`, where in HD starts 4 bytes earlier than in HD and it's not fixable at the moment, probably due to the alignment forced being each 16 bytes.

The addresses match though, despite the data length not matching the code. I am really confused on what's happening here.

sozud: It's expected that they aren't going to match because they replaced the voice acting for Alucard/Richter and the servants

Xeeynamo: Shall we merge it as it is? The HD build is not even building on PC and we can fix this later if we find a solution. But I do not think I have enough knowledge to understand what's going on unfortunately

sozud: Put a `TODO lengths are wrong. see vab1Len etc.` or something?

sozud:
Actually shouldn't the variables be removed from stubs.c? I suppose this is working now since the new bss files don't get compiled/linked. That could be a follow-up

Xeeynamo:
> Actually shouldn't the variables be removed from stubs.c? I suppose this is working now since the new bss files don't get compiled/linked. That could be a follow-up

It is because I do not have BBS.C linked. I have a follow-up PR ready, but one of the sound tests fail. I need to dig into it

PR #1474, opened 2024-08-04
Decompile Axelord Shield EntityWeaponShieldSpell

bismurphy:
Next one! Mostly copied and pasted from other shield spells.


Xeeynamo (approved):
Impressive work!

PR #1478, opened 2024-08-05
DRA reorganise BSS section

Xeeynamo:
A follow-up to #1473 where I imported the whole BSS section by respecting the symbol offset but ignoring their actual type. The work here consists into checking all the almost 500 symbols in DRA one by one. Each type in `bss.c` will be changed to respect the one actually defined in `dra.h` and document the file that actually holds reference to the symbol. If the symbol is only used in one file it will be removed from `dra.h`. Once the work is done I am planning to get rid of `src/dra/bss.c` and have each file to hold the reference.

The reason why I created `dra_bss.h` is because when GCC does not order BSS by how they are declared, but when it finds them first in the code. So doing `extern s32 b; s32 a; s32 b;` will order them as `b` then `a`. `dra_bss.h` needs to not be included in `src/dra/bss.c` but all the structs and enums should.

Tons of fake symbols have been addressed as part of this PR.


src/dra/bss.c
```diff
@@ -1,461 +1,470 @@
...
+u_long* D_80138784[0x700];
+#else
+u8 g_SeqTable[800];
+u_long* D_80138784[488];
+#endif
+s32 g_CurCdPos;
+u8 g_CdMode[2];
+padding u8 D_80138F26[2];
+s32 g_KeyOffChannels;
+u8 g_CdCommandResult[56];
```
sozud, line 295: Length seems too long but it won't hurt anything
image

hohle:
> Once the work is done I am planning to get rid of src/dra/bss.c and have each file to hold the reference.

Yes!

src/dra/5D5BC.c
```diff
@@ -754,14 +762,13 @@ s32 HandleDamage(DamageParam* damage, s32 arg1, s32 amount, s32 arg3) {
...
     do {
-        if (*var_v1 != 0) {
-            (*var_v1)--;
+        if (*p) {
+            (*p)--;
         }
-        var_v1++;
-    } while ((s32)var_v1 < (s32)&D_80139828[0x10]);
+        p++;
+    } while ((long long)p < (long long)&D_80139828[LEN(D_80139828)]);
```
hohle, line 771: This shouldn't block merging, but the convention I've typically seen for comparing to the end of a string is `(str + len(str))` as opposed taking a reference to one past the end of the array. The advantage to the addition is static analyzers will see out of bounds array access and flag it as a bug, when really the goal is to make sure the pointer doesn't go out of bounds to begin with.

src/st/np3/bss.c
```diff
@@ -1,4 +1,4 @@
-#include "common.h"
+#include <game.h>
```
hohle, line 1: It's not worth pulling this out into a header here, but part of the header refactor PR was to avoid signatures like this where previously a few dozen lines of type defs were being included, and now 5K+ lines of type defs, global refs, and all kinds of unrelated things are now part of compiling this file.

I don't think it's a blocker and this file with eventually go away anyway.

hohle (commented):
Just nits and rants. This is awesome.

src/st/np3/bss.c
Xeeynamo: Yeah I am not a fan of this either. But unfortunately there's so much we can do by going into hypothesis on how the game was originally structured. It's like the current C files with weird names: we will slowly find a meaningful organisation for them.

sozud (approved):
Tested and not noticing any regressions, seems ok to me

PR #1479, opened 2024-08-05
Major SFX reference update

synamaxmusic:
We've hit the first beta sound ID `0x640`, which is the old door open sound effect (final is `0x642`).  How do we want to implement the beta sfx IDs in the enum?


sozud:
> We've hit the first beta sound ID 0x640, which is the old door open sound effect (final is 0x642). How do we want to implement the beta sfx IDs in the enum?

Is every SFX in MAD offset by 2? If so you could do something like
```
#define STAGE_MAD
#include "sfx.h"
```

Then in sfx.h

```
// this removes a couple of sfx to get MAD to align
#ifndef STAGE_MAD
    SFX_HARPY_WING_FLAP = 0x601,
    SFX_RIC_WHIP_RATTLE_A,
#endif
```

If the offsets are random then you could just have a second enum

sozud:
Ok I think I'd just have two enums, one with `#ifdef STAGE_MAD` just for that overlay, then the other for everything else

include/sfx.h
```diff
@@ -426,6 +414,49 @@ enum Sfx {
...
+    SFX_CANDLE_HIT,
+    SFX_TELEPORT_BANG_A, // Used when teleport starts
+    SFX_TELEPORT_BANG_B, // Used when teleport ends, but is also Warp Enter
+    SFX_SUC_APPEAR,      // Succubus reveal
+    SFX_UNK_SCRAPE_A,
+    SFX_UNK_SCRAPE_B,
+    SFX_UNK_SCRAPE_C,
+    SFX_UNK_UI_SELECT,
+    SFX_START_SLAM_A, // Dark Shield
+    SFX_START_SLAM_B, // "Press Start", Door Open, Dracula Stomp, etc.
```
hohle, line 426: Having just gone through this list to find a particular symbol, does it make sense to add a value marker comment every 0x10 entries or so?

```c
    ⋮
    // 0x630
    SFX_UNK_BETA_630, // MAD, TE1, …
    ⋮
    // 0x640
    SFX_ANIME_SWORD_B,
    ⋮
```

hohle (commented):
This is really nice.

include/sfx.h
sozud: I think this is fine but I'd put it to the right so it doesn't conflict with the vab info 
```SFX_ANIME_SWORD_B, // 0x640```

**Comment by sozud** [comment]
This seems good except the change to MAD should be reverted unless the second enum is added

File: include/sfx.h (hunk 1 above)
Comment (synamaxmusic) [reply include/sfx.h:426]: That's a great idea, I'll be sure to do that.

**Comment by synamaxmusic** [comment]
Sure thing, I'll revert those MAD changes.  

Looking through MAD.BIN, I made a list of all the sound IDs that I could find.  I searched through the binary data by looking up `060434` and came up with this:

```
0x635
0x611
0x630 = WEAPON_STAB_B_MAD
0x6DB = WEAPON_HIT_MAD
0x640 = DOOR_OPEN_MAD
0x64D = DOOR_CLOSE_MAD
0x670
0x69D
0x672
```
I'll have to look at the codebase to see what calls PlaySfx() for these other IDs.

Since there's not that may sound IDs referenced in MAD, maybe `#ifdef` instead of another large enum would be the better option?

**Comment by sozud** [comment]
MAD is a special case so I don't think it really matters if it's an enum or individual defines. An enum doesn't have to be big though
```
enum Sfx
{
WEAPON_STAB_B = 0x630,
WEAPON_HIT = 0x6DB,
};
```
etc.

**Comment by synamaxmusic** [comment]
Figured out `0x635` is supposed to be the Candle Hit, not the Teleport bang.  

```
0x635 = SFX_CANDLE_HIT_MAD
0x611
0x630 = WEAPON_STAB_B_MAD
0x6DB = WEAPON_HIT_MAD
0x640 = DOOR_OPEN_MAD
0x64D = DOOR_CLOSE_MAD
0x670
0x69D
0x672
```

Comparing the decompiled MAD code with the final overlay code is very useful!

File: include/sfx.h (hunk 1 above)
Comment (synamaxmusic) [reply include/sfx.h:426]: I'll be sure to add this in the next commit.

**Review by sozud — APPROVED** [review]

**Comment by synamaxmusic** [comment]
> MAD is a special case so I don't think it really matters if it's an enum or individual defines. An enum doesn't have to be big though
> 
> ```
> enum Sfx
> {
> WEAPON_STAB_B = 0x630,
> WEAPON_HIT = 0x6DB,
> };
> ```
> 
> etc.

Here's what I got so far for the second enum: 

```
enum MAD_Sfx
{
SFX_MAD_CANDLE_HIT = 0x635,
SFX_MAD_UNK_611 = 0x611,
SFX_MAD_WEAPON_STAB_B = 0x630,
SFX_MAD_WEAPON_HIT = 0x6DB,
SFX_MAD_DOOR_OPEN = 0x640,
SFX_MAD_DOOR_CLOSE = 0x64D,
SFX_MAD_UNK_670 = 0x670,
SFX_MAD_UNK_69D = 0x69D,
SFX_MAD_UNK_672 = 0x672
};
```

PR #1482, opened 2024-08-06
Added MAD and e_red_door SFX

synamaxmusic:
By looking at how the gold pickup sfx are handled using `#ifdef`, I chose to do the same thing here with swapping the sfx references in `e_red_door.h`.

`SFX_METAL_CLANG_E` is the only sound effect that seems to be the same between MAD and the final collision handler.

### DIFF
--- a/include/sfx.h
+++ b/include/sfx.h
@@ -180,16 +180,9 @@ typedef enum { MONO_SOUND, STEREO_SOUND } soundMode;
 #define JP_VO_SH_SONO_TEIDO 0x530 // Shaft: Sono teido no chikara de tatakai...
 #endif
 
-#if defined(VERSION_BETA)
-#define SFX_OPEN_DOOR 0x640
-#define SFX_DOOR_UNKNOWN 0x64D
-#else
-#define SFX_OPEN_DOOR 0x642
 #define SFX_UNK_647 0x647
 #define SFX_UNK_64B 0x64B
 #define SFX_UNK_64C 0x64C
-#define SFX_DOOR_UNKNOWN 0x64F
-#endif
 
 #define SFX_UNK_641 0x641
 #define NA_SE_EN_COG_CLICK 0x642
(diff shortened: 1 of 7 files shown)


include/sfx.h
```diff
@@ -459,4 +452,16 @@ enum Sfx {
     SFX_EXPLODE_SMALL, // Zombie death explosion
 };
 
+// MAD uses an earlier build and has different sfx IDs
+enum MAD_Sfx {
```
sozud, line 456: The way this should be done is
```
#ifdef VERSION_BETA
enum Sfx {
  SFX_WEAPON_STAB_B = 0x630,
...
#else
enum Sfx {
...
#endif
```

This will get if of `#ifdef VERSION_BETA` in several places. The plan is to deduplicate things more in the future so the same functions ought to have the same code

synamaxmusic:
I'm getting a checksum error with the PSP wrp overlay.  I'm not sure what I'm missing here.

synamaxmusic:
Okay, I see what's going on, I'll fix it later tonight.

synamaxmusic:
I fixed the checksum for PSP, but now I got a checksum error for `MAD.BIN`.  Is there a way to look at the generated BIN file to compare with the original?

sozud:
See first_diff.py and vbindiff here: https://github.com/Xeeynamo/sotn-decomp/wiki/Importing-Data

synamaxmusic:
I'm using Codespaces for my dev environment, I don't have any data to extract (other than what I have locally).

sozud:
I don't understand what that means but another method would be reverting all the changes to MAD until the CI is green, then adding them back one by one to figure out the issue

synamaxmusic:
I'll give that a shot.

synamaxmusic:
Okay, sweet.  Looks like `SFX_CANDLE_HIT` is working now.  I'll slowly add back the remaining sfx references.

synamaxmusic:
`mad/collision.c` is definitely the problem child here.  Tried swapping out `stage.h` with `mad.h`, tried defining `VERSION_BETA`, both no go.

At first I was getting checksum errors and then when I started messing around with the header files, I was getting conflicting symbols.  Now we're back to a bad `MAD.BIN` checksum.

synamaxmusic:
In `mad.h` there's a bunch of externs:
```
extern u16 g_testCollEnemyLookup[];
extern u8 g_testCollLuckCutoff[];
extern u8 g_testColluCoords[];
extern u8 g_testCollvCoords[];
extern u16 g_testCollElementLookup[10];
extern u8 g_testColliFrames[10];
extern u16 g_testCollPrizeTable[];
extern u16 g_testCollRandTable;
```
These are already in collision.c, so when I try including `mad.h` instead of `stage.h`, I get conflicts.  I'm wondering if I remove this, will it build properly?

synamaxmusic:
Yes!  That did it!

Xeeynamo (approved):
It is looking good. It is fine having the two `Sfx` from MAD and the retail version separate for now. I am sure we will find a way to merge them down the line.

PR #1483, opened 2024-08-07
Import RIC data & the rest of rodata

sonicdcer:
co-authored-by: @DerpPrincess


src/ric/d_18C40.c
```diff
@@ -0,0 +1,330 @@
...
+    0x0000, 0x0000, 0x0000, 0x015F, 0x0100, 0x0100, 0x7000, 0x0000, 0x0031,
+    0x0200, 0x0000, 0x0000, 0x0000, 0x015F, 0x0200, 0x0800, 0x0800, 0x7007,
+    0x0031, 0x0200, 0x0000, 0x0000, 0x0000, 0x0102, 0x0100, 0x0100, 0x0100,
+    0x0000, 0x0031, 0x0200, 0x0000, 0x0000, 0x0000, 0x015F, 0x00C0, 0x00C0,
+    0x0100, 0x0010, 0x0031, 0x0200, 0x0008, 0x0000, 0x0008, 0x0165, 0x0100,
+    0x0100, 0x0100, 0x7007, 0x0051, 0x0200, 0x0000, 0x0000, 0x0000, 0x0168,
+    0x7100, 0x7100, 0x0200, 0x7007, 0x0031, 0x0200,
+};
+
+Point16 D_8015519C[] = {
```
sozud, line 199: This and  `D_80155244` seem out of sync despite referring to the same data? I'm thinking D_8015519C may be shorter than this and only goes up to before 0x18
```
    loadedParams = &D_8015519C[upperparams >> 8];
    rScale = loadedParams[2];
    gScale = loadedParams[3];
    bScale = loadedParams[4];
    gOffset = loadedParams[0];
    bOffset = loadedParams[1];
```
It would be good to check what the possible values for upperparams are.

src/ric/d_18C40.c
```diff
@@ -0,0 +1,330 @@
...
+    {256, 0},       {0, 0},       {256, 28672}, {28672, 512}, {0, 0},
+    {28928, 28928}, {256, 4096},  {0, 0},       {28928, 256}, {28672, 2048},
+    {0, 0},         {256, 28928}, {256, 1024},  {1280, 2560}, {256, 256},
+    {256, 256},     {1280, 2560}, {256, 512},   {28928, 256}, {-2, -24},
+    {0, -8},        {2, 8},       {0, 24},      {6, -12},     {8, 0},
+    {7, 12},        {-8, -12},    {-9, 0},      {-7, 12},     {-14, -6},
+    {-15, 7},       {13, -7},     {16, 8},      {-7, -22},    {6, -21},
+    {-8, 21},       {7, 22}};
+
+Point16* D_80155244[] = {
```
sozud, line 210: Usage looks like this:
```
        offset = D_80155244[i * 3];
            size = SquareRoot12(
                ((offset->x * offset->x) + (offset->y * offset->y)) << 0xC);

            offset = D_80155244[i * 3 + 1];
            size = SquareRoot12(
                ((offset->x * offset->x) + (offset->y * offset->y)) << 0xC);
       
            offset = D_80155244[i * 3 + 2];
            size = SquareRoot12(
                ((offset->x * offset->x) + (offset->y * offset->y)) << 0xC);
```

Not totally sure how to reconcile this with `D_8015519C`. I'm thinking starting from `&D_8015519C[0x18]` it's not actually a part of D_8015519C and should be separate elements

tools/maspsx
sozud, line 1: Are maspsx changes intended to be here?

src/ric/d_18C40.c
bismurphy: > It would be good to check what the possible values for upperparams are.

I did some analysis, here's my full thinking. Skip to the bottom for my personal conclusion on the size.

The EntityShrinkingPowerUpRing is made by an entity factory.

EntityEntFactory controls its child's params like this:

newEntity->params = self->ext.factory.unkA0

if self->ext.factory.unkA2 != 0{
    newEntity->params += self->ext.factory.unkA6
} else {
    newEntity->params += i
}

First let's look at that second portion. What's unkA2? That comes from the blueprint. We're using RIC blueprints 36 through 39.
unkA2 comes from the start of EntityEntFactory, and appears to be unk2, bit 6. For those blueprints, that value is TRUE. So we're doing += unkA6. That doesn't appear to be set anywhere, so it would always be zero. So all we worry about is the unkA0.
That directly comes from the FACTORY() call. If we use FACTORY(0x400), then the upperparams becomes 0x400.

For calls to FACTORY in RIC, I've searched for blueprints 36 through 39, and it appears that they are only 0, 0x100, 0x200, 0x300, and 0x400.

This indicates that D_8015519C has a length of 5 or more, since we're using elements 0, 1, 2, 3, and 4.

It's possible I missed something somewhere in this analysis, but that's at least my starting point for anyone else to build on.

src/ric/d_18C40.c
```diff
@@ -0,0 +1,330 @@
...
+    0x0001, 0x000A, 0x0001, 0x000B, 0x0001, 0x000C, 0x0001, 0x000D, 0x0001,
+    0x000E, 0x0001, 0x000F, 0x0001, 0x0010, 0x0001, 0x0011, 0x0001, 0x0012,
+    0x0001, 0x0013, 0x0001, 0x0014, 0x0001, 0x0015, 0x0001, 0x0016, 0x0001,
+    0x0017, 0x0001, 0x0018, 0xFFFF, 0x0000,
+};
+
+u8 D_80154CE4[] = {0x02, 0x09, 0x03, 0x0A, 0x01, 0x08, 0x04, 0x0B,
+                   0x00, 0x07, 0x05, 0x0C, 0x06, 0x0D, 0x00, 0x00};
+
+u8 D_80154CF4[] = {
```
sozud, line 25: Seems this is supposed to be length 10
```
            for (i = paramsLo * 2; i < 10; i++) {
                if (g_Player.colliders3[D_80154CF4[i]].effects & 3) {
                    break;
                }
            }
```

src/ric/d_18C40.c
```diff
@@ -0,0 +1,330 @@
...
+
+    0x01, 0x01, 0xFE, 0xFF, 0x33, 0x00, 0x07, 0x00, 0x00, 0x00, 0x80, 0x0C,
+    0x06, 0xC0, 0xC0, 0xC0,
+
+    0x02, 0x02, 0x02, 0x00, 0x7B, 0x00, 0x08, 0x00, 0x00, 0x00, 0x80, 0x08,
+    0x10, 0x7F, 0x7F, 0x7F,
+
+    0x01, 0x01, 0x02, 0x00, 0x33, 0x00, 0x09, 0x00, 0x00, 0x00, 0x80, 0x08};
+
+unkStr_8011E4BC* D_80154DA0[] = {
```
sozud, line 58: I think this one is also wrong, if the first array is size 10, then I think 0xc would be the next offset. So the previous array starting from offset 0xc should be `unkStr_8011E4BC`s

src/ric/319C4.c
```diff
@@ -2106,6 +2106,9 @@ void EntitySubwpnBible(Entity* self) {
     }
 }
 
+const SVECTOR D_80156C50 = {0x0000, 0x0000, 0x0000, 0x0000};
+const SVECTOR D_80156C58 = {0xFFA0, 0x0000, 0x0000, 0x0000};
+
```
hohle, line 2111: `static` and remove from `ric.h`?

src/ric/319C4.c
```diff
@@ -2148,7 +2151,7 @@ void EntityGiantSpinningCross(Entity* self) {
         self->ext.giantcross.unk7E = 0x400;
         g_api.PlaySfx(0x661);
         self->step++;
-        primUVCoords = &D_8015635C[0][0];
+        primUVCoords = &D_8015635C[0];
```
hohle, line 2154: This should be assignable without a reference to the first element - `primUVCoords = D_8015635C`.

src/ric/d_00000.c
```diff
@@ -0,0 +1,17 @@
...
+PlayerOvl RichterBelmont[] = {
+    /* 0x00 */ UpdateEntityRichter,
+    /* 0x04 */ func_80156F40,
+    /* 0x08 */ func_801603C4,
+    /* 0x0C */ func_8015E7B4,
+};
+
+s32* D_8013C010[4] = {
+    &D_8013C010[LEN(D_8013C010)], &D_8013C010[LEN(D_8013C010)],
+    &D_8013C010[LEN(D_8013C010)], &D_8013C010[LEN(D_8013C010)]};
```
hohle, line 17: I believe the symbol `g_richter_spritesheet` resolves to `0x8013C020` (i.e. `&D_8013C010[LEN(D_8013C010)]`) at build time.

src/ric/2A060.c
```diff
@@ -700,26 +700,16 @@ void func_8016779C(Entity* entity) {
...
-     */
     if (g_Player.unk46 != 0) {
         if (entity->step == 0) {
             entity->flags = FLAG_UNK_20000 | FLAG_UNK_40000 |
                             FLAG_UNK_04000000 | FLAG_UNK_10000;
         }
         if (!(entity->params & 0xFF00)) {
-            *(&PLAYER.palette +
-              (*(&D_80155D30 + (entity->animFrameDuration)) * 0x5E)) = 0x8140;
+            g_Entities[D_80155D30[entity->animFrameDuration]].palette = 0x8140;
```
hohle, line 710: `PAL_OVL` macro

config/symbols.us.ric.txt
```diff
@@ -1,2 +1,11 @@
...
 g_RicFactoryBlueprints = 0x80154A6C;
+D_80154CF4 = 0x80154CF4;
+D_80154E5C = 0x80154E5C;
+D_8015519C = 0x8015519C;
+D_80155368 = 0x80155368;
+D_80155964 = 0x80155964;
+D_80155D9C = 0x80155D9C; // type:s32 size:0x40
+D_80155DDC = 0x80155DDC; // type:u16 size:0x60
 UpdateEntityRichter = 0x80157BFC;
+D_80155EFC = 0xD_80155EFC; // type:s16 size:0x180
```
Xeeynamo, line 11:
```suggestion
g_RicFactoryBlueprints = 0x80154A6C;
UpdateEntityRichter = 0x80157BFC;
```
Why are those symbols needed?

src/ric/d_18C40.c
```diff
@@ -0,0 +1,330 @@
...
+    0xFFFD0000,
+    0xFFFDAAAB,
+};
+
+s32 D_80154ED4[] = {
+    0x00010004, 0x00020004, 0x00030004, 0x00040004, 0x00050004,
+    0x00060004, 0x00070004, 0x00080004, 0x00000000,
+};
+
+s32 D_80154EF8[] = {
```
Xeeynamo, line 154:
```suggestion
s32 D_80154EF8[] = {
```
This is an array of `AnimationFrame`

src/ric/d_18C40.c
```diff
@@ -0,0 +1,330 @@
...
+    {0, 0},         {256, 28928}, {256, 1024},  {1280, 2560}, {256, 256},
+    {256, 256},     {1280, 2560}, {256, 512},   {28928, 256}, {-2, -24},
+    {0, -8},        {2, 8},       {0, 24},      {6, -12},     {8, 0},
+    {7, 12},        {-8, -12},    {-9, 0},      {-7, 12},     {-14, -6},
+    {-15, 7},       {13, -7},     {16, 8},      {-7, -22},    {6, -21},
+    {-8, 21},       {7, 22}};
+
+Point16* D_80155244[] = {
+    &D_8015519C[0x18], // + 0x60
+    &D_8015519C[0x1C], // + 0x70
```
Xeeynamo, line 212:
```suggestion
    &D_8015519C[0x1C],
```
any reason to leave those comments? They seem to be redundant with the index value already defined

src/ric/d_19C70.c
```diff
@@ -0,0 +1,114 @@
...
+    0x0002, 0x0002, 0x0002, 0x0003, 0x0002, 0x0004, 0x0000, 0x0000};
+
+s16 D_80155D64[4][6] = {
+    {
+        0x0002,
+        0xFFFE,
+        0x0000,
+        0xFFFC,
+        0x0000,
+        0x0000,
```
Xeeynamo, line 43:
```suggestion
        2, -2, 0, -4, 0, 0,
```
all the values from this array are signed

src/ric/d_19EE0.c
```diff
@@ -0,0 +1,362 @@
...
+s32 D_80155EE0[] = {0x00000F80, 0x00000100, 0x00000700, 0x00000880};
+
+s16 D_80155EF0[] = {0x0400, 0x06AB, 0x0955, 0x0C00, 0x0EAB, 0x1155};
+
+SVECTOR D_80155EFC = {
+    0xFFF8,
+    0xFFB8,
+    0xFFFC,
+    0x0000,
+};
```
Xeeynamo, line 12:
```suggestion
SVECTOR D_80155EFC = { -8, -72, -4 };
```
same for the other `SVECTOR`s

Xeeynamo (commented):
Looks good so far. There are quite a good amount of feedback so far, I hope you do not find it overwhelming. Let us know if you need additional help.

src/ric/d_18C40.c
sonicdcer: Yes, this is definitely wrong, it was a pain to make it work tho, you have to be very careful with the array and struct sizes otherwise the compiler will generate padding for the missing data. I'll take a second look.

src/ric/319C4.c
DerpPrincess: It's good to have an indicator that this is an array at first glance

src/ric/d_00000.c
DerpPrincess: It can't be used because g_richter_spritesheet needs to be defined in symbols.us.ric.txt for the linker to even know which address it corresponds to; which we did, at first, but the linter removes it from the txt file, so we were instructed by Xeeynamo to do the above

src/ric/d_18C40.c
DerpPrincess: Sonic will resolve

src/ric/d_18C40.c
sonicdcer: As you can see, and Sozud pointed out, there's something wrong with that array configuration so i'm leaving those comments there to help resolution.

tools/maspsx
DerpPrincess: I think that was an error on my part... wasn't intended no

src/ric/d_19EE0.c
```diff
@@ -0,0 +1,154 @@
...
+
+SVECTOR D_80155F84 = {24, -32, -4, 0};
+
+SVECTOR D_80155F8C = {-24, -16, -4, 0};
+
+SVECTOR D_80155F94[] = {24, -16, -4, 0, -24, 0, -4, 0};
+
+SVECTOR D_80155FA4 = {-8, 0, -4, 0};
+
+SVECTOR D_80155FAC[] = {8, 0, -4, 0, 24, 0, -4, 0};
```
sozud, line 45: SVECTOR is
```
typedef struct {
    short vx, vy, vz; // Vector coordinates
    short pad;        // System reserved
} SVECTOR;            // size = 0x8
```

The usage looks like
```
    vectors_ptr = &D_8015607C;
    gte_stsxy2(&prim->x0);
    gte_stszotz(&z);
    self->hitboxOffX = prim->x0 - self->posX.i.hi;
    self->hitboxOffY = prim->y0 - self->posY.i.hi;
    for (i = 0; i < 46; i++, prim = prim->next, vectors_ptr += 4) {
        gte_ldv3(vectors_ptr[0], vectors_ptr[1], vectors_ptr[3]);
        gte_rtpt();
        temp_a3 = vectors_ptr[2];
```

image

So I don't think these ones are arrays, probably unused data

sonicdcer: Addressed

sozud (approved):

PR #1487, opened 2024-08-08
Miscelaneous Header Cleanup

hohle:
* forward declaration for `struct Entity` prior to `PfnEntityUpdate`
* definition guard around `VERSION_PC` in `pc.h`
* various libc includes to avoid implicit signatures
* `const` string arguments to avoid dropping a const qualifier

This cleans up most of the warnings from the cmake build. There are still a few from `libgte` and one more `const` to mutable assigment but most have been resolved.

### DIFF
--- a/include/entity.h
+++ b/include/entity.h
@@ -4,6 +4,7 @@
 
 #include "common.h"
 
+struct Entity;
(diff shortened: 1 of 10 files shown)


Xeeynamo (approved):
Looks good!

PR #1488, opened 2024-08-08
More SFX reference updates

synamaxmusic:
I think we're all caught up with the current sfx enum so the next PR will have some new sfx IDs added.


Xeeynamo (commented):
I did review half of the changes, but I stopped at the stage changes. Have you check that those sound effects are consistent across every overlay despite the same SFX ID?

sozud (approved):

sozud:
> I did review half of the changes, but I stopped at the stage changes. Have you check that those sound effects are consistent across every overlay despite the same SFX ID?

Stage sfx are in vabid 3, I think all these are vab 0 so there's no conflict

synamaxmusic:
Yup, once we get to `0x709` then we'll have to start analysis how the sound effects are used in the stages.

Xeeynamo:
Oh wow I did not know. It's crazy how this project grew so big that there are no people that understand every portion of the game.

sozud:
This got conflicted, can you rebase it?

synamaxmusic:
I see what happened, fixing now.

PR #1489, opened 2024-08-08
compile ric

sozud:
This gets RIC compiling, crashes immediately in `func_8015C920` since `g_CurrentEntity` is not set for some reason


Xeeynamo (approved):
Impressive work! I am fine with the `Ric` prefix for now and we can introduce small renaming over time. But so far I'd like to get this in ASAP!

PR #1490, opened 2024-08-08
some ric fixes

sozud:
This fixes a few RIC crashes, gets it up to the point that the blueprints need to be loaded


src/pc/sotn.c
```diff
@@ -204,12 +206,9 @@ bool InitGame(void) {
         return false;
     }
 
-    // TODO different between RIC and ARC
-    if (!FileAsString(
-            InitBlueprintData, "assets/dra/factory_blueprint.json", NULL)) {
-        ERRORF("failed to init blueprint data");
-        return false;
-    }
+    InitBlueprintData("assets/dra/factory_blueprint.json", g_FactoryBlueprints);
```
Xeeynamo, line 209: Why not keep using `FileAsString` and pass either `g_RicFactoryBlueprints` or `g_FactoryBlueprints` as `param` instead of NULL` as a destination? Then instead of doing `FactoryBlueprint* item = &g_FactoryBlueprints[i];` you can do

```c
FactoryBlueprint* array = (FactoryBlueprint*)file->param;
...
FactoryBlueprint* item = array[i];
```

sozud: Ok seems to work

src/ric/d_194B0.c
```diff
@@ -209,5 +209,15 @@ Point32 D_80155AE4[] = {
...
+    2, 5, 2, 0, 7, 2, 0, 2, 0,
+    3, 6, 0, 3, 8, 0, 3, 0, 3,
+    4, 0, 7, 8, 4, 0, 0, 4, 4,
+    5, 5, 5, 0, 0, 5, 1, 2, 0,
+    6, 6, 0, 6, 0, 1, 6, 0, 3,
+    7, 0, 7, 0, 7, 1, 0, 7, 4,
+    8, 0, 0, 8, 8, 4, 3, 7, 8,
+    0, 0, 0, 0, 0, 0, 0, 0, 0
+    // clang-format on
 };
```
Xeeynamo, line 223:
```suggestion
s32 D_80155B2C[][9] = {
    {1, 1, 5, 6, 0, 1, 1, 0, 0},
    {2, 5, 2, 0, 7, 2, 0, 2, 0},
    {3, 6, 0, 3, 8, 0, 3, 0, 3},
    {4, 0, 7, 8, 4, 0, 0, 4, 4},
    {5, 5, 5, 0, 0, 5, 1, 2, 0},
    {6, 6, 0, 6, 0, 1, 6, 0, 3},
    {7, 0, 7, 0, 7, 1, 0, 7, 4},
    {8, 0, 0, 8, 8, 4, 3, 7, 8},
    {0, 0, 0, 0, 0, 0, 0, 0, 0},
};
```

Xeeynamo (approved):
Just a minor suggestion on `D_80155B2C`, but this is very exciting.

PR #1492, opened 2024-08-09
Move Some cen Data to Logical Compile Unit

hohle:
This change moves BSS variables into their most appropriate compile
unit, "decompiles" several data files, and splits out some existing data
into its logical unit.

BSS data has all been removed from `bss.c` and moved to a file where the
data is either first used or where it can be made static.

Several data files have been converted to appropriate types and added
to the most appropriate file. This converts the header related
references and some of the other early data.

Finally, some data like has been rearranged and moved into logical
compile units like data from `entity.c` into `holyglassescutscene.c`.
Logical boundaries were chosen based on where data could be made static.


Xeeynamo (approved):
This is great! I was just looking into this. I think the next step would be to make CEN resemble the same code structure in NZ0 and WRP

PR #1495, opened 2024-08-09
Ric docs, import subweapons

sozud:
I wasn't seeing an advantage to having the subweapons in JSON so I imported it as C.


bismurphy:
The main advantage to having the subweapons in JSON is being able to have the fields named.

For example, if I want to see the number of stun frames that a subweapon has, or see which attackElements are used, the JSON is far more readable.

If I'm researching the Rebound Stone (subweapon 7), then the JSON tells me its fields are:
```
{
            "id": 7,
            "id_hex": "7",
            "ram_addr": "80154714",
            "attack": 40,
            "heartCost": 1,
            "attackElement": [
                "HIT",
                "HOLY"
            ],
            "unk6": 3,
            "nFramesInvincibility": 32,
            "stunFrames": 4,
            "unkA": 0,
            "blueprintNum": 50,
            "hitboxState": 2,
            "hitEffect": 1,
            "crashId": 25,
            "unk11": 0,
            "entityRoomIndex": 0
        },
```
While the C tells me:
`{40, 1, 4128, 3, 32, 4, 0, 50, 2, 1, 25, 0, 0},`

Trying to interpret the second form is much harder and requires cross-referencing the struct, as well as cross-referencing the bit fields for the different attack elements. In contrast, directly reading off HIT and HOLY is much faster to understand.

If there is any way we can preserve the names of the fields in the C version of this, I would strongly request that we do so. The same information was lost when we created config_us.h and lost the names of the fields for all the Equipment. All the research I've done ever since has been much harder since I spend far more time counting indices and comparing numbers to offsets into structs.

sozud:
Ok I put the C version in stubs.c and added the json extraction back

Xeeynamo (approved):
Looks good so far.

For the RIC specific stubs, can we move them into `src/pc/pl_ric.c`?

PR #1496, opened 2024-08-10
Make init recipe

sonicdcer:
for faster cleaning and rebuilding from scratch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,12 +120,22 @@ SOTNASSETS_SOURCES := $(shell find tools/sotn-assets -name '*.go')
 
 CHECK_FILES := $(shell cut -d' ' -f3 config/check.$(VERSION).sha)
 
+N_THREADS ?= $(shell nproc)
+
 .PHONY: build
 
 all: build check
+
+init: 
+	$(MAKE) clean
+	$(MAKE) extract -j $(N_THREADS)
+	$(MAKE) build -j $(N_THREADS)
+	$(MAKE) check
+
 build: build_$(VERSION)
 build_us: main dra weapon ric cen dre mad no3 np3 nz0 sel st0 wrp rwrp tt_000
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -120,12 +120,22 @@ SOTNASSETS_SOURCES := $(shell find tools/sotn-assets -name '*.go')
 
 CHECK_FILES := $(shell cut -d' ' -f3 config/check.$(VERSION).sha)
 
+N_THREADS ?= $(shell nproc)
+
 .PHONY: build
 
 all: build check
+
+init:
```
sozud, line 129: You want it to continue if a command fails? I think you want more like

```
init:
	$(MAKE) clean && \
	$(MAKE) extract -j $(N_THREADS) && \
	$(MAKE) build -j $(N_THREADS) && \
	$(MAKE) check
```

Makefile
```diff
@@ -120,12 +120,22 @@ SOTNASSETS_SOURCES := $(shell find tools/sotn-assets -name '*.go')
...
+
 .PHONY: build
 
 all: build check
+
+init: 
+	$(MAKE) clean
+	$(MAKE) extract -j $(N_THREADS)
+	$(MAKE) build -j $(N_THREADS)
+	$(MAKE) check
```
hohle, line 133: the default target will run `build` and `check` and `check` _should_* have appropriate dependencies to only run after `build` is compete (it depends on all of the files it checks, which in turn depend on the targets to build those files), especially with a freshly cleaned build dir. 

\* this can fail when a checked file is present (previously built), but the target dependency graph isn't complete.

Xeeynamo:
```
# sotn-decomp aliases
alias sotn="make clean && make -j extract && make -j build && make expected"
alias dec="./tools/decompile.py"
alias differ="python3 ./tools/asm-differ/diff.py -mow3 --overlay"
```

This. And we can avoid to pollute the Makefile.

I'd close the PR.

PR #1497, opened 2024-08-10
Stage BSS Organizing, Padding Macros

hohle:
Maps `create_entity` and `e_collect` data, text, and bss for all stages.

Adds macros for defining padding for various sections by size.

Tables for section sizes have been added to common includes to make calculating offsets easier when segments.


hohle:
Will check PSP build failures tomorrow.

Xeeynamo (approved):
The defines at `common.h` almost gave me a jumpscare. But then I saw how you're using `STATIC_PAD_*` and it started to make sense.  This is such a smart idea.

What's the planned use of `UNKNOWN_CRITICAL_PAD_TYPE`, `UNKNOWN_CRITICAL_PAD_FIELD` and `PAD_FIELD`? I usually remove macros, structs and enums that are no longer used in the projects. But I can take note if you plan to use these in specific situations soon.

Is the `Section Sizes` ascii table at the top temporary like notes? I am fine to put them there temporarily. Keep in mind the same information can be retrieved from `build/us/stcen.map` and similar.

This is such a good work and research on the code structure. Great job!

hohle:
> Is the Section Sizes ascii table at the top temporary like notes? I am fine to put them there temporarily. Keep in mind the same information can be retrieved from build/us/stcen.map and similar.

Because the individual stage headers are imported differently across stages, I've found myself having to recompute sizes and offsets for the splat config. As new stages are brought in, I thought having documentation to say how that import will affect offsets would reduce the friction of replacing stage specific code with shared headers. It made it pretty easy to get figure out the offset prior to building and get it right the first time.

> What's the planned use of `UNKNOWN_CRITICAL_PAD_TYPE`, `UNKNOWN_CRITICAL_PAD_FIELD` and `PAD_FIELD`? I usually remove macros, structs and enums that are no longer used in the projects. But I can take note if you plan to use these in specific situations soon.

I can remove them from this commit since it ended up going in a different direction, but the idea was to start replacing pads with macros for code segments, but also within struct fields. the `UNKNOWN` convention was to indicate that it's possible that the pad could be dropped on "new" platforms since it may have been used for alignment that's no longer necessary, but it's not known yet. `CRITICAL` was for things where the padding was necessary, across all versions, even thought it seems like it could be dropped (cases where a field is accessed as a smaller size typically, but then there's the odd one-off code that decides to load it as a larger type). It might be better to make those fields a union portability.

I did try using `__attribute__((aligned(4)))` for cases like BSS chars padded to 32-bit, but I couldn't get GCC to comply. I might be missing something there, though.

Xeeynamo:
I think `__attribute__(aligned(4))` doesn't work for old versions of GCC.

For the unused macros, let's keep them for now. But if we find no use of them let's delete them in the future. I'll merge this as it is :slightly_smiling_face:

PR #1498, opened 2024-08-11
Attempt to fix cross shadow

sozud:
We have a problem where pointers across the various entity structs do not align. This causes the game to crash because it tries to reference the wrong offsets. I don't think this should be the final version of this fix but I'm putting it up for feedback.

This can be tested with this change to give the cross:
```
             // Richter gets a random subweapon.
             if ((g_StageId != STAGE_ST0) && (g_StageId != STAGE_NO3)) {
+                g_Status.subWeapon = SUBWPN_CROSS;
             }
```
The cross spawns a transparent object behind it using the entity factory system. We go from:

EntityCrossBoomerang which creates Entity factory 5, which calls RicCreateEntFactoryFromEntity

The parent pointer there has to align with the children
```
        // the parent pointer must align for anything the factory creates
        entity->ext.factory.parent = source;
```

Then we go through RicEntityEntFactory, which creates EntityCrossShadow.

EntityCrossShadow requires the parent pointer to get unk84 data from the parent, which is used to set the x and y positions of the shadow.

This is currently nonmatching.

### DIFF
--- a/src/pc/sotn.c
+++ b/src/pc/sotn.c
@@ -121,7 +121,7 @@ bool InitGame(void) {
     api.g_pfn_800EA5AC = func_800EA5AC;
     api.func_801027C4 = NULL;
     api.func_800EB758 = NULL;
-    api.CreateEntFactoryFromEntity = NULL;
+    api.CreateEntFactoryFromEntity = CreateEntFactoryFromEntity;
     api.func_80131F68 = func_80131F68;
     api.func_800EDB08 = NULL;
     api.func_80106A28 = func_80106A28;
--- a/src/ric/21250.c
+++ b/src/ric/21250.c
@@ -1328,7 +1328,7 @@ PfnEntityUpdate g_RicEntityTbl[] = {
     EntityCrossBoomerang,
     func_80169C10,
     func_8016147C,
-    func_80169D74,
+    EntityCrossShadow,
     RicEntityHolyWater,
     RicEntityHolyWaterFlame,
     func_80161C2C,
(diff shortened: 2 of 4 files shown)


sozud:
I'm not sure if the rule will hold true everywhere but it seems like CrossBoomerang and Factory and Shadow (and possibly the other sub-entities of this one) ought to have the same layout? This problem exists in other places so it seems like we need a more detailed system other than "each entity gets a unique extension". We need to consider the relationship of child entities to the parents. One thing you can do is this in C11 to check offsets:

```
#include <stddef.h>
#include <assert.h>

struct A {
    int x;
    char y;
};

struct B {
    int z;
    char w;
};

_Static_assert(offsetof(struct A, y) == offsetof(struct B, w), "Offsets are not the same");
```

bismurphy:
Just for clarity, is this PR still waiting on more changes, or is it ready to merge now that it's matching. Only asking since it's still in Draft mode.

sozud:
I have it in draft mode because it seems like we need to decide on a method to handle parent-child relationships between entities before continuing. Pointers are 8 bytes on 64-bit systems so any usage of a pointer is going to make struct layouts not match

bismurphy:
Got it, thanks for that explanation. To me, it seems like that is a wide-reaching change that needs to happen on the repo, and could be its own PR to develop that, but in the meantime, it might be best to merge this improved cross shadow entity, since this is definitely a much better function than what's currently in the repo. I don't think the cross shadow itself should be held up for a big wide change to entity extensions. Analyzing the extensions and figuring out our long term solution should be a big priority moving forward.

bismurphy (approved):

PR #1499, opened 2024-08-11
Fix crashing in RicEntityHolyWaterFlame

sozud:
sp10 isn't large enough to do `i + 8`, tested ingame and seems to look correct


Xeeynamo (approved):
Nice catch. I started looking into it while reviewing your changes and the stack is a bit funny. It does not paint the full picture. `sp10` and `sp20` might have been the same array. Or maybe it was a `u16 sp10[4][4]`. Or even `sp10`, `sp18` and so on with size of four each but they are all functionally equivalent and matching.

TL;DR: good fix

PR #1501, opened 2024-08-12
Decompile `w_005` `EntityWeaponAttack`

hohle:
Decompiles `EntityWeaponAttack` of `w_005`.

Fixes `g_Entities` use in `w_033` to properly use the `PLAYER` macro.


src/weapon/w_005.c
```diff
@@ -9,7 +9,60 @@ extern s32 g_HandId;
...
+        if (g_api.UpdateUnarmedAnim(anim->frameProps, anim->frames) < 0) {
+            DestroyEntity(self);
+            return;
+        }
+    } else {
+        DestroyEntity(self);
+        return;
+    }
+
+    if ((PLAYER.animFrameIdx - 6) < 2U) {
```
sozud, line 58: Is the U required?

sozud (approved):

src/weapon/w_005.c
bismurphy, line 58: These can usually be rewritten as something like `if 5 < PLAYER.animFrameIdx && PLAYER.animFrameIdx < 7`, give something like that a try. Otherwise looks like a good PR, thanks!

bismurphy (approved):

PR #1504, opened 2024-08-13
Decompile and Split NO3 to Match NP3

hohle:
This is a simple synchronization between methods already decompiled in NP3 to their counterparts in NO3. Player water effect has been split to match, but no data has been mapped in this pass.

Decompiles methods based on the NP3 implementation:

* `EntityBackgroundBushes`
* `func_801B94F0`
* `func_801D0A2C`
* `EntityAlucardWaterEffect`
* `EntityMermanSpawner`
* `EntityMermanExplosion`

Splits `e_bone_scimitar`, and `player_water_effect` into a separates file to match `NP3`.


hohle:
There is no new decomp in this PR, just pulling existing implementations and splitting files. I made a few minor changes like using `FIX` instead of a 16-bit hex int but otherwise left the implementations as they were.

sozud (approved):

PR #1510, opened 2024-08-14
Document set of SEL functions for menu graphics and fade-in

bismurphy:
Doing some light research into how SEL works since it has a ton of un-named functions.

This PR represents some of what I've learned. I renamed the functions to descriptive names, and gave them comments describing what they do. I also made a few functions static, where it made sense to do so.

One of them had a do-while in it, I was able to resolve that with the help of PSP. Turns out it was an inverted if-statement.

I'm a little confused about the fact that MainMenuFadeIn doesn't actually return True at any point, it can only return False or exit without a return value. Maybe this is undefined behavior, I'm not sure. Doesn't really matter, just unusual.

I'm also a bit confused at the fact that the PRIM_TILE objects which cover the screen are set to rgb values of 255 (which should be white), but then they are actually rendered as black, and fade away as the rgb values are decremented. This makes it seem like they are using their rgb values just as opacity control, which is interesting. Is that a thing we already know all about? I might take a look at RenderPrimitives and see if I can understand it to see how this works. I imagine it's one of the drawMode flags.


sozud (approved):

Xeeynamo (approved):
Lookin' good

PR #1512, opened 2024-08-14
Compile and enable bat

sozud:


src/servant/tt_000/10E8.c
```diff
@@ -480,10 +482,16 @@ void func_80171ED4(s32 arg0) {
...
+    // i exceeds the size of D_80170448
+    for (i = 0; i < 256 && i < 48; i++) {
+        *dst++ = *src++;
+    }
+#else
     for (i = 0; i < 256; i++) {
         *dst++ = *src++;
     }
-
+#endif
```
hohle, line 494: Instead of reproducing the logic, what do you think about pulling out the upper bound into a macro that's set to 48 for PC, 256 otherwise.

(Odd that this is a inline copy instead of `memcpy`, `memmove`, etc.)

Xeeynamo (approved):
Looking good! I'd follow @hohle suggstion, but other than that I will be happy to have this in!

Maybe in the future what we should do is to put the whole source into a single file, like we're doing with the weapon overlays. Mark everything as static but the OVL export section. It should be easier instead of having to rename the symbols to avoid name collisions

PR #1513, opened 2024-08-14
Document Richter overlay

Xeeynamo:
This is a proposal on how I would like the naming scheme going forward for the player. This includes Alucard and Maria (PSP). Please let's agree on namings and structure so it can be easy to hack and plug new custom characters at any time.

* Animations to have `XXX_anim_yyy` where `XXX` is three letters for `arc`, `ric` or `mar` (`arc` taken from `BIN/ARC_F.BIN`). To mark them `static` whenever it is possible to do so.
* `PL_S_xxx` to mark a player step. Steps might be similar between the three different players, but they must be isolated into the respective headers `dra.h` for Alucard, `ric.h` for Richter, `mar.h` for Maria. I am aware that Alcuard steps are used in weapons, which is why they are currently found in `game.h`. This needs to be addressed at some point.
* PL_T_xxx` for the various timers from `g_Player.D_80072F00`. Those _might_ be need to be shared between all players, we need more research.
* `XxxHandleYyy` for the step handlers. You will see `RicHandleJump` to handle `PL_S_JUMP`.
* `XxxSetYyy` for the functions responsible of transitioning into a specific step (see `RicSetJump`)

Prefer imperative verbs than paste tense or anything else. So `dead` instead of `killed`, `run` instead of `running` and so on.

Functions and animations have `Ric` or `ric_` pre-pended since those symbols are publicly exported and not marked as static. But `#define`s and `enum`s that are isolated into the player's header should be generic.

I cracked the animation system, so I created the macros `A_END`, `A_LOOP_AT` and so on.

In a next pull request I will change `Entity::unk4C` into `Entity::anim`. I did not want to include it here to not change too many files in one go.


sozud (approved):

Xeeynamo:
Thanks for reviewing @sozud . I will wait for a second approval, just so I know we are mostly on-board with my proposal. If there are any objections I'd like to address them now.

bismurphy (approved):

PR #1518, opened 2024-08-15
Near-decompile leather shield EntityWeaponShieldSpell

bismurphy:
This is my attempted decompilation of the leather shield, but the return type mismatch makes it fail.

Scratch here (matching, by using s32 return type): https://decomp.me/scratch/SN7d3


bismurphy:
Found a way to make it compile, we put the Weapon struct at the end of this C file, instead of in weapon_private.h. This means we no longer forward-declare the weapon functions, so their types don't affect anything anywhere.

I like this solution. If we're happy with it, I think this can be a flexible solution for any future weapons with troublesome return type restrictions, or we could even just do this for all weapons moving forward.

sozud (commented):
Seems fine to me but needs approval from Xeeynamo since his solution is different

bismurphy:
update: Putting it at the end wouldn't last long term. This might be okay!

sozud:
This is good to merge now? It's still on draft

bismurphy:
> This is good to merge now? It's still on draft

From my perspective it is ready, but want to wait on xeeynamo's opinion just in case there is an unforeseen issue (like there was with my "put the stuff at the end" attempt)

Xeeynamo (approved):
Let's have this merged for now. It is not too much of an intrusive approach anyway. If we find a more flexible solution later, we will adopt it.

PR #1519, opened 2024-08-15
Decompile Various Weapon Functions

hohle:
* `EntityWeaponAttack` of `w_018`
 * `EntityWeaponAttack` of `w_041`
 * `func_ptr_80170004` of `w_043`

### DIFF
--- a/src/weapon/w_018.c
+++ b/src/weapon/w_018.c
@@ -96,16 +96,16 @@ void EntityWeaponAttack(Entity* self) {
         break;
     // squish and explode
     case 3:
-        self->rotX -= FIX(1.0 / 2048.0);
-        if (self->rotY < FIX(1.0 / 64.0)) {
-            self->rotY += FIX(3.0 / 1024.0);
+        self->rotX -= 32;
+        if (self->rotY < 1024) {
+            self->rotY += 192;
         }
-        if (self->rotX == FIX(1.0 / 1024.0)) {
+        if (self->rotX == 64) {
             self->rotPivotX = 1;
             g_api.PlaySfx(SFX_TELEPORT_BANG_B);
             g_api.CreateEntFactoryFromEntity(self, FACTORY(0xC00, 4), 0);
         }
-        if (self->rotX == FIX(1.0 / 2048.0)) {
+        if (self->rotX == 32) {
             self->palette = PAL_OVL(0x15F);
         }
         if (self->rotX <= 0) {
--- a/src/weapon/w_018.c
+++ b/src/weapon/w_018.c
@@ -32,7 +32,93 @@ static void DebugInputWait(const char* msg) {
         DebugShowWaitInfo(msg);
 }
 
-INCLUDE_ASM("weapon/nonmatchings/w_018", EntityWeaponAttack);
+extern AnimationFrame D_82000_8017A6A8[];
+extern AnimationFrame D_82000_8017A6E4[];
+
+void EntityWeaponAttack(Entity* self) {
+    self->hitboxState = 0;
+    switch (self->step) {
+    // bat appears
+    case 0:
+        self->posX.i.hi = PLAYER.posX.i.hi;
+        self->posY.i.hi = PLAYER.posY.i.hi - 0x38;
+        SetSpriteBank1(g_Animset);
+
+        self->animSet = ANIMSET_OVL(0x10);
+        self->palette = PAL_DRA(0x110);
+        self->unk5A = 0x64;
+        if (g_HandId != 0) {
+            self->animSet += 2;
+            self->palette += 0x18;
+            self->unk5A += 2;
+        }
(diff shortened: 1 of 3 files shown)


sozud (approved):

src/weapon/w_043.c
```diff
@@ -9,17 +9,76 @@ extern s32 g_HandId;
...
-INCLUDE_ASM("weapon/nonmatchings/w_043", func_ptr_80170004);
+s32 func_ptr_80170004(Entity* self) {
+    switch (self->step) {
+    case 0:
+        self->palette = PAL_OVL(0x100);
+        self->zPriority = PLAYER.zPriority;
+        self->facingLeft = PLAYER.facingLeft;
+        // this toggles the high bit, but xor does not
+        // produce the same set of instructions
+        self->animCurFrame = PLAYER.animCurFrame + 0x8000;
```
joshlory, line 24: Looks like this can either be an add or subtract (since it wraps around) but I think it's the same constant as the `ANIMSET_OVL_FLAG`: https://decomp.me/scratch/xyrKw

src/weapon/w_018.c
```diff
@@ -32,7 +32,93 @@ static void DebugInputWait(const char* msg) {
...
+        if (self->rotX == FIX(1.0 / 1024.0)) {
+            self->rotPivotX = 1;
+            g_api.PlaySfx(SFX_TELEPORT_BANG_B);
+            g_api.CreateEntFactoryFromEntity(self, FACTORY(0xC00, 4), 0);
+        }
+        if (self->rotX == FIX(1.0 / 2048.0)) {
+            self->palette = PAL_OVL(0x15F);
+        }
+        if (self->rotX <= 0) {
+            self->rotX = FIX(3.0 / 32768.0);
```
joshlory, line 112: Nit: I don't think rotations are f32.

src/weapon/w_043.c
sozud: 0x8000 seem to indicate that a new animation frame needs to be loaded to vram or something similar https://github.com/Xeeynamo/sotn-decomp/blob/899226c3c305cfa78ab878c2f25fb8e5df1e6234/src/dra/4A538.c#L872

src/weapon/w_018.c
Xeeynamo: I confirm they are not f32, so `FIX` should not be used

src/weapon/w_043.c
Xeeynamo: The `ANIMSET_OVL` should work in this case, I agree

src/weapon/w_043.c
```diff
@@ -9,17 +9,76 @@ extern s32 g_HandId;
...
+        if (PLAYER.animFrameIdx == 5 && PLAYER.animFrameDuration == 1) {
+            self->facingLeft++;
+            self->facingLeft &= 1;
+            g_api.CreateEntFactoryFromEntity(
+                self, ((g_HandId + 1) << 0xC) + FACTORY(0, 62), 0);
+            self->step++;
+        }
+        break;
+    case 2:
+        self->drawMode = 0;
```
joshlory, line 48:
```suggestion
        self->drawMode = DRAW_DEFAULT;
```

src/weapon/w_043.c
hohle: I think using the constant sounds good. I tried a variety of different things to generate matching code because it does seem like a toggle, but this is the only way I found to generate the same assembly. The behavior is a toggle, rather than explicitly setting the value like the macro does, even if that's not what was originally intended.

src/weapon/w_043.c
hohle: Yes!

src/weapon/w_018.c
hohle: Fixed!

src/weapon/w_043.c
```diff
@@ -21,7 +21,7 @@ s32 func_ptr_80170004(Entity* self) {
         self->facingLeft = PLAYER.facingLeft;
         // this toggles the high bit, but xor does not
         // produce the same set of instructions
-        self->animCurFrame = PLAYER.animCurFrame + 0x8000;
+        self->animCurFrame = PLAYER.animCurFrame + ANIMSET_OVL_FLAG;
```
sozud, line 24: I think this has a different meaning than `ANIMSET_OVL_FLAG`, which is for choosing a gfx bank https://github.com/Xeeynamo/sotn-decomp/blob/6c4e63fd1217b61d39bb978ae32f2c11f9f856a9/src/dra/4A538.c#L374

joshlory (approved):
A few rotations are still using the `FIX(...)` macro. Otherwise looks good!

hohle:
Fixed the FIXed rotational values.

PR #1521, opened 2024-08-16
Finish w_045 and import data

joshlory:
Unfortunately this build will fail until the fix is merged to decouple `func_ptr_*` signatures.

* [func_ptr_80170004_w_045](https://decomp.me/scratch/OmJXk)
* [func_ptr_8017000C_w_045](https://decomp.me/scratch/c5Qrm)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -291,12 +291,14 @@ set(SOURCE_FILES_STAGE_WRP
 set(SOURCE_FILES_WEAPON
     src/weapon/w_000.c
     src/weapon/w_002.c
+    src/weapon/w_045.c
     src/weapon/w_046.c
 )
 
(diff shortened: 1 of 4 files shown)


joshlory:
I've added @bismurphy's fix from https://github.com/Xeeynamo/sotn-decomp/pull/1518. If you want to rename the flag or take a different approach, feel free to push changes on top of this PR!

sozud (approved):
Looks good to me but Xeeynamo should approve the approach in https://github.com/Xeeynamo/sotn-decomp/pull/1518 and then we an apply the same approach to this one

Xeeynamo (approved):
Same as #1518: if we start to see different functions requiring the return type changed, instead of adding `#ifdef` everywhere in `weapon_private.h` we just pull the function prototypes out of the header. Until then, this is fine.

PR #1522, opened 2024-08-16
Rename Entity `unk4C` to `anim`

hohle:
`unk4c` is used to store a reference to the entity's current animation. Renaming the field to `anim` match the purpose.


sozud:
I think the other two PRs need changes anyway so if this is rebased we can just merge

Xeeynamo (approved):
:100:

PR #1523, opened 2024-08-16
Document ANIM_FRAME_LOAD

sozud:
I don't think this is the same as ANIMSET_OVL_FLAG, see RenderEntities.


src/weapon/w_053.c
```diff
@@ -92,7 +92,7 @@ static void func_ptr_80170008(Entity* self) {
         self->palette = PAL_OVL(0x100);
         self->zPriority = PLAYER.zPriority;
         self->facingLeft = PLAYER.facingLeft;
-        self->animCurFrame = PLAYER.animCurFrame - ANIMSET_OVL_FLAG;
+        self->animCurFrame = PLAYER.animCurFrame - ANIM_FRAME_LOAD;
```
joshlory, line 95: Might be clearer as `+ ANIM_FRAME_LOAD` based on @hohle's findings. Otherwise looks good!

joshlory (approved):
🎉

PR #1524, opened 2024-08-16
Import data for w_052

joshlory:
Finishes the `.data` section. I thought I could just update sizes and `.bss` would work as well, but I'm running into issues.


sozud (approved):
This looks good for now, we can do the bss later

PR #1529, opened 2024-08-17
Port `EntityMerman` & `EntityMerman` from NP3 to NO3

hohle:
Copied the `EnttityMerman` and `EntityMerman2` implementations from NP3 to NO3. These are identical, however, rely on too many things which require renaming to share at this point.

NO3's `EntityMerman2` was previously `EntityMerman3`, but is identical to NP3's Merman2, so it was renamed to match.

A small amount of cleanup was done to the implementations to bring them closer to current standards.


include/entity.h
```diff
@@ -519,6 +519,7 @@ typedef struct {
     /* 0xA6 */ s16 unkA6;
 } ET_GurkhaSword;
 
+struct draculaPrimitive;
```
sozud, line 522: Is this here by accident?

hohle: I noticed this warning when compiling and get it in my generated context as well. I can move it into a separate PR.

PR #1532, opened 2024-08-17
Improve some TT_000 code quality

bismurphy:
- Rename func_801733D4 to BatFamiliarBlueTrail
- Add PSP symbols for several various variables used in `func_80172120`
- Attempt to get `func_80172120` closer to matching on PSP. Almost got it but not quite working; this is at least an improvement. Scratch: https://decomp.me/scratch/QBytP
- Create `Ext` for BatFamiliarBlueTrail; all it uses is the parent which is the same as Factory, but still better to have its own until we know the real situation for entities.
- General commenting to make flow of functions clearer.


sozud (approved):

include/entity.h
```diff
@@ -488,6 +488,14 @@ typedef struct {
     /* 0xB4 */ struct Entity* follow;
 } ET_Bat;
 
+typedef struct {
+    /* 0x7C */ s32 : 32;
```
Xeeynamo, line 492: I am just realising how this is a much better approach. Nice one!

src/servant/tt_000/10E8.c
```diff
@@ -717,19 +706,19 @@ void func_80172120(Entity* self) {
...
         D_80174B20 = self->ext.bat.target->posY.i.hi;
-        D_80174B0C = func_80173F30(self, (s16)D_80174B1C, (s16)D_80174B20);
+        D_80174B0C = func_80173F30(self, D_80174B1C, D_80174B20);
         D_80174B10 = func_80173F74(D_80174B0C, self->ext.bat.unk86,
-                                   (s16)D_80170658[D_80174C30.level / 10][1]);
+                                   D_80170658[D_80174C30.level / 10][1]);
         self->ext.bat.unk86 = D_80174B10;
-        self->velocityX = rcos(D_80174B10) << 6;
-        self->velocityY = -(rsin(D_80174B10) << 6);
+        self->velocityX = rcos(D_80174B10) << 2 << 4;
```
Xeeynamo, line 720: I wonder if this `<< 2 << 4` is a clue on how the developers used to use rsin/rcos values. Maybe they had a macro or something

Xeeynamo (approved):
Nice work!!

PR #1534, opened 2024-08-18
Enforce ext.*.anim alignment

Xeeynamo:
`#ifdef` ugliness, but there is not much we can do if some of the `ext` structs use 64-bit pointers. With _Static_assert we can ensure, at compile time, that structs are aligned for `g_api.PlayAnimation` to work correctly.

I removed a bit more of `ext.generic` here and there too.

cc. @bismurphy our weapon blacksmith xD

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -37,6 +37,9 @@
 #ifdef VERSION_PC
 #define STATIC_ASSERT _Static_assert
 #define PACKED __attribute__((packed))
+#elif VERSION_PSP
+#define STATIC_ASSERT(x)
+#define PACKED
 #else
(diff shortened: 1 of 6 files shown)


bismurphy (approved):
Seems good to me - I don't really know what the PACKED stuff is doing but I don't see a problem here and it seems to resolve the anim problem which is good

Xeeynamo:
> Seems good to me - I don't really know what the PACKED stuff is doing but I don't see a problem here and it seems to resolve the anim problem which is good

There are very good articles or answered questions online. But in short `__attribute__((packed))` disables a compiler optimisation mechanic where struct fields will not get sorted or padded based on the field size

PR #1535, opened 2024-08-18
Modernize NO3 cavern door code

bismurphy:
Lots of little changes here. Getting away from using `ext.generic`, using various enums/macros, using `prim` instead of `POLY_GT4`, making variables have useful names, etc.

Also, these  functions both match on PSP now.

One interesting thing is that these functions don't overlap in their use of the `ext`, so I had them share one. Kind of interesting.


Xeeynamo (approved):
Exciting times

PR #1536, opened 2024-08-18
Update NZ0 EntityBloodSplatter

bismurphy:
This was an ugly old function with lots of casts and `ext.generic`. I found an identical function in NP3 and copied it out.

I left a comment in the other file about this, for the sake of future potential de-duplication. For now, at least both functions are modern.


Xeeynamo (approved):
gg

PR #1537, opened 2024-08-18
Eliminate remaining uses of ext.generic.unk8C

bismurphy:
My last few PRs have been targeting `ext.generic.unk8C` and removing them one by one. This finishes the job. Next I'll start digging into `unk88` :)


sozud (approved):

Xeeynamo (approved):
Super!

PR #1539, opened 2024-08-18
Finish w_007 and import data

joshlory:
https://decomp.me/scratch/Zn2uO

### DIFF
--- a/src/weapon/w_007.c
+++ b/src/weapon/w_007.c
@@ -241,24 +241,22 @@ void EntityWeaponAttack(Entity* self) {
 
 static void func_ptr_80170004(Entity* self) {
     Entity* parent;
-    s32 temp_v1_3;
-    s32 temp_v1_4;
-    u16 a0;
-    u16 v0;
+    s32 index;
+    u16 offsetX, offsetY;
 
     switch (self->step) {
     case 0:
         parent = self->ext.weapon.parent;
         if (parent->entityId != 0) {
-            temp_v1_3 = (self->params >> 7);
+            index = (self->params >> 7);
             self->facingLeft = parent->facingLeft;
-            a0 = D_35000_8017AC48[temp_v1_3];
-            v0 = D_35000_8017AC48[temp_v1_3 + 1];
+            offsetX = D_35000_8017AC48[index];
+            offsetY = D_35000_8017AC48[index + 1];
             if (self->facingLeft) {
-                a0 = -a0;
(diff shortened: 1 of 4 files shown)


src/weapon/w_007.c
```diff
@@ -67,11 +239,100 @@ void EntityWeaponAttack(Entity* self) {
...
+
+            self->animSet = self->ext.weapon.parent->animSet;
+            self->anim = D_35000_8017AB44;
+            self->zPriority = 0xA8;
+            self->unk5A = self->ext.weapon.parent->unk5A;
+            self->palette = self->ext.weapon.parent->palette;
+
+            SetSpeedX(FIX(4.75));
+            temp_v1_4 = self->params & 0xFF00;
+            if ((temp_v1_4 == 0x300) || (temp_v1_4 == 0x600)) {
```
bismurphy, line 271: Looks like temp_v1_4 is not a real variable, I tried removing it in the scratch and still matched.

src/weapon/w_007.c
```diff
@@ -67,11 +239,100 @@ void EntityWeaponAttack(Entity* self) {
...
+            return;
+        }
+        DestroyEntity(self);
+        return;
+    case 1:
+        if (self->hitFlags) {
+            self->step = 3;
+        }
+        DecelerateX(FIX(7) / 32);
+        DecelerateY(FIX(7) / 32);
```
bismurphy, line 298: This should be FIX(7.0 / 32), or FIX(28.0/128). The idea of FIX is to transform a fixed-point value, so we should include the integer and fractional portions together.

src/weapon/w_007.c
```diff
@@ -67,11 +239,100 @@ void EntityWeaponAttack(Entity* self) {
...
+        self->rotY -= 0xC;
+        if (self->rotX < 0x10) {
+            DestroyEntity(self);
+            return;
+        }
+        // fallthrough
+    case 2:
+        self->velocityY = rsin(self->ext.timer.t);
+        self->posY.val += self->velocityY;
+        self->ext.timer.t += 0x20;
```
bismurphy, line 321: Should use `ext.weapon` in weapon functions

bismurphy (commented):
Looks good! Some comments on various things, these are just style preferences and if you disagree you can ignore them, but generally I think they would be improvements.

sozud (approved):

PR #1541, opened 2024-08-18
Decompile func_ptr_80170010_w_030

joshlory:
https://decomp.me/scratch/TPASg

### DIFF
--- a/src/weapon/w_030.c
+++ b/src/weapon/w_030.c
@@ -9,6 +9,8 @@ extern s32 g_HandId;
 #define g_Animset w_030_1
 #define g_Animset2 w_030_2
 
+extern AnimationFrame D_D6000_8017A514[];
+extern AnimationFrame D_D6000_8017A548[];
 extern WeaponAnimation D_D6000_8017A5E4[];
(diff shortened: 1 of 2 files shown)


src/weapon/w_030.c
```diff
@@ -68,7 +70,100 @@ INCLUDE_ASM("weapon/nonmatchings/w_030", func_ptr_80170008);
...
+        self->rotY = self->rotX;
+        if (self->rotX < 0x40) {
+            self->hitboxState = 0;
+            self->animSet = 0;
+            self->ext.weapon.lifetime = 0x80;
+            self->step++;
+        }
+        /* fallthrough */
+    case 4:
+        if ((--self->ext.weapon.lifetime << 0x10) == 0) {
```
sozud, line 156: if (--self->ext.timer.t == 0) {

sozud: I pasted the wrong thing but you ought to be able to remove the shift

joshlory: Good call, fixed!

PR #1551, opened 2024-08-20
Modernize NO3/NP3 Bone Scimitar

bismurphy:
NO3 and NP3 had a weird mix of Bone Scimitar functions where each was partially complete and missing different parts.

Now both are complete and modernized. Could use another pass to make it PSP-matching, but for now I'll leave this one at this point where the code seems nice.


src/st/no3/e_bone_scimitar.c
```diff
@@ -1,35 +1,276 @@
...
+
+        if (self->animFrameIdx == 7 && self->animFrameDuration == 0) {
+            PlaySfxPositional(SFX_BONE_SWORD_SWISH_B);
+        }
+
+        if (animStatus == 0) {
+            SetStep(BONE_SCIMITAR_WALK_AWAY_FROM_PLAYER);
+            self->ext.boneScimitar.attackTimer =
+                D_80183C18[self->params % 2]
+                          [(++self->ext.boneScimitar.attackTimerIndex) & 3];
```
hohle, line 146: This looks like it was copied from the original, but it's s a bit terse for my taste. The increment of `attackTimerIndex` isn't related to the operation of looking up and assigning the attack timer, so my preference would be that they were separated.

I'd rather get the coverage, however, so if nothing else is blocking feel free to ignore.

hohle (approved):

src/st/no3/e_bone_scimitar.c
bismurphy: Good call, agreed! Updated to split the increment and access into two lines.

sozud:
Looks good, needs a rebase to account for the dedupe PR

bismurphy:
rebase got messed up, fixing

sozud (approved):

PR #1554, opened 2024-08-20
Decompile `EntityWeaponAttack` for `w_054` and `w_056`

hohle:
Similar weapon attack functions.

Thanks to @bismurphy for the help.


bismurphy:
Check out this fork here: https://decomp.me/scratch/GJZHP

This removes the do{}while(0). I discovered this pattern by looking at the PSP version. Turns out there are three cases (i < 7, i & 1, else), but since the first and third have the same result, the PS1 compiler combines the `if`. PSP doesn't do that optimization, which allows us to recover the original structure, and eliminate the do{}while() which happened to change the instruction ordering for PS1.

Bottom line, compilers are weird.

src/weapon/w_054.c
```diff
@@ -9,7 +9,178 @@ extern s32 g_HandId;
...
+        return;
+    }
+
+    if (self->step == 0) {
+        SetSpriteBank1(g_Animset);
+        self->animSet = ANIMSET_OVL(0x10);
+        self->palette = PAL_DRA(0x110);
+        self->unk5A = 0x64;
+        if (g_HandId != 0) {
+            self->animSet = self->animSet + 2;
```
bismurphy, line 54: += 2 here?

sozud (approved):

PR #1558, opened 2024-08-21
Lots of updating on NZ0/311C0.c

bismurphy:
This file was all decompiled, but overall had a lot of "old-looking" traits to it. I went through and updated it in many ways. All the functions are now named with real names (no `func_whatever`), several constants are used in different places, I've added names to the list of entity ID numbers in the .h file, etc.

One thing of interest: I went through starting to document the `.ext` members, and noticed that they seemed to "keep away" from each other, even across unrelated entities. Therefore, at least for now, I decided to create an experimental `ext` which covers all the entities that needed defining in this file. Maybe down the line we can split them up to their own individual entities, but I find it interesting that some entities use the unk80, unk84 offsets, while others only use unk88. I don't see why you would skip to unk88 unless unk80 and unk84 were taken, and the only way they're taken is if they share ext members.

The big thing is that nothing uses generic anymore. It looks like this file was from the era when we were trying to use generic for everything so there is lots of ugly casting, but that's all gone now, which is nice.


sozud (approved):

Xeeynamo (approved):
This is a great initiative, thank you for doing this!

PR #1563, opened 2024-08-22
Decompile ST0 func_801BD8F0

bismurphy:
In addition to decompiling the function, two big changes were needed:

- Decompiling this function uncovered a maspsx bug with generating `nop` padding instructions; that is now fixed and this PR includes a maspsx update.
- We also discovered (thanks @mkst !) that g_Tilemap.bg was not actually a member of g_Tilemap. I renamed it to g_BgLayers, and did a find-and-replace across the repo to change all references. This appears to have no impact on any existing function, but leads to the correct register loading on this function.

It's a weird one, and was tricky to get matching (actually, this was a super old decomp.me browser tab I discovered was still open, which is why I came back to it), but these tricky ones are great for discovering where we have mistakes in our structure of the game's data.


src/st/st0/3D8F0.c
```diff
@@ -1,5 +1,266 @@
...
+        self->ext.GS_Props.pickupFlag = 1;
+        self->step++;
+        return;
+    case 6:
+        SetGeomScreen(0x80);
+        SetGeomOffset(0x80, 0x80);
+        sp38 = (MATRIX*)SPAD(0);
+        sp40 = (SVECTOR*)SPAD(8);
+        sp3c = (VECTOR*)SPAD(10);
+        sp34 = (SVECTOR*)SPAD(14);
```
bismurphy, line 200: Not certain how/if these SPAD calls need to be replaced for PC, would appreciate suggestions.

sozud:
```
typedef struct {
    short m[3][3]; // 3 x 3 matrix coefficient value
    long t[3];     // Parallel transfer volume
} MATRIX;          // size = 0x1E
```
It looks like these overlap and it's all one matrix? You need something like
#ifdef VERSION_PC
MATRIX *sp38;
MATRIX sp38_;
sp38 = &sp38_;
#endif

bismurphy: They don't overlap. But thanks, I think I've made an implementation that should work.

The matrix is 30 bytes long, and the sp40 vector starts 32 bytes deep in the scratchpad, leaving 2 bytes unused. Then the 3 vectors are consecutive.

Xeeynamo (approved):
This is a neat discovery! Appreciate you doing this. Merging~

PR #1566, opened 2024-08-23
Extract w_003-w_010 data and dedupe

joshlory:
Open to ideas on how to better organize the data! The color tables and animations were disjoint across files so it made sense to split them up... I can break things up further for readability/reusability.


config/splat.us.weapon.yaml
```diff
@@ -95,8 +95,7 @@ segments:
     subalign: 4
     subsegments:
       - [0x19040, animset, w_003_1]
-      - [0x19884, animset, w_003_2]
```
sozud, line 98: I think Xeeynamo needs to comment on these since I think this is being handled by the asset manager now?

src/weapon/animations.h
```diff
@@ -0,0 +1,321 @@
+#include "game.h"
+
+#if defined(W_003) || defined(W_007)
```
sozud, line 3: I think all the ifdefs in this file makes it hard to tell what belongs to which weapon. For example the `g_Hitboxes` part is actually nested inside these `#ifs`. I think Xeeynamo needs to comment about the assets tool first, but what if we handled this like the stage overlays and put the individual duplicated parts into .h files and #include those? I think that would be easier than a huge animations.h with a lot of #ifdefs

src/weapon/color_tables.h
```diff
@@ -0,0 +1,153 @@
+#include "game.h"
+
+#if defined(W_002) || defined(W_003) || defined(W_004) || defined(W_005) ||    \
```
sozud, line 3: My comment about animations.h also applies here

config/splat.us.weapon.yaml
joshlory: I took the second animset out for files where it's not used.

src/weapon/animations.h
joshlory: Sure, I can pull them out into separate includes. Do you have a preferred naming convention? Maybe something like what @hohle did with `src/weapon/w_053_056.h`.

sozud: I think `anim_003_007.h` would be fine for example

Xeeynamo (approved):
In theory it is a nice idea. In practice this is coupling data too much across different overlays. The more we apply this pattern to other overlays the more we might be forced to add `#ifdef` for exceptions.
Your initiative revealed how much copy&paste has been done for every weapon it was created. But ultimately each one of them have been crafted and modified ad-hoc. We will need to use the same approach and copy&paste the data across the overlays.

joshlory:
Yeah, I wish we could do `from foo import bar` but alas. I may see if I can still share some of the color table data since it seems to be the most duplicated.

PR #1576, opened 2024-08-27
Decompile `func_ptr_80170004` in `w_001`

hohle:


src/weapon/w_001.c
```diff
@@ -13,7 +13,162 @@ extern s32 g_HandId;
...
+    xOffset = 26;
+    yOffset = -6;
+    if (PLAYER.facingLeft) {
+        xOffset = -xOffset;
+    }
+
+#ifdef VERSION_PSP
+    xOffset = PLAYER.posX.i.hi + xOffset;
+    yOffset = PLAYER.posY.i.hi + yOffset;
+    if (PLAYER.drawFlags & 2) {
```
bismurphy, line 109: I think the flag should be FLAG_DRAW_ROTY like it is in line 115?

src/weapon/w_001.c
```diff
@@ -13,7 +13,162 @@ extern s32 g_HandId;
...
+    s16 x, y;
+    s16 xOffset, yOffset;
+    s16 posX, posY;
+    uvPair* p;
+    s16 animIndex;
+    WeaponAnimation* anim;
+
+    handOffset = 0;
+    if (g_HandId) {
+        handOffset = 1 << 7;
```
bismurphy, line 32: I think this is better as 0x80, since it's being used as a number (moving the V coordinate) rather than as a bit flag (moving a 1 seven positions to the left).

bismurphy (approved):

PR #1577, opened 2024-08-27
RIC blueprint docs

Xeeynamo:
List of the most important changes:
* The two fields in `FACTORY` are now reversed; what used to be `FACTORY(0x2000, 44)` is now `FACTORY(44, 0x20)`
* Introduce `WFACTORY` to imply `((g_HandId + 1) << 0xC)` in it (`((g_HandId + 1) << 0xE)` is not yet cracked)
* Remove `assets/ric/blueprints.json` in favour of baking the info into a C file to easily link enums
* Add `B_MAKE` to define a blueprint entry
* Document almost all RIC entities
* Document almost all RIC blueprints


src/ric/ric.h
```diff
@@ -91,6 +85,160 @@ enum RicTimers {
...
+    E_HIT_BY_HOLY,                   // RicEntityHitByHoly
+    E_CRASH_STOPWATCH_DONE_PARTICLE, // RicEntityCrashStopwatchDoneSparkle
+    E_80170548,                      // func_80170548
+    E_TELEPORT,                      // RicEntityTeleport
+    E_DUMMY_66,                      // RicEntityDummy
+    NUM_ENTITIES,
+};
+
+enum RicBlueprints {
+    BP_BRAKE_SMOKE,
```
sozud, line 161: Is this supposed to be Break? https://en.wikipedia.org/wiki/Brake

sozud (approved):

src/ric/ric.h
Xeeynamo: I am not entirely sure how to say it in English and Brake was within the translation results I found: https://translate.google.it/?sl=it&tl=en&text=frenare&op=translate

What it is meant to represent is when the character is moving at high speed and it suddenly stops, there is the smoke coming from the ground.

bismurphy: Brake works here then! Another option could be "skidding smoke".

bismurphy:
I'm a little unclear about how the new systems work together. What is the process to connect a factory call to the entity being created? For example, if I'm looking at `RicCreateEntFactoryFromEntity(g_CurrentEntity, BP_BRAKE_SMOKE, 0);`, and I want to reference g_RicFactoryBlueprints to see which blueprint is being called, how do I know which blueprint is `BP_BRAKE_SMOKE`? There are a giant list of `B_MAKE` macros, but none of them use that same `BP_BRAKE_SMOKE` string so I can't tell what blueprint is being invoked.

Xeeynamo:
> I'm a little unclear about how the new systems work together. What is the process to connect a factory call to the entity being created? For example, if I'm looking at `RicCreateEntFactoryFromEntity(g_CurrentEntity, BP_BRAKE_SMOKE, 0);`, and I want to reference g_RicFactoryBlueprints to see which blueprint is being called, how do I know which blueprint is `BP_BRAKE_SMOKE`? There are a giant list of `B_MAKE` macros, but none of them use that same `BP_BRAKE_SMOKE` string so I can't tell what blueprint is being invoked.
> 
> Is the answer that it's a multi-step process of counting the index of where BP_BRAKE_SMOKE is in its enum, and then counting that many steps in the blueprints list to find which blueprint lines up? When it's blueprint #0 that's fine, but if I'm looking at a call that makes blueprint #44, that's a bit annoying. It might be nice if we could add comments to the B_MAKE lines which would also include the name of the blueprint. In this case, that might look like `B_MAKE(E_SMOKE_PUFF, 5, 1, true, true, 2, B_DECOR, 0, 0), //BP_BRAKE_SMOKE`.

image

The way I do it is to have the IDE to suggest me which array index the `B_MAKE` corresponds to and have a look at the enum which value is which. This, at least until the enums are well documented. Then it would be possible to know it at a glance that `BP_BLADE_DASH` is the same `B_MAKE` that uses `E_BLADE_DASH`. For the `E_SMOKE_PUFF` it is a bit annoying because it is used by multiple blueprints, so you have to look at the adjacent entities from `B_MAKE` to know which blueprint index is.

I am not sure if there is another easier way to do this.

PR #1583, opened 2024-09-03
Decompile `func_ptr_80170010` of `w_050`

hohle:


src/weapon/w_050.c
```diff
@@ -486,7 +486,122 @@ void func_ptr_8017000C(Entity* self) {
...
+        } else {
+            func_162000_8017BA38(self, &point, false);
+        }
+
+        baseX = point.x;
+        baseY = point.y;
+
+        if (PLAYER.drawFlags & FLAG_DRAW_ROTY) {
+            if ((PLAYER.ext.player.anim - 0x41) == 2 ||
+                (PLAYER.ext.player.anim - 0x41) == 3) {
```
hohle, line 532: This would match PS1 without the subtraction, but only matches PSP with.

src/weapon/w_050.c
```diff
@@ -486,7 +486,122 @@ void func_ptr_8017000C(Entity* self) {
...
+    s32 result;
+
+    if (PLAYER.ext.player.anim < 0x41 || PLAYER.ext.player.anim > 0x47 ||
+        !g_Player.unk46) {
+        DestroyEntity(self);
+        return;
+    }
+
+#ifdef VERSION_PSP
+    params = (self->params & 0x7F00) >> 8;
```
sozud, line 507: psp version matches on ps1 from what I see

src/weapon/w_050.c
sozud: Given it's checking for 0x41 above and returning seems to be deliberate to me and not an optimization

src/weapon/w_050.c
```diff
@@ -86,7 +86,7 @@ void EntityWeaponAttack(Entity* self) {
 
 extern u8 D_162000_8017B000[6][8];
 
-s16 func_162000_8017B784(Primitive* prim, s16 x, s16 y) {
+s32 func_162000_8017B784(Primitive* prim, s16 x, s16 y) {
```
sozud, line 89: The new function showed that the signature of this is wrong?

bismurphy: yep, s16 on this results in an extra sll v0, v0, 16 in the new function. Has to be s32.

hohle: I'm not sure. It's expecting an `s32`, but `0004` is expecting an `s16`. `0004` was happy with a return value cast, but `0010` was not, so moving this back to `s32` (I had just changed it with the `0004` PR yesterday) seemed like the most expedient way to move forward.

src/weapon/w_050.c
hohle: I must have split these when other mismatched caused some divergence. After I had a match on both sides I never went back to check if they still differed.

src/weapon/w_050.c
sozud: I think it's fine for now, we can continue iterating on it

sozud (approved):

PR #1565, opened 2024-08-22
Add AGPL license

Xeeynamo:
From the [public announcement](https://discord.com/channels/1079389589950705684/1079395108501331990/1276310695964835910) on Discord:

> We've been progressing very far with the project and I started wondering what would be the best course of action to protect the work of existing and future contributors. I personally consider contributions to a decompilation project as creative work. Therefore it is my desire to protect this creative work from being re-claimed by other people. This is especially true in the light of the ability of modding the existing decompiled code and create derivative work.
> 
> I talked privately with the current most active contributors. And me, @sozud , @bismurphy and @joshlory so far agreed on pushing through the AGPLv3 license for all the decompiled code in `sotn-decomp/src`, excluding any third party code and decompiled code from the PSX SDK. But I've been wanting to ask all of you if you are fine your code to be licensed and protected under the AGPLv3 terms. 
>
> https://raw.githubusercontent.com/Xeeynamo/sotn-decomp/89686514916cabd0dd88ae0387f749a889c19e05/src/LICENSE
> 
> I put and all the contributors, which can be accessed to the page https://github.com/Xeeynamo/sotn-decomp/graphs/contributors . Again, this will work for existing and future contributors. In the license I put my full name as I will be responsible of any legality. And I understand not every contributor wants to disclose their identity. Please give an approval to my latest PR if you do not have any objection. If not, please write here publicly.
>
> PS. I understand past contributors will either not see the message above or decide to ignore it. If I do not hear from anyone of the contributors for more than one week between those who did not interact with me or the server, I will still go ahead and apply the license. It is not in my intention to ignore past contributors. But I do not want contributors who have been inactive for months or years to be a blocker. Please do not mistake this as an act of malevolence.

PR #1584, opened 2024-09-04
Redact all the lines in src/ from @sonicdcer

Xeeynamo:
```
def redact_author_lines(file_name, author):
    try:
        result = subprocess.run(
            ["git", "blame", "-C", "-C", "-C", "-M", file_name],
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            text=True
        )
        lines = result.stdout.splitlines()
        with open(file_name, 'r') as file:
            original_lines = file.readlines()
        redacted_lines = []
        for i, line in enumerate(lines):
            if author in line:
                redacted_lines.append("REDACTED\n")
            else:
                redacted_lines.append(original_lines[i])
        with open(file_name, 'w') as file:
            file.writelines(redacted_lines)
    except subprocess.CalledProcessError as e:
        print(f"Error running git blame on {file_name}: {e}")
    except Exception as e:
        print(f"Error processing file {file_name}: {e}")

for file_name in file_list:
    redact_author_lines(file_name, "Alejandro")
```

Required before merging #1565 . This only takes account of the files in `src/`. I will create a follow-up PR for `include/`. The build is failing, so we need to push commits to this PR to fix it over time. Some functions might get totally replaced.

The script uses a `git blame -C -C -C` to take account of files that moved via file splits and refactors. This is done programmatically to ensure contributions are accurately removed. I double-checked the results to ensure the script works accurately.


sozud:
My opinion is that this isn't the right approach, we need to go PR by PR and revert each one. If SD originally submitted a function, modifications to that function are a derived work of his. So we can't just do a git blame and remove any line that has his authorship on current master. I'd create a wiki page with a list of his PRs and check each one off as they are reverted. This would also allow an incremental process rather than fixing a monster PR like this

KiritoDv:
I don't wanna create more problems but you can't just remove code from someone since like sozud said, any other code could be a derivate from this work, and could even cause legal problems to you, just saying.

bismurphy:
I am very happy to have the opportunity to close this PR. Lots of good code that will not go to waste.

PR #1585, opened 2024-09-06
EntityStrongWargDeathBeams

Max-Clark:
Thanks again @joshlory @bismurphy for the help!

### DIFF
--- a/src/st/no3/48A84.c
+++ b/src/st/no3/48A84.c
@@ -292,7 +292,7 @@ void EntityStrongWargDeathBeams(Entity* self) {
             while (prim != NULL) {
                 if (prim->drawMode == DRAW_HIDE) {
                     if (self->ext.factory.unk7E & 1) {
-                        func_801CAD28(0x655);
+                        PlaySfxPositional(0x655);
                     }
 
                     if (self->facingLeft != 0) {
--- a/src/st/no3/48A84.c
+++ b/src/st/no3/48A84.c
@@ -296,17 +296,17 @@ void EntityStrongWargDeathBeams(Entity* self) {
                     }
 
                     if (self->facingLeft != 0) {
-                        baseX = self->posX.i.hi -
-                                D_80183080[self->ext.strongWargDeathBeams
-                                               .unk7E &
-                                           0xF];
+                        baseX =
+                            self->posX.i.hi -
+                            D_80183080[self->ext.strongWargDeathBeams.unk7E &
+                                       0xF];
                         prim->x0 = prim->x2 = baseX + 0x10;
                         prim->x1 = prim->x3 = baseX - 0x10;
                     } else {
-                        baseX = self->posX.i.hi +
-                                D_80183080[self->ext.strongWargDeathBeams
-                                               .unk7E &
-                                           0xF];
+                        baseX =
+                            self->posX.i.hi +
+                            D_80183080[self->ext.strongWargDeathBeams.unk7E &
+                                       0xF];
                         prim->x0 = prim->x2 = baseX - 0x10;
                         prim->x1 = prim->x3 = baseX + 0x10;
                     }
@@ -319,8 +319,7 @@ void EntityStrongWargDeathBeams(Entity* self) {
 
                     prim->priority =
                         self->zPriority +
-                        D_801830A0[self->ext.strongWargDeathBeams.unk7E &
-                                   0xF];
+                        D_801830A0[self->ext.strongWargDeathBeams.unk7E & 0xF];
                     prim->drawMode = DRAW_TPAGE2 | DRAW_TPAGE | DRAW_COLORS |
                                      DRAW_UNK02 | DRAW_TRANSP;
                     prim->p1 = (Random() & 3) + 0x10;
--- a/src/st/no3/48A84.c
+++ b/src/st/no3/48A84.c
@@ -285,27 +285,27 @@ void EntityStrongWargDeathBeams(Entity* self) {
         }
         break;
     case 1:
-        if ((self->ext.timer.t == 0) &&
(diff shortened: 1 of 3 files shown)


joshlory:
Congrats on the match! 🎊

src/st/no3/48A84.c
```diff
@@ -246,7 +246,135 @@ INCLUDE_ASM("st/no3/nonmatchings/48A84", EntityStrongWargWaveAttack);
...
+        self->primIndex = primIndex;
+        self->flags |= FLAG_HAS_PRIMS;
+        prim = &g_PrimBuf[primIndex];
+
+        while (prim != NULL) {
+            prim->tpage = temp_s1_u16 / 4;
+            prim->clut = palette;
+            prim->u0 = prim->u1 = (temp_s1 & 1) << 7 | 0x21;
+            prim->v1 = prim->v3 = (temp_s1 & 2) << 6 | 0x59;
+            prim->v0 = prim->v2 = (temp_s1 & 2) << 6 | 0x7F;
```
bismurphy, line 282: Do these still match if you use + instead of | at the end? Since UV coordinates are numerical values, it probably makes more sense to add a physical offset than to bitwise OR it.

src/st/no3/48A84.c
```diff
@@ -246,7 +246,135 @@ INCLUDE_ASM("st/no3/nonmatchings/48A84", EntityStrongWargWaveAttack);
...
+        }
+        break;
+    case 1:
+        if ((self->ext.timer.t == 0) && (self->ext.factory.unk7E < 0x14)) {
+            prim = &g_PrimBuf[self->primIndex];
+
+            while (prim != NULL) {
+                if (prim->drawMode == DRAW_HIDE) {
+                    if (self->ext.factory.unk7E & 1) {
+                        PlaySfxPositional(0x655);
```
bismurphy, line 295: For SFX calls, we usually use the enums in `include/sfx.h`. In this case, this should be SFX_EXPLODE_B.

src/st/no3/48A84.c
```diff
@@ -246,7 +246,135 @@ INCLUDE_ASM("st/no3/nonmatchings/48A84", EntityStrongWargWaveAttack);
...
+                    prim->b1 = prim->b3 = prim->g1 = prim->g3 = prim->r1 =
+                        prim->r3 = prim->b0 = prim->b2 = prim->g0 = prim->g2 =
+                            prim->r0 = prim->r2 = 0x40;
+
+                    priorityOffset = D_801830A0[self->ext.factory.unk7E & 0xF];
+
+                    prim->priority = self->zPriority + priorityOffset;
+                    prim->drawMode = DRAW_TPAGE2 | DRAW_TPAGE | DRAW_COLORS |
+                                     DRAW_UNK02 | DRAW_TRANSP;
+                    prim->p1 = (Random(priorityOffset) & 3) + 0x10;
```
bismurphy, line 321: I think this Random call shouldn't have an argument. The compiler is just reusing the `a0` register, I expect. Try removing `priorityOffset` from this line and see if it still matches.

src/st/no3/48A84.c
```diff
@@ -246,7 +246,135 @@ INCLUDE_ASM("st/no3/nonmatchings/48A84", EntityStrongWargWaveAttack);
...
+                    prim->p2 = 0;
+                    break;
+                }
+
+                prim = prim->next;
+            }
+
+            self->ext.timer.t = 4;
+            self->ext.factory.unk7E++;
```
bismurphy, line 330: You'll probably want to make a new entry in the `Ext` union in `include/entity.h` for this entity. Feel free to ask in discord if you don't know how to do this.

src/st/no3/48A84.c
```diff
@@ -246,7 +246,135 @@ INCLUDE_ASM("st/no3/nonmatchings/48A84", EntityStrongWargWaveAttack);
...
+                } else {
+                    prim->r0 = prim->r2 = prim->r1 = prim->r3 = prim->r3 + 0x10;
+                    prim->g0 = prim->g2 = prim->g1 = prim->g3 = prim->g3 + 0x10;
+                    prim->b0 = prim->b2 = prim->b1 = prim->b3 = prim->b3 + 0x10;
+                }
+
+                prim->y0 = prim->y1 = prim->y1 - prim->p1;
+
+                if (prim->p2 > 0x10) {
+                    prim->drawMode = 8;
```
bismurphy, line 363: drawMode uses enums, in this case 8 should be DRAW_HIDE.

bismurphy:
Looking like a fantastic first PR, well done! Some cleanup comments, but nothing huge. Looking forward to merging this one!

src/st/no3/48A84.c
Max-Clark: I took a shot at it in [a33aa1c](https://github.com/Xeeynamo/sotn-decomp/pull/1585/commits/a33aa1c474971df6f0970bf378fb056006e0eb80), LMK if it needs changed and I'm happy to do so

bismurphy: Looks good, but I'm not sure why you named the variable zPriority. It looks like it's controlling multiple things, including the zPriority. I would recommend keeping it as unk7E unless we know what it is.

Also, I would recommend adding an unk7C, and using that instead of ext.timer.t. When possible, we want an entity to exclusively use its own Ext member, instead of mixing around `self->ext.this` and `self->ext.that`.

sozud (approved):

Xeeynamo:
Congrats @Max-Clark for your first pull request to the project :clap: .

We are currently transitioning to apply a license to the project. Before merging this pull request we will need you to accept the license here: https://github.com/Xeeynamo/sotn-decomp/pull/1565

PR #1587, opened 2024-09-07
Draw Mode Linter

hohle:
Adds a transformer for `drawMode` fields which converts hex or integer values to an OR'd bitmask.

Like `RelicTransformer`, `DrawModeTransformer` reproduces the values and names in the linter.


sozud:
Great work, I think we should able to just run the regular linting after this and you don't have to worry about line breaks?

sozud (approved):

Xeeynamo (approved):
This is great!! I love to see more things being automated

PR #1588, opened 2024-09-07
RIC and Alucard share field docs

Xeeynamo:
Not a big PR. The highlights are `g_Player.timers` and `entity->hitParams`. The last one is a bit weird as it looks like a mix between `attackElement` and `hitEffect`.

The `var_s4` on `RicMain` is uninitialised and it leads to an undefined behaviour. Recompiling the code for PSX by moving things around, that variable will not be `0` at the beginning but some random value. This prevented Richter's run, blade dash and high jump to work. I added an `#ifdef` in case people wants to mod the overlay.

This is probably the last batch of changes for RIC for a while.


sozud (approved):
Nice find

PR #1589, opened 2024-09-07
Enchanced Draw Mode Lint

hohle:
A few updates that didn't make it into the first round:

* masks are now identified and expanded
* AND-EQUALS and OR-EQUALS are supported
* `sotn-lint` is run before `clang-format` to avoid having to run the `format-src` target twice to catch line breaks
* `.h` files are included in `sotn-lint` since they are being used for shared code
* `sotn` lint ignores the `mednafen` directory

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,6 +138,7 @@ clean:
 format: format-src format-tools format-symbols format-license
 
 format-src: bin/clang-format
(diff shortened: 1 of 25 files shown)


bismurphy (approved):
Very cool! It's awesome to see all the lines that this improves. Nice work!

sozud (approved):
Looks good!

PR #1593, opened 2024-09-10
`sotn-lint` support for `drawFlags`, and `flags`

hohle:
`sotn-lint` will now convert `flags` and `drawFlags` fields from integers to enum values.

`sotn-lint` will no longer overwrite files that haven't changed, so running `make format build` only has to rebuild files that are modified.

The logic from `DrawModeTransformer` has been pulled out into `BitFlagLineTransformer` which will handle any type of bitmasked enum type.


sozud (approved):
Really awesome work! Will leave this one up for a bit in case anyone has an issue with the game.h changes

Xeeynamo (changes requested):
The only issue I see is the flags between `FLAG_UNK_1` to `FLAG_UNK_8`. The first 4 bits of `Entity::flags` are not flags but a timer!
In `HitDetection` you can see it being set as `->flags |= 0xF;`. Then in `Update` you can see a `if (!(--e->flags & 0xF))`. We should unfortunately revert those four flags before having this merged in.

Everything else is very well welcomed. I've been wanting to do this for a while but I never had the patience of learning Rust. This is a great initiative!

hohle:
Reopening. The change was easier than I had anticipated. If enum values are not configured for a particular bits a field is being set to, it will be ignored.

There is a heuristic to determine whether a collection of OR'd bits should be inverted or not (are half or more bits set) which may break based on that invariant, but for the common case where a mask consists of 1-3 ignored bits, it works correctly. This can be adjusted if need arises.

Xeeynamo (approved):

PR #1594, opened 2024-09-10
Decompile `func_ptr_80170004` of `w_048`

hohle:


src/weapon/w_048.c
```diff
@@ -116,7 +116,165 @@ void func_154000_8017B810(Entity* ent, Point16* outPoint, bool arg2) {
...
+            func_154000_8017B810(self, &point, true);
+        }
+
+        newXPos = self->posX.i.hi + point.x;
+        newYPos = self->posY.i.hi + point.y;
+    }
+
+    self->facingLeft = PLAYER.facingLeft;
+
+    if (PLAYER.drawFlags & 2) {
```
sozud, line 165: drawFlags enum?

sozud (approved):
Seems good to me, not sure about the mixing of different `ext`s though. I see we have various styles for weapons, some use a mix and others make their own ext like this one. Note that this one still mixes them despite having it's own ext https://github.com/Xeeynamo/sotn-decomp/blob/master/src/weapon/w_006.c

src/weapon/w_048.c
hohle: Arg. I missed the `ext`s. Will fix.

PR #1595, opened 2024-09-10
Decompile `EntityWeaponAttack` of `w_044`

hohle:


include/entity.h
```diff
@@ -240,6 +240,29 @@ typedef struct PACKED {
     /* 0xAE */ s16 equipId;
 } ET_WeaponUnk030;
 
+typedef struct {
+    /* 0x7C */    s16 lifetime;
+    /* 0x7E */    s16 unk7E;
+    /* 0x80 */    s16 unk80;
+    /* 0x82 */    s16 unk82;
+    /* 0x84 */    struct Entity* some_ent;
```
sozud, line 248: Could you add an assert that `anim` is in the right place on 64-bit systems?

src/weapon/w_044.c
```diff
@@ -110,7 +110,229 @@ static s32 g_HandId = HAND_ID;
...
+            }
+
+            if (i == 0xF) { // ???
+                prim->u1--;
+                prim->u3--;
+            }
+
+            self->ext.weapon_044.unk82 += 8;
+            self->posY.i.hi -= 8;
+            self->ext.timer.t = 32;
```
sozud, line 231: mixing exts

src/weapon/w_044.c
```diff
@@ -110,7 +110,229 @@ static s32 g_HandId = HAND_ID;
...
+    if (g_HandId != 0) {
+        clut += 0x18;
+        vCoord += 0x80;
+    }
+
+    doLastblock = false;
+    switch (self->step) {
+    case 0:
+        self->primIndex = g_api.AllocPrimitives(
+            PRIM_GT4, 17); // bug? loops iterate over 16 prims
```
Xeeynamo, line 141: Not a bug, it iterate over 16 primitives yes. But then after the for loop it operates on the last prim

include/entity.h
```diff
@@ -240,6 +240,29 @@ typedef struct PACKED {
     /* 0xAE */ s16 equipId;
 } ET_WeaponUnk030;
 
+typedef struct {
+    /* 0x7C */ s16 lifetime;
+    /* 0x7E */ s16 unk7E;
+    /* 0x80 */ s16 unk80;
+    /* 0x82 */ s16 unk82;
+    /* 0x84 */ struct Entity* some_ent;
```
sozud, line 248: This needs a
```
#ifdef VERSION_PC
u32 padding;
#endif
To fix this since pointers are 8 bytes on 64-bit platforms
```

PR #1596, opened 2024-09-11
Decompile func_ptr_80170008 of w_016

Max-Clark:


include/entity.h
```diff
@@ -1906,6 +1925,7 @@ typedef union { // offset=0x7C
 SYNC_FIELD(ET_Player, ET_Weapon, anim);
 SYNC_FIELD(ET_Player, ET_WeaponUnk006, anim);
 SYNC_FIELD(ET_Player, ET_WeaponUnk012, anim);
+SYNC_FIELD(ET_Player, ET_WeaponUnk016, anim);
```
Max-Clark, line 1928: I'm not 100% sure this is necessary, just following the pattern

Max-Clark: It is in fact, not ;)

include/entity.h
```diff
@@ -353,6 +353,24 @@ typedef struct {
...
 } ET_WeaponUnk014;
 
+typedef struct {
+    s16 unk7C;
+    s16 unk7E;
+    s32 unk80;
+    s16 unk84;
+    s32 : 32;
+    s16 : 16;
+    struct Entity* parent;
```
sozud, line 363: Adding 
```
#ifdef VERSION_PC
u32 padding;
#endif
```
after this point and adding back the assert should fix the issue you were having

Max-Clark: Hmmm... I can't get it to work. Don't want to blow all the pipeline time so going to play around on my build machine

src/weapon/w_016.c
```diff
@@ -133,7 +133,94 @@ s32 func_ptr_80170004(Entity* self) {
...
+            self->animSet += 2;
+        }
+        self->animCurFrame = 0xF;
+        self->facingLeft = PLAYER.facingLeft;
+        self->zPriority = PLAYER.zPriority - 2;
+        self->flags = FLAG_POS_CAMERA_LOCKED | FLAG_KEEP_ALIVE_OFFCAMERA;
+        self->drawFlags = DRAW_COLORS;
+
+        self->posY.i.hi = PLAYER.posY.i.hi + PLAYER.hitboxOffY;
+        if (PLAYER.step != 2) {
```
sozud, line 162: Player_Crouch

sozud (approved):
Nice work

PR #1598, opened 2024-09-12
Decompile `SetVertex` functions in `reg03.c`

JoshSchreuder:
Based this off other Josh's work on https://decomp.me/scratch/ncrbI

Noticed there was already a define in `libgte.h` for Vector0 which I've used, but also filled out the other two.

Hope this is ok for a first PR 😄 
Let me know if anything needs changing.

### DIFF
--- a/include/psxsdk/libgte.h
+++ b/include/psxsdk/libgte.h
@@ -105,6 +105,16 @@ extern long ratan2(long y, long x);
                      "lwc2	$1, 4( %0 )"                                        \
                      :                                                         \
                      : "r"(r0))
+#define gte_ldv1(r0)                                                           \
+    __asm__ volatile("lwc2	$2, 0( %0 );"                                       \
+                     "lwc2	$3, 4( %0 )"                                        \
+                     :                                                         \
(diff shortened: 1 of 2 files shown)


sozud (approved):
Looks good, thanks for your contribution!

JoshSchreuder:
Sorry one thing I just noticed @sozud, do I need to add the new defines to the else block, like so:
https://github.com/JoshSchreuder/sotn-decomp/blob/0d2ed10741284b73b2d0d102774b99c41dc8e65c/include/psxsdk/libgte.h#L209

I'm not 100% clear on what the `#if !defined(VERSION_PC) && !defined(M2CTX) && !defined(PERMUTER)` part of this controls, so not sure what cases would fall under the `else`.

sozud:
As far as I see, there's no functions currently decompiled that use any of these, so you don't need to worry about it. The point of the defines is avoiding including mips assembly in places where it doesn't work. For example like when using the https://github.com/simonlindholm/decomp-permuter project. We can add them if/when it becomes an issue

PR #1602, opened 2024-09-15
Decompile `no3` `EntityUnkId30`

JoshSchreuder:
https://decomp.me/scratch/qg2By


src/st/no3/48A84.c
```diff
@@ -1,506 +1,561 @@
...
-    }
-    entity->posY.val += entity->velocityY;
-    if (AnimateEntity(D_8018383C, entity) == 0) {
-        DestroyEntity(entity);
-    }
-}
-
-#include "../popup.h"
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "no3.h"
```
sozud, line 2: Something is wrong with your whitespace and it's messing up the diff, do you have windows newlines or something?

src/st/no3/48A84.c
```diff
@@ -1,506 +1,561 @@
...
+                    animCurFrame = 13;
+                }
+            } else {
+                if (animCurFrame >= 96) {
+                    animCurFrame -= 48;
+                } else {
+                    animCurFrame -= 1;
+                }
+            }
+            hitboxPtr = &D_80182A4C[animCurFrame * 16 >> 1];
```
sozud, line 229: animCurFrame * 8, same for other usage

src/st/no3/48A84.c
JoshSchreuder: Must be that yeah, I thought `make format` would have fixed that but I guess not, I'll update.

sozud (approved):

src/st/no3/48A84.c
```diff
@@ -179,7 +179,62 @@ void func_801CC90C(Entity* arg0) {
...
-INCLUDE_ASM("st/no3/nonmatchings/48A84", EntityUnkId30);
+void EntityUnkId30(Entity* self) {
+    Entity* entity;
+    u16 animCurFrame;
+    u16* hitboxPtr;
+
+    entity = self - 1;
+    if (self->step == 0) {
+        if (self->params != 0) {
+            InitializeEntity(&D_80180B30);
```
Xeeynamo, line 190:
```suggestion
            InitializeEntity(D_80180B30);
```
`D_80180B30` is already an array, you do not need to get the address with `&`. Same for the other `InitializeEntity` parameters

JoshSchreuder: Thanks, I originally had these not as arrays in the scratch, but then made them arrays in the .h for consistency and forgot to update this.

PR #1603, opened 2024-09-15
Add tool to generate splat configs

Xeeynamo:
As promised, a new tool to automatically generate a splat config for PSP and PSX overlays for multiple game versions. This first version will perform the following tasks:

* Create the splat config with data, rodata, text and bss sections to their right offsets
* Generate a symbol list by cross-referencing duplicate symbols
* Split the overlay by creating the correspondent C file

This tool is not currently:

* Splitting C files
* Importing shared code already decompiled

There is a lot of tech involved, especially to detect the various sections for the PSX overlays. But I will leave it to you to discover how the implementation work if you're interested. Hopefully the code is self-documented enough.

Usage examples:

* `python3 tools/make-config.py mar`
* `VERSION=jp12 python3 tools/make-config.py no2`
* `python3 tools/make-config.py tt_001 --version pspko`

With this tool I aim to add in the project multiple game versions of TT_001, MAR and RBO3 from #557 and NO0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -386,7 +386,6 @@ force_symbols:
 	$(PYTHON) ./tools/symbols.py map build/us/ric.map --no-default > config/symbols.us.ric.txt
 	$(PYTHON) ./tools/symbols.py map build/us/stcen.map --no-default > config/symbols.us.stcen.txt
 	$(PYTHON) ./tools/symbols.py map build/us/stdre.map --no-default > config/symbols.us.stdre.txt
-	$(PYTHON) ./tools/symbols.py map build/us/stmad.map --no-default > config/symbols.us.stmad.txt
 	$(PYTHON) ./tools/symbols.py map build/us/stno3.map --no-default > config/symbols.us.stno3.txt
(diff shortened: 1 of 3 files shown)


sozud (approved):
I ran this on MAR to test and the output looks good to me. Python is super slow so I think a different language for duplicate detection would have made sense but that's not a blocker. Nice work

Makefile
```diff
@@ -386,7 +386,6 @@ force_symbols:
 	$(PYTHON) ./tools/symbols.py map build/us/ric.map --no-default > config/symbols.us.ric.txt
 	$(PYTHON) ./tools/symbols.py map build/us/stcen.map --no-default > config/symbols.us.stcen.txt
 	$(PYTHON) ./tools/symbols.py map build/us/stdre.map --no-default > config/symbols.us.stdre.txt
-	$(PYTHON) ./tools/symbols.py map build/us/stmad.map --no-default > config/symbols.us.stmad.txt
```
sozud, line 389: Is this intentionally removed?

Xeeynamo: Yep! We do not have `us` for MAD. The only "version" that exists for MAD is `beta`. I put it there by mistake long time ago

PR #1604, opened 2024-09-15
Decompile MAD func_80192408

bismurphy:
Derived from a scratch by @hohle :)

Got it down to one FAKE line, which is sad but I'm okay with, especially with how weird and difficult MAD functions seem to be.


Xeeynamo (approved):
Looks clean! It is a bit unfortunate for those two fakes

PR #1605, opened 2024-09-15
Decompile ST0 GetPlayerCollisionWith

bismurphy:
ST0 has a different version of this function, which was not yet decompiled. Other people had some scratches, but they weren't very close.

I took the function that exists in other overlays, and removed code until it matched.

Might try to de-duplicate this with ST0 ifdef stuff, but for now, I wanted to just get the single decompiled function reviewed.


Xeeynamo (approved):
I agree to have this separated for the time being. Great stuff!

PR #1606, opened 2024-09-15
Deduplicate st_common.c across all existing stages

bismurphy:
I noticed that there was large overlap between these, but it was incomplete as different stages had different levels of de-duplication completed.

I went through and completely de-duplicated this file, so it is now st_common.h everywhere.

There were a few leftover functions in MAD and ST0, so I decompiled those in my previous PRs, and now in this one, I'm doing ifdef to make it all match for all stages.

This is pretty cool and represents a large amount of code reduction.

The original C file which was used as the basis for st_common.h was the WRP one, which has PSP support, so this may be a great step to making many overlays match on PSP too.


src/st/st_common.h
```diff
@@ -0,0 +1,221 @@
...
+
+#if !defined(VERSION_PSP)
+#include "../destroy_entities_from_index.h"
+#endif
+
+#include "prevent_entity_from_respawning.h"
+
+#include "animate_entity.h"
+
+#if !defined(VERSION_PSP)
```
sozud, line 59: Once we add more overlays I think we may to run into trouble on PSP. I believe these functions are getting removed due to dead code elimination. Once we add more overlays that use these functions, there's going to be overlays with different mixes of functions in this file. I suppose you could handle it with #ifdefs for each stage. We don't have to solve this now but figured I'd point it out

bismurphy: Understood. Maybe we need to add a "dead code eliminator" to the PSP build chain? That could be the long term solution.

Xeeynamo: I bet it can be automated with a compiler flag. Let's leave it like this and think about it later.

Xeeynamo (approved):
Great stuff! Such a needed change. I do not even remember why this wasn't done in the first place.

PR #1607, opened 2024-09-15
Room Teleport Table Labels and Comments

MottZilla:
I added some of the correct labels to the unknown field on the RoomTeleports. The unknown field is used by the CD Rooms to reload the tileset if you move half way through the room but then decide to turn back. When not used in a CD Room I don't know if it has any effect. But some warps that aren't used in a CD Room do assign this field.


include/game.h
```diff
@@ -712,7 +712,8 @@ typedef struct {
     /* 0x0 */ u16 x;
     /* 0x2 */ u16 y;
     /* 0x4 */ u16 roomId;
-    /* 0x6 */ u16 unk6;
+    /* 0x6 */ u16 unk6; // Current Stage ID to reload Tile GFX if you move
```
sozud, line 715: I think change these to currentStageId and nextStageId? I'd get a second opinion on the names before committing though.

sozud (approved):

include/game.h
MottZilla: I think those are good names. I did not change them in my PR because then everywhere they are referenced would need changing and I wasn't wanting to name the variable on my own.

PR #1608, opened 2024-09-15
Add BOSS/MAR.BIN

Xeeynamo:
Small PR to enable #557 and to let be used as an example for [our wiki](https://github.com/Xeeynamo/sotn-decomp/wiki/Decompilation#add-new-overlay).

This was done with the new `make-config.py` tool.


src/boss/mar/us.c
```diff
@@ -0,0 +1,284 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include <stage.h>
+
+INCLUDE_ASM("boss/mar/nonmatchings/us", func_us_8018AC0C);
```
bismurphy, line 4: I notice the func_placeholder names now have _us_ inserted in them. Is that an intentional change now that will apply to future overlays? Might have just been something I missed in one of the previous PRs.

sozud (approved):
Looks good. Just want to point out to those watching that most of the non-duplicate functions here are already decompiled in https://github.com/Xeeynamo/sotn-decomp/pull/557/files#diff-d5114e378d24d38176bafbb87a89204688e5f27d0b6291224a04b0a98770e629

src/boss/mar/us.c
Xeeynamo: Yep it is intentional

bismurphy (approved):

sozud:
It doesn't have to be part of this PR but just figured I'd add that this should be added to the dups tool eventually https://github.com/Xeeynamo/sotn-decomp/blob/908bd1d2eb64025bf93ddcb688cbf645efc3dc67/tools/dups/src/main.rs#L240

Xeeynamo:
> It doesn't have to be part of this PR but just figured I'd add that this should be added to the dups tool eventually
> 
> https://github.com/Xeeynamo/sotn-decomp/blob/908bd1d2eb64025bf93ddcb688cbf645efc3dc67/tools/dups/src/main.rs#L240

I was actually looking into that. I was thinking it would be a good idea to read all the `config/splat.us.*.yaml` files and use the `src_path` and `asm_path` from there. It would also be cool to leverage the `--disassemble-all` flag and use the `matchings/` folder on top of the `nonmatchings/` to finally deprecate the hacky `make force_extract`. But it is too big for this PR and I need to learn to write Rust code for that.

PR #1609, opened 2024-09-15
Deduplicate st_update.c across overlays

bismurphy:
Renames the file, imports the data, and uses the .h in every overlay.

Given that this one doesn't have any PSP craziness, would this maybe be a place where it would make sense to copy in the function C code? Right now st_update.h is just an array definition and a few #include calls for the .h files for `random.h`, `update.h`, and `update_stage_entities.h`.


Xeeynamo (approved):
Very well welcomed change! I did not realise the data was all similar but in MAD.

Also these batch of changes from you are somewhat coming now that I am about to de-duplicate BOSS/MAR. A good coincidence!

PR #1619, opened 2024-09-16
MAR assets + CutsceneRun

Xeeynamo:
PSX: https://decomp.me/scratch/RL97B
PSP: https://decomp.me/scratch/3WKmX

`func_us_8018B4A0` needed to be decompiled before importing all the assets due to GCC compiling `PfnEntityUpdates[entity->entityId - 1]` into a hardcoded offset landing to the `header.c` data as a fake symbol.

The decompiled function seem to be some kind of object spawner for the cutscenes controller. I have yet to confirm that, so I did not rename any symbols yet.

EDIT: `func_us_8018B4A0` turned out to be a shared function I renamed as `CutsceneRun`. DRE and CEN use slightly different variants and they will not use the new `cutscene.h`. I plan on putting more stuff into `cutscene.h` in a separate PR and maybe rename all the C files that handle cutscenes accordingly.


src/boss/mar/AFC4.c
```diff
@@ -13,7 +13,58 @@ INCLUDE_ASM("boss/mar/nonmatchings/AFC4", func_us_8018B2F0);
...
-INCLUDE_ASM("boss/mar/nonmatchings/AFC4", func_us_8018B4A0);
+struct UnkStruct8018B4A0 {
+    s16 active;
+    u16 index;
+    u8* data;
+};
+extern struct UnkStruct8018B4A0 D_us_8019AEB0;
+extern u32 D_us_8019AF20;
+extern PfnEntityUpdate PfnEntityUpdates[];
+void func_us_8018B4A0(void) {
```
sozud, line 24: I think this is the same function as st0 func_801A8E60. Does it factor out?

sozud (approved):

PR #1620, opened 2024-09-16
Python virtualenv Support

hohle:
Ubuntu/Debian and Python recommend using virtual environments for project-specific Python dependencies and as of Ubuntu 24 this is lightly enforced when installing packages via pip. This is due to pip and the system package manager installing files to the same location which may cause either's internal state to no longer reflect what is actually installed.

This updates the project to use a Python `virtualenv` for project dependencies and updates internal scripts to support both global and virtualenvs, but favors virtualenvs for new workspaces.

All tools that hardcode `/usr/bin/python3` now use `env(1)` to find the first `python3` in the path. For those with a virtualenv configured, this will be the Python managed there. For everyone else, this should be the system Python or whatever other scheme they may have used previously (assuming `python3` already existed in their `PATH`).

The `Makefile` has been updated to prepend `.venv/bin` to the `PATH` and use `python3` from there if it exists and has been configured. It also has a new `python-dependencies` target which will configure the venv and install all python dependnecies there.

The `Dockerfile` has been updated to create an external `.venv` outside of the project directory. Python's `virtualenv`s are not relocatable and may hardcode paths which are mounted differently in the container and host. To deal with differences in paths between the container (which mounts the host's project directory to `/sotn`) host which may be at an arbitarary directory the `VENV_PATH` environment variable is used to override paths in the `Makefile`.

GitHub workflows have been updated to use `.venv`.


hohle:
I'm putting this in draft while I work out the GitHub workflow changes.

Once merged the [Build](https://github.com/Xeeynamo/sotn-decomp/wiki/Build) wiki should be updated to remove the python dependencies step (it looks like the Cargo step can also be removed, unrelated to this change).

hohle:
I plan on changing the wiki to the following. Explicit steps for python or rust should no longer be required.

> # Set-up
> 
> These instructions assume Ubuntu 22.04 LTS, which is what we are testing against in our CI.
> 
> ```bash
> apt-get update
> 
> # install git and clone the repo
> apt-get install -y git
> git clone https://github.com/Xeeynamo/sotn-decomp.git
> cd sotn-decomp
> 
> # update submodules
> git submodule update --init --recursive
> 
> # install debian packages
> apt-get install -y $(cat tools/requirements-debian.txt)
> 
> make update-dependencies
> ```
> 
> Dump your US copy of the game as BIN/CUE, put the files inside sotn-decomp/disks/ name the cue file sotn.us.cue. Then continue with extraction, building and verifying it:
> 
> ```bash
> # extract disk, build and check
> make extract_disk
> make extract -j && make build -j
> make check
> ```

Tested on:

* Ubuntu 22.04 (new setup)
* Ubuntu 24.04 (existing setup)
* Ubuntu 24.10 (new setup)

sozud:
Do current users have to do anything to migrate to this setup? Otherwise seems good to me.

hohle:
> Do current users have to do anything to migrate to this setup? Otherwise seems good to me.

Current users should run `make update-dependencies` _if_ builds start failing, but if they don't fail they can continue on as normal until something goes wrong. If `pip install -r tools/requirements-python.txt` works for current users, it should continue to work.

I've tested this locally with my local Python installing dependencies to `~/.local`.

```bash
$> git checkout master
$> make clean; make extract -j; make -j
⋮ *snip*
build/us/TT_000.BIN: OK
```
build is fine prior to this change

```bash
$> git checkout venv-migration
$> ls .venv
ls: cannot access '.venv': No such file or directory
$> which python3
/usr/bin/python3
$> make clean; make extract -j; make -j
⋮ *snip*
build/us/TT_000.BIN: OK
```

No virtual environment and haven't run `make update-dependencies` yet, so still using the system Python and my `~/.local` directory for PIP packages.

```bash
$> make update-dependencies
$> ls .venv
bin  include  lib  lib64  pyvenv.cfg
$> make clean; make extract -j; make -j
⋮ *snip*
build/us/TT_000.BIN: OK
```

Virtualenv is installed and the build is working, using the virtualenv instead.

```bash
$> rm -r ~/.local
$> make clean; make extract -j; make -j
⋮ *snip*
build/us/TT_000.BIN: OK
```

Deleted my local PIP prefix and the build still works fine, since it's using the virtualenv.

Everything done through `make` should work as expected regardless of how things were set up prior or after running `make update-dependencies`.

The one migration issue I can foresee is the project `splat` will now be installed to `.venv/bin/splat` vs. wherever else it had been installed previously. If people run `splat` or any of the other python tools in `tools` directly, they'll want to run `source .venv/bin/activate` to set the virtualenv for python or add `.venv/bin` to their `PATH`. (TMI `.venv/bin/activate` appends the path and adds some functions to restore the path, but not much else). If Python or any tools are run from `.venv/bin` everything will Just Work™.

sozud:
Ok works for me. Not a comment on this PR in particular (it looks pretty thorough) but we often have issues that appear in the CI when we make changes like this, we should merge this when someone is available to possibly debug the CI or revert until a fix can be made

hohle:
Is it possible to run the report actions without having them publish results? It would be nice to see those come back green prior to merging. Otherwise I don't think there is harm in rolling back if they fail.

sozud:
You could set the CI up in your repo but it's not a big deal to roll back. Will monitor now

PR #1621, opened 2024-09-16
Add `MAR` to `dups`, Remove Path Prefix

hohle:
Adds `boss/mar` to the list of paths evaluated by the duplicate finder.

Removes the `../../` prefix for matching ASM files in the final report.


sozud (approved):

sozud:
Just figured I should point out that if people are looking at big enhancements to the `dups` tool a new similar one came out https://github.com/ethteck/coddog

I don't know how it compares to ours or if it's worth integrating

PR #1624, opened 2024-09-17
 Create enum for g_CastleFlags 

bismurphy:
I went through every use of g_CastleFlags and created enum members for the different flags. This will allow us to better keep track of what flags exist and how they are used. This will also aid in tracking the arrangement of the overlays in the flags, as there seems to be some level of organization. As we add more gameplay overlays we can hopefully learn how these all work together.

Some functions were using local `const int` declarations to give names to flags. For example, there was `const int jewelSwordRoomUnlock = 51;`. For these ones, which had flag purposes identified in code, I went ahead and gave their global castle flag a similar name (in this case, `JEWEL_SWORD_ROOM`). For all other flags, I kept the name as a placeholder of `CASTLE_FLAG_###`; later someone else can identify what the individual flags are.

This should be a good step in making the castle flags more understandable, as well as to connect the different functions across the game and properly see which ones are touching the same flags. When a castle flag is written as 0x32 in some places, and 50 in others, it makes it harder to search. Now that we can use CASTLE_FLAG_50, hopefully we will have better understanding of all the uses of the flag.

### DIFF
--- a/include/game.h
+++ b/include/game.h
@@ -1810,34 +1810,38 @@ extern u8 g_CastleFlags[0x300]; // starts at 0x8003BDEC
 typedef enum {
     CLOCK_ROOM_DOORS, // opened by gold and silver ring; drops down to CEN
     CASTLE_FLAG_2 = 2,
-    CASTLE_FLAG_19 = 19,
+    CASTLE_FLAG_19 = 19, // Randomized by g_RandomizeCastleFlag13; unused
+    // Start NO3/NP3 flags
     CASTLE_FLAG_48 = 48,
     CASTLE_FLAG_49,
     CASTLE_FLAG_50,
     JEWEL_SWORD_ROOM,
     CASTLE_FLAG_52,
     CASTLE_FLAG_53,
-    CASTLE_FLAG_54,
+    CASTLE_FLAG_54, // Unused
     CASTLE_FLAG_55,
     DEATH_STAIRWAY_BROKEN, // Piece of stairs right before Death room
-    SG_RETR_ENTR,        // Slogra/Gaibon retreated from Entrance encounter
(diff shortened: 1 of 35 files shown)


sozud (commented):
This seems good to me, in my opinion some of the names are too abbreviated like `HG_CS_DONE` and `SG_RETR_ENTR`. I would not be able to guess what these mean even knowing the game is SOTN

bismurphy:
> This seems good to me, in my opinion some of the names are too abbreviated like `HG_CS_DONE` and `SG_RETR_ENTR`. I would not be able to guess what these mean even knowing the game is SOTN

Yes; unfortunately these are kind of the consequences of enforcing our strict column limit. I thought it would be better to focus on brevity and then give a full comment in the enum (in this case, elaborating HG_CS as "Holy Glasses cutscene" and SG_RETR_ENTR as "Slogra/Gaibon retreated from Entrance encounter"). 

If you use VS code, these comments also show up on mouseover, as shown here: 
image

sozud:
Ok, I think let's get Xeeynamo's feedback on this one before merging

bismurphy:
> But I would have chosen something that wouldn't have appealed you either.

Does this refer to Sozud or me? I don't have any strong feelings about any particular naming, I just put these in as a starting point. If there are any other preferred names (even if they're longer) I'd be happy to swap them out. For me personally this isn't an opinionated topic. Sozud made a good point about the names being over-abbreviated but I'm not sure what the less-abreviated forms should be. We can agree that full descriptions featuring prepositions and conjunctions (`SLOGRA_AND_GAIBON_HAVE_RETREATED_FROM_ENTRANCE`) would be excessive, but then the amount of trimming is hard to decide. If it should be `SLOGRA_GAIBON_RETREATED_ENTRANCE` that's good with me. Just as a starting point I figured 32 characters for a single enum felt like a lot.

PR #1625, opened 2024-09-17
Decompile `MAR` `func_us_8018CA94`

hohle:


sozud (approved):
We should recombine the files once the other functions are done but looks good to me

src/boss/mar/CA94.c
```diff
@@ -0,0 +1,189 @@
...
+extern u32 g_CutsceneFlags;
+extern s32 D_us_8019AE70;
+
+void func_us_8018CA94(Entity* self) {
+    Entity* entities;
+    Tilemap* tilemap;
+    s32 pan;
+
+    tilemap = &g_Tilemap;
+    entities = &PLAYER;
```
bismurphy, line 24: Any reason to call this "entities" rather than "player"?

Also might be worth a comment mentioning that this pointer is unused. Otherwise, looks good.

bismurphy (approved):
Recombining will likely be mandatory due to rodata padding anyway. File split is great for now.

src/boss/mar/CA94.c
sozud: It doesn't seem like this change was made?

PR #1626, opened 2024-09-18
Decompile `st/no3` `EntityUnkId31`

JoshSchreuder:
https://decomp.me/scratch/TVEAy

### DIFF
--- a/src/st/no3/48A84.c
+++ b/src/st/no3/48A84.c
@@ -236,7 +236,121 @@ void EntityUnkId30(Entity* self) {
     }
 }
 
-INCLUDE_ASM("st/no3/nonmatchings/48A84", EntityUnkId31);
+void EntityUnkId31(Entity* self) {
+    Entity* entity;
+    u16* hitboxPtr;
+    u16 animFrameIdx;
+    u16 animCurFrame;
+    s16 i;
+
+    entity = self - 2;
+    if (self->step == 0) {
+        InitializeEntity(D_80180B24);
+        self->zPriority++;
+        CreateEntityFromCurrentEntity(E_ID_30, self + 1);
+        (self + 1)->params = 1;
+    }
+    if (self->ext.prim) {
+        animFrameIdx = (self->animFrameIdx - 1) * 2;
+        if (entity->step_s == 1) {
+            hitboxPtr = D_80182FC8 + animFrameIdx;
+        } else {
+            hitboxPtr = D_80182FE8 + animFrameIdx;
+        }
+
+        if (self->facingLeft) {
+            self->posX.i.hi = entity->posX.i.hi - *hitboxPtr++;
+        } else {
+            self->posX.i.hi = entity->posX.i.hi + *hitboxPtr++;
+        }
+        self->posY.i.hi = entity->posY.i.hi + *hitboxPtr++;
+    } else {
+        self->posX.i.hi = entity->posX.i.hi;
+        self->posY.i.hi = entity->posY.i.hi;
+    }
+    self->facingLeft = entity->facingLeft;
+    animCurFrame = entity->animCurFrame;
+    if (self->flags & FLAG_DEAD) {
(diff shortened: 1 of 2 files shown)


src/st/no3/48A84.c
```diff
@@ -236,7 +236,121 @@ void EntityUnkId30(Entity* self) {
...
+    u16* hitboxPtr;
+    u16 animFrameIdx;
+    u16 animCurFrame;
+    s16 i;
+
+    entity = self - 2;
+    if (self->step == 0) {
+        InitializeEntity(D_80180B24);
+        self->zPriority++;
+        CreateEntityFromCurrentEntity(ITEM_NO_CAPE, self + 1);
```
bismurphy, line 250: When creating entities, we should use the EntityIDs enum. Since it doesn't have an entry for 0x30, you can leave this as 0x30 for now. But ITEM_NO_CAPE should be reserved for situations where we're dealing with equippable items. If you're not familiar with enums I'm happy to elaborate.

src/st/no3/48A84.c
```diff
@@ -236,7 +236,121 @@ void EntityUnkId30(Entity* self) {
...
+    }
+    if (self->ext.prim) {
+        animFrameIdx = (self->animFrameIdx - 1) * 2;
+        if (entity->step_s == 1) {
+            hitboxPtr = D_80182FC8 + animFrameIdx;
+        } else {
+            hitboxPtr = D_80182FE8 + animFrameIdx;
+        }
+
+        if (self->facingLeft != 0) {
```
bismurphy, line 261: This can just be `if (self->facingLeft)`.

src/st/no3/48A84.c
```diff
@@ -236,7 +236,121 @@ void EntityUnkId30(Entity* self) {
...
+    self->facingLeft = entity->facingLeft;
+    animCurFrame = entity->animCurFrame;
+    if (self->flags & FLAG_DEAD) {
+        hitboxPtr = D_80182F9C;
+        PlaySfxPositional(0x684);
+
+        for (i = 0; i < 3; i++) {
+            entity = AllocEntity(&g_Entities[224], &g_Entities[256]);
+            if (entity) {
+                CreateEntityFromCurrentEntity(ITEM_REVERSE_CLOAK, entity);
```
bismurphy, line 280: Same comment here, shouldn't use the item enums

src/st/no3/48A84.c
```diff
@@ -236,7 +236,121 @@ void EntityUnkId30(Entity* self) {
...
+
+        hitboxPtr = D_80182FA8;
+        for (i = 0; i < 8; i++) {
+            entity = AllocEntity(&g_Entities[224], &g_Entities[256]);
+
+            if (!entity) {
+                break;
+            }
+
+            CreateEntityFromCurrentEntity(E_EXPLOSION, entity);
```
bismurphy, line 300: This is a good enum!

bismurphy (commented):
Looks good! Just a couple style notes.

src/st/no3/48A84.c
JoshSchreuder: @bismurphy think I get you, I'm referencing stuff here in `items.h` which is probably incorrect, but `E_EXPLOSION` is defined in `no3.h` so that one is known correct?

And 0x30 / 0x32 are not in `no3.h` yet so I can leave these as hex values?

bismurphy: Right. You can leave them as hex values, or if you want to be extra cool, you could make new entries in no3.h and call them E_ID_30, E_ID_31, E_ID_32. Your call, we aren't requiring using enums for function calls like this as of right now.

JoshSchreuder: Makes sense, I've added the stub entities as you mentioned, might as well save someone a tiny bit of future effort 😄

sozud (approved):

PR #1628, opened 2024-09-18
De-dupe EntityHeartDrop

bismurphy:
Pulls this function into a .h file.

Since each overlay uses a different offset into g_CastleFlags, we use `#define HEART_DROP_CASTLE_FLAG` before each `#include` statement in order to set these up correctly.

Next we will hopefully be able to dedupe e_misc.c, but that will take a bit longer especially since PSP still needs EntityRelicOrb done, but I'm working on that one now.


bismurphy:
> Just a small potential improvement: Since all the `HEART_DROP_CASTLE_FLAG` between overlays are `0x10` distant each other, wouldn't make sense to have the associated number in hexadecimal instead of decimal?

Good point. My main reason for this is that the castle flags are currently defined as decimal in the enum, so it made sense to me for this to also be decimal here. If you'd like them to be hexadecimal I can make that change.

PR #1629, opened 2024-09-19
decompile EntityUnkId4B

Pyromuffin:
my first decompiled function. no idea who this is but we're setting up their hitbox.


sozud (approved):
Looks good except you need to run `make format` to fix the linter. Thanks for contributing!

PR #1632, opened 2024-09-20
Build QoL

hohle:
Several quality of life changes for building:

* The Makefile has a `help` target which will show common targets. Targets with comments starting with `##@` will be included in the help output.

* The `check` target now has colored output.

* Added `check_disk` target which will check hashes of extracted disk contents (useful for those who dump their own discs)

* Added `dump_disk` target which will create a bin/cue pair from an original disc.


hohle:
`make help`:

Screenshot 2024-09-19 at 9 33 23 PM

`make check`:

Screenshot 2024-09-19 at 9 34 01 PM

sozud (approved):
These changes seem nice to me but this needs a review from Xeeynamo

hohle:
The implementation I chose was @BlackHole1's with some tweaks for portability and terminal compatibility from that same page.

I forgot that splat checks the SHA. I'll make sure it's present in all of the overlays for versions I have access to and remove the targets.

PR #1634, opened 2024-09-21
decompile func_us_8018C90C

Pyromuffin:
not sure where to put these externs, I didnt see a header file for this.

also, the `D_us_8019AF2A[0] ^= 1;` construct is sus for two reasons:
1) this is A bytes away from g_CutsceneFlags, and
2) I could only get a matching decompile by making these arrays and accessing the first element, even though a ptr would make more sense to me.


sozud (approved):
Looks good if Xeeynamo's suggestions are applied

src/boss/mar/AFC4.c
```diff
@@ -21,4 +21,50 @@ INCLUDE_ASM("boss/mar/nonmatchings/AFC4", func_us_8018B74C);
...
+        g_Player.padSim = PAD_RIGHT;
+        g_Player.D_80072EFC = 1;
+        return;
+    case 1:
+        if ((player->posX.i.hi >= 0x21) && !(g_CutsceneFlags & 1)) {
+            g_CutsceneFlags |= 1;
+        }
+        if (player->posX.i.hi >= 0xB1) {
+            g_Player.padSim = PAD_LEFT;
+            self->step += 1;
```
hohle, line 47: These are usually an increment operator (`++`). It _usually_ doesn't matter for PSX, but almost always matters for PSP.

Pyromuffin:
thanks everyone for helping me refine this code!

PR #1649, opened 2024-09-21
Decompile EntityMariaCutscene of boss/mar

hohle:
This is similar to other cutscenes but with a few minor differences. Of
note is that the PSX version uses an uninitialized pointer.

Pulls in the remaining duplicates and `func_us_8018C90C` as well
finishing off the remaining functions in `boss/mar`.

### DIFF
--- a/src/boss/mar/AFC4.c
+++ b/src/boss/mar/AFC4.c
@@ -647,44 +647,4 @@ void EntityMariaCutscene(Entity* self) {
     }
 }
 
-extern u16 D_80180A60[];
-extern u16 D_us_8019AF2A[];
-
-void func_us_8018C90C(Entity* self) {
-    Entity* player = &PLAYER;
-    switch (self->step) {
-    case 0:
-        InitializeEntity(D_80180A60);
-        D_8003C8B8 = 0;
-        g_unkGraphicsStruct.unk0 = 1;
-        g_Player.padSim = PAD_RIGHT;
-        g_Player.D_80072EFC = 1;
-        break;
-
-    case 1:
-        if (PLAYER.posX.i.hi >= 33 && !(g_CutsceneFlags & 1)) {
-            g_CutsceneFlags |= 1;
-        }
-        if (player->posX.i.hi >= 177) {
-            g_Player.padSim = PAD_LEFT;
-            self->step++;
-        } else {
-            g_Player.padSim = PAD_RIGHT;
-        }
-        g_Player.D_80072EFC = 1;
-        break;
-
-    case 2:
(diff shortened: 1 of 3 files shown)


src/boss/mar/AFC4.c
```diff
@@ -11,14 +11,640 @@ extern Dialogue g_Dialogue;
...
+    }
+}
 
 #include "../../st/cutscene_unk6.h"
 
 #include "../../st/cutscene.h"
 
-INCLUDE_ASM("boss/mar/nonmatchings/AFC4", func_us_8018B74C);
+// Animates the portrait size of the actor by enlarging or shrinking it
+void func_us_8018B74C(u8 ySteps) {
```
sozud, line 79: This is a duplicate, could we factor it out? Can wait for a followup PR

src/boss/mar/AFC4.c
```diff
@@ -11,14 +11,640 @@ extern Dialogue g_Dialogue;
 
 #include "../../st/cutscene_unk4.h"
 
-INCLUDE_ASM("boss/mar/nonmatchings/AFC4", func_us_8018B2F0);
+extern char* D_us_80180680[];
+
+void func_us_8018B2F0(u16 actorIndex, Entity* self) {
```
sozud, line 16: Could we factor this out? Doing it in this PR is probably going to make it too big though

sozud (commented):
```
src/boss/mar/AFC4.c:(.text+0x1998): undefined reference to `D_us_801803B4'
mipsel-linux-gnu-ld: src/boss/mar/AFC4.c:(.text+0x199c): undefined reference to `D_us_801803B4'
mipsel-linux-gnu-ld: src/boss/mar/AFC4.c:(.text+0x19a0): undefined reference to `func_us_80191098'
mipsel-linux-gnu-ld: src/boss/mar/AFC4.c:(.text+0x1a88): undefined reference to `func_us_80190388'
```


D_80180A60 is probably incorrectly removed from the symbols. same for func_us_80191098 (InitializeEntity)

hohle:
> D_80180A60 is probably incorrectly removed from the symbols. same for func_us_80191098 (InitializeEntity)

The symbol was removed because I originally had `func_us_8018C90C` as well but saw [1634](https://github.com/Xeeynamo/sotn-decomp/pull/1634) and pulled it at the last minute.

src/boss/mar/AFC4.c
hohle: I'll take a look at these.

sozud (approved):
Thanks for doing the factoring out, looks good!

PR #1651, opened 2024-09-21
De-duplicate CEN

bismurphy:
I went through all of CEN, cross-referencing it with WRP to make as many files match as possible. This involved pulling in data (which had been pulled in on WRP but not on CEN yet), renaming shared variables to not have placeholder names, adjusting file splits, and more.

EntityBreakable seems to be very different here, so I did not do any deduplication on that one yet. I think we need to get a robust system for deduplicating across stages so we can share .h files even when the stages are different. Right now there's no way for an included .h file to know it's part of CEN (at least, that I know of).

### DIFF
--- a/src/st/cen/holyglassescutscene.c
+++ b/src/st/cen/holyglassescutscene.c
@@ -238,7 +238,7 @@ void EntityHolyGlassesCutscene(Entity* self) {
             DestroyEntity(self);
             return;
         }
-        if (CutsceneUnk2(D_801813F0) & 0xFF) {
+        if (CutsceneUnk2(D_801813F0)) {
             self->flags |= FLAG_HAS_PRIMS | FLAG_UNK_2000;
             D_8003C704 = 1;
             g_CutsceneFlags = 0;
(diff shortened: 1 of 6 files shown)


src/st/st_debug.h
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+extern ObjInit g_eBackgroundBlockInit[];
+extern u16 g_eInitGeneric2[];
+
+void EntityBackgroundBlock(Entity* self) {
```
sozud, line 6: EntityBackgroundBlock is considered to be debug related?

sozud (approved):

src/st/st_debug.h
bismurphy: Perhaps we need to rename the files or functions. All I know is WRP had "st_debug.c", with an unnamed function, and I found that same function called EntityBackgroundBlock in CEN, in an unnamed file. Now the combination of the two results in EntityBackgroundBlock being in st_debug.h. Let me try seeing if the files can be shuffled around.

PR #1653, opened 2024-09-22
#740: Fix remaining DRA .ext.generic

JoshSchreuder:
Also cleaned up the function naming and implemented the same generic in RIC `DisableAfterImage` as it looks like similar usage.

Hopefully this isn't too much unrelated stuff in the one PR but I thought it would be worthwhile doing the function rename and using the same entity in RIC - I can move out into separate commit / PR if you prefer.


src/dra/6D59C.c
```diff
@@ -318,12 +318,12 @@ void func_8010DF70(void) {
     }
 }
 
-void func_8010DFF0(s32 arg0, s32 arg1) {
+void func_8010DFF0(s32 resetAnims, s32 arg1) {
```
JoshSchreuder, line 321: I tried renaming this function to `DisableAfterImage` in an earlier commit but was having pipeline issues. Not sure what I need to do to do that, this func seems to be part of `g_api` so maybe I need additional changes to do so. Can be done separately anyway

sozud: When you renamed it I think you missed this spot https://github.com/Xeeynamo/sotn-decomp/blob/7d9f2cd04e9c02d2f71974e58306796230489a17/src/pc/sotn.c#L138

sozud (approved):

src/dra/6D59C.c
JoshSchreuder: Would that be `api.func_8010DFF0 = DisableAfterImage;`, or would you normally go through and rename it as `api.DisableAfterImage = DisableAfterImage` and all the `g_api` calls?

PR #1658, opened 2024-09-23
Decompile `no3` `EntityExplosion3`

JoshSchreuder:
https://decomp.me/scratch/XWQUk

I'm sure there's some non-idiomatic code I've written here so looking forward to rewriting some 😄

### DIFF
--- a/src/st/no3/4AFF0.c
+++ b/src/st/no3/4AFF0.c
@@ -319,7 +319,127 @@ void EntityUnkId31(Entity* self) {
 }
 
 // some sort of explosion
-INCLUDE_ASM("st/no3/nonmatchings/4AFF0", EntityExplosion3);
+void EntityExplosion3(Entity* entity) {
+    Entity* newEntity;
+    u32* ptr32;
+    u8* ptr;
+    Point16* point;
+    Primitive* prim;
+    u16 params;
+    s32 xOffset;
+    s32 yOffset;
+    s16 primIndex;
+    u16 angle;
+    s16 posX;
+    s16 posY;
+    u16 newX;
+    u16 newY;
+
+    params = entity->params;
+    if (!entity->step) {
+        primIndex = g_api.AllocPrimitives(PRIM_GT4, 1);
+        if (primIndex != -1) {
+            InitializeEntity(D_80180B3C);
+            entity->flags |= FLAG_UNK_2000;
+            entity->hitboxState = 0;
+            prim = &g_PrimBuf[primIndex];
+            entity->flags |= FLAG_HAS_PRIMS;
+            entity->primIndex = primIndex;
+            primIndex = entity->unk5A + 3;
+            prim->tpage = primIndex >> 2;
+            prim->clut = entity->palette + 1;
+
+            ptr = &D_80183028[params * 4];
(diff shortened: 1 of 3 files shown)


src/st/no3/4AFF0.c
```diff
@@ -319,7 +319,127 @@ void EntityUnkId31(Entity* self) {
...
+            prim->clut = entity->palette + 1;
+
+            ptr = &D_80183028[params * 4];
+            prim->u0 = prim->u2 = *ptr++ + ((primIndex & 1) << 7);
+            prim->u1 = prim->u3 = *ptr++ + ((primIndex & 1) << 7);
+            prim->v0 = prim->v1 = *ptr++ + ((primIndex & 2) << 6);
+            prim->v2 = prim->v3 = *ptr + ((primIndex & 2) << 6);
+            prim->g0 = prim->g1 = prim->g2 = prim->g3 = prim->r0 = prim->r1 =
+                prim->r2 = prim->r3 = 0x80;
+            prim->b0 = prim->b2 = prim->b1 = prim->b3 = 0x80;
```
JoshSchreuder, line 359: Whoops, I'll inline this as well

JoshSchreuder: Actually do we prefer these inlined or separated out?

Inlined would be
```
            prim->b0 = prim->b2 = prim->b1 = prim->b3 = prim->g0 = prim->g1 =
                prim->g2 = prim->g3 = prim->r0 = prim->r1 = prim->r2 =
                    prim->r3 = 0x80;
```

With the column breaks

vs

```
            prim->r0 = prim->r1 = prim->r2 = prim->r3 = 0x80;
            prim->g0 = prim->g1 = prim->g2 = prim->g3 = 0x80;
            prim->b0 = prim->b2 = prim->b1 = prim->b3 = 0x80;
```

src/st/no3/4AFF0.c
```diff
@@ -319,7 +319,127 @@ void EntityUnkId31(Entity* self) {
...
+            entity->velocityX = *ptr32++;
+        }
+        entity->velocityY = *ptr32;
+    }
+
+    posX = entity->posX.i.hi;
+    posY = entity->posY.i.hi;
+    point = &D_80183054[params];
+    prim = &g_PrimBuf[entity->primIndex];
+    entity->ext.factory.unk7E += D_8018304C[params];
```
sozud, line 382: We're mixing ext.timer and ext.factory, we should probably make an extension

src/st/no3/4AFF0.c
sozud: I don't think those are equivalent on psp?

JoshSchreuder: Yep you're right, looks like PSP is more happy with the latter, so I'll update to that

src/st/no3/4AFF0.c
JoshSchreuder: Added, hope this is what you meant

src/st/no3/4AFF0.c
JoshSchreuder: Updated, and also fixed up a few other things like `if (self->step != 0)` to be a closer match on the PSP side

PR #1660, opened 2024-09-23
Add HD version of the RIC overlay 

Xeeynamo:
I was also able to remove the weird `goto block` on `RicMain`.

The HD version seems to be an older version. I am not entirely sure of the US changes, but it seem to fix a series of bugs when Richter gets grabbed by Gaibon? There's a new debug string which ints `g_Player.unk7A`  being named as `run_disable_f` but I did not yet rename it.

### DIFF
--- a/src/ric/21250.c
+++ b/src/ric/21250.c
@@ -1577,11 +1577,12 @@ void RicUpdatePlayerEntities(void) {
 }
 
 #if defined(VERSION_US)
-#define MAYBE false
+#define NON_CRITICAL false
 #elif defined(VERSION_HD)
-#define MAYBE true
+#define NON_CRITICAL true
 #endif
 FactoryBlueprint g_RicFactoryBlueprints[] = {
+    // clang-format off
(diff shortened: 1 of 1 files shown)


src/ric/21250.c
```diff
@@ -1570,6 +1576,11 @@ void RicUpdatePlayerEntities(void) {
     }
 }
 
+#if defined(VERSION_US)
```
sozud, line 1579: My suggestion here would be to #ifdef the individual entries. What if someone wants to mod and change E_WHIP to true and E_CRASH_HYDROSTORM to false

sozud (approved):

src/ric/21250.c
bismurphy: I think with so many lines, it's better to have a single ifdef.

If someone wants to mod the whip and hydrostorm to be different, they can type "true" and "false" in the two B_MAKE lines, and replace the MAYBE call.

Separately, there should be a different word used for MAYBE, which describes the field it controls. But for the question about how to ifdef, I think one ifdef at the top is better than doing them all over.

bismurphy: Sure that works

Issue #1637, opened 2024-09-21
Add familar overlay TT_001


Xeeynamo:
Refer to the guide on [how to add new overlays](https://github.com/Xeeynamo/sotn-decomp/wiki/Decompilation#add-new-overlay).

The smallest, and probably the easiest, familiar in the game. It might be a very good starting point.

PR #1662, opened 2024-09-24
#1637: Add tt_001 Ghost familiar to docs

JoshSchreuder:


sozud (approved):
Looks OK to me but should probably be checked by Xeeynamo

PR #1664, opened 2024-09-24
Add CEN Overlay to HD

### LINKED ISSUE #1638: Add HD overlay for CEN
Refer to the guide on [how to add new overlays](https://github.com/Xeeynamo/sotn-decomp/wiki/Decompilation#add-new-overlay).

The overlay is complete on the `us` build. It should be straightforward to adapt the `us` code to match the `hd` build.

hohle:
Adds the CEN overaly to the HD version of the game.

Resolves #1638.


src/st/cen/hd_e_room_fg.c
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
```
hohle, line 1: This is a unique version of `e_room_fg` and is not compatible with WRP.

Originally I had this conditionally compiling for `VERSION_HD`, but I would need to add a stage condition as well. I decided to move it to a separate file instead, but can move it back if that's preferred.

Besides the data differences, the condition and operations on lines 27-29 differ from the implementation shared.

Makefile
```diff
@@ -246,7 +247,7 @@ check: config/check.$(VERSION).sha patch $(CHECK_FILES)
             color = 196;   \
         system("tput setaf " color "; printf " $$2 "; tput sgr0"); \
         printf " ]\n"; \
-    }' | column --separator '\t' --table
+    }' | column --separator $$'\t' --table
```
hohle, line 250: This is a minor fix to properly match tab characters when formatting the `check` output.

Makefile
```diff
@@ -142,7 +142,7 @@ extract: extract_$(VERSION)
 build: ##@ build game files
 build: build_$(VERSION)
 build_us: main dra weapon ric cen dre mad no3 np3 nz0 sel st0 wrp rwrp mar tt_000 tt_001
-build_hd: dra $(BUILD_DIR)/WRP.BIN tt_000
+build_hd: dra cen $(BUILD_DIR)/WRP.BIN tt_000
```
sozud, line 145: I'm thinking part of your build issues may be that the only stage overlay we have here is wrp, done as `$(BUILD_DIR)/WRP.BIN`, while you add `cen`

src/st/cen/hd.c
```diff
@@ -0,0 +1,278 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "common.h"
+
+#if defined(VERSION_HD)
+
+INCLUDE_ASM("st/cen/nonmatchings/hd", EntityUnkId12);
```
sozud, line 6: The previous time we added HD to a stage it looked like this https://github.com/Xeeynamo/sotn-decomp/pull/1421/files

Xeeynamo would chime in on whether he wants this done on a branch or incrementally on master

src/st/cen/hd_e_room_fg.c
hohle: I changed this to follow the model of https://github.com/Xeeynamo/sotn-decomp/pull/1421 where the differenced code is local to the stage C file.

Makefile
hohle: The issue appears to be missing tile and sprite entries in the splat config. It looks like I was getting those from a `us` extract. I'm going to pull this down while I work out those offsets.

PR #1667, opened 2024-09-24
Add bsdmainutils to requirements

mkst:
- `column` util was missing from the Docker image, added it and organised the requirements file so it's more human friendly
- rearranged the USER argument in the Dockerfile to avoid the scenario where you didnt specify the user and so lose the layers where apt installed all the packages
- add the python deps at the end to ensure docker can reuse cached layers if a python dep changes.
- update maspsx (I wanted to test latest maspsx against sotn, so this is what led to the other changes!)


sozud (approved):

hohle (approved):

Dockerfile
```diff
@@ -6,22 +6,21 @@
 
 # make extract -j && make build -j
 FROM ubuntu:22.04
+ADD /tools/requirements-debian.txt /tools/requirements-debian.txt
+RUN apt-get update && apt-get install -y $(cat /tools/requirements-debian.txt)
+
+WORKDIR /sotn
```
hohle, line 12: My preference is usually to split repo update from package install for caching. By moving the repo update after adding `requirements-debian.txt` you lose the ability to cache the `apt update` line, which may or may not be intentional.

mkst: It was intentional to avoid having a stale package list - if a new package is added (with the old Dockerfile), the package list could be out of date, whereas now the package list would be updated (because the cache is busted by the change to requirements-debian.txt)

I'm speaking from experience of having a stale package list in one of my own projects 😅

PR #1675, opened 2024-09-25
Update entity IDs to map to enum

JoshSchreuder:
Did my best to give contextual names where possible based on surrounding code.

There's still some remaining ones where it wasn't clear what the entity was referring to so have left some as is.


sozud:
Have you used the debug module? You can use it to spawn entities https://github.com/Xeeynamo/sotn-decomp/wiki/Debug-Module

What's here so far seems reasonable to me but haven't checked any of these ingame

JoshSchreuder:
Thanks, that was quite interesting. I wasn't really able to discern much more.
A combination of not being able to get to specific stages in the debugger (stage select crashes, and I don't know if there's a way to get to NO3 for example after getting the bat card?) and also spawning a lot of these entities or previewing them straight up crashes the emulator.

==================

One interesting aside - I spawned this inside NZ0 https://github.com/Xeeynamo/sotn-decomp/blob/e357fcaeeb20750e30f0595155b924944c1f36e6/src/st/entity_unkId13.h#L13

It does not appear to me to be an explosion - looks more like the dust kick up from running back and forth (hold run for more than 5 frames perhaps?). I'm not sure why the entity ID is mapped as it is, and this is shared across many overlays, but if you have any thoughts I'd be interested.

src/st/np3/32830.c
```diff
@@ -42,7 +43,7 @@ void func_801B28E4(Entity* self) {
         case 0:
             newEntity = AllocEntity(&g_Entities[224], &g_Entities[256]);
             if (newEntity != NULL) {
-                CreateEntityFromCurrentEntity(E_801B5DE8, newEntity);
+                CreateEntityFromCurrentEntity(E_LIGHTNING_THUNDER, newEntity);
```
JoshSchreuder, line 46: This was an interesting one - spawning `func_801B28E4` caused the sky to light up, then spawning `E_801B5DE8` caused a thunder noise to play. There's no SFX call here so I'm not 100% sure where that came from. 

Unpausing crashed the game too 😄

src/st/np3/32830.c
```diff
@@ -21,6 +21,7 @@ void func_801B2830(Entity* self) {
     }
 }
 
+// Maybe related to no3 EntityBackgroundLightning
```
JoshSchreuder, line 24: Didn't see this in the duplicates list, but it seems plausible they'd be related given both of these are entrance.

sozud (approved):
So I tried the debug overlay and it just instantly crashed for me, I think it has regressed. I'm not really set up to dig deeper into this overlays to give good answers to your comments right now unfortunately. I think this is a good incremental improvement even if not everything is figured out or correct. Right now it's gotten conflicted though. Do you want this kept up for a while to see if any other contributors investigate?

JoshSchreuder:
@Xeeynamo are you happy to merge as is if I rebase or would you like me to dig into these in more detail using MAC and keep this open for now?

As mentioned in the original comment, this doesn't cover off the entirety of usages, some of which are still named `E_ID_15` etc.
I've just tried to map these to surrounding code where possible and done a minor spot check of some ones I could access with the debug overlay.

PR #1678, opened 2024-09-26
Decompile libgpu SetDrawEnv

Onenutmcgee:
Decompiled libgpu SetDrawEnv

This function actually packs some of the short length fields from a DRAWENV struct into DR_ENV struct code fields, which are longs.  

IE dr_env->code[7] gets packed with env->clip.x in the hi 2 bytes and env->clip.y in the low 2 bytes.

Perhaps DR_ENV should actually be PACKED_DRAWENV.

Decompilation by sozud <sozud@users.noreply.github.com> 
and brought to the finish line by dezgeg <dezgeg@users.noreply.github.com>

### DIFF
--- a/src/main/psxsdk/libgpu/sys.c
+++ b/src/main/psxsdk/libgpu/sys.c
@@ -306,13 +306,13 @@ extern void SetDrawMode(DR_MODE* p, int dfe, int dtd, int tpage, RECT* tw) {
     p->code[1] = get_tw(tw);
 }
 
-extern void SetDrawEnv(DR_ENV* dr_env0, DRAWENV* env) {
-    RECT temp;
-    s32 var_t0;
-    u16 var_v1;
+extern void SetDrawEnv(DR_ENV* dr_env_in, DRAWENV* env) {
+    RECT clip_rect;
+    s32 offset;
+    u16 calc_clip_height;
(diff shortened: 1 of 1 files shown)


src/main/psxsdk/libgpu/sys.c
```diff
@@ -306,8 +306,65 @@ extern void SetDrawMode(DR_MODE* p, int dfe, int dtd, int tpage, RECT* tw) {
     p->code[1] = get_tw(tw);
 }
 
-extern void SetDrawEnv(DR_ENV* dr_env, DRAWENV* env);
-INCLUDE_ASM("main/nonmatchings/psxsdk/libgpu/sys", SetDrawEnv);
+extern void SetDrawEnv(DR_ENV* dr_env_in, DRAWENV* env) {
```
sozud, line 309: You shouldn't need extern here

sozud (approved):
Looks good, thanks for putting this up.

src/main/psxsdk/libgpu/sys.c
Onenutmcgee: Removed it.  I wasn't sure if it was needed.  My outdated and somewhat limited C knowledge did not contain "extern function"

PR #1679, opened 2024-09-26
Decompile for ServantDesc data for tt_001

Onenutmcgee:
There are 2 symbols that are pointed at functions in the ServantDesc struct.  I wasn't sure how to get those symbols to point into fields of that struct, but breaking it apart works.  Maybe there is a better way to do it.


Onenutmcgee:
I'm going to abandon this for now.  We'll just have to wait until the calling function is decompiled and then we can call it through the function pointer like normal.

PR #1681, opened 2024-09-27
Deduplicate `tt_001` functions with `tt_000`

JoshSchreuder:
* `ProcessEvent`
* `CreateEventEntity`
* `CalculateAngleToEntity` (the function formerly known as func_80173F30)
* `GetTargetPositionWithDistanceBuffer` (the function formerly known as func_80173F74)

There's 4 more duplicates which I'll start on separately so as to not make this PR too big.


src/servant/create_event_entity.h
```diff
@@ -0,0 +1,26 @@
...
+        DestroyEntity(entity);
+        entity->entityId = entityId;
+        entity->zPriority = entityParent->zPriority;
+        entity->facingLeft = entityParent->facingLeft;
+        entity->flags = FLAG_KEEP_ALIVE_OFFCAMERA;
+        entity->posX.val = entityParent->posX.val;
+        entity->posY.val = entityParent->posY.val;
+        // Not necessarily making batFamBlueTrail here, but
+        // that's an Ext that works. Just needs parent at 0x8C.
+        entity->ext.batFamBlueTrail.parent = entityParent;
```
JoshSchreuder, line 23: Should I change this now it's shared between entities? Perhaps a `ext.servant` extension?

src/servant/get_target_position_with_distance_buffer.h
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+s16 GetTargetPositionWithDistanceBuffer(
```
JoshSchreuder, line 2: This seemed like a reasonable name... my guess is maybe it tries to prevent the servant from moving too close to the player character? Though I'm not sure why this only covers the x-axis

src/servant/create_event_entity.h
sozud: I think we can adjust this as we decompile more of the familiars and see what's really in common, this is fine for now.

src/servant/get_target_position_with_distance_buffer.h
sozud: There's a ton of variables involved with this so I think we'd need to start renaming and documenting more to determine exactly what the purpose is, this is good for now

sozud (approved):
There's new scratches popping up on decomp.me re-decompiling these so I'm just going to do a fast merge. Thanks!

PR #1682, opened 2024-09-27
Deduplicate more `tt_001` functions with `tt_000`

JoshSchreuder:
* `CalculateDistance` (fka func_80173FE8)
* `IsMovementAllowed` (fka func_801746A0)
* `CheckAllEntitiesValid` (fka func_801747B8)
* `ServantUnk0` (fka func_80174864)


src/servant/servant_unk0.h
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+s32 ServantUnk0(void) {
```
JoshSchreuder, line 2: Had no idea what to call this, seems like it's some sort of check against an enum value most likely. Calling code didn't shed much light.

sozud (approved):
Looks good, we can continue iterating on the naming.

PR #1685, opened 2024-09-27
Add duplicate match info to function report

JoshSchreuder:
Had an idea, not sure if this is useful or not but I figured it might help give an indication for newer people or those not in the Discord loop if code is already duplicated when looking at the function report.

Here's an example output from the current state (minus the scratches which didn't seem to load for me locally so I skipped them)
[functions.md](https://github.com/user-attachments/files/17162454/functions.md)

@bismurphy guessing this would be your area of expertise 😄 

image

### DIFF
--- a/.github/workflows/validate-and-report.yml
+++ b/.github/workflows/validate-and-report.yml
@@ -191,6 +191,12 @@ jobs:
             make force_extract
             .venv/bin/python3 tools/analyze_calls.py
             git clean -fdx asm/
+      - name: Generate duplicates report
+        run: |
+          make force_symbols
+          make force_extract
+          cd tools/dups
+          cargo run --release -- --threshold .90 --output-file ../../gh-duplicates/duplicates.txt
       - name: Generate function report
         run: |
             git checkout config/
@@ -201,12 +207,6 @@ jobs:
             rm -rf gh-duplicates/function_calls/ || true
             mv function_calls gh-duplicates/
             mv function_graphs.md gh-duplicates
-      - name: Generate duplicates report
-        run: |
-          make force_symbols
-          make force_extract
-          cd tools/dups
-          cargo run --release -- --threshold .90 --output-file ../../gh-duplicates/duplicates.txt
       - name: Commit all reports
         run: |
             git config --global user.name 'GitHub Action'
--- a/tools/function_finder/function_finder_psx.py
+++ b/tools/function_finder/function_finder_psx.py
@@ -141,6 +141,7 @@ if __name__ == "__main__":
 
         wip = ""
         wip_percentage = ""
+        dup = ""
         output.append(
             [
                 ovl_name,
@@ -148,6 +149,7 @@ if __name__ == "__main__":
                 length,
                 branches,
                 jump_table,
+                dup,
                 wip,
                 wip_percentage,
                 f["text"] # local asm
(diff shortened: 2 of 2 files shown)


.github/workflows/validate-and-report.yml
```diff
@@ -191,22 +191,22 @@ jobs:
             make force_extract
             .venv/bin/python3 tools/analyze_calls.py
             git clean -fdx asm/
+      - name: Generate duplicates report
+        run: |
+          make force_symbols
+          make force_extract
+          cd tools/dups
+          cargo run --release -- --threshold .90 --output-file ../../gh-duplicates/duplicates.txt
```
JoshSchreuder, line 199: Since the function report relies on the output of running this, we need to shift the order. I'm not sure if doing so has dependent steps that need to be moved in here like checkout / rm -rf from Generate function report

tools/function_finder/function_finder_psx.py
```diff
@@ -164,13 +172,34 @@ def find_wip(o):
...
+            f = None
+
+        for i, o in enumerate(output):
+            full_match = rf'1.00.+{o[1]}.+{o[0]}';
+            partial_match = rf'0.\d\d.+{o[1]}.+{o[0]}';
+
+            #print(f"Checking full match on {full_match}")
+            if re.search(full_match, dups_text) is not None:
+                #print("Full match found!")
+                o[5] = f"[Full]({base_url}/duplicates.txt#:~:text={o[1]})"
```
JoshSchreuder, line 197: [DOM text fragments](https://developer.mozilla.org/en-US/docs/Web/URI/Fragment/Text_fragments) are pretty fresh and not widely supported but in theory this would let the browser scroll the duplicates list directly to a relevant text match (aka the function name). It works on Chromium browsers but FF seems kind of iffy at the moment

tools/function_finder/function_finder_psx.py
```diff
@@ -180,8 +209,8 @@ def find_wip(o):
 
     # delete asm text
     for o in output:
-        del o[7:]
+        del o[8:]
 
-    headers = ["Ovl", "Function", "Length", "Branches", "Jtbl", "WIP", "%"]
+    headers = ["Ovl", "Function", "Length", "Branches", "Jtbl", f"[Duplicate]({base_url}/duplicates.txt)", "WIP", "%"]
```
JoshSchreuder, line 214: I made the header a link to the base list but this might be redundant. Perhaps this could link to https://github.com/Xeeynamo/sotn-decomp/wiki/Decompilation#duplicate-functions instead

tools/function_finder/function_finder_psx.py
```diff
@@ -164,13 +172,31 @@ def find_wip(o):
...
         )
+
+    if args.match_dups:
+        try:
+            f = open(f"gh-duplicates/duplicates.txt", "r")
+            dups_text = f.read()
+        except:
+            f = None
+
+        if f is not None:
```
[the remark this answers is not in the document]
JoshSchreuder: Done thanks! Been too long since I've done any meaningful Python

tools/function_finder/function_finder_psx.py
```diff
@@ -164,13 +172,31 @@ def find_wip(o):
...
+            for i, o in enumerate(output):
+                full_match = rf'1.00.+{o[1]}.+{o[0]}';
+                partial_match = rf'0.\d\d.+{o[1]}.+{o[0]}';
+
+                if re.search(full_match, dups_text) is not None:
+                    o[5] = f"[Full]({base_url}/duplicates.txt#:~:text={o[1]})"
+                elif re.search(partial_match, dups_text) is not None:
+                    o[5] = f"[Partial]({base_url}/duplicates.txt#:~:text={o[1]})"
```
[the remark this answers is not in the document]
JoshSchreuder: I've renamed to `Part` since `WIP` is already used in the context of scratches and I thought it best not to mix terminology

JoshSchreuder:
Updated with feedback - recommend not merging this unless someone is around to debug the CI 😄

sozud (approved):

sozud:
It looks like this broke the CI, here's the current results saved for reference https://gist.github.com/sozud/a2589a16d059d28d0c6d06c27824f815

To me it looks like it's picking up every function, including the ones already decompiled. This is due to force_extract being used probably. The force_extract needs to be undone before running function_finder since it lists everything it finds in `asm`. We ought to fix these tools to run correctly in splat's `disassemble_all: True` mode so we don't have to extract and rm -rf over and over

JoshSchreuder:
There's also a small bug here where the regex is matching function names partially which I can fix too, eg. `libgte` `NormalColorCol` is showing as a partial duplicate when it's actually `NormalColorCol3`

sozud:
Another thing I notice is that the WIP and % columns are empty, do those show up for you when you run it locally?

JoshSchreuder:
I couldn't get the scratch stuff to work, was getting invalid JSON back from decomp.me, so I had to use the skip flag locally. 
Not sure if it needs auth secrets or what, I didn't play around too much with it

sozud:
I just ran it and it came back fine for me
```
python3 ./tools/function_finder/function_finder_psx.py 
```
It doesn't require secrets. Not sure what's going on in that case

JoshSchreuder:
I think it's a WSL issue most likely, something with the sockets and how many calls it's doing in parallel to decomp.me.
I got it working locally with some exception handling around it.

I was able to work out both the bugs above I think, I'll put a new PR up with that which can be merged after #1689

PR #1687, opened 2024-09-28
Decompile `tt_001` `func_us_80171568`

JoshSchreuder:


src/servant/tt_001/F84.c
```diff
@@ -14,7 +14,34 @@ INCLUDE_ASM("servant/tt_001/nonmatchings/F84", func_us_80171284);
...
+        }
+        entity->zPriority = self->zPriority;
+        entity->facingLeft = self->facingLeft;
+        entity->flags = FLAG_KEEP_ALIVE_OFFCAMERA;
+        entity->posX.val = self->posX.val;
+        entity->posY.val = self->posY.val;
+        entity->ext.factory.parent = self;
+        return entity;
+    }
+}
```
JoshSchreuder, line 44: Seems strange to me that there is a fall-through case here which doesn't return a value, maybe I have something wrong?

sozud: This gets both versions matching with some #ifdefs, maybe it can be cleaned up more?

https://decomp.me/scratch/rbZPM
https://decomp.me/scratch/K9ZjY

sozud: I bet these same differences exist in HD vs. US

JoshSchreuder: Good call, thanks

src/servant/tt_001/F84.c
```diff
@@ -14,7 +14,46 @@ INCLUDE_ASM("servant/tt_001/nonmatchings/F84", func_us_80171284);
 
 void func_us_80171560(void) {}
 
-INCLUDE_ASM("servant/tt_001/nonmatchings/F84", func_us_80171568);
+#if defined(VERSION_PSP)
+void func_psp_092EA460(Entity* self, s32 entityId, s32 params) {
```
JoshSchreuder, line 18: I know this function name can be mapped via symbols later on, but I thought it might be good to explicitly ifdef this because of the different return type and param count

src/servant/tt_001/F84.c
sozud: Also to answer your question there's other spots in the codebase where an undefined return value is apparently required to match, just seems like it was that way originally

sozud (approved):

PR #1689, opened 2024-09-28
Turn on disassemble_all for most yamls

sozud:
This is prep for fixing our tooling and all the force_extract stuff. Weapon0 doesn't seem to match with this setting on for some reason?

### DIFF
--- a/tools/dups/src/main.rs
+++ b/tools/dups/src/main.rs
@@ -120,8 +120,6 @@ Finds duplicates in two asm directories and prints them out in order to identify
 
 Usage:
 
-make force_extract
-
 Do a 2-way compare with ordering
 cargo run --release -- --dir ../../asm/us/st/nz0/nonmatchings/ --dir ../../asm/us/st/np3/nonmatchings/ --threshold .94
 
--- a/tools/dups/src/main.rs
+++ b/tools/dups/src/main.rs
@@ -342,6 +342,10 @@ fn do_dups_report(output_file: Option<String>, threshold: f64) {
         let mut funcs = Vec::new();
         process_directory(&dir, &mut funcs);
 
+        // add matchings dir as well to get decompiled funcs
+        let matchings = dir.replace("nonmatchings", "matchings");
+        process_directory(&matchings, &mut funcs);
+
         // sort functions by vram address
         funcs.sort_by_key(|function| {
             function
--- a/config/splat.hd.dra.yaml
+++ b/config/splat.hd.dra.yaml
@@ -18,6 +18,7 @@ options:
   use_legacy_include_asm: no
   asm_jtbl_label_macro: jlabel
   extensions_path: tools/splat_ext
+  disassemble_all: True
   section_order:
     - ".data"
     - ".rodata"
--- a/config/splat.hd.ric.yaml
+++ b/config/splat.hd.ric.yaml
@@ -20,6 +20,7 @@ options:
   asm_jtbl_label_macro: jlabel
   extensions_path: tools/splat_ext
   symbol_name_format: hd_$VRAM
+  disassemble_all: True
   section_order:
     - .data
     - .rodata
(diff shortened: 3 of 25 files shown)


sozud:
I adjusted the dups tool to use the matchings directory, here's the current output https://gist.github.com/sozud/847e472f78592ca575fe246b80a12345

Weapon is missing since disassemble_all doesn't work with it, but we're almost done with weapon anyway so I think it's fine for now?

sozud:
@JoshSchreuder This is some prep to get your PR merged, once `dups` doesn't require force_extract it should fix your PR

sozud:
@bismurphy Is it possible to make the same change for analyze_calls and we totally get rid of force_extract?

JoshSchreuder:
Should add here too for new overlays?
https://github.com/Xeeynamo/sotn-decomp/blob/master/tools/make-config.py#L427

sozud:
> Should add here too for new overlays? https://github.com/Xeeynamo/sotn-decomp/blob/master/tools/make-config.py#L427

Updated, thanks

JoshSchreuder:
> Weapon0 doesn't seem to match with this setting on for some reason?

Might be something to do with the static values and `symbols.us.weapons.txt.in`?

The object files that don't match are all ones that have these statics at the top of their .c
image

tools/dups/src/main.rs
JoshSchreuder, line 1: https://github.com/Xeeynamo/sotn-decomp/pull/1689/files#diff-24995c27acbb4fd00c28af9a5f389a40e8794e926494f143f10a710619d967c4R123

Small nitpick, this can be removed from usage I think

sozud:
> The object files that don't match are all ones that have these statics at the top of their .c

The files are much larger than previously, maybe https://github.com/encounter/objdiff can say what's different?

tools/dups/src/main.rs
sozud: Fixed

sozud:
Going to merge this, next steps are adjusting analyze_calls to not use force_extract, then removing force_extract from the makefile and CI

JoshSchreuder:
One thing I just noticed since doing this, the decompiled funcs show up both as matching and nonmatching in the duplicates.txt?
https://raw.githubusercontent.com/Xeeynamo/sotn-decomp/gh-duplicates/duplicates.txt

eg.
image

sozud:
> One thing I just noticed since doing this, the decompiled funcs show up both as matching and nonmatchings in the duplicates.txt?

I'm not at a PC where I can look into this for a couple of days. Is that file actually present in both folders? If so seems like a splat issue? Otherwise it could be dups issue

sozud:
Actually now I think about it this is happening due to usage of force_extract probably. We extract normally, which puts files in matchings, then we force_extract, which puts the same file in nonmatchings.

sozud:
I'm wrong apparently. https://github.com/Xeeynamo/sotn-decomp/commit/ea8cb184ad502cd1aab4037e6f0d1d4141fb2d2e

I'll do a debug of this in a couple of days

JoshSchreuder:
Doing that will have the same problem as my other PR as I removed that step there as well. Will need either a `make -j extract` or fix up the function report thingy as you mentioned in your other comment

JoshSchreuder:
Function report is also generating listings for a bunch of .rodata after this
image

I think it's because `disassemble_all` also creates a function for matched rodata because of `migrate_rodata_to_functions` being enabled in most overlays.

Turning `migrate_rodata_to_functions` resolves this for most of the overlays however `tt_001` and `borbo3` need this enabled (also any new overlays before jump tables are migrated?)

Perhaps there's a way to exclude these from the function finder py script

sozud:
I'm not sure about the splat behavior but function_finder should probably have an adjustment to get rid of these like checking if `jr         $ra` exists in the file or similar. I think that's probably the most reliable way to detect a function?

PR #1691, opened 2024-09-28
Add NO0 (Marble Gallery) overlay

bismurphy:
This adds this overlay to the project.

Importantly, we use a new tool (tools/auto_dedupe_new_overlay) to process the new overlay and automatically split its C files according to known duplicate files. As of now, this does not do any decompiling (the whole overlay is kept as INCLUDE_ASM), but it automatically does all the file splits and code copying needed. Hopefully this reduces some amount of duplicated work.

To be clear, this new script is not in any build chain, but is meant to be used after `make-config` to take the overlay from being a single giant lump of `us.c` to being individual files, for the sake of easier deduplicating. It will likely need some revision for future overlays, but at least it should be a good start toward reducing tedious work.

I'll leave this overlay in place like this for about a week in case there are any newcomers who would like to try de-duplicating some of these files and decompiling the new functions; after that week I'll get into doing things myself.


config/splat.us.stno0.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - config/symbols.us.txt
+    - config/symbols.us.stno0.txt
+  undefined_funcs_auto_path: config/undefined_funcs_auto.us.stno0.txt
+  undefined_syms_auto_path: config/undefined_syms_auto.us.stno0.txt
+  find_file_boundaries: true
+  use_legacy_include_asm: false
+  migrate_rodata_to_functions: true
+  asm_jtbl_label_macro: jlabel
+  symbol_name_format: us_$VRAM
+  section_order:
```
sozud, line 22: could we add `disassemble_all: true` for this work? https://github.com/Xeeynamo/sotn-decomp/pull/1689

sozud (approved):

config/splat.us.stno0.yaml
bismurphy: Done, thanks for catching this

PR #1694, opened 2024-09-29
Use collider values to map to enum

JoshSchreuder:
Minor fixes to use the enum values in `dra`, `ric`, `cen`, `no3`, `np3`, `weapon`


sozud (approved):
Looks good, thanks for cleaning this up

PR #1695, opened 2024-09-29
Deduplicate `tt_000` `tt_001` SFX function

JoshSchreuder:
Thought this was going to be more work as duplicate report had it at 93% and with a jump table, as it turns out I think they're the same besides a single global memory address


src/servant/play_sfx.h
```diff
@@ -0,0 +1,71 @@
...
+        break;
+
+    case 4:
+        if (g_api.func_80131F68() == 0) {
+            entity->step++;
+        }
+        break;
+
+    case 5:
+        if ((*(s32*)&D_8003C708.flags & 0x60) != 0) {
```
sozud, line 40: I know this predates your PR but would you mind putting the psp version in a scratch? I suspect it's an optimization and it's actually something like 
```
if(D_8003C708.flags && (D_8003C708.unk2 & 0x60))
```
but that's not it

sozud (approved):
Looks good but would be nice if we can check that one line on PSP

src/servant/play_sfx.h
JoshSchreuder: This function is ifndef'd out of the PSP build so I don't think it's part of that release?

JoshSchreuder: Here's the PSX scratch if that helps
https://decomp.me/scratch/yEfKI

PR #1696, opened 2024-09-29
Decompile `tt_001` `func_us_80170F84`

JoshSchreuder:
PSX: https://decomp.me/scratch/7RyGs
PSP is not being built currently but should match once that's done too: https://decomp.me/scratch/Nbn5X

I've called this one `UpdateEntityVelocityTowardsTarget` based on what it appears to be doing

### DIFF
--- a/src/servant/tt_001/F84.c
+++ b/src/servant/tt_001/F84.c
@@ -6,7 +6,68 @@ extern s32 g_IsServantDestroyed;
 
 #include "../set_entity_animation.h"
 
-INCLUDE_ASM("servant/tt_001/nonmatchings/F84", func_us_80170F84);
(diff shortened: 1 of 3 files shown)


src/servant/tt_001/F84.c
```diff
@@ -6,7 +6,68 @@ extern s32 g_IsServantDestroyed;
 
 #include "../set_entity_animation.h"
 
-INCLUDE_ASM("servant/tt_001/nonmatchings/F84", func_us_80170F84);
+s32 UpdateEntityVelocityTowardsTarget(s32 unused, s32 targetX, s32 targetY) {
```
JoshSchreuder, line 9: There's an unused `arg0` here - maybe a hint that this code is common elsewhere and it is used there

PR #1701, opened 2024-09-29
WRP and CEN cohabitation

Xeeynamo:
Experimental piece of work I feel confident enough to submit. A few things to take in consideration:

* The new `pfn_entity_updates.h` header aims to use the PSP variant of `PfnEntityUpdates` and the two `g_pStObjLayout`, I just tweaked a bit the code from `stage.h`
* Add overlay prefix to the PSX version of `PfnEntityUpdates` and the two `g_pStObjLayout`
* Changed `wrp/create_entity` to `rwrp/create_entity` in RWRP because we need the `RWRP_` prefix to be issued
* Remodernized the WRP header
* I am still not sure what's wrong with `g_HeartDropArray` in `warp.c`. I am pretty sure it is a PSX bug. I opted for the PSP version

There's a `// PfnEntityUpdates = OVL_EXPORT(EntityUpdates);`. If you uncomment and resolve the multiple definitions you will have the two overlays linked. I still need to figure out how to handle the `InitializeEntity` parameters in `e_init` without going mad with the `#ifdef`.

Takeaways from the Xbox 360 build:

* `EntityBreakable` is different in every overlay, so we can use the `OVL_EXPORT` in our decomp
* Every other function seem to be de-duplciated. For those functions that are slightly different between overlays, the developers of the Xbox port opted for `if (g_StageId == XYZ)`

Takeaways from the PSP build:

* god bless the psp


sozud:
Seems reasonable once the compile errors are fixed

sozud (approved):

PR #1707, opened 2024-10-01
Various cleanup to match PSP decomp

Onenutmcgee:
Ran my code for this function against the PSP assembly and fixed a bunch of places where it didn't match.

It's pretty close now, but still missing a bit.  I think it's mostly around the number of parameters and the number of temp variables as it's mostly all in the stack push/pop.

Also removing unnecessary casts.

Scratch
https://decomp.me/scratch/0Gm3x

### DIFF
--- a/src/servant/tt_001/F84.c
+++ b/src/servant/tt_001/F84.c
@@ -408,8 +408,7 @@ void func_us_80171864(Entity* self) {
     self->ext.ghost.unk88 += self->ext.ghost.unk8A;
     if (!(self->ext.ghost.unk88 >= 0x21 &&
           self->ext.ghost.unk88 < 0x5F + 0x21)) {
-        temp_v0 = self->ext.ghost.unk8A;
-        self->ext.ghost.unk8A = -temp_v0;
+        self->ext.ghost.unk8A *= -1;
     }
     self->unk6C = self->ext.ghost.unk88;
     ProcessEvent(self, 0);
--- a/src/servant/tt_001/F84.c
+++ b/src/servant/tt_001/F84.c
@@ -155,8 +155,8 @@ Entity* func_us_80171568(Entity* self, s32 entityId) {
 INCLUDE_ASM("servant/tt_001/nonmatchings/F84", func_us_80171624);
 
 void func_us_80171864(Entity* self) {
-    s32 temp_s4 ;
-    s32 temp_s3 ;
+    s32 temp_s4;
+    s32 temp_s3;
     s32 temp_s2;
     s32 temp_s1;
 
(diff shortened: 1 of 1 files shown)


bismurphy:
Do you have a decomp.me scratch for this? If you share the link we might be able to give some advice on the last bits. I think your temp_v0 is probably a fake variable for example.

Onenutmcgee:
Yes I do:
https://decomp.me/scratch/0Gm3x

I know for sure there is a 2nd parameter (and I'm pretty sure it's before the Entity).  I'm also pretty sure it should get passed to another function.  The rest of the stack push at the top tells me I'm probably missing a temp var, most likely a struct, I think.

bismurphy:
Thanks, looking into it. I can say that I'm pretty sure there is only one argument, what is your indication that there are two?

Onenutmcgee:
The very first line of the target is shifting the stack pointer by 0x30.  I believe it gets shifted once for the call, and once per param.  At least that's what it looks like to me.  I don't know why each parameter adds 16 bytes, but that appears to be the case. 

Then everywhere it's loading from the Entity*, it's doing it at location 0x20(sp) if the entity is the 2nd param.  If you swap them, then it loads from 0x10(sp).  The ASM wants to load the Entity* argument from 0x20(sp).

src/servant/tt_001/F84.c
```diff
@@ -403,7 +411,7 @@ void func_us_80171864(Entity* self) {
         temp_v0 = self->ext.ghost.unk8A;
         self->ext.ghost.unk8A = -temp_v0;
```
JoshSchreuder, line 412: `self->ext.ghost.unk8A *= -1;` matches PSP better I think

bismurphy:
> The very first line of the target is shifting the stack pointer by 0x30. I believe it gets shifted once for the call, and once per param.

While that's true, local variables (in this case, s0 through s4) also shift the stack pointer. We're going to need more locals. Working on figuring out where they belong right now.

src/servant/tt_001/F84.c
Onenutmcgee: Oh, man, that's fantastic.  I was trying to figure that out last night for a while, but I couldn't get it to stop extending the sign flag.  That's the winner for that problem.

bismurphy:
https://decomp.me/scratch/V2ZTq Here's what I was able to get it to, probably pretty close.

Onenutmcgee:
I'll add that to my toolbox.  It's crazy how a little thing like that can cause registers to be used differently.  I wish I understood what causes the compiler to choose register a1 vs v0 as temporary locations.  It seems arbitrary, but it is definitely deterministic.

I'll merge your fixes in.

sozud (approved):
Nice work

PR #1708, opened 2024-10-01
Decompiling function tt_001.func_us_80171624

Onenutmcgee:


src/servant/tt_001/F84.c
```diff
@@ -152,7 +156,89 @@ Entity* func_us_80171568(Entity* self, s32 entityId) {
...
+        e->posX.val = PLAYER.posX.val;
+        e->posY.val = PLAYER.posY.val;
+
+        if (arg0 == 1) {
+            e->entityId = 0xD1;
+            e->posX.val = FIX(128);
+            e->posY.val = FIX(-32);
+        } else {
+            e->entityId = 0xD1;
+            if (D_8003C708.flags & 0x20) {
```
sozud, line 225: STAGE_INVERTEDCASTLE_FLAG

src/servant/tt_001/F84.c
```diff
@@ -152,7 +156,89 @@ Entity* func_us_80171568(Entity* self, s32 entityId) {
...
+        e->animSet = ANIMSET_OVL(20);
+        e->entityId = 0xD1;
+        e->params = 0;
+        e->zPriority = PLAYER.zPriority - 2;
+        e->facingLeft = (PLAYER.facingLeft + 1) & 1;
+        e->posX.val = PLAYER.posX.val;
+        e->posY.val = PLAYER.posY.val;
+
+        if (arg0 == 1) {
+            e->entityId = 0xD1;
```
sozud, line 220: Seems like this is creating an entity within this overlay https://github.com/Xeeynamo/sotn-decomp/blob/cf280c66f5ce0463834aaacf1dc56f4991979d80/src/dra/7A4D0.c#L100

D_80170000 is supposed to point to ServantDesc, so this will create an entity with func_us_80171864 as the update func I think? Assuming I'm correct it would be good to document this

sozud (approved):

src/servant/tt_001/F84.c
Onenutmcgee: I agree.  Also, the bat shares the same entityId as well, which means it's update function is func_80172120.  I was planning on doing some refactoring with things that are the same between tt_000 and tt_001 for my next task.  I'll document this find with that.

This looks like it just fetches the update though.  I don't think it actually creates the entity.  

I think the entity gets created here
https://github.com/Xeeynamo/sotn-decomp/blob/cf280c66f5ce0463834aaacf1dc56f4991979d80/src/dra/loading.c#L130

Which gets called from the Menu and from another part of the game engine.

PR #1709, opened 2024-10-01
Use player status flags where possible

JoshSchreuder:
There's some cases which don't nicely map to the defines in place so I've left as is

- https://github.com/Xeeynamo/sotn-decomp/blob/cf280c66f5ce0463834aaacf1dc56f4991979d80/src/st/dre/succubus.c#L583
    - is this a mask like these? https://github.com/search?q=repo%3AXeeynamo%2Fsotn-decomp%20PLAYER_UNK0C_READY_MASK&type=code


JoshSchreuder:
Perhaps we should rename `PlayerState` `unk0C` -> status or similar? And make an enum instead of `#define`. But can be done separately if it's a good idea

JoshSchreuder:
`tt_000` isn't touched by this I don't think, not sure why PSP build would be failing as a result of this?

JoshSchreuder:
Thanks, that was a remnant from me trying to get deadstripping working last night, fixed

sozud:
> Perhaps we should rename `PlayerState->unk0C` -> `status` or similar? And make an enum instead of `#define`. But can be done separately if it's a good idea

`status` seems good to me. The advantage of an enum is avoiding duplicate values? It's a bit field so we aren't counting up normally like with an enum

sozud (approved):

JoshSchreuder:
> The advantage of an enum is avoiding duplicate values? It's a bit field so we aren't counting up normally like with an enum

I was thinking it would enforce using a value from the enum (more like higher level language enums) but my knowledge of C enums isn't great - I guess under the hood it's still an s32 and people can still use manual values like 0x2 if they want to and the compiler won't enforce anything here.

At the least it would provide tooling support like autocomplete I think. Not sure if the decompiler would be able to use it to infer flag values either, either now or in the future?

sozud:
> I was thinking it would enforce using a value from the enum (more like higher level language enums) but my knowledge of C enums isn't great - I guess under the hood it's still an s32 and people can still use manual values like 0x2 if they want to and the compiler won't enforce anything here.

Yeah you would need C++ strongly typed enums for that https://stackoverflow.com/questions/18335861/why-is-enum-class-considered-safer-to-use-than-plain-enum

We can iterate on the other stuff, I figure let's get this in as an incremental improvement

PR #1712, opened 2024-10-02
Decompile `tt_001` `func_us_80171284`

JoshSchreuder:
PSX: https://decomp.me/scratch/APOek
PSP: https://decomp.me/scratch/T9RKT

Had do some some kinda ugly ifdef PSP stuff here as there is some differences between the two.


src/servant/tt_001/F84.c
```diff
@@ -109,7 +109,104 @@ s32 UpdateEntityVelocityTowardsTarget(s32 unused, s32 targetX, s32 targetY) {
...
+        if (entity->posX.i.hi > 272)
+            continue;
+        if (entity->posY.i.hi > 240)
+            continue;
+        if (entity->posY.i.hi < 0)
+            continue;
+            // Differs from here
+#if defined(VERSION_PSP)
+        if (entity->hitboxState & 8 &&
+            D_us_80170510[D_us_80173810.level / 10][2] == 0)
```
JoshSchreuder, line 146: This is likely some sort of struct, but the access is a bit different between PSP and PSX, so I've just made it like this for now - feedback welcome here if there's a better design that can work for both

src/servant/tt_001/F84.c
```diff
@@ -390,13 +487,13 @@ void func_us_80171864(Entity* self) {
...
-                (D_us_80170508[((D_us_80173810.level / 10) * 3)] - 0x1E)) {
+                (D_us_80170508[D_us_80173810.level / 10][0] - 0x1E)) {
                 self->ext.ghost.unk92 = func_us_80171568(self, 0);
             } else if (self->ext.ghost.unk86 >
-                       D_us_80170508[((D_us_80173810.level / 10) * 3)]) {
+                       D_us_80170508[D_us_80173810.level / 10][0]) {
                 self->ext.ghost.unk86 = 0;
                 g_api.func_8011A3AC(
-                    self, D_us_8017050C[((D_us_80173810.level / 10) * 3)], 1,
+                    self, D_us_8017050C[D_us_80173810.level / 10][0], 1,
```
JoshSchreuder, line 496: These are different memory addresses but the same pattern as my func, likely the same struct. I've updated these for consistency.

My guess is that on PSP side these would also access [2]

include/servant.h
```diff
@@ -79,6 +79,12 @@ extern s16 D_us_801735BC;
 // DistanceToTarget
 extern s16 D_us_801735C0;
 
+extern u32 D_us_801735C4[];
+extern u32 D_us_80170510[][3];
+extern FamiliarStats D_us_80173810;
+extern s32 D_us_80173820;
+extern Entity D_800762D8[256];
```
JoshSchreuder, line 86: PSP indicates this is `g_Entities` but it's a different memory address to that on PSX. Is that normal?

sozud: Not sure I understand, but the memory addresses are totally different everywhere on PSP. https://github.com/Xeeynamo/sotn-decomp/blob/master/config/symbols.pspeu.txt

src/servant/tt_001/F84.c
```diff
@@ -109,7 +109,104 @@ s32 UpdateEntityVelocityTowardsTarget(s32 unused, s32 targetX, s32 targetY) {
...
+
+    if (matches != 0) {
+        index = D_us_80173820 % 128;
+
+        for (i = 0; i < 128; i++) {
+            if (D_us_801735C4[index] != 0) {
+#if defined(VERSION_PSP)
+                entity = &D_800762D8[index + 64];
+#else
+                entity = &D_800762D8[index];
```
sozud, line 198: My guess here is that D_800762D8 is a fake symbol and this is index + 64 on both systems

src/servant/tt_001/F84.c
```diff
@@ -109,7 +109,104 @@ s32 UpdateEntityVelocityTowardsTarget(s32 unused, s32 targetX, s32 targetY) {
...
+        posDelta = self->posX.i.hi - entity->posX.i.hi;
+        if (posDelta < 0) {
+            posDelta *= -1;
+        }
+        if (posDelta >= 49)
+            continue;
+#endif
+
+#if defined(VERSION_PSP)
+        if (abs(self->posY.i.hi - entity->posY.i.hi) >= 49)
```
sozud, line 163: My suspicion is that an abs macro is being used on both systems like
```
#define ABS(x) ((x) < 0 ? -(x) : (x))
```
but I wasn't able to get a match

src/servant/tt_001/F84.c
sozud: I think D_us_80173810 is a fake symbol and the real one is D_us_80173810-0x8. That should make it match in both places

include/servant.h
JoshSchreuder: `g_Entities` on PSX is `0x800733D8` but this array is `0x800762D8`.
https://github.com/Xeeynamo/sotn-decomp/blob/master/config/symbols.us.txt#L973C14-L973C24

The PSP decomp recognised this as `g_Entities` there, but it does not appear to be that same address in the PSX version, unless I'm missing something

sozud: `0x800762D8-0x800733D8 = 0x2F00`
`0x2F00/64 = 0xBC`
sizeof(Entity) is 0xBC
https://github.com/Xeeynamo/sotn-decomp/blob/b9d7e3fd57616a495b1dbdeedd1c01e7ffd63afa/include/game.h#L857
So your address is `&g_Entities[64]`

JoshSchreuder: Aha, thank you

src/servant/tt_001/F84.c
JoshSchreuder: Sorry, tried a few things and couldn't get a match so maybe I'm not understanding exactly what you mean, can you please clarify?

Looks like `D_us_80173810-0x8` (`D_us_8017380C`) is a struct with two s16?
```
glabel D_us_80173808
    /* 3808 80173808 */ .short 0x0000
    /* 380A 8017380A */ .short 0x0000
.size D_us_80173808, . - D_us_80173808
```

sozud: The data gets split up according to what splat detects and isn't necessarily correct. This doesn't work? https://decomp.me/scratch/sc3CM

JoshSchreuder: Yep that worked. The address was already defined and used further on in this file so I think it was conflicting with my other definition in `servant.h` which I've now removed. Thanks!

src/servant/tt_001/F84.c
```diff
@@ -390,13 +479,13 @@ void func_us_80171864(Entity* self) {
...
-                (D_us_80170508[((D_us_80173810.level / 10) * 3)] - 0x1E)) {
+                (D_us_80170508[D_us_80173810.level / 10][0] - 0x1E)) {
                 self->ext.ghost.unk92 = func_us_80171568(self, 0);
             } else if (self->ext.ghost.unk86 >
-                       D_us_80170508[((D_us_80173810.level / 10) * 3)]) {
+                       D_us_80170508[D_us_80173810.level / 10][0]) {
                 self->ext.ghost.unk86 = 0;
                 g_api.func_8011A3AC(
-                    self, D_us_8017050C[((D_us_80173810.level / 10) * 3)], 1,
+                    self, D_us_80170508[D_us_80173810.level / 10][1], 1,
```
JoshSchreuder, line 488: Have corrected this, seems likely they're references into the same struct

src/servant/tt_001/F84.c
JoshSchreuder: Have incorporate this for PSX to clean up the code a little as discussed

PR #1713, opened 2024-10-02
Decompile tt_001 func_us_801720E4

mrmidi:


src/servant/tt_001/F84.c
```diff
@@ -520,7 +520,100 @@ void func_us_801720D4(void) {}
...
+    case 0:
+        self->primIndex = g_api.AllocPrimitives(PRIM_GT4, 8);
+        if (self->primIndex == -1) {
+            DestroyEntity(self);
+            return;
+        }
+
+        self->flags = FLAG_KEEP_ALIVE_OFFCAMERA | FLAG_HAS_PRIMS |
+                      FLAG_UNK_20000; // FLAG_HAS_PRIMS | FLAG_UNK_04000000 |
+                                      // FLAG_UNK_00200000
```
bismurphy, line 548: Can probably remove the comments here?

src/servant/tt_001/F84.c
```diff
@@ -520,7 +520,100 @@ void func_us_801720D4(void) {}
...
+            D_us_801737F0->priority = self->zPriority + 1;
+            D_us_801737F0->blendMode = 0x135;
+
+            D_us_801737F0 = D_us_801737F0->next;
+            D_us_801737E0++;
+        }
+
+        D_us_801737F4 = 0x1E;
+        D_us_801737F8 = 0;
+        g_api.PlaySfx(0x6B0);
```
bismurphy, line 578: All PlaySfx calls should be called with an enum. In this case, this looks like SFX_GLASS_SHARDS.

src/servant/tt_001/F84.c
```diff
@@ -520,7 +520,100 @@ void func_us_801720D4(void) {}
...
+
+    case 1:
+        D_us_801737F8 = (D_us_801737F8 + 0x32) & 0xFFF;
+        D_us_801737F4--;
+        if (D_us_801737F4 < 0) {
+            DestroyEntity(self);
+            return;
+        }
+
+    default:
```
bismurphy, line 589: Is the default label needed here?

src/servant/tt_001/F84.c
```diff
@@ -520,7 +520,100 @@ void func_us_801720D4(void) {}
...
+        self->posX.i.hi = self->ext.factory.parent->posX.i.hi;
+        self->posY.i.hi = self->ext.factory.parent->posY.i.hi;
+
+        D_us_801737E4 = D_us_801737F8;
+        D_us_801737F0 = &g_PrimBuf[self->primIndex];
+        D_us_801737E0 = 0;
+
+        while (D_us_801737E0 < 8) {
```
bismurphy, line 597: This should probably be a for-loop.

bismurphy (commented):
Good start! Just some cleanup comments to bring things more in line with our preferred code style.

src/servant/tt_001/F84.c
```diff
@@ -520,7 +520,91 @@ void func_us_801720D4(void) {}
...
+            D_us_801737F0->v2 = D_us_801737F0->v3 = 0x38;
+
+            D_us_801737F0->r0 = D_us_801737F0->r1 = D_us_801737F0->r2 =
+                D_us_801737F0->r3 = D_us_801737F0->g0 = D_us_801737F0->g1 =
+                    D_us_801737F0->g2 = D_us_801737F0->g3 = D_us_801737F0->b0 =
+                        D_us_801737F0->b1 = D_us_801737F0->b2 =
+                            D_us_801737F0->b3 = 0x80;
+
+            D_us_801737F0->priority = self->zPriority + 1;
+            D_us_801737F0->blendMode = 0x135;
```
bismurphy, line 568: Missed this one. We use drawMode for primitives now (not sure how you even got blendMode here). And that's an enum too.

src/servant/tt_001/F84.c
```diff
@@ -520,7 +520,91 @@ void func_us_801720D4(void) {}
...
+    case 1:
+        D_us_801737F8 = (D_us_801737F8 + 0x32) & 0xFFF;
+        D_us_801737F4--;
+        if (D_us_801737F4 < 0) {
+            DestroyEntity(self);
+            return;
+        }
+    }
+    self->posX.i.hi = self->ext.factory.parent->posX.i.hi;
+    self->posY.i.hi = self->ext.factory.parent->posY.i.hi;
```
bismurphy, line 587: Oh also, we shouldn't be using `ext.factory` for entities that aren't factories.

mrmidi: How to properly write it then?

bismurphy: Add a new entry to the Ext union in `include/entity.h` and use that here, until we can learn more about what this function you've decompiled actually is.

mrmidi: Thanks. How to properly name it then?

bismurphy: You can just name it after the name of this function, look at the other entries in Ext, there are several with numerical names for the unknown functions they're used in

mrmidi: I can't get an OK with this changes

Onenutmcgee: I may already have an ext for this.  Let me check.  I just don't know which Entity* is the correct one.  The ext is self->ext.ghost

bismurphy: Unfortunately that won't work. factory.parent is at offset 0x8C, but that offset is already used in ext.ghost. The entity here is of course related to the ghost, but is not the ghost itself so it uses a different Ext.

Onenutmcgee: Dang.  Yeah, ET_Ghost uses 0x8C as a cycle counter for the shimmer effect (I think).
Looks like there needs to be an entity like batFamBlueTrail, but generic for servants, or the events that servant entities generate such as in [CreateEventEntity](https://github.com/Xeeynamo/sotn-decomp/blob/11631e1475691c68acacedd452c666a7364bd1ea/src/servant/create_event_entity.h#L22)

bismurphy: For now I think it's best to have different Ext entries for different entities. Once we have all the entities decompiled, we can figure out what can be consolidated, and what patterns exist to indicate that two entities are the same.

Onenutmcgee: In the mean time, @mrmidi if you make a copy of [ET_BatFamBlueTrail](https://github.com/Xeeynamo/sotn-decomp/blob/11631e1475691c68acacedd452c666a7364bd1ea/include/entity.h#L661) and call it something like ET_GhostEvent or something like that, add it to the [Ext Union](https://github.com/Xeeynamo/sotn-decomp/blob/11631e1475691c68acacedd452c666a7364bd1ea/include/entity.h#L1893C16-L1893C17), you should be able to replace factory with ghostEvent (or whatever you name it in the union.

bismurphy (approved):
Looks good! Thanks for going through this process.

sozud (approved):

PR #1720, opened 2024-10-03
Refactoring Servant init code

Onenutmcgee:
Lots of cleanup here to make the code easier to read.

1. Adding CLUT index defines to avoid magic numbers
2. Adding InitializeMode enum and hooking that up
3. Updating function names for servant init
4. Cleaning up magic numbers used in tt_000 init function

I wanted to add a enum for the EntityId, but both servants share 0xD1.  Looking at the other EntityIDs definitions, there seems to be some crossover, but it also looks like they may be partly unique across the different binaries.  Probably need more study to figure out how to unify them (or if it's even possible to unify).

We also may want to look at splitting enum definitions and defining flags/fixed values into some sort of enum file structure.  Having them all in game.h is making a large file.  But we would obviously not want to fracture the codebase more.  Either way, that's way beyond the scope of what I wanted to do here.


sozud:
> I wanted to add a enum for the EntityId, but both servants share 0xD1. Looking at the other EntityIDs definitions, there seems to be some crossover, but it also looks like they may be partly unique across the different binaries. Probably need more study to figure out how to unify them (or if it's even possible to unify).

They're most likely going to be specific to each entity like with the stages https://github.com/Xeeynamo/sotn-decomp/blob/6414ab36ab1822401a3b5dcdde16ffae384df4b9/src/st/wrp/wrp.h#L8

sozud (approved):

Onenutmcgee:
> 
> They're most likely going to be specific to each entity like with the stages
> 
Most likely.  But since these two that we have share the same entity id, I'll leave it like this for now, but we can add an enum if other familiars use a different code.

PR #1721, opened 2024-10-03
Rename g_Player.unk0 to g_Player.status

JoshSchreuder:
As discussed in https://github.com/Xeeynamo/sotn-decomp/pull/1709#issuecomment-2386396310


include/game.h
```diff
@@ -1726,7 +1729,7 @@ typedef struct {
 
     /* 80072F24 */ s32 unk04; // copy of the previous field
     /* 80072F28 */ s32 unk08;
-    /* 80072F2C */ u32 unk0C;
+    /* 80072F2C */ PlayerStateStatus status;
```
JoshSchreuder, line 1732: I don't have particularly strong feelings either way, but IMO a small advantage here is using an enumeration might be more self-documenting code than the #define approach.

Doesn't look like we get anything in the way of compile time help by doing this so I'm fine reverting this to #define if preferred.

include/game.h
```diff
@@ -288,40 +288,43 @@ typedef enum {
...
+    PLAYER_STATUS_UNK400000 = 0x400000,
+    PLAYER_STATUS_UNK800000 = 0x800000,
+    PLAYER_STATUS_AXEARMOR = 0x1000000,
+    PLAYER_STATUS_ABSORB_BLOOD = 0x2000000,
+    PLAYER_STATUS_UNK4000000 = 0x4000000,
+    NO_AFTERIMAGE = 0x8000000,
+    PLAYER_STATUS_UNK10000000 = 0x10000000,
+    PLAYER_STATUS_UNK40000000 = 0x40000000,
+    PLAYER_STATUS_UNK80000000 = 0x80000000,
+} PlayerStateStatus;
```
JoshSchreuder, line 327: Not thrilled with this name but there is already an enumeration called `PlayerStatus` used for something else (related to familiars I think?)

sozud (approved):
Seems OK to me but want Xeeynamo's opinion on this one

sozud (commented):
This makes more sense than unk0C, let's just get it in

PR #1723, opened 2024-10-03
More entity cleanup: ET_Generic, renaming, and some Animset improvements

bismurphy:
This is doing a few different things; I'm sort of finding issues and fixing them one by one. I've found more, but don't want this PR to get too bloated.

1) Removal of more uses of ET_Generic, and eliminating a few of its members

2) Renaming EntityUnk15 to EntityGreyPuff

3) Reworking Animset data. It's s16, not u32, so the data in d_2F324.c has been adjusted to match that.


src/dra/6D59C.c
```diff
@@ -797,7 +797,7 @@ s32 CheckChainLimit(s32 itemId, s32 handId) {
         // Hack to load unkAE as an s16 (struct has s8)
         // Longer term, figure out what g_Entites[16-64] are
         // and make dedicated ent extension.
-        if (LOH(entity->ext.generic.unkAE) != itemId) {
+        if (entity->ext.weapon.equipId != itemId) {
```
sozud, line 800: Does weapon.equipId align for all the subweapons? Seems like we want static asserts

bismurphy: This isn't subweapons, this is real weapons. Buffalo stars, etc

sozud: The issue still exists there. We have the anim asserts which mostly covers the problem probably https://github.com/Xeeynamo/sotn-decomp/blob/887e32510bdcaa81acc11748898a4db0d691b405/include/entity.h#L1971

bismurphy:
Realized none of the other weapons even touched their equipId spots, so I just removed them all to avoid them being used inadvertently, so there can never be a conflict and we don't have to worry about making sure they align.

sozud (approved):

PR #1728, opened 2024-10-04
NO0: decompile func_us_801CC8F8

mrmidi:


src/st/no0/4C750.c
```diff
@@ -0,0 +1,106 @@
...
+        entityPtr = (Entity*)((char*)entityPtr + 0xBC);
+    }
+}
+
+extern s32 D_80097408[];
+
+void func_us_801CC8F8(Entity* self) {
+    if (self->step == 0) {
+        InitializeEntity(g_eInitGeneric2);
+        self->animSet = -0x7FFF;
```
sozud, line 63: probably ANIMSET_OVL(some_val)

src/st/no0/4C750.c
```diff
@@ -0,0 +1,106 @@
...
+        self->animCurFrame = 3;
+        self->zPriority = (u16)(D_80097408[0] - 0x54);
```
sozud, line 65: This seems like a fake symbol and part of g_unkGraphicsStruct maybe?

```
g_unkGraphicsStruct = 0x800973F8;
D_80097448          = 0x80097448;
```

src/st/no0/4C750.c
```diff
@@ -0,0 +1,106 @@
...
+    entityPtr = (Entity*)((char*)self + 0xBC);
+
+    for (i = -0x10; i < 0x130; i += 0x60) {
+        DestroyEntity(entityPtr);
+        entityPtr->entityId = 0x16;
+        entityPtr->pfnUpdate = func_us_801CC8F8;
+        entityPtr->posY.i.hi = 0x8E;
+        entityPtr->posX.i.hi = i;
+
+        entityPtr = (Entity*)((char*)entityPtr + 0xBC);
```
sozud, line 54: Looks like entityPtr++?

src/st/no0/4C750.c
```diff
@@ -0,0 +1,106 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "common.h"
+#include "no0.h"
+
+// INCLUDE_ASM("st/no0/nonmatchings/4C750", func_us_801CC750);
```
sozud, line 5: can remove this

src/st/no0/4C750.c
mrmidi: I'm stuck with this a bit. Appreciate the help

Onenutmcgee: I don't think it's enityPtr++, it looks like we're jumping it forward 0xBC.

perhaps
`entityPtr += 0xBC;`

However, the weird cast might make that not correct.  But give that a try.

sozud: sizeof(Entity) is 0xBC https://github.com/Xeeynamo/sotn-decomp/blob/c1b36a102974e1264733e54afec4019b679cdc04/include/game.h#L858

I'm pretty sure it's something like this https://decomp.me/scratch/smg0X
The scracth context is old and needs to be updated to

mrmidi: Not getting OK with that. Something else here I think

PR #1730, opened 2024-10-05
Importing tt_001 clut data

Onenutmcgee:
Again, data is shared with tt_000, so I merged that into a shared header.


bismurphy (approved):
Looks good, nicely done. Great to be able to share code between overlays.

This is totally meaningless, but just from a style point of view "decompile" usually only refers to creating C code for functions. Totally fine though.

Onenutmcgee:
I didn't know what to call it besides decompile.  I agree that is wrong terminology.  I guess maybe importing data sections?

bismurphy:
yeah importing is great; no reason to change this PR or anything, just a terminology nitpick for future, it really doesn't matter though :)

src/servant/tt_000/10E8.c
```diff
@@ -450,12 +448,7 @@ void ServantInit(InitializeMode mode) {
     SpriteParts** spriteBanks;
     Entity* e;
 
-#ifdef VERSION_PC
-    // i exceeds the size of D_80170448
-    const int len = LEN(D_80170448);
```
sozud, line 455: Why was this removed? This is an out of bounds bug

Onenutmcgee: I removed it because if it's there it causes the macOS and linux builds to fail.  I'll add it back in.  Maybe you can help me figure out how to fix

Onenutmcgee: I figured it out.  I had removed an extern variable declaration which was what was breaking those builds.

sozud:
Needs a rebase for conflicts from previous PR

PR #1746, opened 2024-10-05
Add display_animations.py tool for viewing entity animation frames

bismurphy:
This extends upon my work in display_texture.py. I have used this for identifying a few entities already. display_texture is suited for viewing prims, while this one is better for entities that are rendered by RenderEntities.

I think there's enough documentation at the top of the function for others to use it. Hopefully it can be helpful, especially as it evolves over time.

At some point this could be converted to an editor in addition to a viewer, but for now the viewing capability is the core focus.


sozud (approved):
Nice work

PR #1753, opened 2024-10-06
Dedupe `no0` `create_entity` functions

JoshSchreuder:


config/splat.us.stno0.yaml
```diff
@@ -39,6 +39,9 @@ segments:
     subalign: 4
     subsegments:
       - [0x0, data]
+      - [0x778, data, e_laydef]
+      - [0x91C, data, e_init]
+      - [0xC8C, data]
```
JoshSchreuder, line 44: Wasn't technically required for this PR I don't think but I split these while working out the different addresses and lining up with what's in other stage overlays

config/symbols.us.stno0.txt
```diff
@@ -62,11 +62,16 @@ UpdateStageEntities = 0x801C3AB4;
 HitDetection = 0x801C3BBC;
 EntityDamageDisplay = 0x801C4CD4;
 CreateEntityFromLayout = 0x801C5378;
-CreateEntityWhenInHorizontalRange = 0x801C543C;
-FindFirstEntityAbove = 0x801C566C;
-FindFirstEntityBelow = 0x801C56B8;
-CreateEntitiesAbove = 0x801C5710;
-CreateEntitiesBelow = 0x801C580C;
```
JoshSchreuder, line 69: The auto symbol mapping had a hard time matching these up - think @hohle noticed a similar thing on stream with AxeKnightDeath / MakeExplosions.

This is because Above / Right and Below / Left are very close matches so it will just map the first one it finds I think.

sozud: These are similar enough I think you would just need special hardcoded rules in the tooling

sozud (approved):

PR #1756, opened 2024-10-07
Importing TT_001 sbss data

Onenutmcgee:
Importing TT_001 sbss data
Reorganizing the data declaration into ghost.h

### DIFF
--- a/src/servant/tt_001/ghost.h
+++ b/src/servant/tt_001/ghost.h
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
 #include <servant.h>
 
 s16 D_us_801735B0, D_us_801735B0_;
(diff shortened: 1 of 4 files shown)


src/servant/tt_001/ghost.h
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include <servant.h>
+
+s16 D_us_801735B0, D_us_801735B0_;
```
sozud, line 4: What's the reasoning for putting the bss and g_ServantDesc in the header?

Onenutmcgee: I thought it might be a good idea to move the declarations and function prototypes that were ghost specific into it's own header.  I don't have a solid reason for doing so other than that I thought that was how C projects typically are organized.  I can revert that.

Onenutmcgee: I reverted moving those to a header and also found the padding macro.

Onenutmcgee:
I'll mark the data as static in this PR.

PR #1757, opened 2024-10-07
Dedupe `no0` `e_stage_name`

JoshSchreuder:


src/st/no0/prim_helpers.c
```diff
@@ -1,3 +1,2 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
-#include <game.h>
```
JoshSchreuder, line 2: Noticed this randomly, `prim_helpers.h` already includes it so not required

PR #1760, opened 2024-10-07
NZ0 import all data

Xeeynamo:
This comes with a few fake symbols fixed on a few functions, types being corrected when needed and `LEN` being used where I could spot a pattern.

Data in `bossfight` is used in a mix between `bossfight.c`, `slogra.c` and `gaibon.c`, which I believe they were originally one file.

`0x3CB8` is not extracted as an asset. It is an unused blob of data. Using Tile Molester doesn't suggest it is a clear 4bpp image and the decompression tool doesn't work on it. I left it as it is until we know more.

I had to make a few file splits to get a match. I am yet to understand the pattern on how files were originally organised. In some instances it suggests files were split with a C file for every entity. But in other instances it does not seem to be the case. I hope NZ0 will come with a different function order so we will get more hints.

I will probably make a short tool to import the GFX part into a `header.c` for the remaining overlays as it is quite tome consuming to do it manually.

### DIFF
--- a/src/st/nz0/337B4.c
+++ b/src/st/nz0/337B4.c
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
 #include "nz0.h"
 
 // Breakable container holding Skill of Wolf, Bat Card, maybe others
--- a/src/st/nz0/33C3C.c
+++ b/src/st/nz0/33C3C.c
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
 #include "nz0.h"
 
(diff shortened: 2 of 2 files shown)


include/entity.h
```diff
@@ -737,7 +737,7 @@ typedef struct {
 
 typedef struct {
     /* 0x7C */ u8 pad0[0x4];
-    /* 0x80 */ s32 unk80;
+    /* 0x80 */ u8* anim;
```
sozud, line 740: Seems like an alignment bug is possible here but didn't see a clear example in the entity code

sozud (approved):
Nice work!

PR #1762, opened 2024-10-07
Fix some fake symbols and matches in NZ0, NP3 and DRE

Xeeynamo:
I finally solved the mystery behind the `hitbox--; hitbox++` fake match

### DIFF
--- a/src/st/nz0/311C0.c
+++ b/src/st/nz0/311C0.c
@@ -625,7 +625,7 @@ void EntityCannonLever(Entity* self) {
         break;
     }
 
-    if (g_CastleFlags[0x83] != 0) {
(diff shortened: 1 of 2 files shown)


src/st/nz0/311C0.c
```diff
@@ -625,7 +625,7 @@ void EntityCannonLever(Entity* self) {
         break;
     }
 
-    if (D_8003BE6F[0] != 0) {
+    if (g_CastleFlags[0x83] != 0) {
```
bismurphy, line 628: should be `g_CastleFlags[CASTLE_FLAG_131]`. Same for other uses.

bismurphy (approved):
Looks good! Just one enum. Cool to fix this weird fake stuff!

PR #1764, opened 2024-10-07
Adding ghost to buildlist

Onenutmcgee:
Adding ghost to build list

In order to get the build to work, I needed to mark all of the shared functions between tt_000 and tt_001 as static as it was causing linking issues.  Because of this, I needed to combine the two C files for tt_000.  Probably want to look at splitting those up into servant specific and shared, but for now, I just wanted to get this working.  

The issue was msvc didn't like that we were declaring a static function that was not being implemented because part of the shared functions were being implemented in another file.  There may be something we can do with function declarations to clean that up if/when we split it again, but I don't know enough about C static functions to figure it out.


src/servant/servant_private.h
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include <servant.h>
+
+#ifndef VERSION_PSP
+// all of the shared functions need to be marked static
+static s32 AccumulateTowardZero(s32 arg0, s32 arg1);
```
sozud, line 6: I'm thinking something like
```
#ifndef VERSION_PSP
#define STATIC static
#else
#define STATIC
#endif
```
and `STATIC void` etc. would be cleaner?

sozud (approved):
Great work!

src/servant/servant_private.h
Onenutmcgee: yeah, that is way cleaner.  I'll do that.

PR #1766, opened 2024-10-08
Decompile `no0` `func_us_801C2044`

JoshSchreuder:
PSX: https://decomp.me/scratch/jb4U2

PSP: https://decomp.me/scratch/wCOug


src/st/no0/first_c_file.c
```diff
@@ -173,7 +173,37 @@ s16 func_us_801C1F98(Primitive* prim, s16 arg1) {
     return ret;
 }
 
-INCLUDE_ASM("st/no0/nonmatchings/first_c_file", func_us_801C2044);
+s16 func_us_801C2044(Primitive* prim, s16 offset) {
```
JoshSchreuder, line 176: Wasn't too sure what to call this, will try to decompile where it's called from and see if I can work out a good name

src/st/no0/first_c_file.c
```diff
@@ -173,7 +173,37 @@ s16 func_us_801C1F98(Primitive* prim, s16 arg1) {
...
+        prim->v0 = prim->v1 = offset + 6;
+        prim->y0 = prim->y1 = 0x44;
+
+        prim = prim->next;
+        prim->drawMode = DRAW_UNK02;
+        prim->v0 = prim->v1 = 0x50;
+        prim->v2 = prim->v3 = 0x60;
+        prim->y0 = prim->y1 = 0x2C;
+        prim->y2 = prim->y3 = 0x3C;
+        prim->u0 = prim->u2 = 0x22 - offset;
```
JoshSchreuder, line 197: This seems like it may be a bug, or unexpected behaviour if `offset` is > 0x22, since these are `u8`s? 
But maybe that's intentional

sozud (approved):
Looks good, not sure about a name or the 0x22 issue, we can continue documenting as we go

PR #1767, opened 2024-10-08
NO0 e_misc deduplication

mrmidi:
I've stuck with EntityRelicOrb and D_us_801C141C, but all other functions is factored out
#define HEART_DROP_CASTLE_FLAG 0 set to dummy value, I don't know how to get right value yet


sozud:
It looks like you're missing a file
```
src/st/no0/e_misc.c:18:10: fatal error: ../entity_unkId14_spawner.h: No such file or directory
```

sozud (approved):
HEART_DROP_CASTLE_FLAG==0 seems fine to me, it matches up with other overlays that don't have a heart max up I think?

JoshSchreuder:
Yep, seems right to me: https://decomp.me/scratch/QDTfg

PR #1769, opened 2024-10-08
Decompile ST0 func_801B1298

bismurphy:
This one is so dang weird.

It needs to access fields in g_Dialogue which do not match up at all with the g_Dialogue struct. And it accesses some other elements normally, but they show up at the wrong offsets. It's like this one function uses a different g_Dialogue struct. I really don't know what to make of it. Other files in ST0 use normal g_Dialogue, but this one is bizarre.

I went ahead and created an ifdef variable to use to handle this, and that's baked into g_Dialogue, but man. Very weird and yet another thing I haven't seen before in SOTN. ST0 is such a weird beast. Feels like it's from a beta build of the game.

### DIFF
--- a/include/game.h
+++ b/include/game.h
@@ -1455,24 +1455,32 @@ typedef struct {
 } RelicOrb; /* size=0x10 */
 
 typedef struct {
-    /* 0x00 */ u8* scriptCur;         // ptr to dialogue next character
-    /* 0x04 */ s16 startX;            // starting x coord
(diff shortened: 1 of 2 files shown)


sozud (approved):
Looks fine to me, I think the differences make sense since the development process was apparently "copy paste the most recent version of a file and edit it". We can probably trace back which overlays were done first this way

PR #1770, opened 2024-10-09
Refactor/Naming for tt_001

Onenutmcgee:
Cleaning up variable names and function names.
Cleaning up ET_Ghost ext names.

A lot of changes in here, but it's all just renaming mostly inside tt_001.  I also changed out hard coded hex values for decimal values where it made sense to do so (such as frame counts for some of the animation).


Onenutmcgee:
I probably should have broken this up into a couple of PRs, but I was kind of in a groove and had most of it done by the time I thought it may have been too much.

src/servant/tt_001/ghost.c
```diff
@@ -375,62 +390,67 @@ void UpdateServantDefault(Entity* self) {
...
         self->zPriority = PLAYER.zPriority - 2;
     }
     self->hitboxWidth = 0;
     self->hitboxHeight = 0;
+
+    // I think if step is < 2, the target the ghost is moving to is Alucard
     if (self->step < 2) {
+        // I don't believe this is the correct flag.  This appears to be a flag
+        // with the foreground layer, not the stage.
         if (D_8003C708.flags & STAGE_INVERTEDCASTLE_FLAG) {
```
sozud, line 404: Maybe? The main purpose seems to be loading graphics https://github.com/Xeeynamo/sotn-decomp/blob/6301e7d7cec9919bef5da074ee63eb934b0ea605/src/dra/5087C.c#L1664

This place seems odd since it refers to the succubus, so maybe it's not inverted castle related. https://github.com/Xeeynamo/sotn-decomp/blob/6301e7d7cec9919bef5da074ee63eb934b0ea605/src/dra/5087C.c#L997

Xeeynamo would know better

src/servant/tt_001/ghost.c
```diff
@@ -258,18 +273,18 @@ Entity* func_us_80171568(Entity* self, s32 entityId) {
...
+            entity->entityId = ENTITY_ID_CONFUSION;
         }
 #endif
         entity->zPriority = self->zPriority;
         entity->facingLeft = self->facingLeft;
         entity->flags = FLAG_KEEP_ALIVE_OFFCAMERA;
         entity->posX.val = self->posX.val;
         entity->posY.val = self->posY.val;
-        entity->ext.factory.parent = self;
+        entity->ext.ghostEvent.parent = self;
```
sozud, line 287: I haven't checked this case but a common source of bugs is pointer alignment between parent and child entities

sozud (approved):

src/servant/tt_001/ghost.c
Onenutmcgee: Used in this context though, we're setting the target location that the ghost is traveling to.  If that flag is for inverted castle, it essentially hard codes the ghost's target location.  That and ServantUnk0 does some logic that does switch if we are in the inverted castle.  It seems really weird in this context and is probably something to look into at some point, but for now, I just wanted to document it once we have a better handle on what that FgFlags struct is used for elsewhere.

I'm trying to think of a point in the game where the ghost would want to be essentially hard coded to go to the center of the screen, but I'm not sure

src/servant/tt_001/ghost.c
Onenutmcgee: I'm not exactly sure what you are meaning, but I added a static assert in entity.h for that field.  If there is something else you'd like me to do, you'll have to let me know.

config/symbols.pspeu.tt_000.txt
```diff
@@ -32,7 +32,7 @@ IsMovementAllowed = 0x092ebfb8;
 ServantUnk0 = 0x092ec150;
 g_ServantClut = 0x092EC280;
 D_80170040 = 0x092EC2E0;
-D_801704A8 = 0x092EC768;
+DefaultBatAnimationFrame = 0x092EC768;
```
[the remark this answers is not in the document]
Onenutmcgee: whoops, when I was fixing those names I forgot to fix the ones in tt_000

src/servant/tt_001/ghost.c
sozud: We regularly have an issue where things get misaligned like:

```
addr  | parent obj    | child obj
-----------------------
0x10  | child*        | 
----------------------
0x18  |               | parent*
```

So when the child tries to do self->parent it crashes. Was just giving a warning about it, didn't investigate this case deeply

src/servant/tt_001/ghost.c
Onenutmcgee: Ok, I'll add an unknown flag and document that it's the same number as inverted castle flag, but used in a different context.  That way we could track it back if it turns out it should be inverted castle.

Onenutmcgee: I added a couple of flags for where that flag value gets set to the rect.params of the layer definition, but I didn't go replace the hex values everywhere that the flag is checked.  That's probably out of scope for this PR

PR #1772, opened 2024-10-09
Make separate struct for 30030 g_Dialogue

Xeeynamo:
Follow-up to #1769

### DIFF
--- a/src/st/st0/30030.c
+++ b/src/st/st0/30030.c
@@ -469,6 +469,8 @@ void PrologueScroll(void) {
     }
 }
 
+// This file uses g_Dialogue with a different struct than the usual.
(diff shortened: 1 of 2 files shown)


src/st/st0/30030.c
```diff
@@ -471,22 +469,23 @@ void PrologueScroll(void) {
     }
 }
 
+extern Dialogue2 g_Dialogue;
```
bismurphy, line 472: Might be worth a comment here to make it super obvious that g_Dialogue is being used weirdly in this function. Otherwise, this is a good solution to a weird problem. Nice to avoid polluting the main Dialogue struct.

bismurphy (approved):

sozud (approved):

PR #1773, opened 2024-10-10
Dedupe remaining `no0` `e_misc` functions and split data

JoshSchreuder:
Notes for `no0`:

* `unk_collision_func4.h` is not used here
* `unk_recursive_primfunc_1` and `unk_recursive_primfunc_2` and their associated data are not used here

I've stage def'd these out, not sure if this is going to lead to messy shared headers going forwards if we have to keep doing this for other stages


sozud (approved):
I think regarding the `STAGE_IS_NO0` we can just continue seeing how messy it gets and decide to pivot or not

PR #1776, opened 2024-10-11
Decompile `no0` `EntityBreakable`

JoshSchreuder:
PSX: https://decomp.me/scratch/ouTpd
PSP: https://decomp.me/scratch/pZMLg


src/st/no0/first_c_file.c
```diff
@@ -145,7 +145,36 @@ void func_801B0AA4(Entity* self) {
...
+            if (entityDropItem != NULL) {
+                CreateEntityFromCurrentEntity(E_EXPLOSION, entityDropItem);
+                entityDropItem->params =
+                    g_eBreakableExplosionTypes[breakableType];
+            }
+            ReplaceBreakableWithItemDrop(entity);
+        }
+    } else {
+        InitializeEntity(g_eBreakableInit);
+        entity->zPriority = g_unkGraphicsStruct.g_zEntityCenter - 0x14;
```
JoshSchreuder, line 172: This particular line doesn't compile on PSP because `g_zEntityCenter` is a `Multi` union. But seems like this particular struct is still under development so we can probably sort it out later on unless someone knows how to resolve this.

JoshSchreuder:
Added a commit to split and extract the breakable data too as an exercise for myself. Hopefully it doesn't make this too hard to review, I can pull it into a subsequent PR if preferred.

sozud:
Am I missing something or is the sfx part the only difference? https://decomp.me/scratch/hLkzJ
Was something else making it hard to dedupe?

JoshSchreuder:
> Am I missing something or is the sfx part the only difference? https://decomp.me/scratch/hLkzJ
Was something else making it hard to dedupe?

I don't mind deduping in this way, but different stages use different sound effects so I'm not sure if the conditionals will get a bit ugly if I do so.

* [MAD uses SFX_CANDLE_HIT](https://github.com/Xeeynamo/sotn-decomp/blob/af89c6247c65d85eb0ceeda7ee9095a18c8f3928/src/st/mad/D8C8.c#L175)
* [NO3 uses SFX_CANDLE_HIT](https://github.com/Xeeynamo/sotn-decomp/blob/af89c6247c65d85eb0ceeda7ee9095a18c8f3928/src/st/no3/377D4.c#L101)
* [DRE uses SFX_CANDLE_HIT](https://github.com/Xeeynamo/sotn-decomp/blob/af89c6247c65d85eb0ceeda7ee9095a18c8f3928/src/st/dre/11A64.c#L106)
* [The header uses SFX_BREAKABLE_HIT](https://github.com/Xeeynamo/sotn-decomp/blob/af89c6247c65d85eb0ceeda7ee9095a18c8f3928/src/st/e_breakable.h#L21) and is used in MAR, NP3, and WRP

JoshSchreuder:
Oh whoops, I just noticed `SFX_BREAKABLE_HIT` is just a macro to `SFX_CANDLE_HIT`, so all of these could use the header. I'll update this to point to the modified header as suggested

JoshSchreuder:
Updated, thanks for the suggestion!

PR #1779, opened 2024-10-11
Fix issues with running under MSVC

JoshSchreuder:
Fixed some uninitialised pointers / return types which affected a few things at runtime - down on main menu and leaving through door.

Fixed some build errors with WANT_LIBSND_LLE due to unnamed parameters in implemented functions. Still working on getting it building locally with this flag enabled but this gets a little further.


sozud (approved):
Thanks for fixing these, this should be added to the CI build matrix but doesn't have to be part of this PR

PR #1780, opened 2024-10-12
Dedupe `no0` and `nz0` `EntityAxeKnight` shared code

JoshSchreuder:
* Moved `entity_axe_knight` code from `nz0` into shared header
* Dedupe `no0` with shared header
* Split and extract `no0` axe knight data


src/st/no0/no0.h
```diff
@@ -26,46 +26,17 @@ typedef enum EntityIDs {
...
-    /* 0x4C */ E_ZOMBIE = 0x4C,
-    /* 0x59 */ E_ID_59 = 0x59,
-    /* 0x5B */ E_DEATH_5B = 0x5B,
-    /* 0x5D */ E_FALLING_ROCK = 0x5D,
-    /* 0x5E */ E_DEATH_5E,
-    /* 0x60 */ E_FLYING_OWL_UNK60 = 0x60,
-    /* 0x62 */ E_WARG_EXP_OPAQUE = 0x62,
+    /* 0x15 */ E_GREY_PUFF,
+    /* 0x48 */ E_AXE_KNIGHT_AXE = 0x48,
+    /* 0x49 */ E_WARG_EXP_OPAQUE
```
JoshSchreuder, line 33: Cleaned these up as they were copied from `no3` and weren't right - I removed the ones that aren't currently in use so they can be added back later properly

src/st/no0/e_axe_knight.c
```diff
@@ -0,0 +1,42 @@
...
+};
+static u8 hitbox_lookup[] = {
+    0, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 1, 1, 1, 1, 1, 3, 1, 1, 1, 3, 1,
+    1, 1, 1, 1, 1, 1, 1, 3, 3, 3, 3, 3, 1, 1, 1, 1, 1, 3, 1, 1, 1, 3, 1, 0};
+static u8 steps[] = {AXE_KNIGHT_STANDING_THROW, AXE_KNIGHT_DUCKING_THROW,
+                     AXE_KNIGHT_STANDING_THROW, AXE_KNIGHT_DUCKING_THROW,
+                     AXE_KNIGHT_STANDING_THROW, AXE_KNIGHT_DUCKING_THROW,
+                     AXE_KNIGHT_STANDING_THROW, AXE_KNIGHT_DUCKING_THROW};
+static u32 init_velocity_x[] = {FIX(2), FIX(2), FIX(1)};
+static u32 init_velocity_y[] = {FIX(0), FIX(0), FIX(-4)};
```
JoshSchreuder, line 40: This data is the same on `no0` and `nz0` - wasn't sure if data goes into shared headers or not? 
But happy to move that into `entity_axe_knight.h` as well if that's the right thing to do

src/st/no0/41F98.c
```diff
@@ -108,6 +108,22 @@ void func_us_801C3308(Entity* self) {
     DestroyEntity(self);
 }
 
-INCLUDE_ASM("st/no0/nonmatchings/41F98", func_us_801C3334);
+void AxeKnightDeath() {
```
JoshSchreuder, line 111: This feels incorrectly named, looks more likely to be common death code across a few entities like axe knight and warg? But I've just copied this in here for now

src/st/no0/e_axe_knight.c
sozud: It's a tradeoff, it makes deduping harder to do incrementally since you have to do the data at the same time but probably makes more sense long-term

src/st/no0/41F98.c
sozud: If it's this enemy it doesn't seem like the death explosions have much to with the wargs? 

https://www.youtube.com/watch?v=5HnhPNS0rqM&t=1007s

vs.

https://youtu.be/5HnhPNS0rqM?t=438

Looks like there's dedicated code for it in the entity function https://github.com/Xeeynamo/sotn-decomp/pull/1780/files#diff-18a47bb649a82f4e70725d56ace255bf4dd4d6a7a41e758bd3575b7c7d28a98aR357

sozud (approved):

src/st/no0/41F98.c
JoshSchreuder: Was going off the fact this code is using `E_WARG_EXP_OPAQUE` and `entity->ext.wargpuff`, also based on where it sits in memory (pretty far from the other axe knight code which is all grouped up together) so my hypothesis was this was some kind of shared death code, but you may be right

config/symbols.us.stno0.txt
```diff
@@ -52,17 +53,27 @@ g_olroxDroolCollOffsets = 0x80181BFC;
...
+sensors_move = 0x80182478;
+sensors_ground = 0x80182480;
+dead_particle_pos = 0x80182490;
+unused = 0x80182492;
+anim_walk = 0x80182508;
+anim_throw_duck = 0x80182518;
+anim_throw_stand = 0x8018253C;
+anim_die = 0x80182560;
+hitboxes = 0x8018256C;
+hitbox_lookup = 0x8018257C;
```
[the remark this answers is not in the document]
JoshSchreuder: Thought the remove orphans thing in format would do it but I guess not. 

I can do - it's a pity once the data gets excluded with `.data` these symbols don't still appear in the assembly output from splat

include/entity.h
```diff
@@ -1965,6 +1965,18 @@ typedef union { // offset=0x7C
...
+    AXE_KNIGHT_INIT,
+    AXE_KNIGHT_IDLE,
+    AXE_KNIGHT_WALK_TOWARDS_PLAYER,
+    AXE_KNIGHT_WALK_AWAY_FROM_PLAYER,
+    AXE_KNIGHT_STANDING_THROW,
+    AXE_KNIGHT_DUCKING_THROW,
+    AXE_KNIGHT_UNUSED, // Charge Attack missing step from the blue AxeKnight
+    AXE_KNIGHT_ARCING_THROW, // Unused, present in the blue AxeKnight
+    AXE_KNIGHT_DYING,
+} EntityAxeKnightSteps;
```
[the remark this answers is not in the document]
JoshSchreuder: The data and the functions were both using it so I think I would have to move the data into the shared header to do so I think? Sounds like that's the best thing to do here?

config/symbols.us.stno0.txt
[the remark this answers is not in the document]
sozud: > these symbols don't still appear in the assembly output from splat

It doesn't show up in the `matchings` folder? Not sure if you mean something else

PR #1787, opened 2024-10-13
Replace g_HeartDropArray with PrizeDrops

Xeeynamo:
I discovered that `g_HeartDropArray` was always very weirdly used in overlays such as WRP, CEN and DRE where there are no prizes dropped from breakable objects. This lead me to discover that every overlay has a prize array that describes what items to drop. But for WRP CEN and DRE since the array has length zero, the symbol overlaps with what comes right after.

The array is also different for every overlay. The array will be renamed as `OVL_EXPORT(PrizeDrops)`. This PR aims to adjust such array in all overlays:
* [x] CEN
* [x] DRE
* [X] ~MAD~ cannot be moved
* [x] NO0
* [x] NO3
* [x] NP3
* [x] NZ0
* [X] ~ST0~ not existent
* [X] WRP
* [X] RWRP
* [x] MAR

### DIFF
--- a/src/st/wrp_psp/bss.c
+++ b/src/st/wrp_psp/bss.c
@@ -12,7 +12,7 @@ s32 D_80180648;
 u8 g_EntityUnkId12Data[8];
 
 // e_misc.c
-u8 g_HeartDropArray[8]; // different than the g_HeartDropArray in warp.c ???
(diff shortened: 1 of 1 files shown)


sozud (approved):
Nice work figuring this out

PR #1789, opened 2024-10-15
Refactor/renaming in tt 000 part deux

Onenutmcgee:
Finished renaming tt_000.
Code is actually pretty readable.

I may have gotten a little verbose with some of the variable names, but there is a bit of redundancy/crossover between the different sub entities.


Onenutmcgee:
Looks like I borked up the PSP build.  I have a good idea of which change it was.  I'll have to clean it up tomorrow

sozud (approved):

PR #1791, opened 2024-10-15
Ensure decompiled function is treated as the initial comparison basis in dupe report

JoshSchreuder:
This should prevent cases like this:
image

Where the nonmatching function is treated as the comparison for similarity.

After fixing:
image

### DIFF
--- a/tools/dups/src/main.rs
+++ b/tools/dups/src/main.rs
@@ -67,7 +67,7 @@ fn parse_instructions(input: &str, dir: &str, file: &str) -> Function {
         dir: dir.to_string(),
         file: file.to_string(),
         similarity: 0.0,
-        decompiled: false,
+        decompiled: file.contains("/matchings/"),
     }
 }
 
@@ -171,7 +171,6 @@ fn process_directory_for_include_asm(dir: &str) -> Vec<IncludeAsmEntry> {
                     let line_str = line.unwrap();
 
                     if line_str.contains("INCLUDE_ASM") {
-
                         if let Some(captures) = re.captures(&line_str) {
                             let (full, [asm_dir, asm_file]) = captures.extract();
                             output.push(IncludeAsmEntry {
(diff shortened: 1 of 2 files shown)


tools/dups/src/main.rs
```diff
@@ -348,27 +347,20 @@ fn do_dups_report(output_file: Option<String>, threshold: f64) {
...
+        process_asm_directory(&nonmatchings, &mut files);
     }
 
+    files.sort_by(|a, b| {
+        // First, sort by whether the name contains "nonmatching" (false before true)
+        a.name
+            .contains("nonmatching")
+            .cmp(&b.name.contains("nonmatching"))
+            .then_with(|| a.name.cmp(&b.name)) // Then, sort alphabetically by name
+    });
```
JoshSchreuder, line 362: I think this should also give more stable ordering to the duplicate report since it's now alphabetical

sozud:
Could you post a gist of the full new report for comparison?

JoshSchreuder:
@sozud https://gist.github.com/JoshSchreuder/a89a6a689be5964e931187ed04586cde

sozud:
I think this is regressed regarding the naming? Compare
https://github.com/Xeeynamo/sotn-decomp/blob/9079bcc80088f22d0c3b17e674ec78b87c8cd3e3/duplicates.txt#L1403
https://gist.github.com/JoshSchreuder/a89a6a689be5964e931187ed04586cde#file-duplicates-txt-L1433
BlinkItem reverted back to generic names for example

sozud:
I probably know what's happening. Since we prune our symbol files splat doesn't know the names of the functions anymore. I suppose force_symbols before extracting might fix it?

JoshSchreuder:
Yep, think that worked, I pushed an additional commit to add that to the workflow.

Here's the output tested locally:

* make clean
* make force_symbols
* make -j extract
* run report

https://gist.github.com/JoshSchreuder/729a1dd1057e38083013b589ae712e03

Hopefully this doesn't cause knock on effects later on in the workflow 🤞

sozud (approved):

sozud:
It seems it didn't work https://github.com/Xeeynamo/sotn-decomp/blob/50bad9b617e8195b67328dd4ea33252d7a3cb9dd/duplicates.txt#L1433

JoshSchreuder:
Once again I am perplexed by the CI/CD... from the job it seems like the symbols extract ran and off a clean repo this should have worked. Feel free to revert out and I'll take another look at it in a bit

PR #1793, opened 2024-10-16
Finishing refactor for tt_000 ET_Bat and updating game api call

Onenutmcgee:
updating global api function func_8011A3AC with a better name.  It fetches the current familiar stats from the global game variables and will also fetch an attack descriptor if ID is provided.

 Refactoring names for ET_Bat that I forgot to update in my last PR.  Remaining unkXX vars appear to be unused.

Adding enum for g_SpellDefs table.  Some familiar attacks appear to be in that table as well as Alucard's spells.  Some do not such as the bat's fireball.


include/entity.h
```diff
@@ -593,25 +593,25 @@ typedef struct {
...
     /* 0x82 */ s16 batIndex;
-    /* 0x84 */ s16 unk84;
+    /* 0x84 */ s16 randomMovementAngle;
     /* 0x84 */ s16 targetAngle;
-    /* 0x88 */ s16 unk88;
+    /* 0x88 */ s16 randomMovementScaler;
     /* 0x8A */ s16 maxAngle;
     /* 0x8C */ s16 frameCounter;
-    /* 0x8e */ s16 unk8E;
+    /* 0x8e */ s16 doUpdateCloseAnimation;
```
JoshSchreuder, line 601: Is changing stuff like this to `bool` where you're using value `false` possible or does it cause a mismatch?

src/servant/tt_000/bat.c
```diff
@@ -479,13 +479,13 @@ void SwitchModeInitialize(Entity* self) {
                 s_BatPathingPoints[self->ext.bat.batIndex][i].y =
                     PLAYER.posY.i.hi + self->ext.bat.cameraY;
             }
-            self->ext.bat.unkA8 = 0;
+            self->ext.bat.hasShotFireball = 0;
```
JoshSchreuder, line 482: Can this be `false` / `true` also to communicate intent more clearly?

include/entity.h
Onenutmcgee: Type mismatch. I believe that bools map to u32. But that’s just for the PSX compiler. Other compilers could be different I think. Probably better to state the width but use true/false enum to indicate. Perhaps a naming convention to append b or something for values used as bools.

src/servant/tt_000/bat.c
Onenutmcgee: Yes, this should use true/false. I’ll fix that.

include/entity.h
sozud: I'm pretty sure enum maps to s32, you'd have to have a different "short bool" type
https://github.com/Xeeynamo/sotn-decomp/blob/6b544663ff3b78cc2b070e4b3acfceda04dbe5f5/include/types.h#L38

sozud (approved):

include/entity.h
Onenutmcgee: I think that having a short bool type is probably a good idea from a readability standpoint, but if we want to do that, it might be better to locate the places in the codebase that would use it and do it in one shot so we have something consistent.  I'll look into it, but it seems like it could be a lot of work for very little gain.  Proper naming also makes things clear for a bool (either staring with "is" or "has" or something like that)

PR #1797, opened 2024-10-16
Add RBO3

hohle:


sozud (approved):

sozud:
Not sure but it seems like this broke the page 
image

sozud:
ovlmeta.name isn't defined? https://github.com/Xeeynamo/sotn-decomp/blob/81e9afcd6574e297e3c550d0fb660d7fc318139e/src/ChartsVersion.js#L39

PR #1798, opened 2024-10-16
Remove the function calls chart from CI workflow

JoshSchreuder:
Should fix issues with symbols in subsequent steps related to the use of `force_extract`

### DIFF
--- a/.github/workflows/validate-and-report.yml
+++ b/.github/workflows/validate-and-report.yml
@@ -76,12 +76,6 @@ jobs:
         run: make -j build
       - name: Check if they match
         run: make check
-      - name: Analyze calls dry run
-        if: matrix.version == 'us'
-        run: |
-          make force_symbols
-          make force_extract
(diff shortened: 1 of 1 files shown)


.github/workflows/validate-and-report.yml
```diff
@@ -185,16 +179,8 @@ jobs:
...
-      - name: Generate function calls chart
-        run: |
-            make force_symbols
-            make force_extract
-            rm -rf gh-duplicates/function_calls/ || true
-            .venv/bin/python3 tools/analyze_calls.py --output_dir=gh-duplicates/function_calls
-            git clean -fdx asm/
       - name: Generate duplicates report
         run: |
-          make clean
```
JoshSchreuder, line 197: I don't _think_ this is required now, but can add it back if it turns out to be

sozud (approved):

sozud:
I think the make clean needs to stay https://github.com/Xeeynamo/sotn-decomp/actions/runs/11375419531/job/31646103319

PR #1799, opened 2024-10-17
Fix issue with line split INCLUDE_ASM not being picked up by dups report

JoshSchreuder:
Noticed this in the workflow output:

```
 checking "../../src/boss/rbo3/create_entity.c"
Failed to match regex on line: INCLUDE_ASM(
Failed to match regex on line: INCLUDE_ASM(
```

This is because of the linting line breaks inserted to enforce column width
https://github.com/Xeeynamo/sotn-decomp/blob/master/src/boss/rbo3/create_entity.c#L6-L7

For INCLUDE_ASM lines this code checks if it contains a `;`, otherwise it will attempt to buffer subsequent lines until `;` is found to get the full INCLUDE_ASM. Also updated the regex slightly to allow for an optional line break component

Gist output: https://gist.github.com/JoshSchreuder/bf8621628640e9da96de2a565c982d00

### DIFF
--- a/tools/dups/src/levenshtein_hashmap.rs
+++ b/tools/dups/src/levenshtein_hashmap.rs
@@ -122,11 +122,19 @@ mod tests {
             name: String::from("func1"),
             ops: vec![],
             key: vec![1, 2, 3],
+            decompiled: false,
+            dir: String::from("testdir1"),
+            file: String::from("testfile1"),
+            similarity: 0.0,
         };
         let func2 = Function {
             name: String::from("func2"),
             ops: vec![],
             key: vec![1, 2, 3],
+            decompiled: false,
+            dir: String::from("testdir2"),
+            file: String::from("testfile2"),
+            similarity: 0.0,
         };
 
         map.insert(func1.key.clone(), func1.clone());
@@ -144,11 +152,19 @@ mod tests {
             name: String::from("func1"),
             ops: vec![],
             key: vec![1, 2, 3],
+            decompiled: false,
+            dir: String::from("testdir1"),
+            file: String::from("testfile1"),
+            similarity: 0.0,
         };
         let func2 = Function {
             name: String::from("func2"),
             ops: vec![],
             key: vec![4, 5, 6],
+            decompiled: false,
+            dir: String::from("testdir2"),
+            file: String::from("testfile2"),
+            similarity: 0.0,
         };
 
         map.insert(func1.key.clone(), func1.clone());
--- a/tools/dups/src/main.rs
+++ b/tools/dups/src/main.rs
@@ -1,8 +1,6 @@
 use regex::Regex;
-use std::env::*;
 use std::fs;
 use std::fs::File;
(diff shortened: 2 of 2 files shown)


sozud (approved):
We might want a unit test but not required.

I added `make clean` to the CI which fixes the dups report failing, but we don't have symbols since it blows away the `build` directory. force_symbols requires the built elfs to extract the symbols from.

JoshSchreuder:
> I added make clean to the CI which fixes the dups report failing, but we don't have symbols since it blows away the build directory. force_symbols requires the built elfs to extract the symbols from.

Perhaps just an `rm` on what needs to be cleaned up in the workflow is enough? 
What needs to be removed to stop the report failing?

sozud:
It hits this

```
Unable to read directory: No such file or directory (os error 2)
Directory path: ../../asm/us/main/matchings/
```

https://github.com/Xeeynamo/sotn-decomp/blob/c5fb025543e0631f3d711f700ef711f5f912fbb0/tools/dups/src/main.rs#L104

So something is in the repo that is preventing `matchings` from getting extracted?

JoshSchreuder:
Added a couple of tests for this - had to also update one of the other imported files which had failing tests due to missing properties, they're all passing now too.

Good call! I realised in doing this there was a bug with my original code

JoshSchreuder:
> So something is in the repo that is preventing matchings from getting extracted?

My best guess is it might be this?
https://github.com/Xeeynamo/sotn-decomp/actions/runs/11376269287/job/31648454733

Remove clutter step does `Run rm -rf build/us/asm build/us/src build/us/assets`

Then the result of that job gets used as an artifact for generate-duplicates in obtain built binaries step? It's hard to debug this without being able to access the blob @  https://productionresultssa18.blob.core.windows.net/actions-results/e9e85f6d-f4bd-4c03-b5db-6948b47b71cd/workflow-job-run-2df5073e-bfb9-5ea3-c585-7be1eae326da/artifacts/6677237b7523e9d00884ec080bd67f1ffaecc17d8c9c47dfb76f5ef2a8e2af19.zip which would let me run the script on the same folder structure as the workflow is

JoshSchreuder:
The failed build tried to check `2024-10-16T23:27:59.1113085Z checking "../../asm/us/dra/nonmatchings/93290/padding_CdSoundCommand12.s"`

I don't have this as a nonmatching locally, any idea what would have caused this to generate from the CI build?
The one that runs with `make clean` doesn't have any nonmatchings in `nonmatchings/93290`

sozud:
Not sure about your investigations but I think the next step is probably adding a `find .` before extracting to see what the actual state of the repo is in the CI. I tried reworking it but it failed in the same way. https://github.com/Xeeynamo/sotn-decomp/commit/ad21d2497d04c34d0efc40e3d21c74ca267b8303

I'm not sure if more prints could be added to `dups` to figure this out or not

JoshSchreuder:
This should be good to merge now I think

PR #1805, opened 2024-10-18
Decompile RBO3 Functions

hohle:
Decompiles `func_us_80191438`, `func_us_8019179C`, and `func_us_80192B38`.

Imports `func_us_80192020` and `func_us_80192998` from #557.

### DIFF
--- a/src/boss/rbo3/rbo3.c
+++ b/src/boss/rbo3/rbo3.c
@@ -1,13 +1,499 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-#include "common.h"
-#include "stage.h"
+#include "rbo3.h"
 
-INCLUDE_ASM("boss/rbo3/nonmatchings/rbo3", func_us_80191438);
+extern u16 D_us_80180468;
+extern s8 D_us_80190E80;
+extern s8 D_us_80190E8C;
+extern s8 D_us_80190E98;
 
-INCLUDE_ASM("boss/rbo3/nonmatchings/rbo3", func_us_8019179C);
+void func_us_80191438(Entity* self) {
+    s16 params;
+    s16 y;
 
-INCLUDE_ASM("boss/rbo3/nonmatchings/rbo3", func_us_80192020);
+    params = self->params;
+    FntPrint(&D_us_80190E80, params);
+    FntPrint(&D_us_80190E8C, g_Tilemap.left);
+    FntPrint(&D_us_80190E98, g_Tilemap.right);
+
+    switch (self->step) {
+    case 0:
+        InitializeEntity(&D_us_80180468);
+        self->animSet = 2;
+        self->animCurFrame = 1;
+        self->zPriority = 176;
+        break;
+
+    case 1:
+        y = self->posY.i.hi - PLAYER.posY.i.hi;
+        y = abs(y);
+
+        if (y < 32) {
+            switch (params) {
+            case 0:
+                if (g_PlayerX < 384) {
+                    g_Tilemap.x = 384;
+                    g_Tilemap.left++;
+                    self->step++;
+                }
+                break;
+
+            case 1:
+                if (g_PlayerX > 640) {
+                    g_Tilemap.width = 640;
+                    g_Tilemap.right--;
+                    self->step++;
+                }
+                break;
+
+            case 2:
+                if (g_PlayerX < 256) {
+                    g_Tilemap.x = 256;
+                    g_Tilemap.left++;
+                    self->step++;
+                }
+                break;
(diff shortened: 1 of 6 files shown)


src/boss/rbo3/rbo3.c
```diff
@@ -1,13 +1,500 @@
...
+        break;
+
+    case 3:
+        AnimateEntity(&D_us_801805FC, self);
+        if (self->step_s == 0) {
+            self->ext.GS_Props.timer = 64;
+            self->step_s++;
+        }
+        if (GetDistanceToPlayerX() < 104) {
+            self->ext.GS_Props.attackMode = 1;
```
hohle, line 205: My theory is that bosses will use similar extension layouts., but I think it's too early to commit and rename `GB_Props`. This is intended to look out of place to be resolved when more is known about these layouts.

src/boss/rbo3/rbo3.c
```diff
@@ -1,13 +1,499 @@
...
+        if (self->step_s == 0) {
+            if (Random() & 1) {
+                PlaySfxPositional(SFX_RBO3_UNK_7FE);
+            } else {
+                PlaySfxPositional(SFX_RBO3_UNK_7FF);
+            }
+            PlaySfxPositional(SFX_RBO3_UNK_7C8);
+            self->step_s++;
+        }
+        self->ext.stub[9] = 1;
```
sozud, line 250: I think mixing the extensions is going to be problematic since they're not necessarily guaranteed to align

src/boss/rbo3/rbo3.c
```diff
@@ -45,16 +531,153 @@ void func_us_801922EC(Entity* self) {
...
+    switch (self->step) {
+    case 0:
+        InitializeEntity(&D_us_80180444);
+        // fallthrough
+    case 1:
+        entity = &PLAYER;
+        x = entity->posX.i.hi + g_Tilemap.scrollX.i.hi;
+        if (
+#ifdef VERSION_PSP
+            x > 128 && x < 384
```
sozud, line 612: This should match on ps1 too

src/boss/rbo3/rbo3.c
```diff
@@ -1,13 +1,499 @@
...
+
+void func_us_80192020(Entity* self) {
+    Primitive* prim;
+    s32 posX, posY;
+    s16 primIndex;
+    s32 var_a2;
+    s16 angle;
+    u16 var_v0;
+
+    if (self->flags & 0x100) {
```
sozud, line 415: we don't have an enum for this?

src/boss/rbo3/rbo3.c
hohle: Eek. Missed that one. Fixed.

src/boss/rbo3/rbo3.c
hohle: Thanks!

src/boss/rbo3/rbo3.c
hohle: Yep!

src/boss/rbo3/rbo3.c
```diff
@@ -1,13 +1,499 @@
...
 
-#include "common.h"
-#include "stage.h"
+#include "rbo3.h"
 
-INCLUDE_ASM("boss/rbo3/nonmatchings/rbo3", func_us_80191438);
+extern u16 D_us_80180468;
+extern s8 D_us_80190E80;
+extern s8 D_us_80190E8C;
+extern s8 D_us_80190E98;
```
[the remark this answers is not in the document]
hohle: That looks like a rodata offset issue. I'll fix that and move the static strings inline.

src/boss/rbo3/rbo3.c
```diff
@@ -1,13 +1,499 @@
...
+        }
+        if (AnimateEntity(&D_us_8018060C, self) == 0) {
+            SetStep(3);
+            if (g_Player.status & PLAYER_STATUS_STONE) {
+                SetStep(8);
+            }
+        }
+
+        if (self->animFrameIdx == 4 && self->animFrameDuration == 0) {
+            entity = AllocEntity(&D_8007A958, &D_8007A958 + 32);
```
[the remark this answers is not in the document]
hohle: That's interesting.

There is a lot of relative references to entities (`self - 1`, etc.) in these functions that I haven't encountered elsewhere yet.

sozud (approved):

sozud:
Last thing before merging probably
```
-    if (self->flags & 0x100) {
         SetStep(2);
     }
```

PR #1807, opened 2024-10-18
Eliminate `ext.generic.unk80.modeS16.unk2` and other cleanup

bismurphy:
Next member. As usual, while I'm modifying these functions I've cleaned up other things to match more closely with our modern style, such as using `prim` instead of `poly`, naming a couple functions, making weird logic match on PSP, etc.


bismurphy:
will fix the STATIC_ASSERT issue

bismurphy:
okay cool, finally got something that works :)

PR #1808, opened 2024-10-19
Dedupe `no0` clock room entities

JoshSchreuder:


config/symbols.us.bomar.txt
```diff
@@ -1,3 +1,14 @@
...
+bird_cage_pos_y = 0x80181284;
+statue_pos_x_1 = 0x80181288;
+statue_pos_x_2 = 0x8018128C;
+gear_pos_x = 0x80181290;
+stone_door_pos_x_1 = 0x80181294;
+stone_door_pos_x_2 = 0x80181298;
+statue_pos_x_3 = 0x801812A0;
+anim_bird_cage = 0x801812A8;
+anim_gear_1 = 0x801812C4;
+anim_gear_2 = 0x801812D0;
```
JoshSchreuder, line 13: Renamed these just as placeholders - can replace these with better names once the data is extracted, as there is some overlapping array data etc.

src/st/no0/clock_room.c
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "common.h"
+#include "no0.h"
+
+INCLUDE_ASM("st/no0/nonmatchings/clock_room", func_us_801CCC74);
```
JoshSchreuder, line 5: This does share some similarities with `EntityClockRoomController` in `bomar` but there is some significant differences still most likely it will need to be its own function

JoshSchreuder:
Rebased and resolved the conflict

PR #1809, opened 2024-10-19
Major sfxID update

synamaxmusic:
We're finally up to sfxID 0x701!  This means we're back to vabID 1 and we're all caught up with Alucard's voice over sfx.  Currently, we're now on Richter's sfx.


bismurphy (approved):
Nice! Thanks for going through all of these.

PR #1813, opened 2024-10-21
Decompile a couple of `no0` clock room functions

JoshSchreuder:


src/st/no0/clock_room.c
```diff
@@ -4,9 +4,34 @@
...
+    // self + 8 is birdcage door 2
+    self += 1;
+    if (timerMinutes - 30 < 20) {
+        self->ext.birdcage.state = true;
+    } else {
+        self->ext.birdcage.state = false;
+    }
+}
+
+void UpdateClockHands(Entity* self, PlayerStatus* status) {
```
sozud, line 25: These apparently got factored out at some point? https://github.com/Xeeynamo/sotn-decomp/blob/16c3226a0946528a98b0dc0fb02f29a158d71d15/src/boss/mar/clock_room.c#L164

sozud (approved):

src/st/no0/clock_room.c
JoshSchreuder: Yeah, the ClockRoomController in MAR has them inlined, the clock room controller here in NO0 (which is not yet decompiled - https://decomp.me/scratch/4mlgb) calls these as functions.

PR #1815, opened 2024-10-21
Decompile `no0` `func_us_801C2184` elevator function

JoshSchreuder:
Also split all elevator related functions into a separate file

https://decomp.me/scratch/I6bgp


src/st/no0/e_elevator.c
```diff
@@ -0,0 +1,264 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "no0.h"
+
+s16 func_us_801C1F98(Primitive* prim, s16 arg1) {
```
sozud, line 4: I see these dups

```
| 1.00 | true     | func_8019040C                       | asm/us/st/cen/matchings/F890/func_8019040C.s 
| 1.00 | true     | func_us_801C1F98                    | asm/us/st/no0/matchings/41F98/func_us_801C1F98.s 


| 1.00 | true     | EntityUnkId1B                       | asm/us/st/cen/matchings/F890/EntityUnkId1B.s 
| 1.00 | true     | func_us_801C26B8                    | asm/us/st/no0/matchings/41F98/func_us_801C26B8.s 
```

I'm thinking the other functions might also be the same as the CEN elevator but not similar enough to end up in the report? Didn't investigate in detail. You don't need to dedupe these right now, just wanted to bring it up

src/st/no0/e_elevator.c
```diff
@@ -0,0 +1,264 @@
...
+        } else {
+            DestroyEntity(self);
+            return;
+        }
+
+        if (player->posY.i.hi > 0xC0) {
+            self->posY.i.hi = player->posY.i.hi;
+            player->posX.i.hi = self->posX.i.hi;
+            self->animCurFrame = 0xA;
+            g_Entities[1].ext.stub[0x00] = 1;
```
sozud, line 121: This is going to be a portability issue probably. We have an entity for entity slot 1 but not sure if it should be used here or not https://github.com/Xeeynamo/sotn-decomp/blob/16c3226a0946528a98b0dc0fb02f29a158d71d15/include/entity.h#L1886

src/st/no0/e_elevator.c
```diff
@@ -0,0 +1,264 @@
...
+            SetStep(2);
+        }
+
+        break;
+    case 1:
+        if (*(u8*)&self[-1].ext.stub[0x4]) {
```
sozud, line 127: Portability issue

src/st/no0/e_elevator.c
JoshSchreuder: They are reasonably close, I used them as a reference for this decomp but quite a few differences that might make it hard to dedupe

src/st/no0/e_elevator.c
JoshSchreuder: Would this make more sense updating [ET_Elevator](https://github.com/Xeeynamo/sotn-decomp/blob/16c3226a0946528a98b0dc0fb02f29a158d71d15/include/entity.h#L719) to `/* 0x80 */ u8 unk80[0x4];` and making this `if ((self-1)->ext.elevator.unk80[0]) {`?

It still matches

src/st/no0/e_elevator.c
JoshSchreuder: Is the suggestion to update to `g_Entities[UNK_ENTITY_1].ext.entSlot1.unk0`?

src/st/no0/e_elevator.c
sozud: Do we know what object `self[-1]` is? Are we using ET_Elevator for all the associated objects?

src/st/no0/e_elevator.c
sozud: We are doing that for the other elevator so it probably makes sense? https://github.com/Xeeynamo/sotn-decomp/blob/cde3353a32a85000b6aec314d9b4ca3e6aa7b0cc/src/st/cen/F890.c#L130

src/st/no0/e_elevator.c
JoshSchreuder: self-1 should be part of the elevator (I don't know exactly what that is just yet)
https://github.com/Xeeynamo/sotn-decomp/pull/1815/files#diff-8ae1c5fc7b20f674dda925677906970cd0bc621cd02fa25cbde3414a5c344745R81-R82

It's entity 0x29 which is `func_us_801C26B8` and part of this file

sozud: In that case using `ext.cenElevator.unk80` makes sense to me since it will have the right layout in both functions

src/st/entity_explosion.h
```diff
@@ -28,12 +23,7 @@ void EntityExplosion(Entity* entity) {
     } else {
         entity->posY.val += entity->velocityY;
 
-#ifdef MISSING_ANIMATE_ENTITY_PROTOTYPE
-        if (!((AnimateEntityMissingProto)AnimateEntity)(
```
JoshSchreuder, line 32: I thought this would be needed on this PR too - turns out the build will compile just fine without it, and this has not been in use for a bit so can be removed

Then: github.com/Xeeynamo/sotn-decomp/commit/267d6973504f741cbf0f1de580b6624bc672d1fe#diff-ff8ba3c96a3ecaf75e685ba0d5c9d733a6ae1428d1ea8d4b0958da8bd566c47f

Now: https://github.com/search?q=repo%3AXeeynamo%2Fsotn-decomp%20MISSING_ANIMATE_ENTITY_PROTOTYPE&type=code

src/st/no0/e_elevator.c
JoshSchreuder: Done, thank you for working through it with me!

PR #1817, opened 2024-10-21
Tweak the handling of the LayoutEntity files created by sotn-assets

bismurphy:
This changes the formatting of the generated layout entities in order to make the files (in my opinion) more readable. We do the following:

1. change the types in e_laydef to make the offsets make more sense and reduce casts
2. Break up e_layout into blocks to more easily see where the boundaries between each room's data are
3. Add comments between those boundaries with the offsets, to easily cross-reference e_laydef and e_layout
4. Implement reading of the EntityIDs enum so that entities are listed by ID in e_layout.c to make it easier to identify them
5. Include the overlay's .h file rather than common.h, to allow access to the EntityIDs enum.

Sounds like we want to make changes here to move this stuff up to the JSON, so I'll mark this as a draft while we determine the long-term desired functionality.

Also, just a heads-up that this is my first time working with Go, so I might have done some bad code practices.

### DIFF
--- a/src/st/cen/cen.h
+++ b/src/st/cen/cen.h
@@ -9,7 +9,7 @@
 
 #define CASTLE_FLAG_BANK 0x00
 
-typedef enum {
+typedef enum EntityIDs{
     /* 0x00 */ E_NONE,
     /* 0x01 */ E_ID_01,
     /* 0x02 */ E_EXPLOSION,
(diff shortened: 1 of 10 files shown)


sozud (approved):
Seems reasonable to me but ought to be approved by Xeeynamo

PR #1819, opened 2024-10-23
Decompile `no0` `func_us_801C27A4` elevator function

JoshSchreuder:
Decompiled matching both PSP and PSX

PSX: https://decomp.me/scratch/mSJ5k
PSP: https://decomp.me/scratch/z05Dt

Had to do some rather ugly conditionals for `SetStep` related stuff, if someone can work out a better method that matches both that would be good, I tried a whole bunch of stuff like ternery but no luck


src/st/no0/e_elevator.c
```diff
@@ -261,4 +261,113 @@ void func_us_801C26B8(Entity* self) {
...
+                prim->priority = 0x6A;
+                prim->drawMode = DRAW_HIDE;
+                prim = prim->next;
+            }
+        } else {
+            DestroyEntity(self);
+            return;
+        }
+
+#if defined(VERSION_PSP)
```
sozud, line 300: Not sure if I'm missing something but the psp version matches on ps1 for me?

JoshSchreuder: Right you are, I refactored this code so many times I must have not tried this variation for some reason. Updated

PR #1823, opened 2024-10-23
Decompile/decompile tt 002 func us 801746 e8

Onenutmcgee:
PSX: https://decomp.me/scratch/tOtFJ
PSP: https://decomp.me/scratch/L20c5

I believe PSP should match.

Pretty standard init function.  Small difference between PSP and PSX for how it checks the FGLayout flags that I couldn't reconcile together.  

~~Line 130 is a hot mess.  There is either another Entity appended to the Entity struct, or we have some sort of size mismatch, or this is an artifact of when the original devs may have just copied over a struct and modified it.  Definitely building a sub entity.  Sizing is very close to entity++, but not quite.~~


Onenutmcgee:
Ok, that hot mess is actually entity++.  Looks like my context was out of date

bismurphy:
I think I'm getting working scratches with:
```
    entity++;
    DestroyEntity(entity);
```

https://decomp.me/scratch/tcFLB
https://decomp.me/scratch/z27rS

But maybe I'm missing something?

Onenutmcgee:
Yeah, I double checked my context after I posted this and entity++ is correct.  I remember you (or somebody) added that entity to the end of the struct and I didn't update my copy pasta context

src/servant/tt_002/3678.c
```diff
@@ -36,7 +44,101 @@ INCLUDE_ASM("servant/tt_002/nonmatchings/3678", func_us_80173BD0);
...
+    entity->params = 0;
+
+    if (mode == MENU_SWITCH_SERVANT) {
+
+// PSP version does this in 2 chunks, the PSX version uses an lw instruction
+#ifdef VERSION_PSP
+        if (D_8003C708.flags & LAYOUT_RECT_PARAMS_UNKNOWN_20 ||
+            D_8003C708.flags & LAYOUT_RECT_PARAMS_UNKNOWN_40) {
+#else
+        if (LOW(D_8003C708.flags) &
```
sozud, line 99: I suspect this is something like `if ((D_8003C708.flags & 0x60) && (D_8003C708.unk2 == 0)) {` but can't seem to get the optimization to trigger

sozud (approved):

PR #1824, opened 2024-10-24
Dedupe `EntityExplosionPuffOpaque`

JoshSchreuder:
A larger dedupe which spans `no0`, `no3`, `np3` and `nz0`

I've renamed `AxeKnightDeath` to `CreateExplosionPuff` as it does not seem specific to axe knight. 

Similarly instead of this entity being named `EntityWargExplosionPuffOpaque` I've renamed to `EntityExplosionPuffOpaque` - this same entity appears to be used for warg, axe knight and merman so far. Also renamed the associated entity extension.


src/st/e_explosion_puff_opaque.h
```diff
@@ -36,12 +36,12 @@ void EntityWargExplosionPuffOpaque(Entity* self) {
     case 0:
         InitializeEntity(g_EInitParticle);
         params = self->params & 0xF;
-        obj = &D_8018129C[params];
-        self->palette = obj->palette + 0xD0;
+        obj = &D_80180FE0[params];
+        self->palette = obj->palette + E_PUFF_OPAQUE_PALETTE_OFFSET;
```
JoshSchreuder, line 40: Do we have an established pattern or preference for things like this? Currently I've just done stage level `#define`s inside the relevant headers

src/saturn/stage_02.c
```diff
@@ -98,7 +98,7 @@ INCLUDE_ASM("asm/saturn/stage_02/f_nonmat", f60DDF64, func_060DDF64);
 INCLUDE_ASM("asm/saturn/stage_02/f_nonmat", f60DE178, func_060DE178);
 
 // SAT func_060DE2B0
-void AxeKnightDeath(void) {
+void CreateExplosionPuff(void) {
```
JoshSchreuder, line 101: I am not sure if renaming this is correct or will work, so I'm updating, but can revert if required

PR #1825, opened 2024-10-24
Eliminate ET_Generic, and ext.stub

bismurphy:
The quest is done!

This eliminates the ET_Generic, and as a bonus, stub, from the Entity struct. This means no future code will be able to use these as defaults and will have to choose a particular extension, which is nice.

As usual, while doing these cleanups I found various duplications to solve and file splits to normalize. Overall the code becomes more and more readable.


bismurphy:
> My only concern is how mips2c would decompile the `ext.` section for new entities. It will probably start to mix some stuff as an attempt to produce something? You might know better. Depending on that I am wondering if we should replace `ext.stub` with something like `ext.do_not_use_this` or similar.

Good point! I have created a new struct which appears at the top, so it should be used first, if we're lucky. It will show up as `self->ext.ILLEGAL.s32[4]`, for example. Hopefully the all-caps ILLEGAL name will discourage its use and remind people that it needs to be changed. We'll see if M2C picks up on it and uses it; if not, we can tweak this over time. Nice idea!

sozud (approved):

JoshSchreuder:
> I have created a new struct which appears at the top, so it should be used first, if we're lucky. 

Can confirm this works, I just decompiled a new function and it gave me `self->ext.ILLEGAL.u8[0]` in the M2C output
https://decomp.me/scratch/Pagw1

PR #1827, opened 2024-10-25
Break tt 002 on jump tables

Onenutmcgee:
Prep work for decomp of the jump table functions.  There are a lot of them.

I assume a lot of these files can be merged back together after decomp.  This just splits it so we can decomp them easier without needing to do all the jump table stuff for individual functions.


sozud (approved):

Onenutmcgee:
Looks like this is no longer a required step for jump tables

PR #1830, opened 2024-10-26
Decompile `no0` Slinger functions

JoshSchreuder:
That's these guys in Marble Gallery
image

They share quite a bit in common with e_skeleton and e_bone_scimitar, but I decided not to deduplicate these ones at this stage


src/st/no0/e_slinger.c
```diff
@@ -0,0 +1,231 @@
...
+    u8 i;
+
+    if (self->flags & FLAG_DEAD) {
+        self->step = SLINGER_DESTROY;
+    }
+
+    switch (self->step) {
+    case SLINGER_INIT:
+        InitializeEntity(g_EInitSlinger);
+        self->ext.skeleton.attackTimer = 80; // Slinger attack timer cycle
```
JoshSchreuder, line 68: More of a note than anything, but this extension is used now across skeleton, bone scimitar and now slinger.
Since it was already used in bone scimitar and skeleton I decided to also use it here, but I can create a new extension if preferred.

include/sfx.h
```diff
@@ -323,7 +323,7 @@ enum Sfx {
     SFX_ARROW_SHOT_C,
     SFX_ARROW_SHOT_D,
     SFX_SKELETON_DEATH_A,
-    SFX_SKELETON_DEATH_B,
+    SFX_SKELETON_DEATH_B, // Slinger
     SFX_SKELETON_DEATH_C,
```
JoshSchreuder, line 327: C looks to be used for bone scimitar and skeleton. A doesn't appear to be used in our code just yet. I just made a comment here in case we want to rename this down the track

src/st/no0/e_slinger.c
```diff
@@ -0,0 +1,231 @@
...
+            SetStep(SLINGER_WALK_AWAY_FROM_PLAYER);
+            self->ext.skeleton.attackTimerIndex++;
+            self->ext.skeleton.attackTimer =
+                attack_timer_cycles[self->params & 1]
+                                   [self->ext.skeleton.attackTimerIndex & 3];
+            break;
+        }
+
+        if ((animStatus & 0x80) && (self->animCurFrame == 10)) {
+            if (((u32)(((u16)self->posX.i.hi + 16) & 0xFFFF) <= 288) &&
```
sozud, line 121: This is
```
            if ((self->posX.i.hi >= -16 && self->posX.i.hi <= 272) &&
                (self->posY.i.hi >= 0 && self->posY.i.hi <= 240)) {
```
https://decomp.me/scratch/N3g0Z

JoshSchreuder: Thanks, I'll fix this up. Just starting to see a bit of this min <= x <= max pattern in this code for the first time I think

PR #1831, opened 2024-10-26
Extract sprite banks from all overlays

Xeeynamo:
Took me for ever to get through this this but here you go.  All the `assets/st/*/sprite_banks.json` and `assets/boss/*/sprite_banks.json` are getting extracted and built back into `src/`. Not all of them are actually linked but there are some major changes to make before that happens.

Despite the asset-tool being relatively new, it suffered from a major design flaw. The tool expected overlays to store data in a contiguous way. The order of the data type must be the same and no gaps between data are allowed. This assumption started to show the cracks for new overlays

The tool needs to change. Data should be uncoupled but that means offset and type has to be explicitly defined in `config/assets.us.yaml`. I plan to keep maintaining `sotn-assets stage info` to automatically generate the right asset and splat YAMLs, so it should be easier to maintain.


sozud (approved):
Seems reasonable to me once the build errors are fixed

PR #1833, opened 2024-10-26
Decompile ST0 EntityCutscenePhotograph

bismurphy:
Well. Oh my god. This function is finally done. It was brutal. I feel like the status bar face of the Doom Guy when he's at 1 HP.

Anyway, it's decompiled! What a wild function. ST0 is coming to an end just as the other stages did before it. GTE continues to be a monster, but it's done.


sozud (approved):
Nice work!

PR #1843, opened 2024-10-28
Decompiling function func_us_80173BD0

Onenutmcgee:
The 2nd argument to func_us_80173994 is probably either some flags or needs to be an enum.  But I am not sure what they are yet.  I'll come back to clean them up once I get that function decompiled.

### DIFF
--- a/src/servant/tt_002/3678.c
+++ b/src/servant/tt_002/3678.c
@@ -103,7 +103,52 @@ void CreateEventEntity_Dupe(Entity* entityParent, s32 entityId, s32 params) {
     }
 }
 
-INCLUDE_ASM("servant/tt_002/nonmatchings/3678", func_us_80173BD0);
+void func_us_80173BD0(Entity* arg0) {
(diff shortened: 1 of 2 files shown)


src/servant/tt_002/3678.c
```diff
@@ -103,7 +103,52 @@ void CreateEventEntity_Dupe(Entity* entityParent, s32 entityId, s32 params) {
     }
 }
 
-INCLUDE_ASM("servant/tt_002/nonmatchings/3678", func_us_80173BD0);
+void func_us_80173BD0(Entity* arg0) {
+    if (abs(arg0->velocityY) > abs(arg0->velocityX)) {
+        if (abs(arg0->velocityY) < 0x8000) {
```
sozud, line 108: All the velocity numbers can be FIXED

sozud (approved):

src/servant/tt_002/3678.c
Onenutmcgee: Oh yeah. I’ll fix that tonight and update the PR

PR #1845, opened 2024-10-29
PrimitiveType Lint

hohle:
Adds a line transformer to `sotn-lint` for replacing integers with the appropriate `PrimitiveType` enum when the variable is named `prim` and the field being accessed or mutated is named `type`.

The regex for matching statements with enum fields and some binary operation has been moved out to be shared by `EnumLineTransformer` which can find and replace traditional incrementing `enum`s, as well as the existing `BitFlagLingTransformer`, which finds and replaces flag-type enums.


hohle:
<strike>I've changed the type of the `type` field to `PrimitiveType` for `Primitive`, `FakePrimitive`, and `AxePrimitive` with a fixed size of 8-bits. This makes the values more discoverable and works with PS1 and PSP toolchains, however, this may be treated as signed, when those structs previously (and `POLY_GT4` currently) defined the field as a `u8`.</strike> 

This changes has been reverted because modern compilers don't like taking references to bitfields, even when they are aligned on byte boundaries. It's possible that we could get away with using `PrimitiveType` for the field type, but I don't think the C standard gives any guarantees regarding the size, so we would need additional assertions to ensure alignment doesn't break.

sozud (approved):

PR #1847, opened 2024-10-30
decompiling func_us_80177AC4

Onenutmcgee:
Full match both PSP and PSX
PSX: https://decomp.me/scratch/iF9Ax
PSP: https://decomp.me/scratch/bA5Ta

I made a comment in the code that it looks like the extension used by this entity is wrong, but I don't have enough context to build a "correct" one yet.  We'll have a better idea as we get more functions decompiled (especially the functions that spawn these sub entities).

### DIFF
--- a/include/entity.h
+++ b/include/entity.h
@@ -608,10 +608,10 @@ typedef struct {
 } ET_FaerieUnk0;
 
 typedef struct {
-    /* 0x7C */ s16 pad7c;
+    /* 0x7C */ s16 unk7c;
     /* 0x7E */ s16 unk7E;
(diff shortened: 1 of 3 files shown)


sozud (approved):

sozud:
It seems like you committed extra stuff in the new commit, how about I merge the original function now then you can put up a second PR?

Onenutmcgee:
That would work.  I did accidently merge in a bunch of PSP stuff I leave sit.

Onenutmcgee:
Ok, i think I fixed it.

PR #1849, opened 2024-10-30
Add linter for `g_Player.status`

JoshSchreuder:


tools/lints/sotn-lint/src/player_status.rs
```diff
@@ -0,0 +1,168 @@
...
+        let result = DMT.transform_line(input_line);
+        assert_eq!(result, expected_line)
+    }
+
+    // TODO: Implement PLAYER_STATUS_TRANSFORM 0x7 flag
+    //
+    // #[test]
+    // fn test_player_status_combo_flag() {
+    //     let input_line = "g_Player.status = 0x7;";
+    //     let expected_line = "g_Player.status = PLAYER_STATUS_TRANSFORM;";
```
JoshSchreuder, line 164: @hohle as mentioned on Discord, this enum value does not map cleanly in the bitflag. I've left it off this linter for now
https://github.com/Xeeynamo/sotn-decomp/blob/master/include/game.h#L283-L285

Not sure if you have any ideas on how to implement something like this into the BitFlagsTransformer nicely

tools/lints/sotn-lint/src/player_status.rs
```diff
@@ -0,0 +1,168 @@
...
+        (1 << 29, "PLAYER_STATUS_UNK40000000"),
+        (1 << 30, "PLAYER_STATUS_UNK80000000")
+    ];
+}
+
+impl PlayerStatusTransformer {
+    pub fn new() -> Self {
+        Self {
+            transformer: BitFlagLineTransformer::<u32>::new(
+                "status",
```
JoshSchreuder, line 50: This is another pretty generic field name. Might be worth adding a variable name parameter here as was done for the normal enum transformer so we can just target `g_Player`?

PR #1850, opened 2024-10-30
Workling extraction & building of DRA PSP

bismurphy:
Required a lot of shuffling around of makefile, yaml, and mwccgap, but I'm getting an [OK] now with DRA PSP.

Had to move the asm directory to dra_psp rather than just dra, due to list_o_files_psp using "dra_psp" to look for its asm files in the list_drc_files_psp target.


bismurphy:
Weird build error on RBO3 which i didn't touch, looks like I had an old symbols file somehow; fixed now.

sozud (approved):
Looks good to me but needs a review from Xeeynamo

bismurphy:
> I wonder if we should do the same thing we did with WRP PSP, where src_path is a level above it should be and then explicitly mark C files to point to either dra/ or dra_psp/. This way it will be easier to merge C files. What do you think?

Just to be clear, is the idea that we would have `src/12820.c`, and all the other `.c` files for PSP DRA directly in src? I would disagree with that, I think there are enough files in src that they should be carefully organized to keep them grouped by what they are for.

I think we should either have `dra` and `dra_psp` side by side, or perhaps we could have `dra_psp` be within `dra`. But putting `dra_psp` a level above where it is does not seem like a good idea for the rest of the repo.

bismurphy:
Oh wait I'm misunderstanding - you mean keep the files in place, but tell splat to just look in "src", and then each file is in "dra_psp/filename" rather than just being at "filename", with the path set to `src/dra_psp`. That does sound reasonable, and I will make that change.

bismurphy:
I think I made the requested change. Had to change the way INCLUDE_ASM is structured but luckily this was an easy find and replace. Let me know if I did this wrong (recommend looking at just the last commit).

PR #1859, opened 2024-11-01
Decompile `no0` main Ouija Table function

JoshSchreuder:
Match on both
PSX: https://decomp.me/scratch/A79x5
PSP: https://decomp.me/scratch/HJrWp

The functions directly after this are related to the sub-entities such as vase and beaker I believe, so I'll decompile those next. Possible this might lead to changes in the extensions I've added here

Once all that's done I'll split into a separate Ouija Table file with all the related code

### DIFF
--- a/include/entity.h
+++ b/include/entity.h
@@ -1986,6 +1986,21 @@ typedef struct {
     /* 0xAE */ s16 unkAE;
 } ET_BackgroundVortex;
 
+typedef struct {
+    s32 pad0[1];
+    s16 attackTimer;
+    s32 pad6[3];
+    s32 thrownVase;
+} ET_OuijaTable;
+
+typedef struct {
+    s32 pad0[2];
+    s32 spawned;
+    struct Entity* parent;
+    s32 pad16[1];
+    s32 isThrown;
+} ET_OuijaTableContents;
+
 // ====== RIC ENTITIES ======
 
 // ==========================
@@ -2166,6 +2181,8 @@ typedef union { // offset=0x7C
     ET_ClockTower clockTower;
     ET_BackgroundVortex bgVortex;
     ET_MedusaUnk1A medusaUnk1A;
+    ET_OuijaTable ouijaTable;
+    ET_OuijaTableContents ouijaTableContents;
 } Ext;
 
 #define SYNC_FIELD(struct1, struct2, field)                                    \
--- a/src/st/no0/5BDCC.c
+++ b/src/st/no0/5BDCC.c
@@ -1,8 +1,134 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "common.h"
-#include "game.h"
+#include "no0.h"
 
-INCLUDE_ASM("st/no0/nonmatchings/5BDCC", func_us_801DBDCC);
+#define VASE self + 1
+#define BEAKER self + 2
+#define WALKING_TABLE self + 3
+
+extern u16 D_us_80180C14[];
+extern s16 D_us_801825CC[];
+extern u16 D_us_801825D4[];
+extern u8 D_us_801825F4[];
+
+typedef enum {
+    /* 0 */ OUIJA_TABLE_INIT,
+    /* 1 */ OUIJA_TABLE_INIT_SUBENTITIES,
+    /* 2 */ OUIJA_TABLE_COME_ALIVE,
+    /* 3 */ OUIJA_TABLE_ATTACK,
+    /* 4 */ OUIJA_TABLE_UNUSED,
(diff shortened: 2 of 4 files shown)


src/st/no0/5BDCC.c
```diff
@@ -1,8 +1,134 @@
...
+            self->ext.ouijaTable.attackTimer = 192;
+            // Throw out one of the objects on the table
+            if (self->ext.ouijaTable.thrownVase != 0) {
+                entity = VASE;
+            } else {
+                entity = BEAKER;
+            }
+            entity->ext.ouijaTableContents.isThrown = true;
+            // Alternate between throwing vase and beaker
+            self->ext.ouijaTable.thrownVase ^= 1;
```
JoshSchreuder, line 123: Wasn't super confident with the naming here, 
I think it's basically a toggle between which sub-entity gets thrown but wasn't sure what to call it

include/entity.h
```diff
@@ -1986,6 +1986,21 @@ typedef struct {
     /* 0xAE */ s16 unkAE;
 } ET_BackgroundVortex;
 
+typedef struct {
+    s32 pad0[1];
```
bismurphy, line 1990: For pads let's use the "s32 : 32" syntax like you'll see on other entities. Makes it easy to tell what's a pad and what isn't, without needing to name all your pads. Struct offset comments would also be nice.

src/st/no0/5BDCC.c
```diff
@@ -1,8 +1,134 @@
...
+    /* 1 */ OUIJA_TABLE_INIT_SUBENTITIES,
+    /* 2 */ OUIJA_TABLE_COME_ALIVE,
+    /* 3 */ OUIJA_TABLE_ATTACK,
+    /* 4 */ OUIJA_TABLE_UNUSED,
+    /* 255 */ OUIJA_TABLE_DEBUG = 255,
+} OuijaTableSteps;
+
+// Main Ouija Table entity
+void EntityOuijaTable(Entity* self) {
+    Entity* entity;
```
bismurphy, line 25: perhaps name it "other" or something, just so it has its own name instead of being generically "entity".

src/st/no0/5BDCC.c
```diff
@@ -1,8 +1,134 @@
...
+    if (self->flags & FLAG_DEAD) {
+        entity = AllocEntity(&g_Entities[224], &g_Entities[256]);
+        if (entity != NULL) {
+            CreateEntityFromEntity(E_EXPLOSION, self, entity);
+            entity->params = 1;
+        }
+        PlaySfxPositional(SFX_STUTTER_FIREBALL);
+        // Destroy the spinning objects on the table
+        entity = VASE;
+        entity->flags = entity->flags | FLAG_UNK_2000;
```
bismurphy, line 36: Probably should be a `|=`

src/st/no0/5BDCC.c
bismurphy: maybe leave it as an unk for now; circling back later it's a lot easier to identify something that's unfinished (and finish it) than to discover that something is wrong.

bismurphy (approved):
Looks good! Just some style ideas here.

src/st/no0/5BDCC.c
```diff
@@ -1,8 +1,134 @@
...
+            otherEntity->params = 2;
+            otherEntity->ext.ouijaTableContents.parent = self;
+
+            if (self->params) {
+                otherEntity->posX.i.hi += 24;
+                otherEntity->facingLeft = 1;
+            } else {
+                otherEntity->posX.i.hi -= 24;
+            }
+#ifndef VERSION_PSP
```
sozud, line 78: This gets optimized out so you don't need the ifdef

sozud (approved):

src/st/no0/5BDCC.c
JoshSchreuder: On the PSP scratch it seems to add extra instructions for me?
```
lh      v1,6(s0)
sh      v1,6(s0)
```

sozud: I mean you can remove the whole block:
```
#ifndef VERSION_PSP
            // Looks like a coding error
            otherEntity->posY.i.hi = otherEntity->posY.i.hi;
#endif
```

sozud: Nevermind, I tried it again and it introduced reg swaps, made a mistake

JoshSchreuder: Removing the block then causes the PSX to mismatch due to missing instructions. 

```
sh      v0,6(s1)
lhu     v0,0x2c(s0)
nop
```

I thought it would be a no-op too but I guess not?

PR #1872, opened 2024-11-03
Decompile tt_002 func_us_80173994

Onenutmcgee:
This function was used all over the code so I just wanted to see what it was.  It's a very small utility function that just sets animation data from the animation data array.

Matches both PSX and PSP


Onenutmcgee:
Note: I was planning on doing another function in this PR, but the function I picked (and most of what is left) is a bit more complex and heavily uses the data structures, some of which appear to be nested with pointers.  So I'll have to shift gears and do the data next I think to continue to make progress.

sozud (approved):

PR #1873, opened 2024-11-03
Import data tt_002 shared/common

Onenutmcgee:
This is just the shared and or semi-common parts of the data.  The faerie specific data will come in the next PR, mostly because I ran out of time and will be off for a few days.


Onenutmcgee:
Turns out SpriteParts and ClutData not necessarily shared between servants.  Events seem to be though.  Probably need some renaming as the "shared" ones are only shared between ghost and bat right now.  Maybe other will also, but we'll have to come back to that.

I was hoping to get animation data also loaded up, but I kind of ran out of time so this will just have to be good enough for now and I'll get back on it on Thursday.

sozud (approved):

sozud:
This needs conflicts resolved due to the previous PR

PR #1880, opened 2024-11-06
Decompile tt_002 func_us_80176504

Onenutmcgee:
The data structure for the ranked lookup table is messy because the PSP version requires it to be a flat array.

I'm pretty sure I have a union solution for that so the data tables will at least make sense, but until I can get data import done, we'll keep it like this as it will need to be used like this anyway.

PSX: https://decomp.me/scratch/SN6vr
PSP: https://decomp.me/scratch/DM13D

PSP has a register swap that I couldn't figure out, but is largely a match


sozud (approved):

sozud:
Seems to have picked up an issue
```
src/servant/tt_002/3678.c: In function `func_us_80176504':
src/servant/tt_002/3678.c:1087: structure has no member named `unk96'
```

PR #1883, opened 2024-11-07
Decompile tt_002 func_us_80173D60

Onenutmcgee:
This is the primary ability update function for Faerie.  The basic idea is that we some upkeep every frame, and tick down the timer. 

Once the timer has ticked down, it resets and then  searches through it's logic to figure out if the player is in a state that the Faerie  can alliviate.  If a path is found, a luck check is performed and if sucessful, the Entity ID is updated so that a different update function is called on the next update(s) to perform the action.

Note, this is not initiated from the engine, this is called from many of the update functions that the engine does call on every frame.

PSX: https://decomp.me/scratch/3ADW8
PSP: https://decomp.me/scratch/53Vc0

This should be a PSP match.

@hohle got this most of the way, I just figured out the local variables.

Co-authored-by: hohle <jonhohle@gmail.com>

### DIFF
--- a/src/servant/tt_002/3678.c
+++ b/src/servant/tt_002/3678.c
@@ -238,11 +238,11 @@ void func_us_80173BD0(Entity* arg0) {
 }
 
 void func_us_80173D60(Entity* self) {
-    s32 i; 
-    s32 params; 
-    s32 rnd;  
-    s32 unkHpSum; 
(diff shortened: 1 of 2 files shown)


sozud (approved):

hohle:
Thanks for finding it!

PR #1884, opened 2024-11-08
Adding structs for servant ability stat tables

Onenutmcgee:
Adding GhostAbilityValues and BatAbilityValues structs for the ability stats by level lookup tables instead of using a multi dimension array with index defines.

### DIFF
--- a/src/servant/tt_000/bat.c
+++ b/src/servant/tt_000/bat.c
@@ -1,14 +1,8 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
-#include <servant.h>
+#include "bat.h"
 #include <sfx.h>
 #include <psxsdk/libc.h>
 #include "../servant_private.h"
-
-#define DELAY_FRAMES_INDEX 0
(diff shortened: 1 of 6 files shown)


sozud:
Do these changes still match on PSP? A lot of this stuff actually is just arrays

Onenutmcgee:
I'm pretty sure it does.  Let me check though.

Onenutmcgee:
Here's the PSP function for bat where this is used (note that there is another function, but that one is still using the PSP asm due to other register swaps): 
https://decomp.me/scratch/Klemw

And here's the primary use case in Ghost:
https://decomp.me/scratch/0Gm3x

Note that it doesn't match in Ghost (it never did, register swap at ASM 108 and missing moves/swaps at ASM 7c8 and ASM da4).  But those issues were not caused from the struct.

The reason I updated these was @hohle had set up a struct for the Faerie ability stats table D_us_80172408 when doing the decomp for func_us_80173D60, and that matched PSP there.  And I figured if we can use a struct to better define the data, that's great.

sozud (approved):

PR #1885, opened 2024-11-08
Decompile `TT_002` `func_us_80177380`

hohle:
While this is a pretty straight forward function, `unkA4` appears to be either a pointer to a `s16[3]` or a struct. It's more self-documenting for it to be a struct, but there is one case where it behaves like an array. A cast resolves the match.


src/servant/tt_002/3678.c
```diff
@@ -1115,7 +1115,7 @@ void func_us_80176504(Entity* arg0) {
                 }
             }
         }
-        arg0->ext.faerie.unkA8 = arg0->ext.faerie.unkA4[0];
+        arg0->ext.faerie.unkA8 = *((s16*)arg0->ext.faerie.unkA4);
```
sozud, line 1118: unkA4->unk0 doesn't match here?

src/servant/tt_002/3678.c
```diff
@@ -1410,7 +1410,141 @@ void func_us_80176C98(Entity* self) {
...
 
-INCLUDE_ASM("servant/tt_002/nonmatchings/3678", func_us_80177380);
+extern s32 D_us_801792D4;
+extern s32 D_us_801792D8;
+extern Unk2CB0 D_us_80172CB0[];
+
+void func_us_80177380(Entity* self) {
+    char pad[2];
+
+    *((s32*)&self->ext.faerie.left) =
```
sozud, line 1420: left and top are already s32?

```
    /* 0x98 */ s32 left;
    /* 0x9C */ s32 top;
```

sozud (approved):

sozud:
Also this got conflicted

Onenutmcgee:
I don't know how much we actually care about the PSP decomp at this time, but I think it's worth mentioning that when I tried to get the PSP to match, I basically had to make Unk2CB0 and FaerieUnkA4 flat arrays instead of structs.  This isn't the first time I've seen this where I could get a struct to work for PSX, but not for PSP.  It seems as though the PSX optimization gives us quite a bit more wiggle room in regards to data structures, which makes sense as PSP is O0.  

PSP scratch: 
https://decomp.me/scratch/DotJJ

Note that is not a complete match.  There is still 2 blocks that have some register swaps and 2 blocks that have an instruction ordering issue, both of which I was unable to figure out.  But going to flat arrays did resolve a lot of stuff at the cost of readability.

If we're really not too concerned about the PSP decomp, I'm much more in favor structs, but if we are concerned about the PSP, needing to refactor many structs in the future could be significant rework.

This is perhaps a discussion best suited for the discord though.

sozud:
My opinion is we shouldn't do things that make the PSP version worse although it's not high priority at this time

Onenutmcgee:
> any good soul that wants to propose it for TT_002?

As of right now, I know there are other functions in TT_002 that are not a PSP match.  Many are, but there are a handful that have the dreaded v0/a0 swap that I have been unable to figure out.

Anyway, I don't want to derail the conversation in the PR for that.  My suggestion would be to make Unk2CB0 and FaerieUnkA4 arrays as I'm confident those don't match PSP.  But I certainly won't block the PR over that.

Onenutmcgee (approved):

src/servant/tt_002/3678.c
```diff
@@ -1346,7 +1346,7 @@ void func_us_80176504(Entity* arg0) {
                 }
             }
         }
-        arg0->ext.faerie.unkA8 = arg0->ext.faerie.unkA4[0];
+        arg0->ext.faerie.unkA8 = ((s16*)arg0->ext.faerie.unkA4)[0];
```
sozud, line 1349: `arg0->ext.faerie.unkA4->unk0;` ?

PR #1887, opened 2024-11-08
Fix VERSION Warning When Generating Context

hohle:
This updates `m2ctx` to be aware of the `VERSION` env var and sets that env var in the Makefile prior to running `m2ctx` to eliminate the:

    warning: #warning "Version not specified. Falling back to the US version." [-Wcpp]

error caused by `common.h`.

This also removes `#define __STDC_HOSTED__ 0` from the generated context. That macro causes redefinition warnings in decomp.me.


sozud (approved):
Seems reasonable to me, Xeeynamo should probably review though

PR #1890, opened 2024-11-09
Remove PR bot

Xeeynamo:
It was a funny little experiment that it was worth trying. Unfortunately it had to major flaws:

* There is no easy way to update an existing comment without some scripting, use of tokens and PR manipulation through the GitHub API. It was simply not worth it.
* If the opened PR does not have the very latest changes from the default branch, it will report a drop in progress, which is obviously not correct. I could have internally rebased the PR before calculating the changed progress, but it has its own challenges like having to deal with conflicts.


sozud (approved):
I was thinking more like just turning it off in the CI, maybe this is useful for debugging the script? Other than that seems good

PR #1893, opened 2024-11-11
Split more `no0` enemy entities for Ctulhu and Stone Rose

JoshSchreuder:
This should be all of the enemies identified and split now I think.
Also fixed an incorrect split for Slinger


src/st/no0/e_slinger.c
```diff
@@ -229,3 +227,24 @@ void EntitySlingerPieces(Entity* self) {
         self->velocityX = -self->velocityX;
     }
 }
+
+void EntitySlingerRib(Entity* self) {
```
JoshSchreuder, line 231: This was in the next split, have moved it here and renamed with what it represents and a minor tidy up

PR #1896, opened 2024-11-12
Data Import tt_002 

Onenutmcgee:
Importing the remaining data for tt_002
Importing sbss data
Adjusting the data files

The data except for the servant_events data is all unique to the faerie.  I suppose all the data could be put in 1 file, but I had it split up because ghost and bat shared a significant amount of data.  

~~D_us_801724CC[] should actually be an array of Unkstruct_801724CC; however, that doesn't match because it's on a 2 byte barrier and the compiler wants to align it to 4 bytes.  I tried #pragma pack and __attributed__((packed)) and a few other things with no avail.  So that and the cast to Unkstruct_801724CC in func_us_80177F84 are just going to have to be wonky.~~
This issue was figured out by @Xeeynamo 

I didn't really do any refactoring beyond what I needed to in order to get the data in.  I'll work on that next.


src/servant/tt_002/faerie.c
```diff
@@ -1983,7 +1980,11 @@ void func_us_80177F84(Entity* self) {
 
     posX2 = self->posX.i.hi;
     posY = self->posY.i.hi;
-    unkStruct = &D_us_801724CC[LOH(self->params)];
+    // The data in D_us_801724CC is aligned in a way that it needs to be a u16
+    // array So this is a bit ugly to cast it into the struct pointer
+    unkStruct =
```
sozud, line 1985: This seems odd to me, was it an array originally?

src/servant/tt_002/faerie_ability_data.c
```diff
@@ -0,0 +1,127 @@
...
+s16 g_PotionItemsParamMap[] = {
+    ITEM_POTION,
+    0x0010,
+    ITEM_HIGH_POTION,
+    0x0011,
+};
+
+// This is an array of type Unkstruct_801724CC, but due to packing,
+// the raw data needs to be a u16 array.
+u16 D_us_801724CC[] = {
```
sozud, line 108: The problem is too much or too little padding? You can maybe manipulate it https://stackoverflow.com/questions/11770451/what-is-the-meaning-of-attribute-packed-aligned4

src/servant/tt_002/faerie.c
Onenutmcgee: D_us_801724CC Is likely an array and we’re pulling the offset based on param. Ideally, we would use an array of that type, but whenever I would use that type, 2 extra bytes would get padded in to each member.  Since that struct has both an u16 and a s32 in addition to s16, I’m pretty sure it originally was a struct, or they were doing some really weird pointer math.

src/servant/tt_002/faerie_ability_data.c
Onenutmcgee: I tried both #pragma pack(push,1) and then popping it after as well as using __attribute__((packed)) and some various variants of those to no avail. If I built this an an array of the correct type, we always got padded 2 extra bytes. It seems like the compiler just ignored all the packing commands. 

I did try for a few hours before giving in and doing it like this.  If somebody else wants to take a crack at it, I’d be happy to update if they figure it out.

sozud (approved):
Unless someone sees an alternate solution for D_us_801724CC seems good to me

Onenutmcgee:
I kind of wonder if we're missing something with the compile and/or options since the packing compiler directives don't actually seem to do anything.

Onenutmcgee:
I’ll make this fix when I get a chance.

Onenutmcgee:
I didn't quite grok what you were saying until I dug into it, but it makes sense now.  The length of the struct being 0x1A because that was working in the code was wrong since gcc was aligning the struct.  It just happened to match work because of a quirk of gcc was realigning the data.

The data makes significantly more sense now.

PR #1901, opened 2024-11-14
Decompile 'NO0' func_us_801D7DAC

gamezter:
PSX : https://decomp.me/scratch/WCc3T
PSP: https://decomp.me/scratch/9Um4e

I had to convert `Entity.attackElement` to u16 to get a match and it didnt break anything. Let me know if I should revert that and find another solution

### DIFF
--- a/src/st/no0/e_stone_rose.c
+++ b/src/st/no0/e_stone_rose.c
@@ -99,7 +99,11 @@ void func_us_801D7DAC(Entity* self) {
                 i++;
             }
         }
+#ifdef VERSION_PSP
+        newEntity->pfnUpdate = D_us_8018091C[newEntity->entityId - 1];
+#else
         newEntity->pfnUpdate = D_us_8018091C[newEntity->entityId];
+#endif
         newEntity->params = self->hitEffect;
         newEntity->ext.stoneRose.unk8C = self->ext.stoneRose.unk8C;
         newEntity->posX.val = self->posX.val;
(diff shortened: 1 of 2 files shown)


src/st/no0/e_stone_rose.c
```diff
@@ -24,8 +24,104 @@ Entity* func_us_801D7D00(u16 arg0) {
     return NULL;
 }
 
+extern PfnEntityUpdate D_us_8018091C[];
+extern s16 D_us_80182268[][4];
```
JoshSchreuder, line 28: Seems likely this is a struct array?
```
typedef struct {
    s16 unk0;
    s16 unk2;
    s16 unk4;
    s16 unk6;
} Unk_D_us_80182268;

extern Unk_D_us_80182268 D_us_80182268[];
```

Then
```
        self->rotX = D_us_80182268[params].unk0 & 0xFF;
        self->rotY = D_us_80182268[params].unk2 & 0xFF;
        self->rotZ = D_us_80182268[params].unk4 & 0x7F;
```
etc.?

sozud: I'm actually not sure if this is better than not since the data is packed. e.g. unk0 is both rotX and the upper part is hitEffect, so I don't think it can be given a sane name. rotXAndHitEffect? I'd probably get Xeeynamo's opinion before changing it again though

sozud (approved):

PR #1902, opened 2024-11-14
Next funcs on psp DRA

bismurphy:
I noticed that my previous function, `DebugShowWaitInfo`, is actually the first function in the C for ps1. To me, this suggests that all the prior code in PSP (there is a pretty large amount of it, about 0x18000 bytes) is PSP-exclusive. Therefore, I adjusted the file splits to start a file on that function.

Then I did the next function, `DebugInputWait`.

After that is a bunch of debug screen capturing functions. These appear to be unused (presumably used in development but not in the final game code) and therefore I assume they are missing due to deadstripping. Since we know that's an issue on PSP, I decided to make a dedicated macro to flag that code is being removed due to deadstripping - this way, if we get deadstripping to work as intended, we don't have to look through every VERSION_PSP usage to identify what is dead code, versus altered code. We can just find all the uses of `PSP_DEADSTRIP` and eliminate them.

Leaving it there for now, will continue working through DRA in future PRs.


bismurphy:
Good reminder! I'll keep that in mind as I continue :)

PR #1904, opened 2024-11-14
Refactor and cleanup for TT_002

Onenutmcgee:
This is mostly just the low hanging fruit that I kind of already had a good handle on.  The rest of the refactoring is going to take a little more digging in and probably some running the game in a debugger to figure out what's going on.

I think the update for D8 and D9 are for the "hidden path" mode where the Faerie tells you to try to destroy a wall.  DA,DB, and DC i think are actually yousi code for performing the "song".  But I'm not sure on any of that.

DD and DF are for the sub entities created when using an Item.  I think they should use a different extension, but still need to dig into that more.


Onenutmcgee:
I also wanted to mention that PLAYER_STATUS_UNK40000 is for sure player at zero hp.  I'll do that update in it's own PR as it touches a LOT of files and I don't want to mix it in with this.

Onenutmcgee:
This is failing on the i686 build for the not code related issue that's happening.  I think this should be good to review.

PR #1911, opened 2024-11-16
Changing PLAYER_STATUS_UNK40000 to PLAYER_STATUS_ZERO_HP

Onenutmcgee:
Simple change.  Figured out that UNK40000 is either "Player has Zero HP" or "Player is dead", which is largely the same I think.   The Faerie "Life Apple" ability triggers off of this status.  Touches a lot of files so I didn't want to put it in my Faerie refactor.

Also a small renaming in Faerie that I realized I forgot in my refactor


sozud:
This gets set for Richter here https://github.com/Xeeynamo/sotn-decomp/blob/d9bd6a6e4d185d2215a54f9ad29c16b154a07c68/src/ric/1AC60.c#L636
The state is set by https://github.com/Xeeynamo/sotn-decomp/blob/d9bd6a6e4d185d2215a54f9ad29c16b154a07c68/src/dra/5D5BC.c#L75

Onenutmcgee:
I saw that, that's why I wasn't sure if it was "zero hp" or just "dead" because it seems like it's a part of the dead state.  Often in combination with PLAYER_STATUS_UNK80000.  Seems to me that it is when you've taken enough damage to be dead.  Maybe it's splitting hairs.

Do you have a preference for which I use?

PR #1912, opened 2024-11-16
Decompile `no0` Ctulhu shockwave attack

JoshSchreuder:
Planning on doing a cleanup pass over this whole enemy file next and importing data / documenting

PSX: https://decomp.me/scratch/0jcyg
PSP: https://decomp.me/scratch/n4vUq

Thanks to Xeeynamo for help with the UV pointer array


src/st/no0/e_ctulhu.c
```diff
@@ -581,8 +583,217 @@ void EntityCtulhuFireball(Entity* self) {
...
+                }
+                if (!prim->p2) {
+                    prim->p1++;
+                    if (prim->p1 > 13) {
+                        prim->p3 = 0;
+                        prim->drawMode = DRAW_HIDE;
+#ifdef VERSION_PSP
+                        // Looks to be a bug fix on PSP
+                        // PSX can index outside size of D_us_801C11B0
+                        goto primNext;
```
JoshSchreuder, line 770: There's a few ways to accomplish a match here, but I couldn't find a way without PSP defs.

One alternative is doing
```c
if (prim->p1 > 13) {
...
}
#ifdef VERSION_PSP
else {
#endif
...
#ifdef VERSION_PSP
}
#endif
```

which also matches, I can change to that if it feels more intuitive than a goto.

I think this is a bug fix on PSP since `D_us_801C11B0` has 14 elements in the data, and on PSX this code can index past the end of the array after the prim gets hidden

sozud: The non-goto version is more clear to me so that would be my vote

sozud (approved):

src/st/no0/e_ctulhu.c
JoshSchreuder: Updated

PR #1913, opened 2024-11-19
Adding Faerie to CMakeLists

Onenutmcgee:
Adding Faerie to CMakeLists

Seems like SFX not working, but Faerie at least loads.  

~~Possibly also an issue with with the g_Entities array as the wings should occupy g_Entities[5] (set up in line 550 of Faerie.c), but they don't seem to either load or update (unsure which).~~
Seems this was either a graphical bug or something transient as the wings seem to work now...

I'm not particularly interested in debugging this issue at this time, so if you'd rather I abandon the PR, I can do that too.  There wasn't a lot of change.


src/servant/tt_002/faerie.c
```diff
@@ -461,6 +456,10 @@ void CheckForValidAbility(Entity* self) {
     }
 }
 
+#ifdef VERSION_PC
+extern u16 g_FaerieClut[64];
```
sozud, line 460: Why is this needed? The storage doesn't exist elsewhere?

sozud (approved):
The sfx issue is probably a bug in the libsnd code, it seems to have issues with loading and unloading VABs

src/servant/tt_002/faerie.c
Onenutmcgee: I did it this way because that was how the bat and the ghost and bat did it, but yeah, you're right, we can just do that in the actual extern declaration of the array.  The PSX version is fine with it as well.  I made that update in the bat and ghost as well to get rid of that ifdef

PR #1918, opened 2024-11-19
Remaining Dedupe for tt_003

Onenutmcgee:
I think this is the last of the dedupes that are available right now.


Onenutmcgee:
Decided to quickly figure out the FindValidTarget code so I could move the last dupe and the functions in rodata.

sozud (approved):

PR #1919, opened 2024-11-19
Enemy sfxIDs

synamaxmusic:


include/sfx.h
```diff
@@ -553,6 +547,18 @@ enum Sfx {
     SFX_SLOGRA_ROAR_DEFEAT,
     SFX_SLOGRA_PAIN_A, // Used for Gaibon
     SFX_SLOGRA_PAIN_B,
+    //
```
sozud, line 550: What's the empty comment for?

synamaxmusic: This can be removed if need be, but I put the empty comment there as a marker to denote that the sfxIDs after the Slogra sfxIDs are not sequential.

synamaxmusic: I removed it, should be all good now.

PR #1921, opened 2024-11-20
Update mwccgap

mkst:
This PR updates mwccgap. The latest version supports the `INCLUDE_RODATA()` macro which you guys may need at some point. 

It also moves the 'macro.inc' logic into mwccgap itself - as splat does not allow configuration of the asm prelude for the INCLUDE_RODATA files.

.. this version will let you pip a file into mwccgap, but you'll need to figure out how to pre-process the file (`cpp -P foo.c | python3 mwccgap.py ...`) whilst preserving the INCLUDE_ASM macros so mwccpsp doesnt cry - i ran out of steam trying to figure it out..

PS - added the missing sha1 for dra :)

### DIFF
--- a/Makefile.psp.mk
+++ b/Makefile.psp.mk
@@ -89,7 +89,7 @@ OPTIMIZATION = $(if $(findstring dra, $@), $(OPT_HIGH), -Op)
 
 $(BUILD_DIR)/%.c.o: %.c $(MWCCPSP) $(MWCCGAP_APP)
 	@mkdir -p $(dir $@)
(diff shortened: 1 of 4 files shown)


mkst:
Locally this fails when running with --jobs, let me try and understand what i've broken!

sozud (approved):

hohle (changes requested):
`tools/make-config.py` adds `.include "macro.inc"` to the header of any new PSP overlays.

https://github.com/Xeeynamo/sotn-decomp/blob/3809ebf6794b58521405635637ce15f80999e5c1/tools/make-config.py#L538

If this isn't necessary any longer can you remove it from that tool as well?

PR #1925, opened 2024-11-20
Decompile `no0` func_us_801D29F8

gamezter:
PSX : https://decomp.me/scratch/P02IW
PSP: https://decomp.me/scratch/LURHt (not a full match)

This was the largest function that hadn't been decompiled, A few things I've noticed:

- I had to update `func_us_801D2424` to take `Point32*`s instead of `Entity*`s, making me think that Entity position is probably a Point32
- Point32 probably uses `f32`s instead of `s32`s
- `self + 3` is referenced a lot and seems to use a separate extension struct. I named it `ET_UnkPlatelordPlus3`
- I made `ET_PlateLord` `unk88` and `unk9C` a separate inlined struct because they are often passed as an argument to other functions.
- I'm not sure if `unk_PlatelordStruct` `unk10` is an `Entity*` or just a `Point32*`. Both work, but I opted to leave it a `Point32*` for now


sozud (approved):

src/st/no0/e_plate_lord.c
```diff
@@ -58,7 +58,878 @@ bool func_us_801D29F8(s16* arg0, s32 arg1, s32 arg2) {
...
+            self->posX.val += self->velocityX;
+            self->posY.val += self->velocityY;
+            self->velocityY -= self->velocityY >> 8;
+            if (self->facingLeft) {
+                self->velocityX += FIX(0.125);
+            } else {
+                self->velocityX -= FIX(0.125);
+            }
+
+            if (abs(self->velocityX) == 0xC000) {
```
JoshSchreuder, line 572: `FIX(0.75)`

src/st/no0/e_plate_lord.c
```diff
@@ -58,7 +58,878 @@ bool func_us_801D29F8(s16* arg0, s32 arg1, s32 arg2) {
...
+                    DestroyEntity(tempEntity);
+                }
+                return;
+            }
+            if (!(self->ext.plateLord.unk80 & 3)) {
+                posX = self->posX.i.hi;
+                posY = self->posY.i.hi;
+                posX += (Random() & 0x1F) - 0x10;
+                posY += (Random() & 0x1F) - 0x10;
+                tempEntity = AllocEntity(D_8007D858, D_8007D858 + 32);
```
JoshSchreuder, line 891: `tempEntity = AllocEntity(&g_Entities[224], &g_Entities[256]);`

PR #1935, opened 2024-11-23
Stage collision and sensor docs

Xeeynamo:
Document most of the code related to the stage colliders and stage sensors. This is the result of the rabbit hole while checking how the button in NZ0 works.

I am still puzzled on how the various `Collider` fields are supposed to be named, but we're getting there.


sozud (approved):
Seems reasonable once the build is fixed

PR #1943, opened 2024-11-27
Extract `no0` Ctulhu data and some rename/refactor

JoshSchreuder:
Updated `EntityCtulhu` for a closer PSP match (not 100% but for some minor reg swaps)
https://decomp.me/scratch/YeTeE

And extract data, rename variables and fields and document code a bit.


src/st/no0/e_ctulhu.c
```diff
@@ -114,8 +155,9 @@ void EntityCtulhu(Entity* self) {
...
             break;
         case 4:
-            AnimateEntity(D_us_801823B8, self);
-            if (self->animFrameIdx == 3 && self->animFrameDuration == 0) {
+            // Fly horizontally for a period
+            AnimateEntity(anim_wing_flap, self);
+            if (!self->animFrameDuration && self->animFrameIdx == 3) {
                 PlaySfxPositional(SFX_WING_FLAP_A);
             }
             colRet = UnkCollisionFunc2(D_us_8018234C);
```
JoshSchreuder, line 163: Wasn't able to trigger this collision, so unsure what to name it. My guess is maybe it's a roof collision? Didn't seem to happen inside of `no0` anyway

config/symbols.us.stno0.txt
```diff
@@ -79,6 +79,7 @@ init_velocity_y = 0x801825C0;
 bone_projectile_velocity_x = 0x8018272C;
 sensors_special = 0x8018275C;
 sprites_nz0_3 = 0x801BEBB0;
+ctulhu_shockwave_uvs = 0x801C11B0;
```
JoshSchreuder, line 82: This sits in a later data section `0x27E0` so I've not imported this for now, just mapped the symbol.

PR #1947, opened 2024-11-28
Decompile `no0` func_us_801D4CAC

gamezter:
PSX: https://decomp.me/scratch/3hzY4
PSP: https://decomp.me/scratch/jnXaI

I used ext.plateLordUnknown because unkAC in ext.plateLord couldn't apply. However ext.plateLordUnknown.unk80 seems to be used as an angle elsewhere while here it looks like it's used as a timer. Should I make a new ext struct for this entity ?

### DIFF
--- a/src/st/no0/e_plate_lord.c
+++ b/src/st/no0/e_plate_lord.c
@@ -1362,13 +1362,13 @@ void func_us_801D4CAC(Entity* self) {
             x = self->posX.i.hi;
             y = self->posY.i.hi + 0xA;
             g_api.CheckCollision(x, y, &collider, 0);
-            if (collider.effects & 1) {
+            if (collider.effects & EFFECT_SOLID) {
                 self->velocityY = -self->velocityY / 2;
                 self->posY.i.hi += collider.unk18;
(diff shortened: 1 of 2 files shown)


src/st/no0/e_plate_lord.c
```diff
@@ -1333,7 +1333,52 @@ void func_us_801D4AA4(Entity* self) {
...
+        self->velocityY += FIX(0.375);
+        y = self->posY.i.hi + g_Tilemap.scrollY.i.hi;
+        if (y >= self->ext.plateLordUnknown.unkAC) {
+            x = self->posX.i.hi;
+            y = self->posY.i.hi + 0xA;
+            g_api.CheckCollision(x, y, &collider, 0);
+            if (collider.effects & EFFECT_SOLID) {
+                self->velocityY = -self->velocityY / 2;
+                self->posY.i.hi += collider.unk18;
+                self->ext.plateLordUnknown.unk80++;
```
sozud, line 1368: I think the unk80 is fine for now, we can continue documenting and cleaning up as we go

sozud (approved):

PR #1951, opened 2024-11-30
Decompile `no0` `EntityStageDecorations`

JoshSchreuder:
Thanks to @mrmidi and anonymous benefactor Clever Seal for the match
https://decomp.me/scratch/eEHm6

I've imported and documented the data here as well.


src/st/no0/e_stage_decoration.c
```diff
@@ -0,0 +1,119 @@
...
+    lamp_brasier,        clock_face_1,
+    clock_face_2,        clock_face_3,
+    window_highlight_1,  window_highlight_2,
+    window_highlight_3,  window_highlight_4,
+    clock_face_2,        clock_face_3,
+    window_highlight_2,  window_highlight_4,
+    red_moveable_floor,  red_moveable_floor,
+    window_highlight_3};
+
+void EntityStageDecoration(Entity* self) {
```
JoshSchreuder, line 98: This is quite similar to some other functions like `EntityRoomForeground` and `func_801B0958` in `nz0`

With how the data is packed and since this has a `facingLeft` I don't think this can be worked into the existing functions, or use `ObjInit` / `ObjInit2' unfortunately, so this will have to be an almost duplicate

sozud (approved):

PR #1952, opened 2024-11-30
Decompile `no0` `EntityDiplocephalusFireball`

JoshSchreuder:
Wasn't able to reunite the files just yet, maybe once all the Diplo entities are done

### DIFF
--- a/include/entity.h
+++ b/include/entity.h
@@ -2153,6 +2153,19 @@ typedef struct {
 } ET_Diplocephalus;
 
 typedef struct {
+    /* 0x7C */ s32 : 32;
+    /* 0x80 */ s32 : 32;
+    /* 0x84 */ s32 : 32;
+    /* 0x88 */ s32 unk88;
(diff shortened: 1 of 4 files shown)


include/entity.h
```diff
@@ -2152,6 +2152,19 @@ typedef struct {
...
+    /* 0x80 */ s32 : 32;
+    /* 0x84 */ s32 : 32;
+    /* 0x88 */ s32 unk88;
+    /* 0x8C */ s16 unk8C;
+    /* 0x8E */ s16 : 16;
+    /* 0x90 */ struct Primitive* unk90;
+    /* 0x94 */ s32 : 32;
+    /* 0x98 */ s32 : 32;
+    /* 0x9C */ u8 unk9C;
+} ET_DiplocephalusFireball;
```
JoshSchreuder, line 2166: Seems like there's a separate extension used here, 0x90 is a prim not an entity as it is in the other extension, similarly with 0x8C which is an s16 here but an entity in the other.

PR #1953, opened 2024-11-30
`tt_003` func_us_801737F0 decomp

mrmidi:


src/servant/tt_003/demon.c
```diff
@@ -229,7 +229,168 @@ INCLUDE_ASM("servant/tt_003/nonmatchings/demon", func_us_80172EF8);
...
+    if (inverseCounter < 0) {
+        inverseCounter += 7;
+    }
+    newY = posYOffset - (inverseCounter >> 3);
+    prim->y0 = prim->y1 = newY;
+
+    inverseCounter = 0x100 - self->ext.et_801737F0.animationTimer;
+    if (inverseCounter < 0) {
+        inverseCounter += 7;
+    }
```
JoshSchreuder, line 390: I'm pretty sure these are divisions, did not spend too much time playing around with it but this is a closer match for PSP (but farther from PSX)
https://decomp.me/scratch/RWU4x

src/servant/tt_003/demon.c
```diff
@@ -619,7 +744,104 @@ void func_us_80174FD0(Entity* self) {
...
+    case 5:
+        self->velocityX = self->facingLeft ? FIX(0.125) : FIX(-0.125);
+        self->velocityY = FIX(-0.25);
+        self->posX.val += self->velocityX;
+        self->posY.val += self->velocityY;
+
+        self->ext.demon.abilityTimer++;
+
+        if (self->ext.demon.abilityTimer > 30) {
+            self->entityId = 0xD1;
```
Onenutmcgee, line 837:
```suggestion
            self->entityId = DEMON_MODE_DEFAULT_UPDATE;
```

src/servant/tt_003/demon.c
```diff
@@ -240,7 +240,132 @@ INCLUDE_ASM("servant/tt_003/nonmatchings/demon", func_us_80172EF8);
...
+        posXOffset + ((cosine * -(scaledCounter)-sine * (scaledCounter)) >> 12);
+    prim->y2 = posYOffset +
+               ((sine * -(scaledCounter) + cosine * (scaledCounter)) >> 12);
+
+    prim->x3 =
+        posXOffset + ((cosine * (scaledCounter)-sine * (scaledCounter)) >> 12);
+    prim->y3 = posYOffset +
+               ((sine * (scaledCounter) + cosine * (scaledCounter)) >> 12);
+    newColor = ((self->ext.et_801737F0.stepCounter & 1) << 7);
+    PCOL(prim) = newColor;
```
Onenutmcgee, line 354: PSP match with
`PCOL(prim) = (self->ext.et_801737F0.animationTimer * 16) / 256;`

Then you can also get rid of the newColor var

Onenutmcgee (approved):
Just a couple of cleanup suggestions

src/servant/tt_003/demon.c
mrmidi: Not works on PSX - getting lbu instead of lhu instruction

Onenutmcgee: oh, I copied the wrong thing.  That was from a different function.  it should have been

`(self->ext.et_801737F0.stepCounter & 1) * 128`

That gets rid of the local var and does lhu.  Not a big deal.  I'm planning on onboarding this to the PSP build once we're done decomping and refactoring, so I'll catch it then.

mrmidi: It works! I'll include your suggestions with next PR

PR #1960, opened 2024-12-03
Decompile `no0` `func_us_801CE2E0`

JoshSchreuder:
Big thanks to bismurphy for helping me work out the array copy that appears at the beginning of the function (ended up being an inline array declaration)

https://decomp.me/scratch/Fo9ZW

This entity does not appear in the entity list and doesn't seem to be called from anywhere so I think it may be unused and deadstripped out of the PSP.


src/st/no0/4E2E0.c
```diff
@@ -1,6 +1,92 @@
...
+                player->posY.i.hi += offset;
+            }
+        }
+        flags = 15;
+        if (self->ext.et_801CE2E0.unk8A != 0) {
+            self->ext.et_801CE2E0.unk8A--;
+            flags = 11;
+        }
+
+        if (self->ext.et_801CE2E0.unk80 & 15 &&
```
bismurphy, line 43: Probably best to use hex notation for bitmasks - so using `& 0xF` and `& 0xF0` in this case, and for the others in this function.

src/st/no0/4E2E0.c
```diff
@@ -1,6 +1,92 @@
...
+            (self->ext.et_801CE2E0.unk80 & 240)) {
+            flags |= 16;
```
bismurphy, line 45: Same deal here, 0x10 instead.

sozud (approved):

PR #1963, opened 2024-12-03
Add NO1 Outer Wall overlay

bismurphy:
This should handle creating the new overlay. The files are already split out for easy deduping.

With Marble Gallery nearly done, seems to make sense to bring this one in next so it's ready to go ASAP.

Some other small tweaks for things I ran into while setting this up, but nothing huge.


config/splat.us.stno1.yaml
```diff
@@ -0,0 +1,71 @@
+
```
hohle, line 1: Extra line

tools/mwccgap
hohle, line 1: Is this related?

src/st/no1/e_red_door.c
```diff
@@ -0,0 +1,112 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "common.h"
+
+INCLUDE_ASM("st/no1/nonmatchings/e_red_door", EntityIsNearPlayer);
```
sozud, line 4: This file seems like it needs another split but doesn't have to be done now

sozud (approved):
Looks good but needs Xeeynamo's approval

tools/mwccgap
bismurphy: My local mwccgap was older than `master`, so I updated it. `master` was itself behind the most modern version, so I pulled it up to current rather than making my local version match the particular commit that `master` was at.

src/st/no1/e_red_door.c
bismurphy: Yep, the OVL_EXPORT on the red door means it's hard to auto-split on it; I thought about adding that functionality but decided that there wasn't much reason to, since this is the only function that would need it, and we can manually split on our own.

config/splat.us.stno1.yaml
bismurphy: Good catch, will remove

PR #1968, opened 2024-12-04
`Collider.effects` Lint

hohle:
Linter for `Collider.effects`.


hohle:
As discussed in #1961

PR #1970, opened 2024-12-04
Decompile `tt_003` func_us_80176C1C

gamezter:
PSX: https://decomp.me/scratch/Vus3H
PSP: https://decomp.me/scratch/28GaG

### DIFF
--- a/src/servant/tt_003/demon.c
+++ b/src/servant/tt_003/demon.c
@@ -1609,7 +1609,126 @@ void func_us_80176814(Entity* self) {
     }
 }
 
-INCLUDE_ASM("servant/tt_003/nonmatchings/demon", func_us_80176C1C);
+extern s32 D_us_80178680;
+extern s32 D_us_80178684;
+
+void func_us_80176C1C(Entity* self) {
+    Primitive* prim;
+    s32 i;
+
+    switch (self->step) {
+    case 0:
+        if (!self->params) {
+            D_us_80178680 = self->posX.val;
+            D_us_80178684 = self->posY.val;
+        } else {
+            self->posX.val = D_us_80178680;
+            self->posY.val = D_us_80178684;
+        }
+        self->primIndex = g_api.AllocPrimitives(PRIM_GT4, 8);
+        if (self->primIndex == -1) {
+            DestroyEntity(self);
+            return;
+        }
+        self->flags =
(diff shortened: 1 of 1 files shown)


src/servant/tt_003/demon.c
```diff
@@ -1609,7 +1609,126 @@ void func_us_80176814(Entity* self) {
...
+                prim->y0 = prim->y1 = self->posY.i.hi - 4;
+                prim->y2 = prim->y3 = self->posY.i.hi + 4;
+            }
+
+            prim->r0 = prim->r1 = prim->r2 = prim->r3 = prim->g1 = prim->g2 =
+                prim->g2 = prim->g3 = prim->b0 = prim->b1 = prim->b2 =
+                    prim->b3 = 0;
+            prim = prim->next;
+        }
+        g_api.GetServantStats(self, 0x1A, 1, &s_DemonStats);
```
Onenutmcgee, line 1680: 0x1A (ie, dec 26) probably needs to be added to the FamiliarAbilityIds enum.  Not sure what it is yet, but it's an ability identifier.  We also missed a couple of them like 0x18 (24) in func_us_801765A0 and the 25 in the function func_us_80176814, if you wouldn't mind fixing those when you are adding ability ids.

src/servant/tt_003/demon.c
```diff
@@ -1609,7 +1609,126 @@ void func_us_80176814(Entity* self) {
...
+            if (i < 7) {
+                prim->y0 = prim->y1 = self->posY.i.hi - 6;
+                prim->y2 = prim->y3 = self->posY.i.hi + 6;
+            } else {
+                prim->y0 = prim->y1 = self->posY.i.hi - 4;
+                prim->y2 = prim->y3 = self->posY.i.hi + 4;
+            }
+
+            prim->r0 = prim->r1 = prim->r2 = prim->r3 = prim->g1 = prim->g2 =
+                prim->g2 = prim->g3 = prim->b0 = prim->b1 = prim->b2 =
```
Onenutmcgee, line 1676: this is setting prim->g2 twice?  If it matches, it matches, but that's likely a bug in the codebase then, which is neat.  Assuming that's correct, could you add a comment noting that?

Onenutmcgee (approved):
Looks good to me.  Couple of suggestions.

PR #1972, opened 2024-12-04
`tt_003` func_us_801771B0 decompile

mrmidi:
Big shout to @Onenutmcgee for figuring out that hitboxOffX seems to be s16 for Servants overlays.


src/servant/tt_003/demon.c
```diff
@@ -1611,7 +1704,133 @@ void func_us_80176814(Entity* self) {
...
+            CreateEventEntity(self, 0xDE, self->params + 1);
+            self->step += 1;
+        }
+        break;
+
+    case 2:
+        self->posX.val += self->velocityX;
+        self->ext.et_801737F0.animationTriggerCount += 1;
+
+        if ((u16)(self->posX.i.hi + 0x20 & 0xFFFF) >= 0x141U) {
```
Onenutmcgee, line 1795: looks like the cast in necessary, but you can clean it up a bit with
```suggestion
        if ((u16)(self->posX.i.hi + 0x20) > 0x140) {
```

src/servant/tt_003/demon.c
```diff
@@ -1611,7 +1704,133 @@ void func_us_80176814(Entity* self) {
...
+            prim->y1 = posY_hi + D_us_80171B44[i][3];
+            prim->x2 = posX_hi + D_us_80171B44[i][4];
+            prim->y2 = posY_hi + D_us_80171B44[i][5];
+            prim->x3 = posX_hi + D_us_80171B44[i][6];
+            prim->y3 = posY_hi + D_us_80171B44[i][7];
+        }
+
+        colorValue =
+            ((self->ext.et_801737F0.animationTriggerCount & 1) << 6) + 0x40;
+        PCOL(prim) = colorValue;
```
Onenutmcgee, line 1830: I think this can be cleaned up as well
```suggestion
        PCOL(prim) = ((self->ext.ILLEGAL.s16[0] & 1) * 64) + 0x40;
```

Onenutmcgee (approved):
Looks good to me, I have a couple of cleanup suggestions below.  Also, and this probably falls into the nitpicky catagory, but x += 1 is usually just x++.  Just to keep things consistent.

src/servant/tt_003/demon.c
mrmidi: I have not problems with nitpicky reviews. It's all make sense

sozud (approved):

PR #1978, opened 2024-12-06
Fix builtin header definitions

JoshSchreuder:
Should mean no more

```
ctx.c:809: warning: conflicting types for built-in function `strcmp'
ctx.c:811: warning: conflicting types for built-in function `strcpy'
ctx.c:812: warning: conflicting types for built-in function `strlen'
ctx.c:813: warning: conflicting types for built-in function `memcpy'
```

When generating context


sozud (approved):
Seems good to me since people have been complaining about this for a while, but Xeeynamo should probably review due to the incoming psyz stuff

PR #1979, opened 2024-12-07
Decompile `no0` `EntityDiplocephalus`

JoshSchreuder:
PSX: https://decomp.me/scratch/fYrI3
PSP: https://decomp.me/scratch/CYEOp (almost matching but some minor register diffs)

Wasn't able to recombine the 2 Diplo files even after decompiling all of this and the rodata. Possible there is just a file split here as it is a pretty large entity with a lot of code.

### DIFF
--- a/include/entity.h
+++ b/include/entity.h
@@ -2137,10 +2137,30 @@ typedef struct {
 typedef struct {
     /* 0x7C */ struct Entity* parent;
     /* 0x80 */ struct Entity* entity;
+    /* 0x84 */ struct Entity* entity84;
+    /* 0x88 */ struct Entity* entity88;
+    /* 0x8C */ struct Entity* entity8C;
+    /* 0x90 */ struct Entity* entity90;
+    /* 0x94 */ s32 velocityX;
+    /* 0x98 */ s32 velocityY;
+    /* 0x9C */ u8 unk9C;
+    /* 0x9D */ u8 unk9D;
+    /* 0x9E */ u8 unk9E;
+    /* 0x9F */ u8 unk9F;
+    /* 0xA0 */ struct Entity* entityA0;
+    /* 0xA4 */ struct Entity* entityA4;
+    /* 0xA8 */ s32 : 32;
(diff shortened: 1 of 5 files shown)


src/st/no0/no0.h
```diff
@@ -31,9 +31,13 @@ typedef enum EntityIDs {
...
     /* 0x20 */ E_CLOCK_ROOM_SHADOW = 0x20,
-    /* 0x21 */ E_ID_21,
-    /* 0x24 */ E_DIPLOCEPHALUS_TAIL = 0x24,
+    /* 0x21 */ E_DIPLOCEPHALUS,
+    /* 0x22 */ E_ID_22,
+    /* 0x23 */ E_ID_23,
+    /* 0x24 */ E_DIPLOCEPHALUS_TAIL,
     /* 0x25 */ E_DIPLOCEPHALUS_FIREBALL,
+    /* 0x26 */ E_DIPLOCEPHALUS_TORSO,
+    /* 0x27 */ E_OLROX_DROOL,
```
JoshSchreuder, line 40: This entity should probably be renamed something more generic at some point as it's not Olrox related

include/entity.h
```diff
@@ -2137,40 +2137,40 @@ typedef struct {
...
-    /* 0x9F */ u8 : 8;
-    /* 0xA0 */ u32 : 32;
-    /* 0xA4 */ s32 posX;
-    /* 0xA8 */ s32 posY;
+    /* 0x9F */ u8 unk9F;
+    /* 0xA0 */ s32 : 32;
+    /* 0xA4 */ s32 unkA4;
+    /* 0xA8 */ s32 unkA8;
     /* 0xAC */ s32 unkAC;
 } ET_DiplocephalusUnk;
```
JoshSchreuder, line 2175: I'm still not 100% certain about these two structs, I have a feeling like `ET_Diplocephalus` is the main entity struct, and `ET_DiplocephalusUnk` is for the subentities like the torso, tail, feet etc.

But will need to study the code further in debugger to fully decode this

PR #1980, opened 2024-12-07
Decompile `no0` `EntityClockRoomController`

JoshSchreuder:
PSX: https://decomp.me/scratch/4mlgb
PSP: https://decomp.me/scratch/nhV6T (not a full match)

I used PSP as basis here but from what I can tell it was refactored a bit compared to PSX - a number of changes and additions that make ifdefing hard.

Potentially in future it might make more sense to make these separate implementations.
Apologies for how messy all the conditionals are here.


src/st/no0/clock_room.c
```diff
@@ -55,6 +67,356 @@ void UpdateClockHands(Entity* self, PlayerStatus* status) {
...
+        g_Player.D_80072EFC = 1;
+        entity = &PLAYER;
+        posX = entity->posX.i.hi;
+#ifdef VERSION_PSP
+        FntPrint("step=%d\n", self->step_s);
+#endif
+        switch (self->step_s) {
+        case 0:
+#ifndef VERSION_PSP
+            primIndex = g_Player.status; // FAKE
```
JoshSchreuder, line 247: Needed this to fix a stack mismatch, not 100% sure why but this worked

JoshSchreuder:
Good call as always, I have done that and it does seem a lot clearer, thanks

PR #1985, opened 2024-12-08
Garbage collect PSP symbols

Xeeynamo:
Start removing all the `#if !defined(VERSION_PSP)` wraps on unused data and functions

I am not sure about the `symexport` naming scheme and if we want to proceed with this approach. It is an important decision to take before merging this PR as it will shape how future PSP overlays will be written into the code base. Leaving this PR as a draft until we're fully onboard with this design choice.

### DIFF
--- a/config/undefined_syms.pspeu.txt
+++ b/config/undefined_syms.pspeu.txt
@@ -27,7 +27,6 @@ g_randomNext        = 0x091CE6B0;
 g_CurrentEntity     = 0x091E1678;
 PfnEntityUpdates    = 0x08C6BC30;
 g_PlayableCharacter = 0x091CF020;
-D_8018047C          = 0x09246ED0;
 g_PlayableCharacter = 0x091CF020;
 D_091CF6DC          = 0x091CF6DC;
 D_091CF698          = 0x091CF698;


bismurphy:
This design seems okay, but probably should have some documented information somewhere about what this is and how it works.

bismurphy (approved):

src/st/wrp_psp/e_collect.c
```diff
@@ -441,6 +441,6 @@ void EntityPrizeDrop(Entity* self) {
 
 #include "../blink_item.h"
 
-u16 g_ItemIconSlots[ICON_SLOT_NUM];
+u16 _g_ItemIconSlots[ICON_SLOT_NUM];
```
sozud, line 444: Is there a technical reason for this rename?

config/symexport.pspeu.stwrp.txt
```diff
@@ -0,0 +1,4 @@
+EXTERN(_binary_assets_st_wrp_mwo_header_bin_start);
```
sozud, line 1: I think we probably want something on the wiki explaining how this file works

sozud (approved):
Seems reasonable to me

sozud:
Seems fine to me still

PR #1991, opened 2024-12-09
Refactor tt_003 part1

Onenutmcgee:
Got the SFX figured out and refactored most of the functions that use those values.  Still need to go through the special attack update functions to clean those up.

This is about the half way point I think.


src/servant/tt_003/demon.h
```diff
@@ -1,20 +1,54 @@
...
-    DEMON_EVENT_ATTACK_UPDATE,
+    DEMON_SPECIAL_ATTACK_UPDATE,
     DEMON_MODE_UNK_DB,
     DEMON_MODE_UNK_DC,
     DEMON_MODE_UNK_DD,
     DEMON_MODE_UNK_DE
 } EntityIds;
+
+// SFX Ids
+#define NA_VO_DEMON_INTRO_COMMAND 0x4E8 // Command me, my lord and master
```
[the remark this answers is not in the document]
Onenutmcgee: Sounds good.  I clean that up when I get a moment tonight

sozud (approved):

src/servant/tt_002/faerie.h
```diff
@@ -33,32 +33,31 @@ typedef enum {
...
-#define NA_SE_FAERIE_HA1 0x88B
-#define NA_SE_FAERIE_HA2 0x88C
-#define NA_SE_FAERIE_AH 0x88D
-#define NA_SE_FAERIE_UHH 0x88E
-#define NA_SE_FAERIE_UHH2 0x88F
-#define NA_SE_FAERIE_OHH 0x890
-#define NA_VO_FAERIE_OH_NO 0x891      // "OH NOOO!"
-#define NA_VO_FAERIE_ARE_YOU_OK 0x892 // "Are you ok?" possibly unreachable code
+typedef enum {
+    FAERIE_INTRO_LIFE = 0x472, // "Thank you for giving me life"
```
sozud, line 37: As far as I understand we're combining two different things here. The lower numbers are XA audio while the higher ones are sound effects. The higher ones should be part of https://github.com/Xeeynamo/sotn-decomp/blob/0c25f09d9747a5521f282fead949d4d7ca202ea5/include/sfx.h#L564

Onenutmcgee: Originally, when I did the Faerie, I had the SFX ids in the sfx header. 
I think we had a conversation about it here:
https://github.com/Xeeynamo/sotn-decomp/pull/1909

I don't think we really ever game to any hard conclusion.  I think this is a design philosophy question really, ie, do we want the keys to be in the places that they are used or in a single enum.  I don't particularly have a problem with either, but I'm quite sure that they keys here are used only in the Faerie (possibly the japan only fairy as well?).  My opinion would be that it makes more sense to have the keys that are used by one entity in that entity's code and have the shared keys in the sfx header.  I do realize this makes things a bit harder to sus out what is shared and what isn't when we are only part way through the work though.  That being said, I believe all of the keys that are not loaded from the servant specific binary (ie, all the ones under 0x600), are only used inside those servants anyway.

I don't really have a problem with either way, but we should probably see what @Xeeynamo thinks?

sozud: AFAIK the only actual evidence we have is the shifted IDs in MAD. We don't have any info about servants. https://github.com/Xeeynamo/sotn-decomp/blob/0c25f09d9747a5521f282fead949d4d7ca202ea5/include/sfx.h#L264
The shifted IDs there do suggest it was an enum originally though or at least 0x630-0x6DB was

Onenutmcgee: From what I have seen thus far, there are no collisions in the servant keys. At least there isn't in the servants we currently have (unknown if true about sword).  

I think they *could* be all in a single enum.  It's really more a question of if we want to organize it that way.

Onenutmcgee: Sounds good.  I will move the keys that are shared into sfx.h (those are all they keys under 0x600 btw) and will keep the keys that are loaded from the VB/VH in the servant file.

I'll fix this tonight when I get home from work, then this should be good to merge.

PR #2001, opened 2024-12-15
Split no1 entities

JoshSchreuder:
Wasn't able to work out everything through playing and debugging. 

I suspect based on looking at some of the entities that there may be unused enemies, or maybe enemies from reverse castle loaded here and not used which can be worked out later using the debug overlay


sozud (approved):
Seems good, needs a rebase though

PR #2011, opened 2024-12-19
Decompile `tt_004` `ServantInit` and `func_us_80176BF8`

hohle:
This also includes the data segment to leverage the compiler generated offsets instead of manual symbol table entries.

Both `ServantInit` and `func_us_80176BF8` reference some data with fixed offsets that get generated as separate symbols by splat but can otherwise be a contiguous array.

Unlike other servants, the Sword Familiar splits its CLUT into two non-contiguous sections, so the common `g_ServantClut` declaration could not be used. It also does not have an obvious `SpriteParts` analog and draws differently from other familiars.


src/servant/tt_004/sword.c
```diff
@@ -423,7 +423,146 @@ void func_us_80173CB8(Entity* self) {
...
+
+    self->zPriority = PLAYER.zPriority - 2;
+    self->facingLeft = (PLAYER.facingLeft + 1) & 1;
+
+    if (mode == MENU_SWITCH_SERVANT) {
+#ifdef VERSION_PSP
+        if ((D_8003C708.flags & FLAG_UNK_20) ||
+            (D_8003C708.flags & FLAG_UNK_40)) {
+#else
+        if ((LOW(D_8003C708.flags) & (FLAG_UNK_40 | FLAG_UNK_20)) != 0) {
```
sozud, line 504: The PSP version doesn't match on PS1? Seems like an optimization

sozud (approved):

src/servant/tt_004/sword.c
hohle: I'll check again.

hohle: Yeah, I just tried again and couldn't get them to match. The load is different (word vs half) as well. Possibly a bug in the original (loading the word) leading to the logic change. 🤷

PR #2012, opened 2024-12-20
Dedupe `no1` `e_red_door` + `EntityUnkId12`

JoshSchreuder:
~~This will need rebase once #2009 is merged due to file splits.~~ ✅ Done

NO1 has some unique red door code here, I'm not really sure why but there is some new stuff here that updates zPriority for a bunch of things. Also needed a new extension, I couldn't see a way to use the existing `ET_Door`

PSX: https://decomp.me/scratch/V72pD
PSP: https://decomp.me/scratch/klIDQ


src/st/e_red_door.h
```diff
@@ -46,10 +50,22 @@ void OVL_EXPORT(EntityRedDoor)(Entity* self) {
 
     switch (self->step) {
     case 0:
+#ifdef STAGE_IS_NO1
+        self->ext.no1RedDoor.unk88 = self->params & 0x10;
```
sozud, line 54: I'm thinking `self->ext.no1RedDoor` and `self->ext.door` are not going to work well together in the same function. The alignment could be different. I think options would be just copy pasting the whole thing and having e_red_door_no1.h or something like
```
#ifdef NO1
#define EXT no1RedDoor
#else
#define EXT door
#endif
```
```
self->ext.EXT....
```

I'd wait for @Xeeynamo to take a look before making changes though

JoshSchreuder: Yeah it's a weird one. I could also replace them all with `LOH(self->ext.door.sideToPlayer)` for a match but I'm not sure that's actually what's going on here...

sozud: I think `LOH(self->ext.door.sideToPlayer)` would be better probably, since Xeeynamo seems to be busy we can go with that, or we can wait if you prefer

JoshSchreuder: I'm ok waiting and getting some wider thoughts, it is the holidays so no reason to rush this through

JoshSchreuder: Good call, I didn't notice that about those two props. I have made the changes as suggested!

PR #2013, opened 2024-12-20
Minor Servant Related Symbol Names

hohle:
Naming `D_80170000` and updating some constants to reflect values that have already been named.


include/servant.h
```diff
@@ -27,20 +27,20 @@ typedef enum {
...
+    PfnEntityUpdate Unk1C;
+    PfnEntityUpdate Unk20;
+    PfnEntityUpdate Unk24;
+    PfnEntityUpdate Unk28;
+    PfnEntityUpdate Unk2C;
+    PfnEntityUpdate Unk30;
+    PfnEntityUpdate Unk34;
+    PfnEntityUpdate Unk38;
+    PfnEntityUpdate Unk3C;
 } ServantDesc;
```
hohle, line 44: I left the names here, but this may be better described as a:

```lang=c
typedef struct {
     void (*Init)(InitializeMode mode);
     PfnEntityUpdate Update[15];
} ServantDesc;
```

The functions and their purpose are controlled by each familiar and just select by the overloaded `entityId` of the familiar.

hohle:
Checking build failure.

config/symbols.us.dra.txt
```diff
@@ -11,4 +11,5 @@ g_PlOvlAluBatSpritesheet = 0x8013C004;
 g_PlOvlSpritesheet = 0x8013C020;
 g_imgUnk8013C200 = 0x8013C200;
 g_imgUnk8013C270 = 0x8013C270;
+g_ServantDesc = 0x80170000;
```
sozud, line 14: I think the idea is to eventually be able to build without splat, but this makes it so we can't get this symbol at this address without it. Originally there was either a custom linker script or it's just hard coded casts of 0x80170000. @Xeeynamo any opinion here?

hohle: `ld` supports the `--just-symbols` option for providing a symbols file like is currently being used. I'm not sure if it's more or less flexible to continue to have have platform specific symbol tables outside of the code. That's not dissimilar from the current linker command and doesn't affect building with or without splat, AFAICT.

https://github.com/Xeeynamo/sotn-decomp/blob/5fa4630311eed05e31fb5075da490135a39a6b6e/Makefile#L116-L123

PR #2017, opened 2024-12-27
Decompile `no1` `e_blue_axe_knight`

JoshSchreuder:
Apologies for the size, this ended up being bigger than expected the more I picked at it.

Decompiles the blue version of axe knight found in Outer Wall, Castle Keep, Colosseum. This enemy is very similar to the existing green axe knight header we had, but has some additional attacks and an extra function.

As I decompiled it, I found it easier to separate this into a separate entity rather than trying to fit into the current shared header with conditionals.


config/symbols.us.stno1.txt
```diff
@@ -46,6 +48,7 @@ D_80182488 = 0x801827D8;
 D_801824B8 = 0x80182808;
 D_801824C0 = 0x80182810;
 D_80182A4C = 0x80182A4C;
+sprites_3 = 0x801A86BC;
```
JoshSchreuder, line 51: This is called `sprites_nz0_3` in axe knight, but I think this will probably need to be changed to use OVL_EXPORT or something similar in future once the other overlay sprite banks get imported. I've just named this a bit more generically for now and mapped the symbol direct

include/entity.h
```diff
@@ -958,11 +958,12 @@ typedef struct {
 
 typedef struct {
     /* 0x7C */ s8 unk7C;
-    /* 0x7D */ s8 unk7D;
+    /* 0x7D */ u8 unk7D;
```
JoshSchreuder, line 961: This change is needed for the blue axe knight, and did not cause any issues with the green axe knight. If preferred I can pull these into separate extensions but I don't think it is necessary.

src/st/entity_axe_knight.h
```diff
@@ -1,37 +1,67 @@
...
+// *********************
+
 static u8 anim_walk[] = {0x1A, 0x18, 0x0A, 0x19, 0x0A, 0x1A, 0x1A,
                          0x1B, 0x0A, 0x1A, 0x0A, 0x19, 0x00};
 static u8 anim_throw_duck[] = {
-    0x08, 0x1C, 0x08, 0x1D, 0x08, 0x1E, 0x08, 0x1F, 0x0A, 0x20, 0x04, 0x21,
-    0x04, 0x22, 0x04, 0x23, 0x00, 0x00, 0x00, 0x00, 0x08, 0x24, 0x08, 0x25,
-    0x02, 0x26, 0x02, 0x27, 0x02, 0x28, 0x20, 0x29, 0x08, 0x2A, 0x00};
+    0x08, 0x1C, 0x08, 0x1D, 0x08, 0x1E, 0x08, 0x1F, 0x0A, 0x20,
+    0x04, 0x21, 0x04, 0x22, 0x04, 0x23, 0x00, 0x00, 0x00, 0x00};
```
JoshSchreuder, line 50: Cleaned up these data imports a little - the green axe knight data here was combining two animations, only the first of which was in use. You can see this from the empty byte termination in between the two `0x00, 0x00, 0x00, 0x00`

src/st/entity_axe_knight.h
```diff
@@ -1,37 +1,67 @@
...
 static s16 sensors_move[] = {0, 32, 8, 0};
 static s16 sensors_ground[4][2] = {{0, 32}, {0, 4}, {8, -4}, {-16, 0}};
 static s16 dead_particle_pos[][2] = {
     {-8, -4}, {8, -6}, {-2, 4}, {-1, -7}, {2, 8}, {11, -7}, {4, 6}, {-3, 0}};
-static u16 unused[] = {
-    0x0114, 0x020A, 0x030A, 0x0414, 0x030A, 0x020A, 0x0000, 0x0000, 0x0508,
-    0x0608, 0x0708, 0x0D08, 0x0E08, 0x0F02, 0x1402, 0x1502, 0x1620, 0x1708,
-    0x0000, 0x0000, 0x0508, 0x0608, 0x0708, 0x0808, 0x090A, 0x0A04, 0x0B04,
-    0x0C04, 0x0000, 0x0000, 0x0E04, 0x0F04, 0x1004, 0x1104, 0x0000, 0x0000,
-    0x0D08, 0x0E08, 0x0F01, 0x1001, 0x1101, 0x1220, 0x1308, 0x0000};
```
JoshSchreuder, line 24: While decompiling the blue axe knight I found this is actually 5 separate animations. I've moved those changes over here too for completeness, even though these aren't used

sozud (approved):

PR #2018, opened 2024-12-31
Integrate tooling test in the CI pipeline

Xeeynamo:
We have tests for most of the important tools, but they need to be ran manually. It is possible to easily introduce accidental regressions. This PR adds new steps to the CI to run all those tests, plus a regression fix found by @JoshSchreuder


sozud (approved):
Seems good once the CI is passing

PR #2020, opened 2025-01-01
Update/document game logic for handling elemental resistances, absorbs, etc

bismurphy:
Lots of cleanups here with renaming stuff, creating enums, and more.

This was the kind of thing where I kept on having a cycle where I would improve all the uses of a certain field, then in doing so, I'd see where it was used, rename the function, notice more fields, and so on. But I think in the end I tied up the loose ends that I found, and this becomes a nicely contained PR which should help lots of study.


src/dra/menu.c
```diff
@@ -423,65 +423,66 @@ void CalcDefense(void) {
...
     }
 
-    if (CheckEquipmentItemCount(ITEM_MIRROR_CUIRASS, EQUIP_HAND) != 0) {
-        g_Status.D_80097C2C |= 0x200;
+    if (CheckEquipmentItemCount(ITEM_MEDUSA_SHIELD, EQUIP_HAND) != 0) {
+        g_Status.elementsImmune |= ELEMENT_STONE;
     }
-    if (CheckEquipmentItemCount(ITEM_ALUCARD_MAIL, EQUIP_HAND) != 0) {
-        g_Status.D_80097C2C |= 0x8000;
+    if (CheckEquipmentItemCount(ITEM_FIRE_SHIELD, EQUIP_HAND) != 0) {
```
bismurphy, line 456: We were using the wrong item enums here! They're on EQUIP_HAND so they need to be hand items. Glad I caught this eventually! There's a funny coincidence that the mirror cuirass and the medusa shield are the same enum ID, so this code looked reasonable at first glance.

src/dra/72BB0.c
```diff
@@ -792,63 +792,63 @@ void AlucardHandleDamage(DamageParam* damage, s16 arg1, s16 arg2) {
...
         PlaySfx(g_SfxPainGrunts[sfxIndex]);
-        if (damage->effects & EFFECT_UNK_0100) {
+        if (damage->effects & ELEMENT_CURSE) {
             g_Player.timers[1] =
                 GetStatusAilmentTimer(STATUS_AILMENT_CURSE, 0x400);
             g_Player.unk40 = 0x8165;
             CreateEntFactoryFromEntity(g_CurrentEntity, FACTORY(47, 2), 0);
             CreateEntFactoryFromEntity(g_CurrentEntity, FACTORY(44, 0x17), 0);
             PlaySfx(SFX_VO_ALU_WHAT);
-        } else if (damage->effects & EFFECT_SOLID_FROM_BELOW) {
```
JoshSchreuder, line 802: Were these auto-updated earlier from the collider effects linter? Perhaps we need to write something in there to prevent that going forwards (but won't block this PR)
https://github.com/Xeeynamo/sotn-decomp/blob/b778ddf88e5f1a2cb1b591546e38718faa0564cc/tools/lints/sotn-lint/src/collider_effects.rs#L11-L28

bismurphy: Not sure - but either way, the linter isn't being triggered now, so I don't think it's a huge issue, especially given that the DRA engine is essentially done. I don't think this particular linter is going to have to do its job much at all into the future so isn't worth tweaking its behavior.

include/game.h
```diff
@@ -1419,10 +1436,10 @@ typedef struct {
     /* 08 */ s16 attBonus;
     /* 0A */ s16 defBonus;
     /* 0C */ u8 statsBonus[4];
-    /* 10 */ u16 unk10;
-    /* 10 */ u16 unk12;
-    /* 14 */ u16 unk14;
-    /* 10 */ u16 unk16;
+    /* 10 */ u16 weaknessElements;
```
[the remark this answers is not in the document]
bismurphy: Good call, changed

sozud (approved):

PR #2021, opened 2025-01-02
PSP ST0 e_collect

Xeeynamo:
Decompiled most of `e_collect.c` in PSP ST0. The two missing functions are `CollectHeart` (https://decomp.me/scratch/6IYdJ) and `EntityEquipItemDrop`..

The biggest function I had to adapt was `EntityPrizeDrop`, which was a mix between the PSP version of WRP and the US ST0. There are five known variants of the function: US ST0, US MAD, US final, PSP ST0 and PSP final. It was quite challenging to bring everything together in a way that is decent to read. I added some comments on how I speculate the function evolved through the development.

### DIFF
--- a/src/st/collect_heart.h
+++ b/src/st/collect_heart.h
@@ -7,16 +7,8 @@ void CollectHeart(u16 arg0) {
     s8 heartPrizes[2];
     s8 mad_unknown[2];
 
-#if defined(VERSION_PSP)
-    heartPrizes[0] = c_HeartPrizes[0];
-    heartPrizes[1] = c_HeartPrizes[1];
-    mad_unknown[0] = c_HeartPrizes[0];
-    mad_unknown[1] = c_HeartPrizes[1];
-#else
     __builtin_memcpy(heartPrizes, c_HeartPrizes[0], 2);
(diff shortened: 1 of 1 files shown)


src/st/collect_heart.h
```diff
@@ -7,8 +7,16 @@ void CollectHeart(u16 arg0) {
     s8 heartPrizes[2];
     s8 mad_unknown[2];
 
+#if defined(VERSION_PSP)
+    heartPrizes[0] = c_HeartPrizes[0];
+    heartPrizes[1] = c_HeartPrizes[1];
+    mad_unknown[0] = c_HeartPrizes[0];
+    mad_unknown[1] = c_HeartPrizes[1];
```
JoshSchreuder, line 14: Believe these are likely just inline array definitions and not actually a memcpy (probably on ST0 / BETA also)
See https://decomp.me/scratch/Fqn5y

sozud: Are we going to apply this?

sozud (approved):
Seems good although we're pushing the limits on #ifdefs I think

PR #2023, opened 2025-01-02
Pull DRA DrawHudSubweapon into PSP

bismurphy:
Re-decompiled this for PSP and put the matching code into both. Code should be the same in both versions now; on the path toward being able to share code between the two.

I also identified LoadTPage, which required some adjustments to BlitChar to get the right types.


sozud (approved):
Seems good if the linting is fixed

bismurphy:
Thanks, this PR was a last-minute push as I ran out the door, of course I forgot to lint :) Should be all set now.

PR #2029, opened 2025-01-04
PSP GTE & Game Symbols

hohle:
Several GTE and game engine symbols that were not previously mapped for PSP.


hohle:
While working through `tt_004` I came across several of these functions used for PS1 source compatibility.

The `libgte.h` header may need to be replaced completely for PSP since these macros aren't used and this pollutes the PS1 SDK headers with PSP-isms.

PR #2033, opened 2025-01-05
Update `st` `collision` for PSP match

JoshSchreuder:
PSP related fixes for `ST0_HitDetection`
https://decomp.me/scratch/KITuv

Also made the same fixes to the shared `HitDetection` for other overlays as I think the same fixes are likely to be required there for PSP match when the time comes.


src/st/st0_hit_detection.h
JoshSchreuder, line 1: I am not sure if pulling this function into a shared header for PSP is a good approach or not. Thoughts?

PR #2037, opened 2025-01-05
DrawRichterHUD matching for PSP

bismurphy:
Yet another one in sequence. As usual, a bit of tweaking for the match, but looks nice on both systems.

Very curious what the PSP sprite is. Might need to fire up my PSP version and compare side by side. As far as I know there isn't a good debugging-friendly PSP emulator out there, because it would be nice to be able to view the VRAM to see what's in the tpage.

### DIFF
--- a/src/dra_psp/67F0.c
+++ b/src/dra_psp/67F0.c
@@ -7,7 +7,120 @@ INCLUDE_ASM("dra_psp/psp/dra_psp/67F0", func_psp_090E3170);
 
 INCLUDE_ASM("dra_psp/psp/dra_psp/67F0", func_psp_090E31F8);
 
-INCLUDE_ASM("dra_psp/psp/dra_psp/67F0", DrawRichterHud);
+extern PlayerHud g_PlayerHud;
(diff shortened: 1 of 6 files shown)


src/dra_psp/67F0.c
```diff
@@ -7,7 +7,120 @@ INCLUDE_ASM("dra_psp/psp/dra_psp/67F0", func_psp_090E3170);
...
+
+    SetTexturedPrimRect(prim, 18, 38, 8, 8, 0, 0);
+    prim->tpage = 0x1B;
+    prim->clut = 0x102;
+    prim->priority = 0x1F0;
+    prim->drawMode = DRAW_ABSPOS;
+    prim = prim->next;
+
+    // Extra HUD sprite on PSP! Should identify what this is.
+#ifdef VERSION_PSP
```
sozud, line 96: This is set up exactly the same as the previous prim so I'd think it's displaying the same thing?

sozud (approved):

src/dra_psp/67F0.c
bismurphy: Huh. Good catch. Can't imagine why they would possibly do that then.

PR #2038, opened 2025-01-06
PSP GTE & Game Symbols

hohle:
Several GTE and game engine symbols that were not previously mapped for PSP.


src/st/wrp_psp/stage_name_new.c
```diff
@@ -50,17 +45,17 @@ void func_psp_09244020(Primitive* arg0) {
     }
     sp70[2] = LOH(arg0->next->tpage);
     func_8928694(sp70[2], &sp18[8]);
-    func_892880C(&sp18[8], sp60);
+    TransMatrix((MATRIX*)&sp18[8], (VECTOR*)sp60);
```
sozud, line 48: Not a part of this PR but figured I should point out that since `s8 sp18[8] = {0};` something seems wrong about passing `&sp18[8]`. I think the stack layout is wrong or something

sozud (approved):

src/st/wrp_psp/stage_name_new.c
hohle: Yeah, I debated rewriting the function to use the correct types and layout, but wanted the mapping and function signatures for other work.

PR #2040, opened 2025-01-07
Finish a file for PSP DRA, with full code sharing

bismurphy:
Decided to finish out this file since the functions were simple - this PR was initially one function but I decided the changes to finish it were small so I did it all in one go rather than making tons of PRs one after another.


sozud (approved):
How was the file split determined?

src/dra/628AC.c
```diff
@@ -180,490 +176,32 @@ void func_80102D08(void) {
...
-    }
-    return 0;
-}
-
-void func_80103EAC(void) {
-    D_80137E4C = 0;
-    MemcardInfoInit();
-}
+// Next func in US is MemCardSetPort; next in PSP is UpdateCapePalette.
+// Indicates end of this file.
```
bismurphy, line 207: @sozud Comment here - the functions de-sync between the two versions, so that would suggest that this file is finished.

PR #2042, opened 2025-01-08
Decompile `no1` func_us_801B7188

gamezter:
PSX: https://decomp.me/scratch/Nggm8
PSP: https://decomp.me/scratch/2NBXw

I created a `PrimLine(x)` macro because prim is used as both a `Primitive` and a `PrimLineG2`. I'm not a huge fan of this solution so feel free to offer suggestions :)


src/st/no1/unk_35E20.c
```diff
@@ -381,7 +381,259 @@ void func_us_801B6490(Entity* self) {
...
-INCLUDE_ASM("st/no1/nonmatchings/unk_35E20", func_us_801B7188);
+extern u16 D_us_8018095C[];
+extern s16 D_us_80180E8C[];
+extern s16 D_us_80180E9C[];
+extern s16 D_us_80180ED4[];
+extern s32 D_us_80180EEC;
+extern s16 D_us_801D6328[];
+extern s16 D_us_801D6334[];
+
+#define PrimLine(x) ((PrimLineG2*)(x))
```
sozud, line 392: I think prim might have been a union originally which would explain the way it's used in this function. The macro is fine to me as well.

sozud (approved):

PR #2044, opened 2025-01-08
Decompile Several `tt_004` Functions

hohle:
Decompiles `UpdateServantDefault`, `func_us_801746BC`, and `func_us_80176270`.

I've added the `gte_stdp` and `gte_stflg` macros here which are required for `func_us_80172E84`, but don't have a match there yet.


src/servant/tt_004/sword.c
```diff
@@ -659,15 +664,394 @@ void ServantInit(InitializeMode mode) {
...
+        case 1:
+            D_us_80178548 = FIX(0xC0);
+            break;
+        case 2:
+            if (self->posX.i.hi > 128) {
+                var_v1 = FIX(0xC0);
+            } else {
+                var_v1 = FIX(0x40);
+            }
+            D_us_80178548 = var_v1;
```
bismurphy, line 710: Is var_v1 needed here? Usually this pattern matches just by doing direct assignment in both cases of the if-else.

sozud (approved):

src/servant/tt_004/sword.c
hohle: I should have caught that. It was a ternary and there were some other `MAX` conditions I was able to find and fix as well.

PR #2049, opened 2025-01-09
Remove Unncessary Symbols

hohle:
Cleans up symbols that are now determined at compile or link time and no longer references from ASM.


hohle:
I got a little ahead of myself on the first push. The format of `symbols.$VERSION.$OVERLAY.txt` seems to need to be compatible with both splat and `ld`, but `ld` doesn't like modern single line comments.

Moving the symbols to the `undefined_syms` file worked out and may be a great place for overlay boundary symbols.

PR #2052, opened 2025-01-10
Update `create_entity` funcs for PSP

JoshSchreuder:
Matched with codesharing for WRP and ST0


src/st/create_entities_horizontal.h
```diff
@@ -26,17 +27,35 @@ void CreateEntitiesToTheLeft(s16 posX) {
...
+        if (layoutObjHorizontal[LAYOUT_OBJ_POS_X] == LAYOUT_OBJ_START) {
+            return;
+        }
+
+        if (posX > layoutObjHorizontal[LAYOUT_OBJ_POS_X]) {
             return;
         }
 
+        flag = (layoutObjHorizontal[LAYOUT_OBJ_SLOT] >> 8);
+#ifdef VERSION_PSP
```
JoshSchreuder, line 40: I am not very happy with this conditional, but I could not find a way to reconcile the match here for PSP and PSX

See here for the PSP code in PSP decomp which is not matching, if anyone has a good idea here I can resolve this same pattern seen in a few of these files
https://decomp.me/scratch/MH94o

src/st/find_entity_horizontal.h
```diff
@@ -11,10 +11,11 @@
  *    `g_LayoutObjHorizontal` will be positioned at an element with a
  *    x position larger than `posX` or the tail eleement.
  */
-void FindFirstEntityToTheRight(s16 posX) {
+static void FindFirstEntityToTheRight(s16 posX) {
```
JoshSchreuder, line 14: These needed to be marked as static to resolve register swaps on PSP where these functions were called.

This [has been discussed recently on Discord](https://discord.com/channels/1079389589950705684/1079395108501331990/1326286609309630496) and may give us further PSP matches on code which is only just off.

PR #2057, opened 2025-01-11
PSP ST0 DrawCutsceneActorName

Xeeynamo:
Vastly different from `DrawCutsceneActorName` on PSX. I also found the localization function that select between languages.

### DIFF
--- a/src/st/st0_psp/cutscene.c
+++ b/src/st/st0_psp/cutscene.c
@@ -39,11 +39,7 @@ void DrawCutsceneActorName(u16 actorIndex, Entity* self, u16* actorNames,
 
     // Pre-calculate primitives that renders the actor's name
     x = 0x38;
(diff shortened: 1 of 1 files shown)


src/st/st0_psp/cutscene.c
```diff
@@ -1,7 +1,92 @@
...
+    ch = actorNameLength[arg3[actorIndex]];
+    primIndex = g_api.AllocPrimitives(PRIM_SPRT, ch);
+    if (primIndex == -1) {
+        DestroyEntity(self);
+        return;
+    }
+
+    // Pre-calculate primitives that renders the actor's name
+    x = 0x38;
+#if defined(VERSION_PSP)
```
sozud, line 42: Is this intentional?

sozud (approved):

PR #2062, opened 2025-01-12
Factor out register pointers

ser-pounce:
I was getting back into the project and noticed all of the duped registers around 0x235E8; I reckon the library devs might have accidentally made these read/write when they probably could have been de-duped in the .rodata (or even just a macro really...).
The symbol additions are a happy coincidence, if the remaining files had used these registers as well it would have been more annoying unless everything else were decompiled.


src/main/psxsdk/libcd/registers.h
```diff
@@ -0,0 +1,6 @@
+#include "types.h"
+
+static volatile u8* libcd_CDRegister0 = (u8*)0x1F801800;
```
sozud, line 3: What's happening is that c_XXX.c was originally one file, which got split up with a bunch of #ifdefs so that functions that don't get used by the game can be removed by the compiler. I suppose the issue here is that they didn't understand that only BSS will get deduplicated?

We don't have to replicate the #ifdef setup at the moment, just giving some info

Also not sure if other names would be better but the IO ports seem to be multipurpose so naming is hard https://psx-spx.consoledev.net/cdromdrive/#cdrom-controller-io-ports

sozud (approved):

src/main/psxsdk/libcd/registers.h
ser-pounce: Thanks for the insight, wasn't really sure what to call them as well.

PR #2063, opened 2025-01-12
Decompile `no1` EntityMistDoor

gamezter:
PSX: https://decomp.me/scratch/KHdpC
PSP: https://decomp.me/scratch/RdP5H


src/st/no1/e_mist_door.c
```diff
@@ -1,5 +1,72 @@
...
+        self->hitboxState = 1;
+        self->hitPoints = 0x7FFF;
+        self->attack = 0;
+        self->hitboxWidth = 0xC;
+        self->hitboxHeight = 0x18;
+    }
+    if (self->hitFlags) {
+        messageBox = AllocEntity(&g_Entities[224], &g_Entities[256]);
+        if (messageBox != NULL) {
+            CreateEntityFromCurrentEntity(E_SAVE_GAME_POPUP, messageBox);
```
gamezter, line 41: Maybe `E_SAVE_GAME_POPUP` should be renamed `E_MESSAGE_BOX` ? It is only used in one other place as a message box as well

sozud: This one is the same in every overlay I think? https://github.com/Xeeynamo/sotn-decomp/blob/f7086b647f4503a3873730e17d6a13b38d382f46/src/st/wrp/e_init.c#L27C5-L27C21

Here's the only example in the codebase so far I think? For the "magically sealed" text. https://github.com/Xeeynamo/sotn-decomp/blob/f7086b647f4503a3873730e17d6a13b38d382f46/src/st/e_magically_sealed_door.h#L179

I think it's just a generic message box and would make sense to change it.

sozud (approved):

src/st/no1/e_mist_door.c
```diff
@@ -1,5 +1,78 @@
...
+            D_psp_0929A520 = func_psp_092490A8(
+                D_psp_09265A88, D_psp_09265AD8, D_psp_09265AA0, D_psp_09265AF0,
+                D_psp_09265AB8);
+            D_psp_0929A518 = func_psp_092490A8(
+                D_psp_09265B88, D_psp_09265BD0, D_psp_09265BA0, D_psp_09265BF0,
+                D_psp_09265BB8);
+            D_psp_0929A510 = func_psp_092490A8(
+                D_psp_09265B08, D_psp_09265B58, D_psp_09265B20, D_psp_09265B70,
+                D_psp_09265B38);
+            if (g_PlayableCharacter == 2) {
```
JoshSchreuder, line 60: These have defines you can compare against (should probably be an enum, but I digress)

```
#define PLAYER_ALUCARD 0
#define PLAYER_RICHTER 1
#define PLAYER_MARIA 2
```

https://github.com/Xeeynamo/sotn-decomp/blob/master/include/game.h#L175-L177

JoshSchreuder (approved):

PR #2065, opened 2025-01-13
Update for PSP match on `st0` `EntityCutscene` file

JoshSchreuder:
Mostly already matching, just needed a few tweaks including the `static` keyword on a nearby func.


src/st/st0/2A218.c
```diff
@@ -17,8 +17,13 @@ void func_801AA218(s16 arg0) {
 extern s32 g_SkipCutscene;
 extern u32 g_CutsceneFlags;
 extern s32 g_IsCutsceneDone;
-static u8 D_80180830[] = {
-    0x08, 0x05, 0x08, 0x06, 0x08, 0x07, 0xFF, 0x00, 0x20, 0x01, 0xFF, 0x00};
+static u8 D_80180830[] = {0x08, 0x05, 0x08, 0x06, 0x08, 0x07, 0xFF, 0x00};
+
+#ifndef VERSION_PSP
+// This appears to be an unused animation and deadstripped on PSP
+static u8 unused[] = {0x20, 0x01, 0xFF, 0x00};
```
JoshSchreuder, line 24: Looking at the final two bytes here it seems likely this is an unused animation on PSX, since it does not exist on the PSP side and ends in the `0xFF, 0x00` terminator like the other animations here.

sozud (approved):

src/st/st0/2A218.c
JoshSchreuder: It works, very cool. Just pushed change

PR #2066, opened 2025-01-13
Decompile `no1` func_us_801C10F4

gamezter:
PSX: https://decomp.me/scratch/fOzMw

@mrmidi had found a 100% match on `func_us_801C1090`, I made a small change + rename variables so that it matches more closely similar code:
https://decomp.me/scratch/C3RbT

This one was a bit tougher than usual because I could not find the PSP code.

### DIFF
--- a/src/st/no1/unk_3FA34.c
+++ b/src/st/no1/unk_3FA34.c
@@ -103,7 +103,253 @@ s32 func_us_801C0E98(Primitive* prim) {
 }
 
 // support function for `func_us_801C10F4`
-INCLUDE_ASM("st/no1/nonmatchings/unk_3FA34", func_us_801C1090);
+void func_us_801C1090(s32* value, s32 target, s32 step) {
+    if (*value < target) {
+        *value += step;
+    }
+    if (*value > target) {
+        *value -= step;
+    }
+
+    if (*value < 0) {
+        *value = 0;
+    }
+
+    if (*value > 0x80) {
+        *value = 0x80;
+    }
+}
 
 // "Elevator activatd." text
-INCLUDE_ASM("st/no1/nonmatchings/unk_3FA34", func_us_801C10F4);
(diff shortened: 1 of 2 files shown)


sozud (approved):

src/st/no1/unk_3FA34.c
```diff
@@ -103,7 +103,253 @@ s32 func_us_801C0E98(Primitive* prim) {
...
+extern s32 D_us_801D7008;
+extern s32 D_us_801D700C;
+extern s32 D_us_801D7010;
+extern s32 D_us_801D7014;
+extern s32 D_us_801D7018;
+extern s32 D_us_801D701C;
+extern s32 D_us_801D7020;
+extern u16 g_EInitInteractable[];
+
+void func_us_801C10F4(Entity* self) {
```
mrmidi, line 137: I have a strong feeling that @hohle did it on the stream.

gamezter: Oh you're right : https://decomp.me/scratch/08mke I didnt catch it because the name of his decomp.me 'project' isn't `func_us_801C10F4`

mrmidi: I appreciate your pull request and have no objections to it. However, I believe it would be helpful to establish some form of communication to coordinate who is working on what. This would help us avoid duplicating efforts.

hohle: Yeah, sorry about that. I usually don't rename my PSP scratches.

When we were doing weapons, which all have functions named for their field offsets, we agreed to add `-wXX` or something similar to make them more easily identifiable.

Since the PSP versions have wildly different offsets, and it would be nice to have them findable when PSP work starts in earnest, I'm not sure what the best option is. :shrug: For functions that take more than one session, I'll often add comments linking them, but I'm not always good about that.

PR #2076, opened 2025-01-14
Update for full PSP match on `st0` `e_collect`

JoshSchreuder:
Updated with code sharing between PSX and PSP

Was able to resolve some of the data loading inside `collect_heart` which on ST0/beta is just an inline array init

I'm pretty sure I can bring in WRP PSP here too but I will do it in a separate PR


src/st/e_collect.h
```diff
@@ -100,14 +100,20 @@ u8* g_SubweaponAnimPrizeDrop[] = {
...
 #endif
 
+// from another file
+extern u16 g_EInitObtainable[];
+
+#include "prize_drop_fall.h"
+
+#include "prize_drop_fall2.h"
+
+#include "collect_heart.h"
```
JoshSchreuder, line 115: Had to shift some of the code around so it's interleved with data, since ST0/beta load their data inline so this needs to happen before `g_ExplosionYVelocities` etc. below.

I couldn't find any other way to accomplish this, any other way would put `heartPrizes` / `mad_unknown` at the end of the data imports

src/st/entity_equip_item_drop.h
```diff
@@ -83,10 +82,21 @@ void EntityEquipItemDrop(Entity* self) {
...
+#endif
         prim->u1 = prim->u3 = prim->u0 + 0xE;
+
+#ifdef VERSION_PSP
+        prim->v0 = prim->v1 = (i & 0x18) * 2 + 0x81;
+#else
         prim->v0 = prim->v1 = (u8)(i & 0x18) * 2 + 0x81;
+#endif
         prim->v2 = prim->v3 = prim->v0 + 0xE;
+
```
JoshSchreuder, line 99: Couldn't find a way to remove these casts on PSX. If someone can figure it out I can incorporate, see

// PSX: https://decomp.me/scratch/DQZyZ
// PSP: https://decomp.me/scratch/h5Klh

sozud (approved):

PR #2088, opened 2025-01-17
Update for matching `st0` `e_bg_vortex`

JoshSchreuder:
Few PSP specific changes but overall pretty similar


src/st/st0/e_bg_vortex.c
```diff
@@ -122,7 +135,7 @@ void EntityBackgroundVortex(Entity* self) {
         prim->priority = 2;
         prim->drawMode = DRAW_UNK_1000;
         prim = prim->next;
-        dr_env = g_api.func_800EDB08(prim);
+        dr_env = g_api.func_800EDB08((POLY_GT4*)prim);
```
JoshSchreuder, line 138: Not really related to this PR but wondering if I should clean up these headers to use `Primitive` instead? Seems to match fine on PSX and we could remove a few casts around the place like this
https://github.com/Xeeynamo/sotn-decomp/blob/master/include/game.h#L1570

If it's a good idea I can make a separate PR

sozud: Probably a question for @Xeeynamo, I don't really get all the prim variants

sozud (approved):

PR #2092, opened 2025-01-19
Big PSP decomp with code-sharing

bismurphy:
As established in my previous PR, I took a full C file, decompiled every function on PSP, backported the changes to US, and then enabled code-sharing. Ultimately this increases code quality for US. I also took the liberty of improving some symbol names, making an enum for the g_Player.timers[], and such. It's nice that this process means a new set of eyes is examining every function and generally finding improvements to make.


src/dra/6E42C.c
```diff
@@ -212,8 +213,11 @@ s16 g_SfxAttackGrunts[] = {
     SFX_VO_ALU_ATTACK_B,
     SFX_VO_ALU_ATTACK_C,
     SFX_VO_ALU_ATTACK_D,
+#ifndef VERSION_PSP
```
bismurphy, line 216: This is weird. I don't think these 0x0000 are real data, I think they're just padding between this array and the next chunk of data. PSP doesn't have it, but I couldn't get US to match without these 0x0000's explicitly written out, so we get an ifndef.

sozud: If you add padding after this variable does it match or does it have to be part of g_SfxAttackGrunts? If padding matches after the variable it's probably something unused that got stripped on PSP.

sozud (approved):

src/dra/6E42C.c
bismurphy: Great call! That matches and I've updated to do that instead.

PR #2093, opened 2025-01-19
Update for PSP match on `st_0` `2A8DC` secret room entities

JoshSchreuder:


include/stage.h
```diff
@@ -38,7 +38,6 @@ void Update(void);
 void HitDetection(void);
 
 s32 Random();
-void CreateEntityFromEntity(u16 entityId, Entity* source, Entity* entity);
```
JoshSchreuder, line 41: Having this defined causes an extra `andi` instruction on PSP converting between the u32 and u16 here. See https://decomp.me/scratch/qQwps

Removing it does not seem to cause any issues on PSX, but I can add it back conditionally if preferred

sozud: Does this add extra warnings in the cmake builds? If so maybe a `#ifdef VERSION_PC`? All those warnings ought to be fixed someday

JoshSchreuder: It does, on MSVC anyway. I'll wrap this one for now anyway, even though there are quite a few that aren't defined that are popping the same error (UnkPrimHelper, CreateEntityFromCurrentEntity, PlaySfxPositional, UnkPolyFunc0, UnkPolyFunc2, SetStep, GetSideToPlayer, AnimateEntity, etc.)

PR #2096, opened 2025-01-20
Fix dupes report

JoshSchreuder:
I think I found the reason why @Xeeynamo had to remove `disassemble_all` from the yaml configs to get the `make_config` tools test passing

I think there is a bug with `splat` where it's supposed to prioritise the command line argument over the yaml value but it's not working right now, as if you pass it in via argument it treats the yaml value as an unread option and fails a validation check since the option never gets parsed
https://github.com/ethteck/splat/blob/main/src/splat/util/options.py#L558-L560

### DIFF
--- a/tools/make-config.py
+++ b/tools/make-config.py
@@ -436,7 +436,7 @@ def get_splat_config(
             "ld_bss_is_noload": bss_is_no_load,
             "disasm_unknown": True,
             "include_macro_inc": False,
-            "disassemble_all": True
+            "disassemble_all": True,
         }
     }
 
--- a/tools/make-config.py
+++ b/tools/make-config.py
@@ -436,6 +436,7 @@ def get_splat_config(
             "ld_bss_is_noload": bss_is_no_load,
             "disasm_unknown": True,
             "include_macro_inc": False,
+            "disassemble_all": True
         }
     }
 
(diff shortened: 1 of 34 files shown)


tools/make-config.py
```diff
@@ -642,7 +643,10 @@ def find_dups(threshold, dir1, dir2) -> dict[str, str]:
 
 
 def split(splat_config_path: str, disassemble_all: bool):
-    if disassemble_all:
+    with open(splat_config_path) as f:
+        c = yaml.load(f, Loader=yaml.SafeLoader)
+        config_disasm_all = c["options"]["disassemble_all"]
+    if disassemble_all and not config_disasm_all == True:
```
JoshSchreuder, line 649: This is a bit of a hack to workaround the splat bug, basically check if it's already set to true in the yaml and if so don't try to set the command argument. Doesn't seem to slow down make_config too much

The alternative would be what xee was testing which is making this option passed in via command line always in the Makefile

tools/make-config.py
```diff
@@ -436,6 +436,7 @@ def get_splat_config(
             "ld_bss_is_noload": bss_is_no_load,
             "disasm_unknown": True,
             "include_macro_inc": False,
+            "disassemble_all": True
```
JoshSchreuder, line 439: Default this to true for new overlays

PR #2108, opened 2025-01-22
Dedupe `lib` `st_update`

JoshSchreuder:
Also codeshared with PSP


tools/make-config.py
```diff
@@ -891,7 +891,7 @@ def hydrate_stage_export_table_symbols(splat_config, version: str, export_table)
     # add_symbol(splat_config, version, "NULL", export_table[7]) # ??????
     add_symbol(splat_config, version, "g_TileLayers", export_table[8])
     add_symbol(splat_config, version, "g_EntityGfxs", export_table[9])
-    add_symbol(splat_config, version, "UpdateStageEntitiesAlt", export_table[10])
+    add_symbol(splat_config, version, "UpdateStageEntities", export_table[10])
```
JoshSchreuder, line 894: I am not certain this is a correct change, but there is no function called `UpdateStageEntitiesAlt` and this appears to have incorrectly mapped PSP symbol names? I can pull this out if it's incorrect

sozud (approved):
@Xeeynamo should probably comment on the make-config changes but the rest looks good

PR #2109, opened 2025-01-22
Update PSP `wrp` for match on `e_collect` funcs

JoshSchreuder:
[As discussed in Discord](https://discord.com/channels/1079389589950705684/1079395311350452255/1331207771844247552), had to move `EntityRelicOrb` out of `e_misc` and into `e_collect` so it can make use of a statically defined `BlinkItem` function which fixes a register mismatch.

I've applied this via conditional logic for now since this requires more work for other overlays which are compiled `e_collect, blit_char, e_misc` (`blit_char` would also need to move into `e_collect`). This can be done later on if we work out this is a common occurrence on PSP.


src/st/blink_item.h
```diff
@@ -1,7 +1,11 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "game.h"
 
-void BlinkItem(Entity* self, u16 renderFlags) {
+#ifdef STATIC_FUNC
+static
+#endif
```
JoshSchreuder, line 6: There's a few different ways this could be applied conditionally, I am open to suggestions if there's a preference.
One other option is to define something more globally like

```
#ifdef USE_STATIC
    #define STATIC static
#else
    #define STATIC
#endif
```

And this function would just always be `STATIC void BlinkItem(Entity* self, u16 renderFlags)` with WRP PSP defining `USE_STATIC`.

JoshSchreuder: Alternatively I could just pull this directly into WRP PSP and make it static

sozud: Does making it static everywhere cause a mismatch? Why have this be conditional?

JoshSchreuder: Yeah, because `EntityRelicOrb` is [defined in `e_misc.h`](https://github.com/Xeeynamo/sotn-decomp/blob/master/src/st/e_misc.h#L3) it's not possible to make it static always without also moving that function from `e_misc.h` to `e_collect.h`, since doing so means `EntityRelicOrb` can no longer see `BlinkItem` as it's not in the same file.

You end up with a bunch of compilation errors
```
mipsel-linux-gnu-ld: build/us/src/st/rwrp/e_misc.c.o: in function `LM227':
mipsel-linux-gnu-ldsrc/st/rwrp/e_misc.c:: (.text+0x8e0): undefined reference to `BlinkItem'
build/us/src/boss/rbo3/e_misc.c.o: in function `LM227':
src/boss/rbo3/e_misc.c:(.text+0x8e0): undefined reference to `BlinkItem'
mipsel-linux-gnu-ld: build/us/src/boss/mar/e_misc.c.o: in function `LM227':
src/boss/mar/e_misc.c:(.text+0x8e0): undefined reference to `BlinkItem'
mipsel-linux-gnu-ld: mipsel-linux-gnu-ldbuild/us/src/st/no3/e_misc.c.o: : in function `LM227':
src/st/no3/e_misc.c:(.text+0x8e0): undefined reference to `BlinkItem'
build/us/src/st/np3/e_misc.c.o: in function `LM227':
src/st/np3/e_misc.c:(.text+0x8e0): undefined reference to `BlinkItem'
mipsel-linux-gnu-ld: build/us/src/st/nz0/e_misc.c.o: in function `LM227':
src/st/nz0/e_misc.c:(.text+0x8e0): undefined reference to `BlinkItem'
```

So it is possible to do that but it would be a larger piece of work and require updating every overlay which is using `e_collect` and `e_misc`, which is most of them. There is also added complexity because for many overlays it would also involve merging `blit_char` into `e_collect` also, since that sits square in between `EntityEquipItemDrop` (e_collect) and `EntityRelicOrb` (e_misc) for overlays that use it.

sozud (approved):
Seems ok to me, we can continue iterating on this as we figure out more

JoshSchreuder:
> Would MWCC allow to do the following?

Great idea! This works, I have updated to use this and it is much more self-contained, thanks.

> What I am trying to say is that probably those three source files are actually part of the same file. Maybe to get static void BlinkItem matching on PSX as well would need for the three source files to get merged?

Yes I think this is likely, we will know more once we start to try to match `e_collect` on other PSP overlays I think.

JoshSchreuder:
I'm not sure how but it worked even without removing the `game.h` include from BlinkItem. Not sure how the `static` keyword ended up on the func, but anyway it's not required to include this in BlinkItem so I removed it as suggested also.

sozud:
Do you still want Xeeynamo to review this? I think his comments are resolved

JoshSchreuder:
@sozud if you're happy to merge this that's good with me, agree that I think feedback is addressed

PR #2111, opened 2025-01-22
`libcd` Decomp & Cleanup

hohle:
Decompiles several functions in `bios.c` and `sys.c`.

Cleans up several function signatures and types that conflicted with the SDK docs and headers, but otherwise provided implementations that produced correct assembly (e.g. functions which immediately call and return other functions with the same signature and all args in registers).

The remaining `bios.c` functions may require small changes to some types, but it's not clear yet what those might be.


hohle:
checking pc build failures

sozud (approved):

PR #2112, opened 2025-01-22
Reduce reliance on hand-coded asm for syscalls

ser-pounce:
No idea if this is useful or necessary, but less is usually more!


sozud (approved):
I still don't know how these objects were generated originally, whether it was C+inline asm, asm+aspsx or asm+asmpsx or something else. Other than that seems fine to me

PR #2120, opened 2025-01-24
Shop Inventory Data & Types

hohle:
Adding the default shop inventory tables and associated data along with
types and enums. I've updated @gamezter's function to use the updated
types and enums as well.

Thanks to @MottZilla and @eldri7ch for helping with offsets, structs,
and inventory info.


src/st/lib/e_shop.c
```diff
@@ -0,0 +1,691 @@
...
+typedef enum {
+    INVENTORY_HAND,
+    INVENTORY_HEAD,
+    INVENTORY_BODY,
+    INVENTORY_CAPE,
+    INVENTORY_ACCESSORY,
+    INVENTORY_RELIC,
+    INVENTORY_DOCUMENT,
+} InventoryCategory;
+
```
hohle, line 16: The bulk of new work is at the top of `e_shop.c`. The rest is either updating fields and constants in `func_us_801B2BE4` or moving functions to new segments.

src/st/lib/e_shop.c
```diff
@@ -0,0 +1,691 @@
...
+            if (g_Status.gold <
+                    D_us_801D415C[itemId] * D_us_801D4364[itemId].price ||
+                !D_us_801D415C[itemId]) {
+                g_api.PlaySfx(SFX_UI_ERROR);
+            } else {
+                g_Status.gold -=
+                    D_us_801D415C[itemId] * D_us_801D4364[itemId].price;
+                switch (D_us_801D4364[itemId].category) {
+                case INVENTORY_HAND:
+                    tempVar = 0;
```
sozud, line 457: I think you could use EquipKind here? https://github.com/Xeeynamo/sotn-decomp/blob/33c2cb1ed6d6cb8eb8368c0af79164ca6f75b63a/include/game.h#L960

sozud (approved):

src/st/lib/e_shop.c
hohle: Removed most of the `InventoryCategory` and using `EquipKind` instead. Added a `ShopEquipKind` for the additional kinds that the shop uses.

hohle: 🤦 sorry, missed the meaning. reverted and fixed.

PR #2121, opened 2025-01-25
Dedupe `lib` `e_room_fg` and `st_common`

JoshSchreuder:
`e_room_fg` was just one func and straightforward

`st_common` required some rematching on the PSP side but was able to make the code a better match overall and cleaner now it's shared between PSP and PSX.


config/splat.us.stlib.yaml
```diff
@@ -63,9 +65,10 @@ segments:
       - [0x41898, c, e_red_door]
       - [0x42548, c, e_room_fg]
       - [0x42634, c, st_common]
+      - [0x43AE4, c, e_misc]
       - [0x45E80, c, e_collect]
       - [0x473EC, c, blit_char]
-      - [0x4769C, c, e_misc]
+      - [0x4769C, c, e_misc_2]
```
JoshSchreuder, line 71: This is more evidence to me that `BlitChar`, `EntityRelicOrb`, `EntityHeartDrop`, `EntityMessageBox` likely reside within `e_collect`.

src/st/lib/e_misc_2.c
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "common.h"
+
+INCLUDE_ASM("st/lib/nonmatchings/e_misc_2", EntityRelicOrb);
+
+INCLUDE_ASM("st/lib/nonmatchings/e_misc_2", EntityHeartDrop);
+
+INCLUDE_ASM("st/lib/nonmatchings/e_misc_2", EntityMessageBox);
```
JoshSchreuder, line 8: This is more evidence to me that `BlitChar`, `EntityRelicOrb`, `EntityHeartDrop`, `EntityMessageBox` likely reside within `e_collect`.

The other functions in the file begin to be bloody zombie etc.

src/st/np3/np3.h
```diff
@@ -85,9 +85,11 @@ extern void func_801BB044(s16);
 extern void func_801BB140(s16);
 extern void func_801BB2F8(s16);
 extern void func_801BB3F4(s16);
-extern s32 UnkCollisionFunc(u16* hitSensors, s16 sensorCount);
+extern s32 UnkCollisionFunc(s16* hitSensors, s16 sensorCount);
 extern s32 GetPlayerCollisionWith(Entity* entity, u16 w, u16 h, u16 flags);
-extern s32 GetDistanceToPlayerX(void);
+#ifdef VERSION_PC
+extern s16 GetDistanceToPlayerX(void);
```
JoshSchreuder, line 91: Defining this header causes mismatches on PSX, I've left it in but conditional for PC only to avoid the warning there

src/st/lib/e_misc_2.c
sozud: What is the evidence? Not understanding

sozud (approved):

src/st/lib/e_misc_2.c
JoshSchreuder: The [currently defined `e_misc.h`](https://github.com/Xeeynamo/sotn-decomp/blob/master/src/st/e_misc.h#L14) starts with `EntityRelicOrb`, `EntityHeartDrop`, `EntityMessageBox` then `CheckColliderOffsets`.

LIB on the other hand puts `CheckColliderOffsets` earlier in the overlay at `0x43AE4`, then `e_collect`, then `blit_char` then these 3 funcs. 

```
      - [0x43AE4, c, e_misc] # CheckColliderOffsets etc.
      - [0x45E80, c, e_collect]
      - [0x473EC, c, blit_char]
      - [0x4769C, c, e_misc_2] # EntityRelicOrb / EntityHeartDrop / EntityMessageBox
```

So the current `e_misc.h` does not work for LIB, but it will if we move `EntityRelicOrb / EntityHeartDrop / EntityMessageBox` and `BlitChar` into `e_collect` instead, which fits with what we saw with the statically defined function in WRP PSP.

I don't necessarily think we should commit to that change across all overlays right now, but I think it makes it more plausible?

PR #2124, opened 2025-01-25
Fun with flags

ProjectOblivion:
I added most of the flags from the normal castle.  It's a huge amount of stuff, so I split it out into it's own file to make it a little bit easier (imo) to update going forward.  Many of the offsets have individual bits used so I included all of them in comments, but I left the structure unchanged.  I expect that handling the flags can be improved, but this is a start.

### DIFF
--- a/include/castle_flags.h
+++ b/include/castle_flags.h
@@ -1,0 +1,855 @@
+// Holds flags that maintain the persistant "state" of the castle.
+// These cover a wide range of uses, but what they have in common is a need to
+// track something across rooms or stages.
+extern u8 g_CastleFlags[0x300]; // starts at 0x8003BDEC
+
+// All offsets and values included for organizational convenience.
+typedef enum {
+    /* 0x0 */ CEN_OPEN = 0, // Wear the gold and silver rings in the clock room
+    /* 0x1 */ NO4_OPEN = 1, // Floor switch behind the blue door in NO0
+    /* 0x2 */ NO0_BANNER = 2,
+    /* 0x10 */ NO1_ELEVATOR_ACTIVATED = 16, // Set by switch in outer wall
+    /* 0x11 */ CASTLE_FLAG_17 = 17,         // Unknown trigger
+    /* 0x12 */ NO1_SECRET_WALL_BROKEN = 18, // Wall chicken wall fully broken
+
+    // Used for the weather in NO1. An option is randomly selected
+    // from g_NO1WeatherOptions when changing overlays.
+    // A value of 0x80 is rain
+    // A value of 0x81 is clear
+    // A value of 0x82 is fog
+    /* 0x13 */ NO1_WEATHER = 19,
+
+    // Used for the bird lifecycle under the bottom room of NO1.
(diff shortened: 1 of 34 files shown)


include/castle_flags.h
```diff
@@ -0,0 +1,783 @@
...
+// All offsets and values included for organizational convenience.
+typedef enum {
+    /* 0x0 */ CEN_OPEN = 0, // Set by wearing the gold and silver rings together
+                            // in the clock room
+    /* 0x1 */ NO4_OPEN = 1, // Floor switch behind the blue door in NO0
+    /* 0x2 */ NO0_BANNER = 2,
+    /* 0x10 */ NO1_ELEVATOR_ACTIVATED = 16, // Set by switch in outer wall
+    /* 0x11 */ CASTLE_FLAG_17 = 17,         // Unknown trigger
+    /* 0x12 */ NO1_SECRET_WALL_BROKEN = 18, // Wall chicken wall fully broken
+    // Offset 0x13 holds the NO1 weather state and is set for the first time
```
sozud, line 15: Could we adjust these for readability? e.g.

```
// Offset 0x13 holds the NO1 weather state and is set for the first time
// when entering the DG10 room. An option is randomly selected from
// g_NO1WeatherOptions when changing overlays.
// A flag value of 0x80 is rain, 0x81 is clear, and 0x82 is fog
```
More of the comments have this issue

include/castle_flags.h
```diff
@@ -0,0 +1,783 @@
...
+    /* 10000000 */ // NOVA_SKELETON_TERMINUS_EST
+    /* 01000000 */ // OROBOUROUS_LAPIS_LAZULI
+    /* 00100000 */ // WHITE_DRAGON_NONE
+    /* 00010000 */ // FIRE_WARG_KARMA_COIN
+    /* 00001000 */ // ROCK_KNIGHT_PLATINUM_MAIL
+    /* 00000100 */ // SNIPER_OF_GOTH_BRILLIANT_MAIL
+    /* 00000010 */ // SPECTRAL_SWORD_3_MABLUNG_SWORD
+    /* 00000001 */ // GHOST_DANCER_STONE_MASK
+    /* 0x1BF */ ENEMY_LIST_RAREDROP_1BF = 447,
+    /* 10000000 */ // WARG_RIDER_NONE
```
sozud, line 752: I'm thinking a newline would help readability, since it's not obvious if WARG_RIDER_NONE is part of ENEMY_LIST_RAREDROP_1BF or ENEMY_LIST_RAREDROP_1C0 at a glance

sozud (commented):
Seems OK if rebased and formatted. It's currently conflicted

include/castle_flags.h
ProjectOblivion: Oh, understand now.  The comments seem to have gotten messed up by running make format (I only forgot to run it at the very end right before the PR).

sozud (approved):
Seems good to me now, I think this should get a second review from @Xeeynamo before merging

PR #2125, opened 2025-01-25
Finish decomp of intr.c, including segment allocation.

ser-pounce:
Big thanks to @dezgeg, @sozud and @mkst for this one!

Still needs some work but it at least builds correctly.


src/main/psxsdk/libetc/intr.c
```diff
@@ -1,47 +1,217 @@
...
+    u16 enabledInterruptsMask;
+    u16 savedMask;
+    int savedPcr;
+    jmp_buf buf;
+    s32 stack[1024];
+};
+
+static intrEnv_t intrEnv = {0};
+static struct Callbacks callbacks = {
+    "$Id: intr.c,v 1.73 1995/11/10 05:29:40 suzu Exp $",
```
hohle, line 48: The RCS ID is part of this struct? Usually this is a `static char const []` at the start of a file (maybe not `const` in this case to interleave the pointer in DATA instead of RODATA).

That also may or may not suggest that `intrEnv` was in a separate file.

src/main/psxsdk/libetc/intr.c
```diff
@@ -1,47 +1,217 @@
...
 int VSyncCallbacks(int ch, void (*f)()) {
-    return D_8002D340->VSyncCallbacks(ch, f);
+    return pCallbacks->VSyncCallbacks(ch, f);
 }
 
-int StopCallback(void) { return D_8002D340->StopCallback(); }
+void* StopCallback(void) { return pCallbacks->StopCallback(); }
 
-int RestartCallback(void) { return D_8002D340->RestartCallback(); }
+void* RestartCallback(void) { return pCallbacks->RestartCallback(); }
```
hohle, line 80: `StopCallback`'s return signature in docs is `void`. `RestartCallback`'s is `int`.

src/main/psxsdk/libetc/intr.c
dezgeg: In the COFF versions of the libs (which reveals names of the static symbols also) no symbol looks like it should belong to the string, so it really should be something of this form.

Conversely, if intrEnv were not static, the symbol name should show up in the psyq version of the libs, but it doesn't, so intrEnv should be a static symbol.

ser-pounce: I guess it does make sense to give the string its own array if that matches the layout of other files, I'm just being lazy 😅.

src/main/psxsdk/libapi/libapi_internal.h
```diff
@@ -0,0 +1,13 @@
+#ifndef LIBAPI_INTERNAL_H
+#define LIBAPI_INTERNAL_H
+
+#include <types.h>
+
+void ResetEntryInt();
+void HookEntryInt(u16*);
```
dezgeg, line 7: According to https://psx-spx.consoledev.net/kernelbios/ HookEntryInt takes an jmp_buf pointer.

src/main/psxsdk/libetc/intr.c
```diff
@@ -1,47 +1,217 @@
...
+    "$Id: intr.c,v 1.73 1995/11/10 05:29:40 suzu Exp $",
+    0,
+    setIntr,
+    startIntr,
+    stopIntr,
+    0,
+    restartIntr,
+    &intrEnv};
+struct Callbacks* pCallbacks = &callbacks;
+static volatile u16* i_mask = (u16*)0x1F801070;
```
dezgeg, line 57: `i_mask` should be named `i_stat` (probably my mistake)

src/main/psxsdk/libetc/intr.c
ser-pounce: Thanks, somehow missed that these were public.

src/main/psxsdk/libapi/libapi_internal.h
ser-pounce: Nice, I've updated the signatures.

src/main/psxsdk/libetc/intr.c
ser-pounce: Fixed

src/main/psxsdk/libetc/intr.c
```diff
@@ -1,47 +1,217 @@
...
+        return NULL;
+    }
+    *i_mask = *g_InterruptMask = 0;
+    *d_pcr = 0x33333333;
+    memclr(&intrEnv, sizeof(intrEnv) / sizeof(s32));
+    if (setjmp(intrEnv.buf) != 0) {
+        trapIntr();
+    }
+    intrEnv.buf[JB_SP] = (s32)&intrEnv.stack[1004];
+    HookEntryInt((u16*)intrEnv.buf);
```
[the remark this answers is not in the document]
ser-pounce: Fixed

src/main/psxsdk/libetc/intr.c
```diff
@@ -1,47 +1,200 @@
...
+        }
+    } else {
+        D_8002D350 = 0;
+    }
+    intrEnv.inInterrupt = 0;
+    ReturnFromException();
+}
 
-INCLUDE_ASM("main/nonmatchings/psxsdk/libetc/intr", restartIntr);
+Callback setIntr(s32 arg0, Callback arg1) {
```
dezgeg, line 131: Everything else has nice names, so I guess might as well do this also...
I'd say:

arg0 -> irq
arg1 -> handler
temp_s4 -> prevHandler
var_s3 -> mask

temp_v1 and the 0xffff mask is not needed, `var_s3 = *g_InterruptMask;` is enough

src/main/psxsdk/libetc/intr.c
```diff
@@ -1,47 +1,200 @@
...
+            intrEnv.handlers[arg0] = arg1;
+            var_s3 = var_s3 | (1 << arg0);
+            intrEnv.enabledInterruptsMask |= (1 << arg0);
+        } else {
+            intrEnv.handlers[arg0] = 0;
+            var_s3 = var_s3 & ~(1 << arg0);
+            intrEnv.enabledInterruptsMask &= ~(1 << arg0);
+        }
+        if (arg0 == 0) {
+            ChangeClearPAD(arg1 == 0);
```
dezgeg, line 151: `arg1 == 0` is comparing function pointer to null so `arg1 == NULL`, or `handler == NULL` after the previous renaming

src/main/psxsdk/libetc/intr.c
```diff
@@ -1,47 +1,200 @@
...
+                *i_stat = ~(1 << i);
+                if (intrEnv.handlers[i] != NULL) {
+                    intrEnv.handlers[i]();
+                }
+            }
+        }
+        mask = (intrEnv.enabledInterruptsMask & *i_stat) & *g_InterruptMask;
+    }
+    if (*i_stat & *g_InterruptMask) {
+        if (D_8002D350++ > 0x800) {
```
dezgeg, line 119: D_8002D350 would be named `nintr` but that is such an horrible name I wonder if that should be used :/

src/main/psxsdk/libetc/intr.c
ser-pounce: Nice, fixed

src/main/psxsdk/libetc/intr.c
ser-pounce: I agree, something like `trapMissedCount` ?

src/main/psxsdk/libetc/intr.c
ser-pounce: Fixed

sozud:
I tried matching this against the original lib and found some differences. Here's a PR although it's gotten sort of messy since it looks like some renaming was done or whatever. https://github.com/ser-pounce/sotn-decomp/pull/1

This is a diff of the two objs after my adjustments: https://gist.github.com/sozud/0d11373029558b242b7af2bf2f5dffb6

The main things are:

intrEnv gets initialized as 56 bytes then 4144 bytes. So as far as I can tell it needs to be done like this
```
struct intrEnv_t  {
    u16 interruptsInitialized; // 2
    u16 inInterrupt; //2 
    Callback handlers[11]; // 44
    u16 enabledInterruptsMask; // 2
    u16 savedMask; // 2
    struct Temp {
        int savedPcr; // 4
        jmp_buf buf; // 48
        s32 stack[1024];
    } temp;
};

static struct intrEnv_t intrEnv = {
    0,                         // interruptsInitialized
    0,                         // inInterrupt
    {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0},  // handlers (explicit zeros for each element)
    0,                         // enabledInterruptsMask
    0,                         // savedMask
    {0}, // temp
};
```

i_mask etc. need to be lower in the file.

D_8002D350 needs to be lower in the file.

Callbacks is odd with the rcs info and the &intrEnv but that's apparently real judging from the reloc info

sozud:
I figured out a matching version, there's probably improvements possible https://github.com/sozud/psy-q-decomp/commit/78fbe9e37995f2f7e6af4e800bd22e96ce22d797

sozud (approved):

PR #2127, opened 2025-01-26
Next file PSP code-sharing update

bismurphy:
Another file where I re-decompiled everything to match on PSP, back-ported the changes, and then enabled full code-sharing. Generally resulted in an increase in code quality. Nice to be able to see the diff and recognize that all the changes are quite minor and don't affect behavior really.


src/dra/6D584.c
```diff
@@ -247,59 +250,66 @@ void PlayAnimation(s8* frameProps, AnimationFrame** frames) {
     g_CurrentEntity->animCurFrame = animFrame->unk2 & 0x1FF;
 }
 
+// Nasty casting. This is just
+// g_CurrentEntity->anim[g_CurrentEntity->animFrameIdx] But PSP is weird and
+// does odd stuff with the struct indexing. So we cast the pointer to u16, index
+// off of animFrameIdx*2, and cast back to AnimationFrame.
+#define CURRANIM                                                               \
+    (*((AnimationFrame*)(&(                                                    \
+        ((u16*)g_CurrentEntity->anim)[g_CurrentEntity->animFrameIdx * 2]))))
```
sozud, line 259: Do you have the scratches for UpdateAnim handy? I'm wondering if this should just be a raw u16 array

bismurphy: Sure. PSP here: https://decomp.me/scratch/tdwym

And PS1 here: https://decomp.me/scratch/kUzMj

I did try doing raw u16 array, but that broke a bunch of other stuff in other places (including in tt_000), and I ended up changing so much that I decided it was more readable to make this one-off weirdness get handled by a clumsy define.

PR #2128, opened 2025-01-26
Minor renames, function realignment, and clarifying/identifying comments

ProjectOblivion:
Fixed e_mouse function attribution, made a couple minor function moves, and added some identifying information for some functions.


src/st/no1/e_wereskeleton.c
```diff
@@ -25,6 +26,7 @@ extern s16 D_us_80182930[];
 extern s16 D_us_80182940[];
 extern s32 D_us_80182950[][2];
 
+// Wereskeleton
 void func_us_801CDE20(Entity* self) {
```
JoshSchreuder, line 30: Is this unused? Based on the bestiary it seems like this should only be in Catacombs, would explain why I could track down some of these entities in the debugger.

sozud: It may be a a file with multiple skeleton variants. For example in the entrance there's fire warg included even though it's unused

PR #2129, opened 2025-01-26
Add fully matching CHI overlay

nyxojaele:
The work for this was largely completed back in Oct '24, and everything was brought up to current standards at the time. However, this branch has sat around since then while master was being updated. When I merged master into this branch again to create this PR, I did a cursory overview to bring things up to the new standards, but I fully recognize that I may have missed something.

This should be a fully matching, fully analysed, every byte accounted for (short of BSS), CHI overlay that should mostly be able to drop right into the project.

It's worth noting that this branch was started before any scripts existed to automate the creation of new overlays in any way, so I wouldn't be surprised if there is something missing that the new scripts do for us.

I generally endeavoured to make minimal changes to shared code, often opting to provide comments describing CHI's use-case, rather than actually renaming anything and risking this branch bleeding into other stage overlays.

I'm realizing that my filename naming scheme doesn't necessarily match that in the rest of the stages, however I'd like to present mine as a possible alternative. More than happy to switch over if it's undesired for the project tho:
* dt_  : "Code" data-only
* bn_  : Binary data
* en_  : Entity-specific functionality
* st_  : Stage-related general functionality
* ut_  : Utility, or otherwise general functionality


include/sfx.h
```diff
@@ -238,6 +238,17 @@ enum SfxModes {
 // plays every 10 frames while using bible subweapon
 #define BIBLE_SUBWPN_SWOOSH 0x8C3
 
+// STAGE CHI
+#define NA_SE_EN_VENUS_WEED_ATTACK_TENDRIL_SPIKE 0x6D2
```
sozud, line 242: we have a sfx enum, would you mind moving these in it?

src/st/chi/en_breakable.h
```diff
@@ -0,0 +1,209 @@
...
+            self->velocityX = FIX(-1);
+        } else {
+            self->velocityX = FIX(1);
+        }
+
+        temp = 0x8000;
+        temp_v0_2 = Random() << 8;
+        self->velocityX = self->velocityX + temp - temp_v0_2;
+        self->velocityY = FIX(-3);
+        self->velocityY = (self->params >> 1) * 0x6000 - 0x30000;
```
sozud, line 133: FIX for the constants?

src/st/chi/en_breakable_wall.c
```diff
@@ -0,0 +1,351 @@
...
+        self->animCurFrame = animCurFrame;
+        if (self->rotZ & 1) {
+            self->facingLeft = true;
+            self->rotZ &= 0xFFF0;
+        }
+
+        velX = (Random() & 0xF) << 0xC;
+        self->velocityX = velX;
+        if (self->animCurFrame == 0xD) {
+            self->velocityX = velX + 0x4000;
```
sozud, line 50: FIX?

src/st/chi/en_breakable_wall.c
```diff
@@ -0,0 +1,351 @@
...
+        }
+
+        velY = ((Random() & 7) << 0xB) - 0x4000;
```
sozud, line 53: FIX?

src/st/chi/en_breakable_wall.c
```diff
@@ -0,0 +1,351 @@
...
+        self->velocityY = velY;
+        if (self->animCurFrame < 0xB) {
+            self->velocityY = velY + 0xFFFF0000;
```
sozud, line 56: FIX?

src/st/chi/en_breakable_wall.c
```diff
@@ -0,0 +1,351 @@
...
+                        CreateEntityFromEntity(
+                            E_BREAKABLE_WALL_DEBRIS, self, entity);
+                        entity->params = ((Random() & 3) + 9) | 0x100;
+                    }
+                }
+                DestroyEntity(self);
+                return;
+            }
+            newVelY = self->velocityY;
+            if (newVelY <= 0x7FFF) {
```
sozud, line 97: FIX?

src/st/chi/en_demon_switch_wall.c
```diff
@@ -0,0 +1,303 @@
...
+        MoveEntity();
+#else
+        if ((self->ext.demonSwitchWall.unk80 & 7) == 0) {
+            g_api.PlaySfx(SFX_WALL_DEBRIS_B);
+        }
+        MoveEntity();
+#endif
+
+        if (self->velocityX < FIX(0.25)) {
+            self->velocityX += 0x200;
```
sozud, line 215: FIX

tools/splat_ext/cmp.py
```diff
@@ -1,35 +1,22 @@
...
 from splat.segtypes.n64.segment import N64Segment
+from splat.segtypes.segment import Segment
 from splat.util import options, log
 
 
 class PSXSegCmp(N64Segment):
     def __init__(self, rom_start, rom_end, type, name, vram_start, args, yaml):
         super().__init__(
             rom_start, rom_end, type, name, vram_start, args=args, yaml=yaml
-        ),
```
sozud, line 11: What's the changes here about?

sozud (approved):
This PR is gigantic so I just did a basic pass. This should get a second review probably before merging. Good work!

sozud:
I forgot to mention this got conflicted

src/st/chi/dt_entity.c
```diff
@@ -0,0 +1,211 @@
...
+    EntityEquipItemDrop,
+    EntityRelicOrb,
+    EntityHeartDrop,
+    EntityEnemyBlood,
+    EntityMessageBox,
+    EntityDummy,
+    EntityDummy,
+    EntityBackgroundBlock,
+    EntityUnkId12, // EntityCameraLock
+    EntityUnkId13, // EntityParticleTrail
```
JoshSchreuder, line 75: Just a suggestion, it might be good to add this as a comment to these two .h files and we can consider renaming them later on (I saw you added a comment about the params already)

src/st/chi/en_breakable.h
JoshSchreuder: Probably good to do for `temp` also

src/st/chi/en_gremlin.c
```diff
@@ -0,0 +1,382 @@
...
+                    entity->params = 1;
+                }
+                DestroyEntity(self);
+                return;
+            }
+        }
+        break;
+
+    case DEBUG:
+        FntPrint("charal %x\n", self->animCurFrame);
```
JoshSchreuder, line 210: I think this can import `src/st/pad2_anim_debug.h` for a little bit of dedupe

src/st/chi/en_salem_witch.c
```diff
@@ -0,0 +1,1020 @@
...
+        // fallthrough
+    case SHADOW_WAIT:
+        self->unk6C -= ShadowTickSpeed;
+        if (self->unk6C < ShadowDestroyThreshold) {
+            DestroyEntity(self);
+        }
+        return;
+
+    case DEBUG:
+        FntPrint("charal %x\n", self->animCurFrame);
```
JoshSchreuder, line 574: `src/st/pad2_anim_debug.h`

include/sfx.h
nyxojaele: Moved. I wasn't sure about stage-specific SFX since I didn't really see any in there

src/st/chi/en_gremlin.c
nyxojaele: Ah, good call. that didn't exist when I first did this.

tools/splat_ext/cmp.py
nyxojaele: So this was a sorta "two birds with one stone" thing:
1. Addressing the comment left in the original code: "does splat provide this without extracting the symbols for every file?"
2. Gives support to supplying custom names defined within the YAML
Not really sure how important this script even is anymore, with the asset manager coming online (and introducing cmpgfx)

PR #2130, opened 2025-01-26
Dedupe `lib` `prim_helpers`

JoshSchreuder:
Some rematching done for PSP but overall pretty minimal changes to dedupe


src/st/prim_helpers.h
```diff
@@ -79,100 +79,102 @@ void UnkPrimHelper(Primitive* prim) {
...
-            LOH(prim->u2) = tempUv + (new_var2 << 8);
-            ++prim->p1;
+            prim->p1++;
             return 0;
         }
     }
 
+    prim->p2--;
+#ifndef VERSION_PSP
     retVal |= 1;
```
JoshSchreuder, line 122: PSX is being finicky here, needs an extra ORI but I'm not sure why. If anyone else can work it out let me know

PSX: https://decomp.me/scratch/GDPlq
PSP: https://decomp.me/scratch/OEQHX

src/st/prim_helpers.h
```diff
@@ -183,8 +185,13 @@ Primitive* PrimToggleVisibility(Primitive* firstPrim, s32 count) {
         }
 
         prim = prim->next;
-        if (prim == NULL)
-            return 0;
+        if (prim == NULL) {
+            // Required for PSP match
+            if (true) {
```
JoshSchreuder, line 190: Another weird one needed for registers, there's a few things that work here, such as wrapping this as `do { return NULL } while(0);` or even `return NULL; do {} while(0);` after the return matches PSP. 

This seems the "least weird" to me, but I can change it if there's a fix or preference
https://decomp.me/scratch/8l5Dr

sozud (approved):

PR #2131, opened 2025-01-26
Decompile VSync + map libetc data

ser-pounce:
This should hopefully be everything in libetc, although as before it may not be a 1:1 match for the standalone libs themselves.


sozud (approved):
Matched this here https://github.com/sozud/psy-q-decomp/commit/52d83ddb82e9163f0b3c45356785d1f6fc55f56c
main differences are v_wait is static and Hcount is not static I think

PR #2132, opened 2025-01-26
Next PSP code sharing file

bismurphy:
Just one function in this file which is interesting. Lots of little updates especially with using FIX() where appropriate.


src/dra/6FDF8.c
```diff
@@ -84,20 +83,15 @@ bool func_8010FDF8(s32 branchFlags) {
...
-
-            if (g_Player.unk44 & 0x10) {
+            } else if (g_Player.unk44 & 0x10) {
                 func_8010E6AC(1);
-                goto label;
-            }
-
-            if (abs(PLAYER.velocityX) > 0x20000) {
+                PlaySfx(SFX_STOMP_SOFT_B, SFX_arg, 0);
+            } else if (abs(PLAYER.velocityX) > 0x20000) {
```
sozud, line 89: FIX?

sozud (approved):

src/dra/6FDF8.c
bismurphy: Good call

PR #2140, opened 2025-01-28
More `lib` dedupe

JoshSchreuder:
Did `e_bloody_zombie` - looking at the bestiary I think this is an unused entity in LIB, and as such appeared to be stripped from PSP version as I could not find it.

Did `BlitChar` and some related misc functions for both PSP and PSX. PSP uses a custom `BlitChar` which was present in WRP already so I pulled into a shared header for LIB and WRP to use (and likely other overlays going forwards)


JoshSchreuder:
Actually looking at it further I was only partially right about `e_bloody_zombie`, as `EntityBloodSplatter` and `EntityBloodDrips` do appear on PSP but the functions in between are stripped.

I will fix this up in a subsequent PR. I've mapped the symbols for now though.

sozud (approved):

PR #2142, opened 2025-01-28
Separate libpress from stream.c

dezgeg:
Latter half of stream.c is actually LIBPRESS.LIB/LIBPRESS.OBJ from PSY-Q. Separate it and use symbol names that are available in the COFF versions of the lib.

The location of libpress.c in the repo is not yet optimal, as it can't be moved to the psxsdk/ directory as that one is subdirectory of main/. Something will be need to be done about that later.


sozud (approved):
I matched this against the psy-q object, changes are optional https://github.com/sozud/psy-q-decomp/pull/10

The differences are various funcs/vars are static, some vars moved around. Last function is named timeout.

Data split is this I think if you want to import
```
      - [0xDBD4, data]
      - [0x162F4, data]
      - [0x16438, data]
      - [0x176C0, rodata]
 ```

File: src/st/sel/libpress.c [hunk 1]
```diff
@@ -0,0 +1,215 @@
...
+
+    return arg0;
+}
+
+s32* DecDCTPutEnv(DECDCTENV* arg0) {
+    s32 i;
+    s32 *src1, *dst1;
+    s32 *src2, *dst2;
+
+    dst1 = &mdec_iq[1];
```
sozud, line 64: Actually looking at the data I think these are wrong.

```
static s32 mdec_iq[] = {
    0x40000001, 0x13101002, ...

static s32 mdec_coef[] = {
    0x60000000, 0x5A825A82, ...
```

The first number of those two arrays don't fit in as coefficients? 33 is an odd count that doesn't fit the rest of the code either

dezgeg: The first word is a command word to the hardware whether to set iq/coefs

dezgeg:
> I matched this against the psy-q object, changes are optional [sozud/psy-q-decomp#10](https://github.com/sozud/psy-q-decomp/pull/10)
> 
> The differences are various funcs/vars are static, some vars moved around. Last function is named timeout.
> 
> Data split is this I think if you want to import
> 
> ```
>       - [0xDBD4, data]
>       - [0x162F4, data]
>       - [0x16438, data]
>       - [0x176C0, rodata]
> ```

Great! I'll give the unk's names as wel.

dezgeg:
Updated

PR #2156, opened 2025-01-31
Import/no1/assets

gamezter:
I tried importing sprite_banks, palette_def and graphics_banks but ran into too many issues. 

- The current tools to import sprite_banks expects 24 spriteBanks but no1 has 25
- The current tools to import palette_def expects that there is a single pal_def, but no1 has 4


sozud (approved):
Someone familiar with the asset system should follow up but seems good to me

PR #2158, opened 2025-02-01
Dedupe `lib` `e_misc` funcs

JoshSchreuder:
Some rematching done for PSP


src/st/make_entity_from_id.h
```diff
@@ -4,7 +4,12 @@
 void MakeEntityFromId(u16 entityId, Entity* src, Entity* dst) {
     DestroyEntity(dst);
     dst->entityId = entityId;
+#if defined(VERSION_PSP) && defined(STAGE_IS_LIB)
+    dst->pfnUpdate = OVL_EXPORT(EntityUpdates)[entityId - 1];
```
JoshSchreuder, line 8: Not sure why just yet, but PSP LIB uses `LIB_EntityUpdates` @ `0x09275F18` here, not the default `PfnEntityUpdates` @ `0x08C6BC30`

Other overlays on PSP are not doing this, like WRP / ST0, so I've special cased it for now, we can see if it happens on another stage overlay.

sozud: Are these not the same thing? https://github.com/Xeeynamo/sotn-decomp/blob/664965264e973eed27fe294519cbd79da255fd29/src/st/pfn_entity_update.h#L20

JoshSchreuder: Yeah that's for PSX though, PSP uses this
https://github.com/Xeeynamo/sotn-decomp/blob/664965264e973eed27fe294519cbd79da255fd29/src/st/pfn_entity_update.h#L8

If I use the PSX definition it breaks elsewhere like ST0/WRP `CreateEntityFromCurrentEntity` / `CreateEntityFromEntity`
I'm not really sure at this stage why this overlay uses a different entity list to the others here

JoshSchreuder: Actually WRP and ST0 don't use `MakeEntityFromId`, so it's possible this function just always uses `OVL_EXPORT(EntityUpdates) ` on PSP, but we won't know until we have more overlays on PSP that use this function I guess

JoshSchreuder: I've updated this now, after thinking on it a bit I don't think we need any conditional here at all and we can just use `OVL_EXPORT(EntityUpdates)` directly. I've updated this accordingly and it should continue to match on both PSP and PSX

sozud (approved):

PR #2161, opened 2025-02-02
Dedupe `lib` `corpseweed` / `thornweed` entities

JoshSchreuder:
Added codesharing for PSP

Needed some minor rewriting for PSP match


src/st/en_thornweed_corpseweed.h
```diff
@@ -563,6 +563,8 @@ void EntityCorpseweed(Entity* self) {
             self->ext.corpseweed.stalkDoneGrowing = false;
             self->hitboxState = 0;
             self->drawFlags = FLAG_DRAW_ROTZ;
+            // Need a no-op self access here for PSP match
+            self->step_s;
```
JoshSchreuder, line 567: Don't think we have a way of determining what exactly, but there is a line here that accesses `self`.
My hypothesis would be something like `self->step++` that was supposed to increment but didn't and is a bug, but no way to know for sure. 

There are many things that will match here on PSP, found this via the permuter as `if (step && step) {}` empty statement, but I think something like this is a little clearer.

https://decomp.me/scratch/ZHUBn

src/st/en_thornweed_corpseweed.h
```diff
@@ -677,15 +679,17 @@ void EntityCorpseweed(Entity* self) {
         t = self->ext.corpseweed.bobbingLeavesXT += BobbingSpeedX_Leaves;
         x = (rcos(t) * 2) >> 0xC;
         pos = (prim->x2 + prim->x3) / 2;
-        prim->x0 = pos - BobbingOriginOffsX_Leaves + x;
-        prim->x1 = pos + BobbingOriginOffsX_Leaves + x;
+        // 0xC represents BobbingOriginOffsX_Leaves
+        prim->x0 = pos - 0xC + x;
+        prim->x1 = pos + 0xC + x;
```
JoshSchreuder, line 684: A MWCC oddity perhaps - using a `const int` here that is function scoped leads to different registers than just inlining the constant value. 

I'm in two minds about how to solve this, one suggestion from Dezgag on Discord was to put these inside an enum, but I'm not sure it's much clearer and close to the original code, so I've chosen to just document this via comments.

https://decomp.me/scratch/ZHUBn

sozud: I think most likely these were just raw constants, or maybe #defines. So I think this is fine

PR #2163, opened 2025-02-02
Dedupe `lib` `e_skeleton`

JoshSchreuder:
According to bestiary this shouldn't be in LIB, but it has some custom code so perhaps was cut at some point in development.

Did a bunch of rematching here for PSP as well


src/st/e_skeleton.h
```diff
@@ -72,6 +86,18 @@ void EntitySkeleton(Entity* self) {
...
+        if (self->params & 0x10) {
+            self->flags |= FLAG_UNK_2000;
+            self->hitboxState = 0;
+            self->animCurFrame = 0;
+            self->step = SKELETON_LIB_UNK;
+            return;
+        }
+        break;
+    case SKELETON_LIB_UNK:
+        func_us_801BB8DC(sensors_ground);
```
JoshSchreuder, line 99: Not sure what this function is just yet but it's used in a few places in LIB, looks like a collision function.

sozud (approved):

PR #2164, opened 2025-02-02
Upgrade splat to 26.0

sozud:
Supposedly I need to upgrade splat to fix the hasm preamble stuff. 26.0 removes the include_macro_inc option and fails if it's present in the yaml. The project still seems to build without it

```
* BREAKING: Removed `include_macro_inc` option, as it was never required and often detrimental
```


hohle:
I was working on a similar change - https://github.com/Xeeynamo/sotn-decomp/pull/2165

PR #2165, opened 2025-02-03
Update Splat to 0.27.3

hohle:
A few things to keep up with breaking splat changes in more recent versions:

* extensions are now derived from `Segment` instead of `N64Segment`. the latter did not add any additional behavior to `Segment` so only the reference needed to be updated.
* removed the `include_macro_inc` option from all splat config.
* `macro.inc` was being included in every extracted ASM file, so include guards have been added to avoid duplicate symbols.

0.27.2 is the latest version which continues to work without additional changes. 0.26.1-0.27.2 include a bug triggered by HD DRA where a segment becomes its own sibling. 0.28.0 requires additional changes that may be trivial, but increase the scope of what needs to be changed. 0.27.3 seemed to be a nice version for a reasonable incremental change.


sozud:
I think let's see if mine passes the full CI then rebase this version

sozud:
Upgrading broke the duplicates report and progress report, those need to be verified before merging this. No need to rebase on my PR, I'm not going to resubmit it

config/splat.us.stare.yaml
```diff
@@ -0,0 +1,44 @@
+options:
+  platform: psx
```
sozud, line 2: I think the stare stuff is included by accident?

config/symbols.us.stare.txt
```diff
@@ -0,0 +1,117 @@
+func_801A7D64 = 0x801B5910;
+EntityUnkId12 = 0x801B59D8;
```
sozud, line 2: stare

tools/dups/src/splat.rs
```diff
@@ -0,0 +1,206 @@
+extern crate yaml_rust;
```
sozud, line 1: I don't see this file used in main.rs? Also there's no changes to skip nonmatchings in main.rs?

config/splat.us.stare.yaml
hohle: Bad `git add`s.

hohle:
I tried to make the additional changes for `dups` and `progress.py` minimal. The main issue with them was `splat` no longer eagerly creating `matchings` and `nonmatchings` directories. Both `dups` and `progress.py` sanity checked those directories existing, but for various reasons they may not exist (e.g. when everything is decompiled or nothing is decompiled).

The change for `dups` was just to treat the check as a warning and non-fatal.

The change for `progress.py` was a little more involved because it uses a heuristic of the ASM directory layout to determine where assembly files are. Since those directories could no longer be used as guidestones, I chose to base the ASM directory calculation on the splat config first, and fall back to trying to calculate the directory, which was really only relevant for weapons.

I'm open to other ways of going about it, but was trying to keep the scope of the changes small.

sozud:
I'll merge and see what happens in CI

PR #2172, opened 2025-02-04
Decompile `e_stage_name` for PSP `wrp`/`lib`

JoshSchreuder:
A lot of conditionals required here, so I can pull this into separate PSP specific file if preferred.
https://decomp.me/scratch/DqmOT

PSP appears to use the "jp" version of the stage name code which is different to PSX's.


src/st/entity_stage_name_popup_jp.h
```diff
@@ -1,24 +1,40 @@
...
             return;
         }
-
+#ifdef STAGE_IS_LIB
+        func_psp_0923C0C0();
+#endif
         InitializeEntity(g_EInitInteractable);
         self->ext.stpopupj.unk8C = 0;
+#ifdef VERSION_PSP
+        primIndex = g_api.AllocPrimitives(PRIM_GT4, 183);
```
sozud, line 34: Could do 
```
#if VERSION_PSP
#define NUM_PRIMS 183
else
#define NUM_PRIMS 181
#endif
```
but either is fine

sozud (approved):

src/st/entity_stage_name_popup_jp.h
JoshSchreuder: Good idea, updated

PR #2177, opened 2025-02-04
Decompile Library Background Parallax Entities

hohle:
Decompiles background bookcase, chairs & desk, & shadows behind the ectoplasms.

Split library background into a separate segment.


sozud (approved):
Looks good but needs to be rebased

src/st/lib/e_library_bg.c
```diff
@@ -0,0 +1,277 @@
...
+
+    switch (self->step) {
+    case 0:
+        InitializeEntity(g_EInitInteractable);
+        primIndex = g_api.func_800EDB58(PRIM_TILE_ALT, 10);
+        if (primIndex != -1) {
+            self->flags |= FLAG_HAS_PRIMS;
+            self->primIndex = primIndex;
+            prim = &g_PrimBuf[primIndex];
+            self->ext.prim = prim;
```
JoshSchreuder, line 202: Would this make sense as a `Primitive*` at 0x7C on `self->ext.libraryShadow`? Since that is stubbed anyway.
I'm not 100% clear on when we use `self->ext.prim` vs. not.

hohle: I was using it out of convenience in the previous two functions to avoid creating a new extension struct, but should have used the LibraryShadow struct in that last function. Thanks for catching.

JoshSchreuder (approved):

PR #2180, opened 2025-02-05
Next PSP DRA file

bismurphy:
Small one. Finishes out the subweapons, I believe.


sozud (approved):

src/dra/8A0A4.c
```diff
@@ -23,19 +24,20 @@ void EntityStopWatchExpandingCircle(Entity* self) {
         self->flags = FLAG_POS_CAMERA_LOCKED | FLAG_KEEP_ALIVE_OFFCAMERA |
                       FLAG_HAS_PRIMS | FLAG_UNK_20000;
         prim = &g_PrimBuf[self->primIndex];
-        for (i = 0; i < 16; i++, prim = prim->next) {
+        for (i = 0; i < 16; prim = prim->next, i++) {
```
JoshSchreuder, line 27: Very minor nit and probably just personal taste, but I think in these cases it is more natural to have `prim = prim->next` at the end of the loop rather than here? Think it should match either way

bismurphy: I think that `prim=prim->next` is just as much an incrementer as `i++` is, so it makes sense to put both in the loop. The other thing is that we just as often see loops with `for (i = 0; prim != (0); i++, prim = prim->next) {` where the `i++` comes first, and that should definitely end up in the `for` line rather than at the end.

PR #2186, opened 2025-02-06
More fun with flags

ProjectOblivion:
Added inverted castle flags, cleaned up some names for consistency, and added a couple missing normal castle flags.

Converted stage name entities and stage headers to use g_CastleFlags instead of CASTLE_FLAG_BANK.

This will need to be rebased when pending approved changes have been merged.


src/dra/play.c
```diff
@@ -24,7 +24,7 @@ u8 g_StageSelectOrder[] = {
     STAGE_RBO8,       STAGE_TOP_ALT,  STAGE_ENDING,
 };
 
-// The effect of this array is unknown
+// This is used for NO1 Weather
 // 0: 25% of chance
 // 1: 62.5% of chance
 // 2: 12.5% of chance
```
JoshSchreuder, line 30: Might be good to rename this variable to `g_RandomizeOuterWallWeather`. Think it's only used in this file so impact should be minimal

src/st/e_stage_name.h
```diff
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "stage.h"
 
-#ifndef CASTLE_FLAG_BANK
-#define CASTLE_FLAG_BANK 0
+#ifndef STAGE_FLAG
+#define STAGE_FLAG NULL_STAGE_FLAG
```
JoshSchreuder, line 5: Doesn't matter too much but maybe it makes more sense to make the default `OVL_EXPORT(STAGE_FLAG)` and override it to `NULL_STAGE_FLAG` for those specific overlays?

src/dra/play.c
JoshSchreuder: Or I think your comment in `castle_flags.h` calls this `g_NO1WeatherOptions`.

ProjectOblivion: Makes sense to me.  I've been naming things by overlay for consistency, so I'll do that here as well.

src/st/e_stage_name.h
ProjectOblivion: It's tough to say without having the inverted castle headers figured out what the best option would be, the inverted castle doesn't have flags associated with going to an OVL for the first time like the normal castle.  I think for now leaving it as is will be the least effort option.

I have no real attachment either way, I can do whatever is needed.

JoshSchreuder (approved):

PR #2189, opened 2025-02-06
Refactor PSP `wrp` and import some additional data

JoshSchreuder:
Was attempting to import all the remaining data here but the asset support remains a bit of a mystery to me

I think I've been able to identify most of the data offsets for the sprites, graphics banks, etc. and documented them on the splat config for now until I can work out what needs to change in the assets tool.


config/splat.pspeu.stwrp.yaml
```diff
@@ -62,73 +62,33 @@ segments:
...
-      - [0xF970, data] # D_psp_09246FF0
+      - [0xF4D8, data] # WRP_spriteBanks
+      - [0xF5F8, data] # WRP_rooms_layers
+      - [0xF700, data] # WRP_cluts
+      - [0xF708, .data, wrp_psp/graphics_banks]
+      - [0xF790, .data, wrp_psp/e_init]
+      - [0xF8A0, data] # WRP_rooms
+      - [0xF920, .data, wrp/e_breakable]
+      - [0xF970, data] # [0xF970, cmpgfx, stage_title_jp, 128, 128, 4]
+      - [0xFCB8, data] # [0xFCB8, cmpgfx, stage_placeholder, 128, 128, 4]
```
JoshSchreuder, line 80: I was able to export these successfully to png, but re-linking them back into the build output caused a mismatch so I've commented this out for now. I might put up a WIP PR and see if I can get a hand on the linking side to understand where I'm going wrong.

src/st/wrp_psp/graphics_banks.c
```diff
@@ -0,0 +1,39 @@
...
+extern u8 stage_title_jp[];
+extern u8 stage_placeholder[];
+static u_long* wrp_gfxbank_1[] = {
+    (u_long*)GFX_BANK_COMPRESSED,
+    GFX_ENTRY(0x100, 0x040, 128, 128, stage_title_jp),
+    GFX_ENTRY(0x100, 0x060, 128, 128, stage_placeholder),
+    GFX_TERMINATE(),
+};
+
+void* OVL_EXPORT(gfxBanks)[] = {
```
JoshSchreuder, line 18: This is one of the generated outputs of the assets tool and is not usually committed to the repo, but this data was already imported so I have kept it that way, just standardised with the asset tool output from WRP PSX.

sozud (approved):
I don't really understand the asset tool so can't comment on that part but seems like an improvement to me

src/st/wrp_psp/graphics_banks.c
sozud: I think this import predates the tool? Maybe test it with the tool and see if it matches? I think the idea is to migrate the manually done stuff to the tool

JoshSchreuder: It does yeah. I basically took `graphics_banks` that is auto imported from WRP PSX and applied it here for now for consistency. I'm playing around with the asset pipeline now to see if I can get it working for PSP as well, in which case this file will go away and be auto-generated.

sozud: If you get it working we can do a followup I think

PR #2193, opened 2025-02-09
Saturn: AddToInventory and neighbours

Xeeynamo:
On a quest to understand what is the real function order of DRA, I started exploring once again the Saturn build. With the incredible work from @bismurphy we can copy the current updated functions from the PSP/PSX code back to the Saturn port with little effort. The code is mostly identical.

NOTE:

For every code iteration I had to execute `cp src/saturn/* build/saturn/; make build_saturn && VERSION=saturn make check && git add .` as it was the only way to have `build/saturn/*` up-to-date before building. A new make rule would have done the trick, but I am waiting to review #2188 before doing any change.

`cp src/saturn/* build/saturn/; VERSION=saturn python3 ./tools/asm-differ/diff.py -mow3 --overlay game -f build/saturn/game.o _func_800FD6C4` was used to run differ locally (thanks @ProjectOblivion for the suggestion)


sozud (approved):
Looks good, cool to see some Saturn progress

PR #2200, opened 2025-02-11
Decompile `lib` func_us_801CC054

gamezter:
PSX: https://decomp.me/scratch/1rWl0
PSP: https://decomp.me/scratch/3MO3f


JoshSchreuder:
If you move `func_us_801CC984` back up into e_dhuron including the rodata it should match on PSX
image

Got this from splat which is warning about rodata misalignment
```
The rodata segment 'unk_4C984' has jumptables that are not aligned properly file-wise, indicating one or more likely file split.
File split suggestions for this segment will follow in config yaml format:
      - [0x2DBC4, .rodata]
```

PSP is matching right with the current splits, so this may resolve itself once `func_us_801CC984` is matched

src/st/lib/e_dhuron.c
```diff
@@ -1,11 +1,210 @@
...
+    self->hitboxHeight = *hitboxPtr++;
+}
 
 // Dhuron death parts
 INCLUDE_ASM("st/lib/nonmatchings/e_dhuron", func_us_801CC6B0);
 
 // Dhuron death
 INCLUDE_ASM("st/lib/nonmatchings/e_dhuron", func_us_801CC7BC);
+
+INCLUDE_ASM("st/lib/nonmatchings/e_dhuron", func_us_801CC984);
```
JoshSchreuder, line 210: Would be good to move this up for the PSP too if possible, otherwise I can do it a bit later on if you don't have it setup

sozud (approved):

JoshSchreuder (approved):

PR #2202, opened 2025-02-12
Just renaming

snowboundmage2:
Just renaming functions based on what I can see they do/what calls them. 
Trying to Follow Naming/Style Docs
Some late regrets: AllocPrimBuffers should be AllocatePrimitivesRecursively(WIP name)


JoshSchreuder:
Run `make format` to fix the linting

sozud:
This currently touches 220 files, can we break this up into more than 1 PR? Good candidates to split up would be the g_api changes which touch a million files

snowboundmage2:
> This currently touches 220 files, can we break this up into more than 1 PR? Good candidates to split up would be the g_api changes which touch a million files

Not sure how to do that, this was the side effect of replacing all instances of a functions name, after a few days reading/learning this project. I just went through these files:
src/dra/4A538.c
src/dra/5D5BC.c
src/dra/5F60C.c
src/dra/6BF64.c
src/dra/6D584.c
src/dra/6E42C.c
and changed the names of all functions I could understand, and did a replace all to make sure every use of that function was going to the right place.

ProjectOblivion:
It looks like limiting it by overlay would help out a lot with the number of files touched, probably being able to combine some of the less touch overlays.

In general, to reduce the number of files you touch you want to reduce number of operations (in this case number of different renames) or reduce scope (in this case the overlays).  Sometimes a combination of both, but also sometimes (rarely) it's unavoidable.

sozud:
One example of reducing scope would be changing AllocPrimitives back. You can do this by adding a second commit and github will automatically update the diff. Then that change could be discussed in #naming in discord for example. Changing something that was already named has a higher bar than going from func_XXXXXXX to some sort of name. 

This PR is going to get conflicted multiple times probably since it touches a lot of files. We already have a conflicting PR in the queue here https://github.com/Xeeynamo/sotn-decomp/pull/2203/files#diff-5f608dee41a44a8814d9520f5e3f5636846de0c9ea7182be5b60fa6523824ab1R321

PR #2221, opened 2025-02-15
Updated "Build saturn binaries" to use VERSION env

ProjectOblivion:
#2188 will fail until this is merged.


sozud:
Fails on master https://github.com/Xeeynamo/sotn-decomp/actions/runs/13349139223/job/37283350332

PR #2222, opened 2025-02-15
Revert "Updated "Build saturn binaries" to use VERSION env"

sozud:
Reverts Xeeynamo/sotn-decomp#2221


sozud:
I think this is still using the yaml from master so merging should fix

PR #2223, opened 2025-02-15
Update build saturn binaries

ProjectOblivion:
I forgot that the st%.elf pattern breaks saturn building with VERSION.
That has been added in and should allow the CI step to be updated.


ProjectOblivion:
I didn't push the new commit to github, I'll be doing that momentarily.

ProjectOblivion:
The fix has been properly committed now, this should be ready for review.

PR #2224, opened 2025-02-16
Dedupe `lib` `cutscene` for PSP

JoshSchreuder:
`LIB_EntityCutscene` ended up being a beast on PSP which is going to take a while to decomp, so I'm going to put up what I've got so far and get this merged


src/st/cutscene_unk3.h
```diff
@@ -4,17 +4,17 @@
 #endif
 
 #if defined(VERSION_PSP)
-extern void* D_pspeu_0927B6A8;
+extern s32 cutsceneUnk3Unknown[0x380]; // bss
 #endif
 
-static void CutsceneUnk3(s16 arg0) {
+static void CutsceneUnk3(s16 nextCharY) {
```
JoshSchreuder, line 10: Renamed based on usage, this is called from `CutsceneUnk4` with `g_Dialogue.nextCharY` so seemed sensible to renam

PR #2225, opened 2025-02-16
Decompiled func_us_801BAF60 and func_psp_0926DAA8, merged, and imported data

ProjectOblivion:
PSX: https://www.decomp.me/scratch/wzVJc
PSP: https://www.decomp.me/scratch/ZOm1H

Also needed to adjust pad2_anim_debug.h to accommodate PSP.


src/st/lib/e_candle_table.c
```diff
@@ -0,0 +1,90 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "lib.h"
+
+// EInit D_us_801808FC = {ANIMSET_OVL(0x08), 0x00, 0x50, 0x256, 0x005};
```
sozud, line 4: What's this for?

src/st/lib/e_candle_table.c
```diff
@@ -0,0 +1,90 @@
...
+    0x0103, 0x0203, 0x0303, 0x0403, 0x0503, 0x0603, 0x0703, 0x0000};
+
+#ifdef VERSION_PSP
+static u16 D_us_80181A78[] = {0x0903, 0x0A03, 0x0B03, 0x0C03, 0x0D03, 0x0E03,
+                              0xF03,  0x1003, 0x0000, 0x0000, 0x0000, 0x0000};
+extern u32 D_psp_092A5590;
+extern s32 D_8B42050;
+// I expect this symbol will change depending on the function that is including
+// pad2_anim_debug.h
+#define BUTTON_SYMBOL D_8B42050
```
sozud, line 15: Not sure how this is working but we have other instances of this symbol https://github.com/Xeeynamo/sotn-decomp/blob/2a9c93e6dd5bd1fce4a576bbf0f2d9210684aa0f/src/st/lib/unk_3C57C.c#L579

Also overlays start at 0x09237700 and DRA starts at 0x090DCA00 so I think this is actually part of the SDK or PSP operating system or something

src/st/lib/e_candle_table.c
ProjectOblivion: It's the import for the EInit used.  It's correct, but doesn't work for PSP because of data alignment issues.  Since I had already pulled it in, I just commented it out.  I probably should have put that in a comment with it.

src/st/lib/e_candle_table.c
ProjectOblivion: This is the first PSP function to import pad2_anim_debug.h, which is probably why.  LIB and DRA are the only ones that use this particular symbol, which is why I suspected other overlays would use different symbols.

ProjectOblivion: I can adjust it however is needed.

src/st/lib/e_candle_table.c
```diff
@@ -0,0 +1,93 @@
...
+
+// This is the definition for the EInit for this function
+// I've verified it for PSX, but PSP needs e_init.c
+// EInit D_us_801808FC = {ANIMSET_OVL(0x08), 0x00, 0x50, 0x256, 0x005};
+
+static u16 D_us_80181A68[] = {
+    0x0103, 0x0203, 0x0303, 0x0403, 0x0503, 0x0603, 0x0703, 0x0000};
+
+#ifdef VERSION_PSP
+static u16 D_us_80181A78[] = {0x0903, 0x0A03, 0x0B03, 0x0C03, 0x0D03, 0x0E03,
```
gamezter, line 12: I noticed `D_us_80181A78` and `D_us_80181A68` are passed to `AnimateEntity`, these are actually `u8[]`s

PR #2228, opened 2025-02-17
Next code-share file between PSP and US

bismurphy:
Lots of knock-on effects here, but mostly just improving code and ending up closer to what we would guess the real code looked like.

We have a merge conflict; will resolve.


sozud (approved):

sozud:
Conflicted

bismurphy:
Thanks for the ping. I'm removing all the `g_api_` forward declarations since they shouldn't be used anyway.

sozud:
I thought those were there so the m2c uses them?

bismurphy:
Understood - they are now back.

PR #2234, opened 2025-02-19
Removed debian requirement that doesn't play well with 24.04 and added tools/dosemu-deb to git submodules

ProjectOblivion:
Also found out 24.04 doesn't like to install .deb files through apt so I removed them.


JoshSchreuder:
`gcc-mips-linux-gnu` isn't needed for decomp-permuter though? I don't have this installed in mine and it runs fine using `mipsel-linux-gnu-as` in the permuter options

ProjectOblivion:
I couldn't get objcopy to work without it.  If there was another method, I couldn't find it.

JoshSchreuder:
Right, I have that installed from `binutils-mips-linux-gnu`

Not sure where that came from, maybe I installed it manually at some stage.

Though I think it would make more sense to update permuter to also use `mipsel-linux-gnu-objdump` which we should be installing already

.gitmodules
```diff
@@ -13,3 +13,9 @@
 [submodule "tools/mwccgap"]
 	path = tools/mwccgap
 	url = https://github.com/mkst/mwccgap.git
+[submodule "tools/decomp-permuter"]
+	path = tools/decomp-permuter
+	url = https://github.com/simonlindholm/decomp-permuter.git
+[submodule "tools/dosemu-deb"]
```
sozud, line 19: The original point of this was that dosemu put up breaking changes and we needed to fix the CI. I think you ought to be able to run the project just from vanilla debian dosemu?

ProjectOblivion: Yes, you can use dosemu2 from apt, but it doesn't have the dependencies in it.

ProjectOblivion: I take that back, 22.04 did not have them, but 24.04 does.  I'll check if using the apt packages cause issues and update this PR accordingly.

ProjectOblivion:
No, there is no benefit that I'm aware of.
I was simply under the impression that it was expected to be the same as all of the other submodules.  All instructions both given and written seemed to assume that it was already there.

bismurphy:
Personally I like the idea of having a standardized permuter configuration, in a known directory location, so that we can assist newcomers in how to use it. Knowing where it's installed (within the Tools directory) would help with that.

ProjectOblivion:
I would like to have that as well, but some changes need to the code itself before it is accessible to newcomers.  I don't know if submitting PRs to the main repo or creating our own fork where we can make those changes is a better option.

hohle:
Agreed on permuter.

ProjectOblivion:
Since I submitted #2270 , which is a more sensible place with a much better reason for adding permuter as a submodule, I removed it from this PR.  This PR is now only the removal of the debian requirements that 24.04 doesn't like and adding dosemu as a submodule.  Having dosemu as a submodule will make it much simpler to get up and going with the saturn version.

ProjectOblivion:
That makes sense to me, I don't really see any sense in making it a unilateral dependency if not everyone actually needs it.
This is minimal enough that it could be handled on a more case by case basis in the saturn makefile, similiar to how the tar.gz files are handled.

ProjectOblivion:
I'll close this PR and deal with removing the packages from the debian requirements on one of the future make passes.

PR #2235, opened 2025-02-19
Clean up `lib` imported data

gamezter:
Data seems to be aligned to a 4 byte boundary, so a lot of imported data has trailing zeros that are actually padding between data. Here I mostly remove trailing zeros from `u8[]`s passed to `AnimateEntity` since that function ignores everything after a `-1, 0` or a `0, 0`. 

I also adjusted some imported arrays so their size matches what the code expects their size to be. I had to add a  `STATIC_PAD_DATA(4);` but perhaps that could be a clue to adjust the file boundary.


src/st/lib/unk_49870.c
```diff
@@ -5,19 +5,18 @@ static s16 D_us_8018280C[] = {0, 23, 0, 4, 8, -4, -16, 0};
...
-    24, 7, 24, 8, 24, 9, 24, 8, 0, 0, 0, 0, 18, 8, 18, 9, -1, 0, 0, 0};
-static u8 D_us_80182868[] = {8, 7, 8, 10, 4, 7, 32, 10, 0, 0, 0, 0};
-static u8 D_us_80182874[] = {
-    8, 1, 8, 11, 8, 12, 8, 13, 4, 14, 2, 15, -1, 0, 0, 0};
-static u8 D_us_80182884[] = {2, 16, 2, 17, 0, 0, 0, 0};
-static u8 D_us_8018288C[] = {2, 18, 2, 19, 2, 18, 2, 19, 0, 0, 0, 0};
+static u8 D_us_8018283C[] = {20, 1, 20, 2, 20, 3, 20, 2, 0, 0};
+static u8 D_us_80182848[] = {24, 4, 24, 5, 24, 6, 24, 5, 0, 0};
+static u8 D_us_80182854[] = {24, 7, 24, 8, 24, 9, 24, 8, 0, 0};
+static u8 D_us_80182860[] = {18, 8, 18, 9, -1, 0};
```
gamezter, line 11: notice I split `D_us_80182854` in two here since `0,0` signals the end of this data

src/st/lib/e_candle_table.c
```diff
@@ -5,20 +5,19 @@
...
-extern s32 D_8B42050;
+static s32 D_us_80181A8C = 0;
 // I expect this symbol will change depending on the function that is including
 // pad2_anim_debug.h
-#define BUTTON_SYMBOL D_8B42050
+#define BUTTON_SYMBOL D_psp_08B42050
 #else
-static u16 D_us_80181A78[] = {0x0903, 0x0A03, 0x0B03, 0x0C03, 0x0D03, 0x0E03,
-                              0xF03,  0x1003, 0x0000, 0x0000, 0x0003, 0x0000};
+static s32 D_us_80181A8C = 3;
```
gamezter, line 20: These were actually `u8[]`s since they are passed to `AnimateEntity` with an unused s32 at the end

src/st/lib/unk_49870.c
JoshSchreuder: I found a few of these when importing PSP data, since the unused symbols get stripped there, but it causes a mismatch if they're part of a data address that is actually used. Nice work

sozud (approved):

PR #2237, opened 2025-02-20
Refactor `EntityLockCamera` code

JoshSchreuder:
There were multiple implementations of this function with various changes through overlays:

* `EntityUnkId12` was used by rbo3, cen, chi, dre, mad, no1, no3, no4, np3, rwrp, wrp
* no0, nz0 had customised implementation called `func_801B0AA4` 
* st0, lib had customised implemention called `EntityLockCamera`

All of this has been renamed to `EntityLockCamera`.
I've updated the shared header `entity_lock_camera.h` to be the previous `EntityUnkId12` implementation since it is the most common, and moved the old customised implementation into `st0` which was the only place it was imported.

I have also removed two extension structs `ET_Entity12` and `ET_801B0AA4` which are no longer required, since all of this code can use `ET_LockCamera`


config/splat.us.stno0.yaml
```diff
@@ -41,7 +41,7 @@ segments:
       - [0x778, data, e_laydef]
       - [0x920, .data, e_init]
       - [0xC8C, .data, e_stage_decoration]
-      - [0xE94, data, 419EC]
+      - [0xE94, .data, e_lock_camera]
```
JoshSchreuder, line 44: Imported this data while I was here

config/splat.us.stnz0.yaml
```diff
@@ -151,7 +151,8 @@ segments:
       - [0x30900, .rodata, e_stage_name]
       - [0x30934, .rodata, prim_helpers]
       - [0x3093C, .rodata, e_life_up]
-      - [0x30958, c]
+      - [0x30958, c, st_debug]
+      - [0x30A20, c, e_lock_camera]
```
JoshSchreuder, line 155: Split these while I was here, and refactored `EntityBackgroundBlock` to be named the same as its common implementation elsewhere

src/st/nz0/e_lock_camera.c
```diff
@@ -0,0 +1,142 @@
...
+#include "../player_is_within_hitbox.h"
+
+extern u16 g_EInitLockCamera[];
+static u8 lockCamera_hitbox[] = {32, 32, 32, 32};
+static u8 lockCamera_data[] = {0, 0, 0, 0};
+static u16 lockCamera_mapScroll[] = {
+    0x0000, 0x00FB, 0x0100, 0x01FB, 0x0000, 0x00FB, 0x0100, 0x01FB,
+    0x0000, 0x00FB, 0x0600, 0x01FB, 0x0000, 0x0000, 0x0600, 0x01FB,
+};
+void EntityLockCamera(Entity* self) {
```
sozud, line 13: I'm not sure this is the same thing as the other versions, have we checked what this does ingame?

sozud: Actually I think this is fine for now, if it's not the same we can adjust the naming later

PR #2243, opened 2025-02-20
Decompiled LIB func_us_801AFE0C/func_psp_0925D8D8

ProjectOblivion:
func_us_801AFE0C/func_psp_0925D8D8
PSX: https://decomp.me/scratch/L5A5D
PSP: https://decomp.me/scratch/ydeqP

Code is shared but not merged.

Added PAD_NONE (value of 0) in game.h macros to make this function a little more self documenting.
Also fixed a tiny error with where I put castle flag offset 0x29 previously.


config/symbols.pspeu.stlib.txt
```diff
@@ -1,3 +1,6 @@
+D_80097928 = 0x091CE5B0;
+D_80073510 = 0x091E17B8;
+D_8003BE5E = 0x092347A2;
```
JoshSchreuder, line 3: These are all outside the overlay, should be in `symbols.pspeu.txt` I think?

I think `0x091E17B8` would be `g_Entities[1].ext.something` but unclear what

src/st/lib/unk_2FA80.c
```diff
@@ -287,7 +287,228 @@ INCLUDE_RODATA("st/lib/nonmatchings/unk_2FA80", D_us_801ACD3C);
...
+// I expect these two sounds to be the same, but 0x202 has not yet been defined.
+// This leads me to think that the macro that has been defined for sfx 0x302
+// is only accurate for PSX and does not align with the sfx for PSP here.
+#ifdef VERSION_PSP
+            g_api.PlaySfx(0x302);
+#else
+            g_api.PlaySfx(0x202);
+#endif
+        }
+        if (D_8003BE5E) {
```
JoshSchreuder, line 315: I think this is `g_CastleFlags[0x72]`?

ProjectOblivion: Sure was, thanks for pointing it out.

src/st/lib/unk_2FA80.c
```diff
@@ -287,7 +287,227 @@ INCLUDE_RODATA("st/lib/nonmatchings/unk_2FA80", D_us_801ACD3C);
...
+    Entity* entity = g_Entities;
+
+    switch (self->step) {
+    case 0:
+#ifdef VERSION_PSP
+        func_psp_0925D4D0();
+#endif
+        InitializeEntity(&D_us_80180824);
+        if (entity->posX.i.hi < 0x100) {
+// I expect these two sounds to be the same, but 0x202 has not yet been defined.
```
sozud, line 305: This is playing the librarian seq music most likely. On PSP they probably recorded it as baked audio rather than sequenced music hence the # difference 
https://github.com/Xeeynamo/sotn-decomp/blob/1bf088d7d4007c9559c9d0851434158166689eb4/src/dra/94F50.c#L20

src/st/lib/unk_2FA80.c
```diff
@@ -287,7 +287,227 @@ INCLUDE_RODATA("st/lib/nonmatchings/unk_2FA80", D_us_801ACD3C);
...
+        SetStep(2);
+        break;
+    case 2:
+        if (entity->posX.i.hi > 0xE8) {
+            if (g_Player.status & PLAYER_STATUS_TRANSFORM) {
+                g_Player.padSim = PAD_NONE;
+                if (g_Timer & 1) {
+                    if (g_Player.status & PLAYER_STATUS_BAT_FORM) {
+#ifdef VERSION_PSP
+                        g_Player.padSim = PAD_R1 | PLAYER_STATUS_UNK20000;
```
sozud, line 354: player status seems wrong here? Should be a button?

ProjectOblivion: My line of thought was that this is related to the form status, so it made sense.  I have no objection to altering it to whatever, though, that was just my best guess.
The literal code is g_Player.padSim = 20200;  which lines up perfectly with PAD_R1 | 0x20000; which in turn lines up with the rest of the function structure.

sozud: There's some sort of flag that gets checked here https://github.com/Xeeynamo/sotn-decomp/blob/fbfef65aed0da1f02e7cc3ba3f070839babc88b2/src/dra/75F54.c#L236

Seems like it forces the player to stand? https://github.com/Xeeynamo/sotn-decomp/blob/fbfef65aed0da1f02e7cc3ba3f070839babc88b2/src/dra/6E42C.c#L51

I think `PAD_SIM_FORCE_STAND` or something would make sense

PR #2246, opened 2025-02-21
Rename `lib` symbols

gamezter:


src/st/lib/e_init.c
```diff
@@ -170,10 +170,10 @@ EInit g_EInitLockCamera = {ANIMSET_DRA(0x00), 0x00, 0x00, 0x0000, 0x001};
 EInit g_EInitCommon = {ANIMSET_DRA(0x00), 0x00, 0x00, 0x0000, 0x003};
 EInit g_EInitDamageNum = {ANIMSET_DRA(0x00), 0x00, 0x00, 0x0000, 0x003};
 EInit g_EInitBloodyZombie = {ANIMSET_OVL(0x05), 0x01, 0x48, 0x0207, 0x00D};
-EInit D_us_80180878 = {ANIMSET_OVL(0x06), 0x01, 0x49, 0x020A, 0x00F};
-EInit D_us_80180884 = {ANIMSET_OVL(0x06), 0x00, 0x49, 0x020A, 0x00E};
+EInit g_EInitFlyingZombie2 = {ANIMSET_OVL(0x06), 0x01, 0x49, 0x020A, 0x00F};
+EInit g_EInitFlyingZombie1 = {ANIMSET_OVL(0x06), 0x00, 0x49, 0x020A, 0x00E};
```
gamezter, line 174: Any ideas for a better name for these 2 ?

include/entity.h
```diff
@@ -3112,12 +3112,12 @@ typedef union { // offset=0x7C
     ET_Marionette marionette;
     ET_801B6F30 et_801B6F30;
     ET_801B15C0 et_801B15C0;
-    ET_801D064C et_801D064C;
-    ET_801D2274 et_801D2274;
+    ET_Mudman mudman;
+    ET_SpellbookMagicTome spellbookMagicTome;
```
gamezter, line 3116: Is this too long ? Would simply `book` be better ?

src/st/lib/e_init.c
JoshSchreuder: Think the enemy splits in two and I would guess it's top and bottom half but I haven't tested in game to know for sure

hohle (approved):
I think the names are great, even if they may be tweaked in the future. Just having readable names makes things easier to follow.

sozud (approved):
Conflicted, otherwise looks good

PR #2247, opened 2025-02-21
Add size hints for PSP globals and fix some references

JoshSchreuder:
Should help with decomp because when doing this, splat can provide offsets from the base address in the ASM, for example
```
    /* 23C8C 0910060C 1D09023C */  lui        $v0, %hi(g_Status + 0x240)
    /* 23C90 09100610 40F6428C */  lw         $v0, %lo(g_Status + 0x240)($v0)
```

```
    /* 1A29C 090F6C1C 2309023C */  lui        $v0, %hi(g_CastleFlags + 0x96)
    /* 1A2A0 090F6C20 C64743A0 */  sb         $v1, %lo(g_CastleFlags + 0x96)($v0)
```

This should stop accidentally mapping these addresses as extern symbols when decompiling [like this](https://github.com/Xeeynamo/sotn-decomp/pull/2243#discussion_r1964505269).

While I was here I fixed a few of the existing cases where that had happened, using the splat hints as a guide.

I think the same should be possible on the PSX side but I need to verify a few things.


config/symbols.pspeu.txt
```diff
@@ -70,19 +70,11 @@ g_CdStep = 0x91CE5E0;
 g_unkGraphicsStruct = 0x091CE5F8;
 D_80097488 = 0x091CE688;
 g_randomNext = 0x091CE6B0;
-g_Settings = 0x091CEEBC;
+g_Settings = 0x091CEEBC; // size:0x110
 g_PlayableCharacter = 0x091CF020;
 g_Servant = 0x091CF028;
-g_Player = 0x091CF030;
-g_Player_pl_vram_flag = 0x091CF380;
```
JoshSchreuder, line 77: One of the downsides to this approach is that you will no longer get direct symbol hints as to the property by mapping them in this manner.

However most structs are not mapped out in much detail this way, and personally I think the benefits of having the whole range of the struct covered rather than just the members we map manually outweighs this.

src/st/e_particles_maria1.h
```diff
@@ -11,8 +11,8 @@ void func_psp_0923AD68(Entity* self) {
     u16 direction;
     Entity* pl;
 
-    if (D_psp_091CF3DC && D_psp_091CF3A0) {
-        pl = D_psp_091CF3A0;
+    if (g_Player.unk5C && g_Player.unk20) {
+        pl = (Entity*)g_Player.unk20;
```
JoshSchreuder, line 15: Possible this is just an `Entity*` on PSP, but elsewhere on PSX/HD it is used as an s32. Need more PSP decompiled to know for sure.

include/game.h
```diff
@@ -1796,7 +1796,7 @@ typedef struct {
     /* 80072F3C */ s32 unk1C;
     /* 80072F40 */ s32 unk20;
     /* 80072F44 */ s32 unk24;
-    /* 80072F48 */ s32 unk28;
+    /* 80072F48 */ void (*unk28)(Entity*);
```
hohle, line 1799: Any reason not to use the `PfnEntityUpdate` type? `func_psp_0923B2F0` looks like a typical entity state machine function.

JoshSchreuder: No reason, just copied out of the old definition https://github.com/Xeeynamo/sotn-decomp/pull/2247/files#diff-4dbceb2f69232a451b571009bbb93e8c561300b5c08d314cafe8a39f5c49d6d2L1932

Thanks for the feedback, have updated

sozud (approved):

PR #2248, opened 2025-02-22
Speed Up `dups` & Fix Grouping

hohle:
Peeling the speedup and grouping bug fix off from the other changes in the `speedy-dups` PR (#2205).

Looking at the difference of size of two functions is a constant time approximation to the Levenshtein distance. Checking this prior to calculating the true edit distance provides a significant speedup without any modification to behavior.

This also fixes an issue where a "similarity" calculation (one minus the ratio of edits to size) was being returned but treated as a "distance" in some calculations. This meant that functions where grouped with the least similar function which was above the threshold instead of the most similar. This affects a few groups, like `*_EntityCutscene` where the two groups now contain functions that match more closely than before.

Some make recipes are also fixed: Finally, `MAD` symbols are extracted and `force_extract` will no longer fail if there are no linker scripts in the build directory.

Without the approximation shortcut:

```
$> time target/release/dups --threshold 0.9 --output-file ../../gh-duplicates/duplicates-current.txt
⋮
real    3m27.978s
user    3m26.470s
sys     0m1.493s
```

With the approximation shortcut:
```
$> time target/release/dups --threshold 0.9 --output-file ../../gh-duplicates/duplicates-new.txt
⋮
real    0m10.266s
user    0m9.739s
sys     0m0.526s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -453,7 +453,7 @@ endif
 # Force to extract all the assembly code regardless if a function is already decompiled
 force_extract:
 	mv src src_tmp
-	rm $(BUILD_DIR)/*.ld
+	find $(BUILD_DIR) -type f -name "*.ld" -delete
 	make extract -j
 	rm -rf src/
 	mv src_tmp src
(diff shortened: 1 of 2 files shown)


sozud:
Can you post the results as a comparison?

hohle:
https://gist.github.com/hohle/67ce761727b725353d27e2ecb3f55500

I've included a diff in a comment on that gist as well. The biggest change is that `MAD` has matched symbols now so sorting changes a bit.

One grouping change is `EntityCutscene` related - `NO3_EntityCutscene` and `NZ0_EntityCutscene` swap places and are more similar in their new clusters. There's also a bit of regrouping with (`def_cbread`, `def_cbready` , `def_cbsync`, `GsGetVcount`, `GsClearVcount`) and (`func_800F6A48`, `SpuSetNoiseVoice`).

sozud:
I don't see the comment?

hohle:
I must not have submitted submitted. Diff added to gist.

sozud (approved):

PR #2259, opened 2025-02-25
Fix MAD funcs not appearing in function report

JoshSchreuder:
EDIT: Approach updated, below is no longer applicable. See comments for more details.

This restores two MAD functions to the function report which went missing

Output from local:
https://gist.github.com/JoshSchreuder/c146abedc057b18d588b44446b00b596

```
| mad    | func_8018D8C8                     |      250 |         45 |        |                                                                                                    |       |     |
| mad    | func_80197B94                     |      491 |         44 | Yes    |                                                                                                    |       |     |
```

This was because since #2248 added MAD to the force_symbols extract, it was mapping 5 labels in these functions as symbols which was [causing invalid splat disasm after extracting with these symbols](https://gist.github.com/JoshSchreuder/adb4dc20da482d23838fb50b56580a17)
```
.L80197ED4 = 0x80197ED4; // allow_duplicated:True
.L80197F8C = 0x80197F8C; // allow_duplicated:True
.L8019803C = 0x8019803C; // allow_duplicated:True
.L801980DC = 0x801980DC; // allow_duplicated:True
.L80198198 = 0x80198198; // allow_duplicated:True
```

I have added a simple exclude for .L. This does not appear to have any impact on any other overlays or functions. I am not clear why these .L symbols were being added to the MAD elf in the first place, so if anyone has a better solution please let me know.

### DIFF
--- a/config/splat.us.stmad.yaml
+++ b/config/splat.us.stmad.yaml
@@ -17,7 +17,7 @@ options:
   undefined_syms_auto_path:  config/undefined_syms_auto.stmad.txt
   find_file_boundaries: yes
   use_legacy_include_asm: no
-  migrate_rodata_to_functions: no
+  migrate_rodata_to_functions: yes
   asm_jtbl_label_macro: jlabel
   extensions_path: tools/splat_ext
   string_encoding: SHIFT-JIS
@@ -76,12 +76,12 @@ segments:
       - [0x1774, data]
       - [0x19A4, data]
       - [0x1D14, data]
-      - [0xD794, rodata]
+      - [0xD794, .rodata, D8C8]
       - [0xD7CC, .rodata, e_red_door] # EntityRedDoor
       - [0xD83C, .rodata, e_collect]
       - [0xD85C, .rodata, e_collect] # EntityEquipItemDrop
       - [0xD874, .rodata, e_misc]
-      - [0xD8AC, rodata]
+      - [0xD8AC, .rodata, 17B94]
       - [0xD8C0, .rodata, prim_helpers]
       - [0xD8C8, c]
       - [0xE5AC, c, st_debug]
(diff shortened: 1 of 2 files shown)


ProjectOblivion:
Wouldn't it be better to revert just the single change that was causing the issue, rather than a quick workaround that we're unsure whether there are consequences?

JoshSchreuder:
Yeah that is another option, it would mean losing MAD from the duplicates report though which was just recently fixed, so ideally we could fix both but I'm fine with whatever @sozud and @hohle are happy witth

ProjectOblivion:
This fixes the problem symbols from being added to the map in the first place and therefore won't be added with make force_symbols.
```
--- a/config/splat.us.stmad.yaml
+++ b/config/splat.us.stmad.yaml
@@ -20 +20 @@ options:
-  migrate_rodata_to_functions: no
+  migrate_rodata_to_functions: yes
@@ -79 +79 @@ segments:
-      - [0xD794, rodata]
+      - [0xD794, .rodata, D8C8]
@@ -84 +84 @@ segments:
-      - [0xD8AC, rodata]
+      - [0xD8AC, .rodata, 17B94]
```

hohle:
I have a local change that I haven't pushed yet to ignore anything but `.text` sections. This filters out these vars without string matching on labels:

```
// parse .s file to get instructions and function name
 // parse .s file to get instructions and function name
 fn parse_instructions(input: &str, dir: &str, file: &str) -> Function {
     let mut instructions = Vec::new();
     let mut func_name = "";
+    let mut section = ".text";

     for line in input.lines() {
         let parts: Vec<&str> = line.split_whitespace().collect();

         // find the function name
         if parts.len() == 2 {
-            if parts[0] == "glabel" {
-                func_name = parts[1];
+            match parts[0] {
+                "glabel" => func_name = parts[1],
+                ".section" => section = parts[1],
+                _ => (),
             }
         }

+        if section != ".text" {
+            // ignore non-code sections
+            continue;
+        }
+
```

sozud:
Based on the slack convo it sounds like importing the rodata is the cleanest fix?

JoshSchreuder:
@sozud agree, have updated PR with the above approach from @ProjectOblivion 

Works for me locally, here's a diff on function report with those changes
https://gist.github.com/JoshSchreuder/20597fec1f965802e3dfc3715d859ef4

PR #2261, opened 2025-02-25
Decompile Doppelganger Animation Functions

hohle:
These are duplicates or near duplicates of functions for controlling Alucard or Richter in `DRA` or `RIC`.

These have been pulled out into a new segment in `DRA` to match the grouping in `BO4`.


hohle:
I started out assuming they were the same because all of the functions match. What's interesting is they seem to emulate controller button presses to control doppelgänger instead of a typical entity that's controlled directly.

If I were going to pull it out to make it common, I'd probably make a new macro, maybe `CHARACTER` or something similar that could be defined. It looks like the `DOPPELGANGER` macro is going to be used a lot more in other functions.

sozud:
I think we can investigate factoring these out as a follow-up

PR #2264, opened 2025-02-26
Add size hints for PSX globals and fix some references

JoshSchreuder:
Like what I had done earlier on PSP in #2247, and this has similar benefits but for PSX.

Using this I fixed up some incorrect symbol mappings


tools/sotn-debugmodule/Makefile
```diff
@@ -19,21 +19,25 @@ CC_FLAGS        += -c -march=r3000 -fno-stack-protector -mtune=r3000 -EL -G0 -O2
...
+	@echo "Processing: $(CONFIG_DIR)/symbols.$(VERSION).txt -> $(CONFIG_DIR)/symbols.$(VERSION).ld.txt"
+	grep -o '^[^/]*' "$(CONFIG_DIR)/symbols.$(VERSION).txt" > "$(CONFIG_DIR)/symbols.$(VERSION).ld.txt"
 
 $(BUILD_DIR)/$(MODULE_NAME).bin: $(BUILD_DIR)/$(MODULE_NAME).elf
 	$(OBJCOPY) -O binary $< $@
 	printf '\x00' | dd of=$@ bs=1 seek=40959 count=1 conv=notrunc
 
 $(BUILD_DIR)/%.elf: $(BUILD_DIR)/tools/%/inject.o $(BUILD_DIR)/tools/%/debugmode.o $(BUILD_DIR)/tools/%/menu.o $(BUILD_DIR)/tools/%/fontmanager.o $(BUILD_DIR)/tools/%/debug_flags.o $(BUILD_DIR)/tools/%/entity_spawn.o $(BUILD_DIR)/tools/%/sfx_player.o $(BUILD_DIR)/tools/%/flag_checker.o
-	$(LD) -o $@ -Map $*.map -T $*.ld -T $(CONFIG_DIR)/symbols.$(VERSION).txt \$^
+	$(LD) -o $@ -Map $*.map -T $*.ld -T $(CONFIG_DIR)/symbols.$(VERSION).ld.txt \$^
```
JoshSchreuder, line 40: `mipsel-linux-gnu-ld` doesn't like symbols files with comments, unlike splat. 
I've tried to keep this self-contained, copies to a new file and strips the comments.

tools/make-config.py
```diff
@@ -436,6 +436,7 @@ def get_splat_config(
             "ld_bss_is_noload": bss_is_no_load,
             "disasm_unknown": True,
             "disassemble_all": True,
+            "global_vram_start": HexInt(0x80010000),
```
JoshSchreuder, line 439: This is overridden already for PSP, I tested this locally and it works as expected

config/symbols.us.txt
```diff
@@ -872,69 +839,7 @@ _svm_vh = 0x8006C3C0;
 g_Servant = 0x8006CBC4;
 _svm_tn = 0x8006CBC8;
 g_Clut = 0x8006CBCC;
-g_Player = 0x80072BD0;
```
sozud, line 875: So now we have to manually figure out the field name from the offset in decomp.me right? Like `g_Player+0x31C == g_Player.padPressed`

JoshSchreuder: Here's an example of that which I just extracted off this branch. 

You get `g_Player + 0x320` in the ASM, but M2C is pretty good at determining the property which is in use, and correctly works out this is `g_Player.padTapped` etc.
https://decomp.me/scratch/6SrOz

If it's less useful this way I can revert the changes to some of these where we've manually mapped members and just keep it contained to ones which we didn't yet.

gamezter: IMO base symbol + offset is a lot clearer when decompiling in decomp.me. I think the fake symbols have been confusing to newcomers as well.

sozud: If contributors prefer it this way I don't have an issue with it, @Xeeynamo do you have an opinion?

ProjectOblivion: The fake symbols have helped me significantly in working the functions, I didn't find them confusing at all.

That being said, I found the example from this branch to be just as easy to understand and seems like it would be a bit more convenient.  Where I found the branch version particularly nice is in final double check. It's really easy to compare offset to offset, but not so much fake symbol to offset.

tools/sotn-debugmodule/Makefile
JoshSchreuder: @Xeeynamo this didn't seem to work for me, get a whole heap of undefined references. Even if I copy over the whole `symbols.us.txt` file into `undefined_syms.us.txt` it still fails compiling DRA

PR #2270, opened 2025-02-27
Created a loader for decomp-permuter

ProjectOblivion:
I created a loader for decomp-permuter so that its code doesn't need to be edited for it to work.
    - There are 4 symlinks I needed to create in order to deal with a few things about decomp-permuter
      - The link to src/ and default_weights.toml are because of the pathing import.py expects
      - The link to import.py is needed because import.py cannot be imported
      - The link to decomp-permuter/ is needed because Python doesn't allow `-` in variables/modules

I noticed that some python was missing for make format, so I added those to make along with this script.

I added the decomp-permuter default `nonmatchings` directory and the loader's default directory to `.gitignore`.

I added decomp-permuter to git submodules and disabled updates to it since this loader is probably very fragile and any updates to that project could easily break its functionality.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ sotn_calltree.txt
 function_calls
 
 gmon.out
+
+nonmatchings/
+tools/sotn_permuter/permute.me/
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,7 @@
 [submodule "tools/mwccgap"]
 	path = tools/mwccgap
 	url = https://github.com/mkst/mwccgap.git
+[submodule "tools/decomp-permuter"]
+	path = tools/decomp-permuter
+	url = https://github.com/simonlindholm/decomp-permuter
+	update = none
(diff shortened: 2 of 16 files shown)


JoshSchreuder:
Looks good, haven't yet tried this locally but two suggestions (don't have to do this as part of same PR, we can iterate on it)

- combine import and permute commands? I do this locally with my permuter; I'm not sure there's much value to having a two step process, given you presumably always want to do both?

- further to above, add a Makefile target so we can call this from root. eg. `make permute path/to/file.c path/to/asm.o`. Just to make it super easy to call, and ideally it can use the same $VERSION variable.

sozud (approved):
Seems fine to me, would be good if a couple of users can test this out before merging

ProjectOblivion:
> Looks good, haven't yet tried this locally but two suggestions (don't have to do this as part of same PR, we can iterate on it)
> 
> * combine import and permute commands? I do this locally with my permuter; I'm not sure there's much value to having a two step process, given you presumably always want to do both?
> * further to above, add a Makefile target so we can call this from root. eg. `make permute path/to/file.c path/to/asm.o`. Just to make it super easy to call, and ideally it can use the same $VERSION variable.

These are ideas I also had for the future, I was intending to create a couple of make targets with permute having the dependency of the imported function, effectively eliminating the need for import at any point.

PR #2271, opened 2025-02-28
Decompile `st0` `EntityClouds`

JoshSchreuder:
This should be the last function for PSX ST0 🎉 

Matching both
PSX: https://decomp.me/scratch/afe2R
PSP: https://decomp.me/scratch/ZJDcJ

With thanks to @dezgeg who helped me work out a struct usage.


include/psxsdk/libgte.h
```diff
@@ -216,6 +216,10 @@ extern long ratan2(long y, long x);
                      :                                                         \
                      : "r"(r0), "r"(r1), "r"(r2))
 
+#define gte_ldtx(r0) __asm__ volatile("ctc2   %0, $5;" : : "r"(r0))
```
sozud, line 219: As far as I can these aren't actually part of psy-q? It's fine to leave it here, just wondering what these are. It's possible they made their own macros from parts of the headers. https://github.com/sozud/psy-q/blob/9f541dbade4b25bcba449aea0e3dac9ab3cde14b/3.6/PSX/INCLUDE/INLINE_O.H#L196

sozud (approved):

include/psxsdk/libgte.h
JoshSchreuder: These are the 3 instructions from `gte_ldtr` above split out
That would be this one
https://github.com/sozud/psy-q/blob/9f541dbade4b25bcba449aea0e3dac9ab3cde14b/3.6/PSX/INCLUDE/INLINE_O.H#L338-L340

Seems plausible they might have made their own macros from it based on how this code is structured, since it calls all 3 at independent parts of the code rather than all together.

PR #2273, opened 2025-02-28
Import `lib` e_shop data

gamezter:


src/st/lib/e_shop.c
```diff
@@ -3161,13 +3176,18 @@ void func_us_801B5068(Entity* self) {
...
-const char D_us_801AD0C8[] = _S("Magic scroll 2");
-const char D_us_801AD0D8[] = _S("Magic scroll 1");
-const char D_us_801AD0E8[] = _S("Castle map");
+#ifdef VERSION_PSP
+extern char* D_us_801816A4;
+extern const char** D_us_801816AC;
+extern const char** D_us_801816B0;
+#else
+static char D_us_801816A4[] = {
+    CH('S'), CH('O'), CH('L'), CH('D'), CH(' '), CH('O'), CH('U'), CH('T')};
```
gamezter, line 3185: There are a few places where a `_S(...)` macro that doesnt append an `0xFF` would be useful in this file, this is an example

src/st/lib/e_shop.c
```diff
@@ -4160,6 +4218,56 @@ void func_us_801B6F30(Entity* self) {
...
+                              0x0200, 0x0080, 0x0100, 0x0040, 0x0020, 0x0000};
+
+#ifdef VERSION_PSP
+extern const char** D_us_801818C0;
+extern const char* D_us_801818EC[];
+extern const char* D_us_801818F0[];
+#else
+static const char* D_us_801818C0[] = {
+    _S("LV"),         _S("HP"),       "$EF",       _S("Exp"),
+    _S("Drop Items"), _S("No."),      "\x1F\x1F",  _S("Strong VS."),
```
gamezter, line 4265: "$EF" and "\x1F\x1F" are actually `_S("Def")` and `_S("??")` without the trailing 0xFF. I did this until we find a solution

PR #2275, opened 2025-03-02
Update `st0` PSP for sharing of Cutscene entities

JoshSchreuder:
Brought in the PSP side for this file

Made a few refactors to some types, and a couple of changes to the PSX code required to match on PSP


src/st/st0/3AB08.c
```diff
@@ -283,67 +288,54 @@ static ClockTowerData2 D_8018219C[] = {
...
+    {0x0041, 0xFF9D, 0xFFBF, 0x0000}, {0x0063, 0xFF9D, 0xFF9D, 0x0000},
+    {0x0063, 0xFF9D, 0xFFBF, 0x0000},
 };
-static u8 D_80182220[] = {
-    0x53, 0x00, 0x5D, 0xFF, 0x52, 0x00, 0x00, 0x00, 0x42, 0x00,
-    0x9D, 0xFF, 0x42, 0x00, 0x00, 0x00, 0x42, 0x00, 0x9D, 0xFF,
-    0x63, 0x00, 0x00, 0x00, 0x63, 0x00, 0x9D, 0xFF, 0x42, 0x00,
-    0x00, 0x00, 0x63, 0x00, 0x9D, 0xFF, 0x63, 0x00, 0x00, 0x00,
+static SVECTOR D_80182220[] = {
+    {0x0053, 0xFF5D, 0x0052, 0x0000}, {0x0042, 0xFF9D, 0x0042, 0x0000},
```
bismurphy, line 302: I don't know anything, but if the members of SVECTOR are signed values, would it make more sense for FF5D to be presented as a negative value than as a value starting with FF? Doesn't hugely matter either way and shouldn't affect anything being merged, just a thought.

JoshSchreuder: Yeah believe they are short (s16), I can update these

PR #2282, opened 2025-03-04
Add codesharing for `st0` `EntityDracula` entities part one

JoshSchreuder:
Thanks to @ProjectOblivion for help matching `func_801ABBBC`.

This brings in the first batch of `EntityDracula` entities for `st0` with codesharing back to PSP.

As part of this I have removed the `draculaPrimitive` struct and replaced it with some calls to `Primitive` members which I will comment inline.


src/st/st0/2C564.c
```diff
@@ -96,86 +100,89 @@ static u32 func_801ABBBC(s32 step, Entity* dracula) {
...
-            prim->y1_f32.i.lo = 0;
-            prim->y1_f32.i.hi = prim->y1;
-            // FAKE variable reuse
-            color = (i >> 1);
-            prim->y1_f32_velocity.val =
-                ((color * color * color) << 8) + 0x80000;
+            F(prim->x2).i.lo = 0;
+            F(prim->x2).i.hi = prim->y1;
+            color = i >> 1;
+            F(prim->x3).val = ((color * color * color) << 8) + 0x80000;
```
JoshSchreuder, line 185: These are the only usage of anything unique to `draculaPrimitive`. Since this can be achieved with the `F()` macro on `Primitive`, I did not see much reason to keep this struct. IMO it didn't seem to be giving much additional context or clarity to this code, but I can revert this change if required.

src/st/st0/2C564.c
```diff
@@ -42,52 +46,52 @@ static u8 D_80180A20[] = {
...
-static Point16 D_80180A60[] = {
-    {12, 12}, {12, -12}, {-12, 12}, {-12, -12}, {80, 64}, {48, 34},
-    {26, 20}, {18, 16},  {15, 13},  {13, 12},   {12, 11}, {11, 10},
-    {10, 10}, {9, 9},    {9, 8},    {8, 8},     {8, 4}};
+static Point16 D_80180A60[] = {{12, 12}, {12, -12}, {-12, 12}, {-12, -12}};
+// Possible this is an unused u8 animation from the next file.
+// It is stripped on PSP
+static Point16 unused[] = {
+    {80, 64}, {48, 34}, {26, 20}, {18, 16}, {15, 13}, {13, 12}, {12, 11},
+    {11, 10}, {10, 10}, {9, 9},   {9, 8},   {8, 8},   {8, 4}};
```
JoshSchreuder, line 56: This is stripped on PSP, and as such isn't part of `D_80180A60`. Given the next lot of data is animations from `2DAC8`, I suspect this is likely an unused animation but will need to dig into that further to confirm.

include/sfx.h
```diff
@@ -229,7 +229,11 @@ enum SfxModes {
 #define NA_SE_EV_WATER_SPLASH 0x7C2
 #define NA_SE_VO_DR_TAUNT_1 0x84F
 #define NA_SE_VO_DR_TAUNT_2 0x850
+#ifdef VERSION_PSP
+#define NA_SE_VO_DR_HURT_1 0x85A
```
JoshSchreuder, line 233: This seems to be an intuitive value on PSP:
HURT_1 = 0x85A
HURT_2 = 0x85B
HURT_3 = 0x85C
HURT_4 = 0x85D

Strangely it is not sequential like this on PSX.

sozud: Doesn't have to be done now but the fact that the numbers are shifted on psp suggests these are part of the sfx enum

sozud (approved):

PR #2283, opened 2025-03-04
Fix a couple of PSP build warnings

JoshSchreuder:
Nothing major, just a couple of things that were bugging me while working on PSP.


src/st/st_common.h
```diff
@@ -56,7 +56,7 @@ static s16 g_SineTable[] = {
 #include "move_entity.h"
 #include "fall_entity.h"
 
-#if defined(VERSION_BETA) || STAGE == STAGE_ST0
+#if defined(VERSION_BETA) || (STAGE == STAGE_ST0 && !defined(VERSION_PSP))
```
JoshSchreuder, line 59: This function is stripped on PSP but still goes through the compiler, we can just exclude it.

```
### mwccpsp.exe Compiler:
#      In: src\st\st_common.h
#    From: src\st\st0\st_common.c
# -------------------------------
#      61: u8 unkState;
# Warning:    ^^^^^^^^
#   variable 'unkState' is not initialized before being used
```

sozud (approved):

PR #2289, opened 2025-03-05
Next file of code sharing for DRA PSP

bismurphy:
Almost every change here is simple, reasonable, and straightforward. Nice to see a clean diff; I'm really glad we're using this workflow of pre-aligning the file splits, and then decompiling in the next PR.

Getting a merge conflict, will fix before merging.

Also should resolve the matter of `BTN_WOLF`, I raised the question in this in #naming in the Discord.


include/game.h
```diff
@@ -185,13 +185,22 @@ typedef enum {
...
     GAMEBUTTONS = (~(PAD_START | PAD_SELECT)),
 } PlayerPad;
 
+// PSP only has one shoulder button on each side, so its default transform
+// controls are different. This captures those.
+#ifdef VERSION_PSP
+#define BTN_WOLF PAD_L1
+#else
+#define BTN_WOLF PAD_R2
+#endif
```
hohle, line 202: This seems like it will redefine `BTN_WOLF` if `VERSION_PSP` is not defined (once on 188 and again on 201)

bismurphy: Good catch, I had it up above thinking I should reuse the same ifdef, but then I realized it was inside an enum which didn't feel right to be defining vars in that location, so I moved them down below, but looks like I didn't properly remove the first one. Thanks for the notice!

PR #2306, opened 2025-03-11
Add myself to codeowners

gamezter:


JoshSchreuder:
I don't think this is actually working for us

> The people you choose as code owners must have write permissions for the repository. When the code owner is a team, that team must be visible and it must have write permissions, even if all the individual members of the team already have write permissions directly, through organization membership, or through another team membership.

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

PR #2307, opened 2025-03-11
Decompile `lib_psp` func_us_801B8A00

gamezter:
Depends on https://github.com/Xeeynamo/sotn-decomp/pull/2305
I realized PSX `func_us_801B8A00` was matched with the wrong PSP function, here I decompile the true PSP function.
PSX: https://decomp.me/scratch/jiK9K
PSP: https://decomp.me/scratch/CgyWa


src/st/lib/e_shop.c
```diff
@@ -5595,13 +5669,13 @@ void func_us_801B8A00(Entity* self) {
         break;
 
     case 9:
-        g_CutsceneFlags &= ~0x400;
 #ifdef VERSION_PSP
         if (self->ext.et_801B6F30.unk86) {
-            g_api.PlaySfx(MU_LOST_PAINTING_LOOP_POINT);
+            g_api.PlaySfx(0x302);
```
sozud, line 5674: MU_LOST_PAINTING_LOOP_POINT

sozud (approved):

src/st/lib/e_shop.c
gamezter: I reverted this back to 0x302 based on this comment : https://github.com/Xeeynamo/sotn-decomp/blob/5be631b01c0be0b301433e2f09b792cf222f1288/src/st/lib_psp/e_shop.c#L66

sozud: You're right, the 0x200 range on ps1 is the sequenced music so this is most likely starting the librarian song. My guess is that they recorded it to PCM on psp for simplicity so it's in the 0x300 range there

sozud:
This is conflicted

PR #2308, opened 2025-03-12
Update `st0` PSP prologue cutscene for match

JoshSchreuder:
I have chosen not to merge these files for now. There was quite a bit of difference between the special `g_Dialogue` used on PSP and PSX side, and many differences in the implementations due to PSP's 5 languages etc.

Where possible I have "ported" the code cleanups back to the PSX version. Also imported the PSX script for the prologue cutscene roll which is what this code is responsible for.

Think this leaves just one remaining ST0 function for PSP (`ST0_EntityCutscene`)


src/st/st0_psp/3101C.c
```diff
@@ -1,16 +1,352 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
-#include "common.h"
+#include "../st0/st0.h"
+#include "../st0/dialogue2.h"
 
-INCLUDE_ASM("st/st0_psp/psp/st0_psp/3101C", func_pspeu_09242A20);
+typedef struct {
+    Primitive* prim;
+    float y0;
```
JoshSchreuder, line 7: This is quite peculiar, don't think there are many places outside DRA PSP using floats, but here is one. It doesn't seem to have much purpose as far as I can tell, just gets set here from a function below.

src/st/st0/dialogue2.h
```diff
@@ -12,9 +38,17 @@ typedef struct {
     /* 0x10 */ u16 portraitAnimTimer; // portrait animation timer
     /* 0x12 */ u8 unk12;              // unknown
     /* 0x13 */ u8 clutIndex;
-    /* 0x14 */ Primitive* prim[1]; // for dialogue graphics rendering
-    /* 0x18 */ s32 primIndex[8];   // primIndices: unk, actorName, unk
```
JoshSchreuder, line 16: Didn't see much evidence these were arrays (if they are the other indexes are not used). I've updated this to be closer to the PSP struct which just has a single entry here and padded out the remainder.

src/st/st0_psp/3101C.c
```diff
@@ -1,16 +1,352 @@
...
+    switch (self->step) {
+    case 0:
+        g_Dialogue2.script = GetLangAt(
+            0, (u8*)prologue_script_en, (u8*)prologue_script_fr,
+            (u8*)prologue_script_es, (u8*)prologue_script_de,
+            (u8*)prologue_script_it);
+        if (func_801B101C(g_Dialogue2.script)) {
+            func_8925F7C(0x100, 0xF1, 0x10, 1);
+            func_8925F7C(0x110, 0xF2, 0x10, 1);
+            func_8925F7C(0x100, 0xF0, 0x10, 1);
```
JoshSchreuder, line 184: Not sure what this is but it sits between these two?
```
FntPrint = 0x08925e98;
SsSetMVol = 0x89261B4;
```

Any idea how we found out what these system functions were?

sozud: DRA starts at 0x090DCA00 so I think it's probably part of the psp equivalent of main? We'd probably want to add that file although I'm not sure what it is named

sozud (approved):

PR #2309, opened 2025-03-12
Update mwccgap + import psp strings as Shift-JIS

gamezter:
mwccgap now allows us to set a target encoding for strings, which seem to be encoded as Shift-JIS. Otherwise strings are encoded as utf-8 : https://discord.com/channels/1079389589950705684/1079395311350452255/1349511246256406555


sozud (approved):
Does this work with the string import stuff you're doing? Right now I'm not sure there's any psp shift-jis examples to confirm with? If so seems fine to me

ProjectOblivion:
It does, I've tested it out as well.

PR #2311, opened 2025-03-14
Import `lib_psp` e_shop strings

gamezter:
Import e_shop strings on PSP + decompile more PSP functions, I had to add a build SOTN_STR build step (Thanks @ProjectOblivion !)

This is still a work in progress, but this PR was becoming huge so I think this is a good place to have it reviewed.

So far it looks like PSP uses Shift-JIS encoding, but replaces range 0xA1 - 0xDF (halfwidth katakana characters) with accent characters. Maybe in the future we could add a pass in SOTN_STR to convert all accent characters on PSP to their Shift-JIS representation ?

I also removed the `MENUCHAR(x)` macro since it is identical to `CH(x)`, and added the `CIRCLE`, `CROSS`, `SQUARE` and `TRIANGLE` defines for their printable 8x8 font representation


src/st/lib_psp/e_shop.c
```diff
@@ -20,8 +48,2222 @@ typedef enum {
...
+    _S("Prぅlogo"),
+    _S("Baile de ilusiones"),
+    _S("Plegaria"),
+    _S("Castillo de Dracula"),
+    _S("Baile de oro"),
+    _S("Pasillo de mロrmol"),
+    _S("Torre de niebla"),
+    _S("Nocturno"),
+    _S("Partita tallada en madera"),
+    _S("Puerta de esp妖ritussagrados"),
```
sozud, line 1705: Is this right? 妖 should be coded as 0x9760, it's not half-width. There's other examples of the same issue but different characters. Given that this is in `_S` they probably changed the sotn_str encoding. If we look at the psp vram we can probably see the font texture?

ProjectOblivion: Those characters were used because they represent the correct byte literals for accent characters.

gamezter: This is what SOTN_STR accepts to convert to the proper bytes, I tried using the expected halfwidth characters but SOTN_STR will fail since those characters are not in the utf8_to_index dictionary.

sozud: It doesn't have to be a part of this PR but we ought to make a second jp.py character table for the psp version so these get converted correctly 
image

PR #2312, opened 2025-03-15
Scattered progress on PSP DRA, especially -O4 functions

bismurphy:
Decided to try tackling the 1E50.c file. Got all the functions matching, except the ones that are unique to PSP.

Since those unique ones are using lots of unidentified symbols, I started to look to other functions that also touched those symbols. So I ended up doing a bunch of individual functions, which ended up revealing a ton more symbols, so that's nice.

Seems like I hit a reasonable number of changes for one PR for now. We'll see what's next.

No backports on any of these functions at this point. Seems like it makes more sense to do them file by file.


Makefile.psp.mk
```diff
@@ -113,12 +113,12 @@ $(BUILD_DIR)/%.s.o: %.s
...
 OPT_HIGH = -O4,p #need this because otherwise the comma breaks the if-statement
 # Allow override. Any file in this list will get O4.
-OPT_HI_OVERRIDES = 33F0.c.o A710.c.o C0B0.c.o EC60.c.o 186E8.c.o 61F30.c.o 624DC.c.o 628AC.c.o 63C90.c.o 64EE0.c.o
+OPT_HI_OVERRIDES = 80.c.o 1E50.c.o 33F0.c.o A710.c.o C0B0.c.o EC60.c.o F4D0.c.o 186E8.c.o 61F30.c.o 624DC.c.o 628AC.c.o 63C90.c.o 64EE0.c.o
 OPTIMIZATION = $(if $(filter $(notdir $@),$(OPT_HI_OVERRIDES)), $(OPT_HIGH), -Op)
 
 $(BUILD_DIR)/%.c.o: %.c $(MWCCPSP) $(MWCCGAP_APP)
 	@mkdir -p $(dir $@)
-	$(MWCCGAP) $< $@ --mwcc-path $(MWCCPSP) --use-wibo --wibo-path $(WIBO) --as-path $(AS) --asm-dir-prefix asm/pspeu --target-encoding sjis --macro-inc-path include/macro.inc $(MWCCPSP_FLAGS) $(OPTIMIZATION)
+	$(MWCCGAP) $< $@ --mwcc-path $(MWCCPSP) --use-wibo --wibo-path $(WIBO) --as-path $(AS) --asm-dir-prefix asm/pspeu --target-encoding sjis --macro-inc-path include/macro.inc $(MWCCPSP_FLAGS) $(OPTIMIZATION) -opt nointrinsics
```
JoshSchreuder, line 121: Should we make an issue for decomp so mkst can add this to the PSP preset? 
Or it's only applicable for O4 do you think?

bismurphy: Yeah it only applies when an O4 function calls a builtin function like abs, I think for now we can leave it out and if it becomes more relevant it can join the preset. Not worth the trouble for now.

PR #2316, opened 2025-03-16
Reworking sotn_str

sozud:
Since sotn_str is going to get more complicated to handle the new PSP cases I started reworking it in rust with a plan to add the psp handling. This doesn't produce a matching build yet but handles most things. Putting this up as a draft for awareness


sozud:
I added the french accents, I figure we don't have to fix all the languages at once

PR #2317, opened 2025-03-16
Streamlined Makefile.psp.mk and cleaned up Makefile.saturn.mk some

ProjectOblivion:
Incorporates changes from #2300 and requires #2315
new_link and new_list_o_files are intended to be renamed to link and list_o_files respectively once all targets using the old versions have been converted.


ProjectOblivion:
I just found that I had gotten my wires crossed and changes that I thought I had submitted in #2315 were actually in a totally separate branch and not submitted.  Since this PR may be expecting some those changes, I'm going to close this and resubmit after those get reviewed.

PR #2320, opened 2025-03-16
Decompile small `no1` funcs

JoshSchreuder:
A couple more smaller `no1` funcs

`DebugInputWait` looks like a direct copy from `dra`.
Doesn't appear to be called within this overlay and thus I think is stripped on PSP side.

`func_us_801B8F50`
PSX: https://decomp.me/scratch/688Qe
PSP: https://decomp.me/scratch/LCQNp

### DIFF
--- a/src/st/no1/unk_381E8.c
+++ b/src/st/no1/unk_381E8.c
@@ -39,16 +39,13 @@ void func_us_801B81E8(Entity* self) {
     g_api.UpdateAnim(NULL, NULL);
 }
 
-static s32 D_us_801D637C;
-
-void func_us_801B832C(s8* msg) {
-    s32 temp_v0;
-
+// Likely copied out of DRA. This code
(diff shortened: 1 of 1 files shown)


src/st/no1/unk_381E8.c
```diff
@@ -333,8 +334,29 @@ void func_us_801B8B00(Entity* self) {
 
 INCLUDE_ASM("st/no1/nonmatchings/unk_381E8", func_us_801B8D30);
 
-// entering doppleganger's room
```
JoshSchreuder, line 336: I think this comment was a red herring from debugging because the doppleganger room moves into RBO4 and this function probably sits at a similar address in memory. Given the decompiled code references the elevator flag, and the code directly above this is all elevator related I think it's safe to say this is probably part of an elevator code file.

PR #2321, opened 2025-03-16
asm-differ: watch psp source directory

Xeeynamo:
Simple hack to append the `_psp` suffix to a source directory. This will allow asm-differ to watch the PSP source files when diffing locally.

### DIFF
--- a/diff_settings.py
+++ b/diff_settings.py
@@ -115,5 +115,16 @@ def apply(config, args):
             apply_psx_base(config, version, name)
         else:
             apply_psx_bin(config, version, name)
+
+        # HACK: allow to watch for the 'src/*_psp' directory
+        if version.startswith("psp"):
+            source_directories = config["source_directories"]
+            new_source_directories = []
+            for source_dir in source_directories:
+                new_source_directories.append(source_dir)
+                if source_dir.startswith("src/"):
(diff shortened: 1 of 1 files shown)


sozud (approved):

diff_settings.py
```diff
@@ -115,5 +115,16 @@ def apply(config, args):
...
+
+        # HACK: allow to watch for the 'src/*_psp' directory
+        if version.startswith("psp"):
+            source_directories = config["source_directories"]
+            new_source_directories = []
+            for source_dir in source_directories:
+                new_source_directories.append(source_dir)
+                if source_dir.startswith("src/"):
+                    new_source_directories.append(source_dir + "_psp")
+            config["source_directories"] = new_source_directories
```
ProjectOblivion, line 127: This could be simplified to:
```
        if version.startswith("psp") and config["source_directories"]:
            add_dirs = [f"{dir}_psp" for dir in config["source_directories"] if dir.startswith("src/")]
            config["source_directories"].extend(add_dirs)
```

This is assuming config["source_directories"] can be assumed to be a list.

ProjectOblivion (changes requested):

ProjectOblivion:
I also just discovered that _psp is already accounted for in apply_psx_stage() and apply_psx_boss()

PR #2322, opened 2025-03-16
PSP RIC matches most of functions from the file entry point

Xeeynamo:
Matched almost everything from what I plan to renaming `pl_main.c`. I am waiting to decompile `RicMain` too but it will happen in a separate PR.


sozud:
Needs conflicts resolved

PR #2325, opened 2025-03-17
Decompile remaining `no1` `e_spear_guard` funcs and import data

JoshSchreuder:
Block entity:
PSX: https://decomp.me/scratch/BYhHr
PSP: https://decomp.me/scratch/b1GKw

Unknown helper:
PSX: https://decomp.me/scratch/2nZsL
PSP: https://decomp.me/scratch/rqQUB


src/st/no1/unk_54400.c
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "no1.h"
 
+// Possible this might be part of e_spear_guard
 void func_us_801D4400(Entity* self) {
```
JoshSchreuder, line 5: I believe this is `E_ID_58` which is created from spear guard, but it's in a code path I couldn't trigger in the debugger to 100% confirm, so not clear on what this is. Might not be possible with the spear guard that's in NO1

JoshSchreuder: Was able to trigger this in game. I don't think it is possible within NO1 because of the distance required but this is spear guard related. When the player is far enough away from the spear guard horizontally, the spear guard spawns and throws a spear entity at the player, which is what this is.

Have updated the code accordingly.

JoshSchreuder: Was able to trigger this in game by updating the code and building an disk image. 

I don't think it is possible within NO1 because of the distance required but this is spear guard related. When the player is far enough away from the spear guard horizontally, the spear guard spawns and throws a spear entity at the player, which is what this is.

Have updated the code accordingly.

PR #2334, opened 2025-03-18
Decompile Several `bo4` and `rbo5` Functions

hohle:
Decompile several functions that are duplicated between `bo4`, `rbo5`, and frequently `dra`.

Adds several more global symbols for mapping Doppleganger state fields.


config/symbols.us.bobo4.txt
```diff
@@ -196,9 +196,12 @@ g_Dop = 0x801D4A20;
 g_Dop_padPressed = 0x801D4D38;
 g_Dop_padTapped = 0x801D4D3C;
 g_Dop_padHeld = 0x801D4D40;
+g_Dop_pl_vram_flag = 0x801D4D70;
```
JoshSchreuder, line 199: Thoughts on removing these and just leaving `g_Dop = 0x801D4A20; // size:0x3D0`?

Means you end up with `g_Dop + 0x350` in the asm instead so not as convenient in terms of naming, but you also don't need to map every single struct member like this

hohle: Adding the size is a great idea. Does splat support both simultaneously? I like being able to just replace the underscore with a dot, but see the advantage of covering the whole range.

hohle: It does look like you can have both at the same time. Sweet! Will update.

JoshSchreuder: Hmm I thought I tried both and splat didn't like it but if it works that's great. In that case we might be able to do the same for the g_Entities array also (I left this off because g_Entities[0] is PLAYER and that does seem meaningful)

Maybe it was PSP that I tried and it didn't work...

sozud:
Needs a rebase

sozud:
Looks like this was incompatible with the renaming PR, would you mind rebasing one more time? Thanks
```
src/boss/bo4/doors.c:59: structure has no member named `D_80072EFC'
```

PR #2338, opened 2025-03-19
Decompile and code-share `lib` e_spellbook_magic_tome

gamezter:


src/st/lib/e_spellbook_magic_tome.c
```diff
@@ -1,22 +1,229 @@
...
+    D_us_801830C0, D_us_801830BC, D_us_80183090, D_us_80183098,
+    D_us_801830C8, D_us_801830C8, D_us_801830C8, D_us_801830C4};
+static u8* D_us_8018310C[] = {
+    D_us_8018309C, D_us_801830A4, D_us_801830C0, D_us_801830C0,
+    D_us_801830C0, D_us_801830BC, D_us_801830A0, D_us_801830A8,
+    D_us_801830C8, D_us_801830C8, D_us_801830C8, D_us_801830C4};
+static u8* unused[] = {
+    D_us_801830AC, D_us_801830B4, D_us_801830C0, D_us_801830C0,
+    D_us_801830C0, D_us_801830BC, D_us_801830B0, D_us_801830B8,
+    D_us_801830C8, D_us_801830C8, D_us_801830C8, D_us_801830C4};
```
gamezter, line 74: UVs for an unused book. this gets stripped on PSP

src/st/lib/e_spellbook_magic_tome.c
```diff
@@ -275,14 +482,14 @@ void EntitySpellbook(Entity* self) {
...
             self->ext.spellbookMagicTome.unk82 = 0;
         }
-        FntPrint("x:%4x\n", self->ext.spellbookMagicTome.unk9C);
-        FntPrint("y:%4x\n", self->ext.spellbookMagicTome.unk9E);
-        FntPrint("z:%4x\n", self->ext.spellbookMagicTome.unkA0);
+        FntPrint("x:%4x\n", self->ext.spellbookMagicTome.unk9C.vx);
+        FntPrint("y:%4x\n", self->ext.spellbookMagicTome.unk9C.vy);
+        FntPrint("z:%4x\n", self->ext.spellbookMagicTome.unk9C.vz);
         FntPrint("hz:%4x\n", self->ext.spellbookMagicTome.unk8A);
         FntPrint("w:%4x\n", self->ext.spellbookMagicTome.unk82);
```
JoshSchreuder, line 494: Not essential, but we could rename these struct members according to the debug statements too possibly

include/psxsdk/libgte.h
```diff
@@ -308,6 +312,9 @@ extern long ratan2(long y, long x);
         : "r"(r0)                                                              \
         : "$12", "memory")
 
+#define gte_stotz(r0)                                                          \
+    __asm__ volatile("swc2   $7, 0( %0 )" : : "r"(r0) : "memory")
+
```
JoshSchreuder, line 317: @sozud these are in psy-q so is this the right approach? I don't really know enough about how the psy-q stuff fits in

https://github.com/sozud/psy-q/blob/9f541dbade4b25bcba449aea0e3dac9ab3cde14b/3.6/PSX/INCLUDE/INLINE_O.H#L761-L765

https://github.com/sozud/psy-q/blob/9f541dbade4b25bcba449aea0e3dac9ab3cde14b/3.6/PSX/INCLUDE/INLINE_O.H#L950-L953

src/st/lib/e_spellbook_magic_tome.c
```diff
@@ -1,22 +1,229 @@
...
+    for (i = 0; i < 2; i++) {
+        for (j = 0; j < 6; j++) {
+            if ((j == 5) && !g_CurrentEntity->ext.spellbookMagicTome.unk82) {
+                vecPtr += 4;
+                prim->drawMode = DRAW_HIDE;
+                prim = prim->next;
+            } else {
+                gte_ldv3(vecPtr[0], vecPtr[1], vecPtr[2]);
+                gte_rtpt();
+                if (g_CurrentEntity->palette & 0x8000) {
```
JoshSchreuder, line 116: Not sure if it makes sense but `PAL_OVL(0)`?

src/st/lib/e_spellbook_magic_tome.c
gamezter: unk9C seems to be used as x/y/z rotation vector higher up, idk if it is a position or rotation based off the debug strings. maybe they just represent the 3 components of SVECTOR ?

JoshSchreuder: Was more thinking `self->ext.spellbookMagicTome.ext8A` -> `self->ext.spellbookMagicTome.hz` and `self->ext.spellbookMagicTome.unk82` -> `self->ext.spellbookMagicTome.w` rather than the vectors but not sure if it clarifies the rest of the code much?

include/psxsdk/libgte.h
sozud: So I forget how it works exactly but at some point in the toolchain the values get patched. e.g. like this https://github.com/Gemini-Loboto3/dmpsx/blob/1abd56f740b410ae3052f39437bdc865aa46f07d/src/dmpsx.cpp#L217

The values are obfuscated to hide how the gte works. So 0x11FF becomes 0x4B68002E. I don't think we have this set up in our repo so I don't think using the original avsz4 macro is going to work.

```
original psy-q macro
#define gte_avsz4() { \
  __asm__ volatile ("nop   ": : :"$12","$13","$14","$15","memory"); \
  __asm__ volatile ("nop   ": : :"$12","$13","$14","$15","memory"); \
  __asm__ volatile (".word 0x000011ff": : :"$12","$13","$14","$15","memory"); \
}
```

stotz seems to have different versions but 3.5 seems the same
```
original psy-q macro
#define gte_stotz( r0 ) __asm__ volatile (			\
	"swc2	$7, 0( %0 )"					\
	:							\
	: "r"( r0 )						\
	: "memory" )
```
https://github.com/sozud/psy-q/blob/9f541dbade4b25bcba449aea0e3dac9ab3cde14b/3.5/PSX/INCLUDE/INLINE_C.H#L1042

PR #2341, opened 2025-03-19
Decompile `rbo3` `EntityClouds`

JoshSchreuder:
Appears to be the scrolling clouds in the background of the Medusa fight. Luckily this was quite similar to ST0's clouds which I worked on in #2271 as that was quite a tricky one to match.

This should be the last function in RBO3 🎉 

PSX: https://decomp.me/scratch/Uba2H
PSP: https://decomp.me/scratch/o6MUb


include/cloud_struct.h
```diff
@@ -0,0 +1,13 @@
...
+    s16 unk4;
+    u16 priority;
+} cloudData;
+
+typedef struct {
+    /* 0x00 */ s32 unk0;
+    /* 0x04 */ s32 unk4;
+    /* 0x08 */ s32 unk8;
+    /* 0x0C */ s32 unkC;
+} cloudStructUnk;
```
JoshSchreuder, line 13: Wasn't sure of a good place to put these. Open to suggestions... /include is probably too generic?

Maybe I am over optimising here for duplicates and I should just copy these back into the ST0 and RBO3 cloud source files as duplicate code?

sozud: I think I'd probably think of it like this: if the two functions are sufficiently different that they shouldn't be recombined it's fine to consider the data to be separate too. It's not a big deal either way though

PR #2343, opened 2025-03-20
Clock Tower Documentation

hohle:
The clock tower uses triangle and quad drawing "scripts" and small interpreters to draw objects. It uses unique vertex sets and indexes into those sets to define the geometry of each of the clock tower features.

Macros have been added which make it straight forward to both read and create new versions of these. Variables have been updated to reflect their purpose.


bismurphy:
This is incredible. I really love when the code becomes more readable, in this case by a factor of 10 or more! This is a fantastic PR and I'll keep it in mind for the future as a great thing to point to for people who say "I want to help but I don't know assembly". Awesome work!

PR #2344, opened 2025-03-20
Small code cleanup

gamezter:
- Rename `pl_demo_timer` => `demo_timer`
- Rename `pl_vram_flag` => `vram_flag`
- Rename `pl_high_jump_timer` => `high_jump_timer`
- 0x8140 is double-wide space in shift-jis

PR #2347, opened 2025-03-21
Decompile CD and Memory Card functions for DRA PSP

bismurphy:
Pretty well-contained in their files. Lots of functions here are very different from their original versions, so no code sharing for this PR. Some functions are the same but there are enough differences to make it reasonable to have these as separate files, in my opinion.


src/dra_psp/13BD0.c
```diff
@@ -1,8 +1,136 @@
...
 INCLUDE_ASM("dra_psp/psp/dra_psp/13BD0", func_psp_090F0550);
 
-INCLUDE_ASM("dra_psp/psp/dra_psp/13BD0", func_psp_090F06D8);
+extern u8* D_8013794C;
+extern s32 D_80137950;
+extern s32 D_80137954;
+
+void func_800F99B8(char* str, s32 arg1, s32 arg2) {
+    // See src/st/blit_char.h
+    const u16 MINSCODE = 0x8140;
```
JoshSchreuder, line 13: There's some pending cleanup here you might want to bring in for consistency too
https://github.com/Xeeynamo/sotn-decomp/pull/2344/files#diff-26dbcbf3b7d37e52c5b69787fb846d3fea0f4323bfea30d35e18076e37fe0741R2315

bismurphy: Good call, done

sozud:
Needs a rebase

bismurphy:
Rebase successful!

PR #2350, opened 2025-03-22
Rename functions related to revealing secret passages

SestrenExsis:
There are a handful of functions that are related to drawing pixels on the map to indicate hidden passageways that the player has revealed (like the breakable wall and floor in Alchemy Lab). The array at `D_800A2BC0` lists out the possible points and what castle flags they pertain to.

There does not appear to be an entity that reacts to the presence or absence of the Hidden Stairs in Reverse Keep, so toggling this flag appears to do nothing. Nevertheless, the pairing of bytes in the `D_800A2BC0` array implies that castleFlags `0x94` and `0x9B` are related. The same goes for the castleFlags pertaining to the Elevator in Colosseum.

Also, I apologize for disabling the formatter around D_800A2BC0, but with it on it made the grouped relationship of the fields much harder to see.


include/castle_flags.h
```diff
@@ -102,9 +102,7 @@ typedef enum {
     /* 0x98 */ RTOP_SECRET_WALL_1_BROKEN = 152, // Lower right of large room
     /* 0x99 */ RTOP_SECRET_WALL_2_BROKEN = 153, // Upper right of large room
 
-    // Since 0x9B and 0xB9 are set at the same time, this goes by code order for
-    // what they likely are
-    /* 0x9B */ ST0_STAGE_FLAG = 155,
+    /* 0x9B */ RTOP_SECRET_STAIRS = 155,
```
sozud, line 105: @ProjectOblivion Could you comment on the flag name changes?

include/game.h
```diff
@@ -1607,7 +1615,7 @@ typedef struct {
     /* 8003C860 */ void (*func_800F53A4)(void);
     /* 8003C864 */ u32 (*CheckEquipmentItemCount)(u32 itemId, u32 equipType);
     /* 8003C868 */ void (*GetPlayerSensor)(Collider* col);
-    /* 8003C86C */ void (*func_800F1FC4)(s32 arg0);
+    /* 8003C86C */ void (*revealSecretPassageAtPlayerPositionOnMap)(s32 arg0);
```
sozud, line 1618: Capital R for coding standard

src/dra/5087C.c
```diff
@@ -769,7 +772,7 @@ void func_800F1B08(s32 x, s32 y, s32 arg2) {
     LoadTPage(bitmap, 0, 0, x + VramPosX, y * 4 + VramPosY, 8, 5);
 }
 
-void func_800F1D54(s32 x, s32 y, s32 arg2) {
+void drawSecretPassageOnMap(s32 x, s32 y, s32 direction) {
```
sozud, line 775: Capital D

src/dra/5087C.c
```diff
@@ -782,60 +785,62 @@ void func_800F1D54(s32 x, s32 y, s32 arg2) {
...
 }
 
-void func_800F1EB0(s32 playerX, s32 playerY, s32 arg2) {
-    s32 x;
-    s32 y;
-    s32 data_2;
-    s32 data_3;
-    s32 data_4;
-    u8* ptr;
+void revealSecretPassageOnMap(s32 playerMapX, s32 playerMapY, s32 flagId) {
```
sozud, line 803: R

include/castle_flags.h
ProjectOblivion: The original names were an educated guess based on the circumstances, but I think the logic here is sound.
I checked and it doesn't look like ST0 uses the stage popup (e_stage_name.h), so that also lines up.

src/dra/5087C.c
ProjectOblivion: Leading upper for functions, leading lower for vars, right?
I only mention it because there are two d's and my understanding is:
`DrawSecretPassageOnMap `and `direction`

sozud: I meant DrawSecretPassageOnMap https://github.com/Xeeynamo/sotn-decomp/blob/master/docs/STYLE.md#naming

include/castle_flags.h
SestrenExsis: As in the description above, I'm basing it on the fact that both flags are paired together in the D_800A2BC0 array, implying that one is simply the Reverse Castle version of the other.

src/dra/5087C.c
SestrenExsis: Thank you for catching the casing mismatch. I'm rusty, so I forgot all about parts of the naming convention.

include/castle_flags.h
ProjectOblivion: Not sure if I wasn't clear, or I'm misreading this comment @SestrenExsis .
I named them based on an educated guess, but I think your logic for renaming them is sound and I agree with it.

Something that the OVL_STAGE_FLAGS are used for, which isn't immediately obvious, is providing an offset to e_stage_name.c/.h.  I checked that and ST0 doesn't seem to use it, so there is no issue with renaming the flag from that perspective.

sozud (approved):

PR #2351, opened 2025-03-23
Decompile EntityAlucard for PSP, and all other functions in its file

bismurphy:
Nice to get this pulled in. I aligned the files between versions and decompiled everything in this file.

No code-sharing yet; there is a lot of data in this file and I'm having some trouble importing it. Down the line the data can be pulled in and the code back-ported (by me or someone else). For now, just getting the decompilation in.


bismurphy:
switching to draft while I try to figure out if we can make the data work. I think ultimately PS1 probably has some data in the wrong files and hopefully the PSP file order will reveal the issue.

bismurphy:
Well this one was a beast, but it works! Full codeshare!

PR #2352, opened 2025-03-23
Delete docs/STYLE.md

Xeeynamo:
Not needed any more. Styling has been consolidated and existing code can be used as a style reference. This guide was old and did not fully conform to our code base. Most of the styling is fixed by the formatter anyway.


sozud:
So are we just ignoring style now? Like no PR reviews like this? https://github.com/Xeeynamo/sotn-decomp/pull/2350#discussion_r2008896099

sozud (approved):
I think it makes sense to have a style guide but the wiki would be a better place for it. I think if there's no style guide at all then we shouldn't comment on style on PRs though since it's arbitrary. So I think the options are:

No style guide => PRs are a free-for-all
Style guide => enforced in PRs

ProjectOblivion:
Having some baseline style enforcement is important IMO, otherwise someone will need to do all that cleanup down the line.
There are things like function/variable/macro naming and bool over 1/0 that is fairly easy to decide and enforce.  I think requiring these things and continuing to require make format be run would go a long way toward styling without being opinion based.

The way I see it is that it's the subjective styling points that get really murky and don't really contribute much to overall code quality, so my suggestion is to trim down the style guide and instead call it "PR guidelines" or something like that.

I could put together an example of what I mean and submit a PR for it if you want me to.

PR #2354, opened 2025-03-23
Improve permute loader support for PSP overlays

Xeeynamo:
cc @ProjectOblivion 

### `isinstance(x, list)` fix

PSP splat configs have a `{type: bss, vram: 0x92D6600}` at the end. We need to filter them out.

### `subsegments[0]` fix

Since the subsegment path is matched with `x[-1].endswith(args.source)`, if `x[-1]` contains a path with directories involved then we cannot simply use `args.source`. This led me to issue where `permute ric 6DB0 RicMain` was trying to open `src/6DB0.c`. And `permute ric ric_psp/6DB0 RicMain` was trying to open `asm/pspeu/ric_psp/nonmatchings/ric_psp/ric_psp/6DB0/RicMain.s` (notice the double `ric_psp/` in the path).

### DIFF
--- a/tools/sotn_permuter/permuter_loader.py
+++ b/tools/sotn_permuter/permuter_loader.py
@@ -205,7 +205,7 @@ if __name__ == "__main__":
                             if isinstance(x, list)
                             and len(x) > 1
                             and x[1] == "c"
-                            and x[-1].endswith(args.source)
+                            and str(x[-1]).endswith(args.source)
                         ]
                     )
             if subsegments and len(subsegments) == 1:


ProjectOblivion:
Did you mean `import ric 6DB0 RicMain`?  What you are describing lines up with that, rather than `permute`

tools/sotn_permuter/permuter_loader.py
```diff
@@ -202,7 +202,8 @@ def sotn_finalize_compile_command(cmdline: importer.List[str]) -> str:
                         [
                             x[-1]
                             for x in segment["subsegments"]
-                            if len(x) > 1
+                            if isinstance(x, list)
```
ProjectOblivion, line 205: This is a good sanity check even though it shouldn't ever happen.  We all know how well "shouldn't ever happen" holds up.

tools/sotn_permuter/permuter_loader.py
```diff
@@ -226,7 +227,7 @@ def sotn_finalize_compile_command(cmdline: importer.List[str]) -> str:
             else:
                 srcPath = os.path.join(os.getcwd(), splatConfig["options"]["src_path"])
 
-            cFilename = os.path.join(srcPath, f"{args.source}.c")
+            cFilename = os.path.join(srcPath, f"{subsegments[0]}.c")
```
ProjectOblivion, line 230: This is not where the issue is and I have concerns about unintended effects, since it is slightly different logic then what I had in mind when I wrote it.  The issue is actually a couple lines up on lines 222-228.  I boneheadedly put the exact same srcPath string for both conditions, instead of the proper one for pspeu.  The right fix is to move that block up to line 213 and add in ssPath[0].

```
@@ -213,2 +214,9 @@ if __name__ == "__main__":
                     nonmatchingsPath = os.path.join(nonmatchingsPath, ssPath[0])
+                    srcPath = os.path.join(
+                        os.getcwd(), splatConfig["options"]["src_path"], ssPath[0]
+                    )
+                else:
+                    srcPath = os.path.join(
+                        os.getcwd(), splatConfig["options"]["src_path"]
+                    )
             elif len(subsegments) > 1:
@@ -220,10 +228,2 @@ if __name__ == "__main__":
 
-            if args.version == "pspeu":
-                srcPath = os.path.join(
-                    os.getcwd(),
-                    splatConfig["options"]["src_path"],
-                )
-            else:
-                srcPath = os.path.join(os.getcwd(), splatConfig["options"]["src_path"])
-
             cFilename = os.path.join(srcPath, f"{args.source}.c")
```

ProjectOblivion (changes requested):

tools/sotn_permuter/permuter_loader.py
```diff
@@ -226,7 +227,7 @@ def sotn_finalize_compile_command(cmdline: importer.List[str]) -> str:
...
             sys.argv = [cFilename if a == args.source else a for a in sys.argv]
```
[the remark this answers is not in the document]
ProjectOblivion: This is important because it replaces the shortened filename with the long filename to send on to decomp-permuter when using -p.

ProjectOblivion (approved):

sozud (approved):

PR #2360, opened 2025-03-24
Fix a couple of small permuter issues related to sotn_str changes

JoshSchreuder:


tools/sotn_permuter/permuter_settings.us.toml
```diff
@@ -1,7 +1,7 @@
...
+compiler_command = "mipsel-linux-gnu-cpp -Iinclude -Iinclude/psxsdk -undef -Wall -fno-builtin -Dmips -D__GNUC__=2 -D__OPTIMIZE__ -D__mips__ -D__mips -Dpsx -D__psx__ -D__psx -D_PSYQ -D__EXTENSIONS__ -D_MIPSEL -D_LANGUAGE_C -DLANGUAGE_C -DNO_LOGS -DHACKS -DUSE_INCLUDE_ASM -D_internal_version_us -DSOTN_STR -lang-c | ./tools/sotn_str/target/release/sotn_str process | iconv --from-code=UTF-8 --to-code=Shift-JIS | ./bin/cc1-psx-26 -G0 -w -O2 -funsigned-char -fpeephole -ffunction-cse -fpcc-struct-return -fcommon -fverbose-asm -msoft-float -g -quiet -mcpu=3000 -fgnu-linker -mgas -gcoff | python3 tools/maspsx/maspsx.py --expand-div --aspsx-version=2.34 | mipsel-linux-gnu-as -Iinclude -march=r3000 -mtune=r3000 -no-pad-sections -O1 -G0"
```
JoshSchreuder, line 2: @ProjectOblivion the same needs to be done on the PSP settings but I didn't feel super confident working out the piping / new command line to mwccgap. Is that something you are able to fix up? Otherwise I can keep playing around with it

ProjectOblivion: Permuter doesn't use sotn_str for PSP.  I didn't add it in when mwcc was updated because I didn't think anybody would be permuting before importing data.
It should be just changing the beginning of the string to:
`tools/sotn_str/target/release/sotn_str -f $INPUT "python3 tools/mwccgap/mwccgap.py $OUTPUT --src-dir $(dirname "${OUTPUT}") --mwcc-path`

I didn't test that out, so feel free reach out on discord if it doesn't work

JoshSchreuder: Yeah fair, it is probably not worth the overhead anyway, we can add later if required. I am going to add the no-intrinsics parameter though

ProjectOblivion (approved):

PR #2373, opened 2025-03-27
Cleaned up Makefile utility targets

ProjectOblivion:
I will be doing a cleanup/review pass after the 3 primary makefiles (Makefile, psx, and psp) PR's get merged to look at all of the changes in context.  I expect that there will be a few non-functional loose ends to tie up when they've all been put back together.


Makefile.psx.mk
```diff
@@ -35,6 +35,8 @@ CC1PSX          := ./bin/cc1-psx-26
 CC              := $(CC1PSX)
 AS              := $(CROSS)as
 CPP             := $(CROSS)cpp
+SOTNSTR			:= $(SOTNSTR_APP) process
```
ProjectOblivion, line 38: This will be removed and simply written inline in my Makefile.psp.mk PR.

ProjectOblivion:
How about I just remove function finder and duplicates report from the Makefile entirely, since they are neither build related nor intended to be run as a part of the build process?

ProjectOblivion:
I have similar feelings about some of the other root directories that you mentioned about tools/
Currently, there are the following directories that are defined as variables that I don't foresee changing.  It doesn't matter to me whether they are variables or hardcoded as long as they are consistent throughout (previously, sometimes the var was used and sometimes hardcoded).  It is quite redundant to have a variable that is literally `<name.upper>_DIR`.
BIN_DIR := bin
SRC_DIR := src
INCLUDE_DIR := include
ASSETS_DIR := assets
TOOLS_DIR := tools

Makefile.psp.mk
ProjectOblivion, line 1: Changes to this file were only made to allow it to be properly and are ultimately expected to be superseded by #2365 .

Makefile.psx.mk
ProjectOblivion, line 1: Changes to this file were only made to allow it to build properly and are ultimately expected to be superseded by a Makefile.psx.mk specific PR.

Makefile
```diff
@@ -114,521 +105,312 @@ endef
...
+disk-prepare: build $(SOTNDISK)
+	mkdir -p $(BUILD_DISK_DIR); cp -r $(EXTRACTED_DISK_DIR)/* $(BUILD_DISK_DIR)
+	cp $(BUILD_DIR)/main.exe $(BUILD_DISK_DIR)/SLUS_000.67
+	$(call echo,cp $(BUILD_DIR)/*.BIN $(BUILD_DISK_DIR)/) $(foreach item,$(DISK_PREPARE_FILES),cp $(BUILD_DIR)/$(notdir $(item)) $(BUILD_DISK_DIR)/$(item);)
+disk-debug: disk-prepare
+	cd tools/sotn-debugmodule && $(MAKE)
+	cp $(BUILD_DIR:$(VERSION)=)/sotn-debugmodule.bin $(BUILD_DISK_DIR)/SERVANT/TT_000.BIN
+	$(SOTNDISK) make $(BUILD_DIR:$(VERSION)=)/sotn.$(VERSION).cue $(BUILD_DISK_DIR) $(CONFIG_DIR)/disk.$(VERSION).lba
+disk: disk-prepare
+	$(SOTNDISK) make $(BUILD_DIR:/$(VERSION)=)/sotn.$(VERSION).cue $(BUILD_DISK_DIR) $(CONFIG_DIR)/disk.$(VERSION).lba
```
ProjectOblivion, line 241: I redid the disk targets because it was already unwieldy and it will only get worse as more overlays are added.  This has main, dra, ric, and sel hardcoded because there was no benefit to dynamically building those.  It uses the enabled build targets to add all of the other overlays.

I did just realize that the two weapon .BIN files are not accounted for in here (they aren't in master either).

Makefile
```diff
@@ -114,521 +105,312 @@ endef
...
+            --datafile sotn.$*.bin \
+            --device "$$DEVICE" \
+            --driver generic-mmc-raw \
+            sotn.$*.toc && \
+        toc2cue sotn.$*.toc sotn.$*.cue && \
+        rm sotn.$*.toc
+$(addprefix dump-disk,.eu .hk .jp10 .jp11 .saturn .us .usproto): disks/sotn.$(VERSION).cue
+dump-disk.%: PHONY# Last resort
+	$(error Automated dumping of $* is not supported)
+dump-disk: dump-disk.$(VERSION)
```
ProjectOblivion, line 260: I made no functional changes here, only reworked the target logic so that it is more version agnostic and will return the same message for any version that is not matched by a specific target.

Makefile
```diff
@@ -114,521 +105,312 @@ endef
...
+	$(PIP) install -r tools/requirements-python.txt && touch $@
+python-dependencies: tools/python-dependencies.make.chkpt
 
-define list_o_files_psp
-	$(foreach file,$(call list_src_files_psp,$(1)),$(BUILD_DIR)/$(file).o)
-endef
-### End old header ###
+tools/graphviz.make.chkpt: tools/python-dependencies.make.chkpt
+	sudo apt update && sudo apt-get install -y graphviz && touch $@
+graphviz: tools/graphviz.make.chkpt
```
ProjectOblivion, line 317: I didn't really do anything major to the dependency targets, only cleaned up and reworked logic so that they should all now be based on a real file so it will only update dependencies that had had their prerequisites change.  Compiled programs that are a part of the repo should all have their source files as prerequisites.

sotndisk and sotnassets were moved from being installed in `home/go/bin/` to being created in their directory to keep sotn-decomp contained in the repo folder and also because make didn't seem to properly rebuild them on prereq change in the home directory.

Added a few targets to consolidate the dependency related things together.  In combination with the relevant prereqs for targets, it should be possible to simply install os dependencies and make (needs vetting, though) with one caveat; I haven't gotten `make build` to trigger `make extract` correctly yet.

I removed m2ctx because it is already in the repo, so it doesn't make any sense to download it.  Additionally, what is downloaded is different than what is in the repo so if it should be in Makefile, then it should be removed from the repo and vice versa.
 
If you're okay with function finder being removed from Makefile, then graphviz and it's chkpt target should also be removed.

Makefile
```diff
@@ -114,521 +105,312 @@ endef
...
 
-ifneq (,$(filter $(VERSION),us hd)) # Both us and hd versions use the PSX platform
+$(GO):
+	curl -sSfL -O https://go.dev/dl/go1.22.4.$(OS)-$(ARCH).tar.gz
+	$(muffle)tar -C $(HOME) -xzf go1.22.4.$(OS)-$(ARCH).tar.gz; rm go1.22.4.$(OS)-$(ARCH).tar.gz
```
ProjectOblivion, line 321: Doesn't really matter overall, but since I was already cleaning up the web download strings I added download of the correct go file for the system.

Makefile
```diff
@@ -114,521 +105,312 @@ endef
...
+		else if($$0 ~ /^#/) \
+			printf "\n%s\n", $$2; \
+		else if($$1 == "") \
+			printf "     %-20s%s\n", "", $$2; \
+		else { \
+            system("tput setaf 4; printf \"\n    %-20s \" " $$1 "; tput sgr0"); \
+            printf "%s\n", $$2; \
+        }; \
+	}'
+
```
ProjectOblivion, line 355: No functional change here, just moved things so that all of the help text is in one location.  While working through all of this, mainly at the beginning, I felt as though having the help text interspersed throughout the targets was more hindrance than help.  I didn't review to see if anything needed to be added, so there are probably some things that need to be adjusted with the help text.

Makefile
```diff
@@ -114,521 +105,312 @@ endef
...
-get-debug: get-phony get-silent
-get-phony:
-	echo ".PHONY:" > make.phony.targets
-	$(foreach target,$(PHONY_TARGETS),echo $(target) >> make.phony.targets;)
-get-silent:
-	echo ".SILENT:" > make.silent.targets
-	$(foreach target,$(MUFFLED_TARGETS),echo $(target) >> make.silent.targets;)
+MUFFLED_TARGETS += $(PHONY_TARGETS) $(MASPSX_APP) $(MWCCGAP_APP) $(MWCCPSP) $(SATURN_SPLITTER_DIR) $(SATURN_SPLITTER_APP) $(EXTRACTED_DISK_DIR) $(ASMDIFFER_APP) $(PERMUTER_APP) $(dir $(M2C_APP)) $(M2C_APP)
+MUFFLED_TARGETS += $(DOSEMU_DIR) tools/dosemu.make.chkpt tools/python-dependencies.make.chkpt tools/graphviz.make.chkpt $(SOTNDISK) $(SOTNASSETS) $(VENV_DIR) $(VENV_DIR)
+.PHONY: $(PHONY_TARGETS) expected clean force-symbols disk disk-prepare disk-debug
```
ProjectOblivion, line 400: I began looking at targets to see where I can trim all of this down.  I was able to move some targets directly to .PHONY and clean it up a bit, but I only did it in a few select places for now.

Makefile
```diff
@@ -114,521 +105,312 @@ endef
...
+force_symbols: force-symbols
+	$(info make: warning: force_symbols has been renamed to force-symbols and will be removed in the future.)
+extract_disk: extract-disk
+	$(info make: warning: extract_disk has been renamed to extract-disk and will be removed in the future.)
+disk_prepare: disk-prepare
+	$(info make: warning: disk_prepare has been renamed to disk-prepare and will be removed in the future.)
+disk_debug: disk-debug
+	$(info make: warning: disk_debug has been renamed to disk-debug and will be removed in the future.)
+dump_disk: dump-disk
+	$(info make: warning: dump_disk has been renamed to dump-disk and will be removed in the future.)
```
ProjectOblivion, line 416: I renamed these targets so that all target naming was consistent with `-` as a word break.

Xeeynamo:
> I have similar feelings about some of the other root directories that you mentioned about tools/ Currently, there are also the following directories that are defined as variables that I don't foresee changing either. It doesn't matter to me whether they are variables or hardcoded as long as they are consistent throughout (previously, sometimes the var was used and sometimes hardcoded). Although, it is quite redundant to have a variable that is literally named `<value.upper>_DIR`. BIN_DIR := bin SRC_DIR := src INCLUDE_DIR := include ASSETS_DIR := assets

I agree with everything but for `ASSETS_DIR`. I have the feeling in the future we might need to have `ASSETS_DIR := assets/$(VERSION)` to cross-build between versions successfully.

PR #2379, opened 2025-03-28
RFC: File listing "module" for build tool

ProjectOblivion:
Something I put together for moving the inline python out of Makefile.  The changes to the makefiles are only to allow it to build correctly with the script in place.  This PR is primarily to get feedback on what I have so far.

Instead of trying to micromanage directory listings I've instead just built the files directly from the splat configs.

I leaned toward the side of overcommenting on the expectation that the reader may not be strong in Python or programming in general.  My aim was to make it performant, but still clean and maintainable.

Any comments or criticism are welcome, even if it is subjective or related to my programming itself.  What I'm looking for out of this PR is essentially peer review of where my head is at as far as expanding building beyond the walls of make.  Python is what I know, so it was the smart place to start, but I don't have any plans as to whether I will continue doing it in Python.


tools/sotn_build/get_file_list.py
```diff
@@ -0,0 +1,311 @@
...
+# This function still needs a bit of refinement, but it is serviceable for now
+def GetAssetFiles(
+    options: dict, segments: list[list], assetFilters: tuple[tuple]
+) -> set[str]:
+    """Constructs a list of asset files using a directory listing or the segments"""
+    assetPath = options["asset_path"].rstrip("/")
+
+    # This conditional needs to be cleaned up and probably relocated for easier code maintenance
+    if (
+        options["platform"] == "psx"
```
JoshSchreuder, line 174: Should this be `us`?

tools/sotn_build/get_file_list.py
```diff
@@ -0,0 +1,311 @@
...
+        choices=("src_files", "o_files", "merged"),
+        help="What to list",
+    )
+    parserList.add_argument(
+        "-c",
+        "--config-type",
+        type=str,
+        choices=list(CONFIG_PATTERNS.keys()),
+        required=True,
+        help="Type of config to reference",
```
JoshSchreuder, line 247: Am I misunderstanding, or is this only `splat` currently?
What would be the future use of this parameter?

tools/sotn_build/get_file_list.py
```diff
@@ -0,0 +1,311 @@
...
+        "ci4",
+        "palette",
+        "raw",
+        "assets",
+        "bin",
+    ),
+}
+
+
+# This cannot handle returning only shared src files, but since there aren't any currently it isn't really worth putting time into it.
```
JoshSchreuder, line 34: I'm not clear on this comment, is this some sort of antiquity that is no longer used in the codebase but was present in the Makefile so it's made its way here too?

From my understanding all the shared src files are coming via .h imports from overlay specific .c files?

tools/sotn_build/get_file_list.py
```diff
@@ -0,0 +1,311 @@
...
+
+    assetFilters: tuple = tuple(
+        tuple(aFilter.split("*")) for aFilter in NormalizeArgs(args.asset_filter)
+    )
+    fileSegments: list = GetFileSegments(config["segments"])
+    asmFiles: set = GetAsmFiles(config["options"], fileSegments)
+    srcFiles: set = GetSrcFiles(config["options"], fileSegments)
+    assetFiles: set = GetAssetFiles(config["options"], fileSegments, assetFilters)
+
+    # These are currently sorted for ease of comparison/debugging, but it isn't relevant for actual function
```
JoshSchreuder, line 289: Just for performance reasons it might be worth removing this sorting before merging since it will have some small overhead (or perhaps make sorted output an option for debugging purposes).

tools/sotn_build/get_file_list.py
```diff
@@ -0,0 +1,311 @@
...
+        dataSegTypes = SEG_TYPES["data"] + tuple(
+            f".{type}" for type in SEG_TYPES["data"]
+        )
+
+    dataFiles: set = {
+        FILENAME_TEMPLATE.substitute(
+            path=dataPath,
+            # If a segment has 3 or more elements, then it has a name
+            # If it only has 1 or 2 elements, we use the memory address as the name
+            filename=seg[2] if len(seg) >= 3 else f"{seg[0]:X}",
```
JoshSchreuder, line 119: My main concern with this approach would be that it might be somewhat brittle to changes on the splat side

Noting that I am raising a problem and not a solution here 😄

tools/sotn_build/get_file_list.py
ProjectOblivion: This applies to both us and hd (they both use psx as platform in the splat configs). Really just a different way of saying "not psp".

tools/sotn_build/get_file_list.py
ProjectOblivion: You are correct on both accounts.  I had already written handling of shared source files before I realized there were none.

ProjectOblivion:
I agree that there is a lot that can be done to clean up and/or improve the options.
It is primarily because of how it evolved through the course of writing it and I'm still not entirely sure what this is trying to be.  That's a big reason for getting some outside feedback to fine tune it a bit.

As for feeding back into splat, I would love to contribute back to it whenever possible.

tools/sotn_build/get_file_list.py
```diff
@@ -0,0 +1,293 @@
...
+
+    return parser.parse_args(args)
+
+
+# This logic is expected to be contained elsewhere if this file is imported as a module
+def main(args: list) -> None:
+    args = parse_args(args)
+    config = get_config(args)
+
+    file_segments = get_file_segments(config["segments"])
```
sozud, line 256: Shouldn't we only do these if the arguments actually call for it? Right now we do the filesystem access then throw away the result depending

sozud (commented):
The maspsx stuff should be removed from this probably but otherwise it seems like an improvement to me to trim down the amount of makefile stuff

tools/sotn_build/get_file_list.py
ProjectOblivion: The subsegments we use are included under the actual segments section of the config, so I just referenced them all as segments.  This code doesn't gather any files from the file system, except for assets.  Assets only look at the filesystem because I haven't yet mapped how some of them go from the splat config to file names.

ProjectOblivion: Yeah, that's more or less where I was at as well.  I understand how this plays into the building, but I don't know if it is aligned with the future direction of the project.

PR #2381, opened 2025-03-28
Doppleganger State Machine Decomp

hohle:
Decompiles one of Doppleganger 10's main state machines. This controls his primary key pad input based on the player distance and state.

While I was here I also updated enums for setting Alucard status to help identify various parts in this function. There's also some cleanup that I forgot to push before the previous `BO4` change.


sozud (approved):

sozud:
Conflicted

PR #2386, opened 2025-03-30
Import `no4` blit_char, collision and e_misc_2

gamezter:


src/st/entity_relic_orb.h
```diff
@@ -263,26 +253,26 @@ void EntityRelicOrb(Entity* self) {
         switch (D_8B42058) {
         case 1:
         default:
-            obtainedStr = obtained_ENG;
+            obtainedStr = "Obtained";
             break;
         case 2:
-            obtainedStr = obtained_FR;
+            obtainedStr = "Obtenu \xB1 ";
```
sozud, line 259: I think B1 is supposed to be mapped to : in sotn_str https://youtu.be/28CeeiJ6Its?list=PLkP_w744i1iRQVwxOe4h5AyRCeIdWROTd&t=1035

gamezter: I can't find a char to replace `/xB1`. I'd expect it to be `ｱ` but that causes a mismatch, and sotn_str won't compile if I use `\xB1`.

PR #2391, opened 2025-03-31
Empty `asm/nonmatchings` folder for fully decompiled overlays

gamezter:
`migrate_rodata_to_functions` will put rodata in asm/nonmatchings folder even if that data is imported. This just tidies up overlays that have been fully decompiled. Depends on https://github.com/Xeeynamo/sotn-decomp/pull/2389 since that is the last `lib` function to be decompiled


sozud:
> migrate_rodata_to_functions will put rodata in asm/nonmatchings folder even if that data is imported. 
This seems like a splat bug? Maybe we should try filing an issue

gamezter:
Made my first github issue : https://github.com/ethteck/splat/issues/454

PR #2398, opened 2025-04-02
Decompile no4 func_us_801D8D44

amelenty:


src/st/no4/unk_55678.c
```diff
@@ -25,7 +27,34 @@ INCLUDE_ASM("st/no4/nonmatchings/unk_55678", func_us_801D85A4);
 
 INCLUDE_ASM("st/no4/nonmatchings/unk_55678", func_us_801D8654);
 
-INCLUDE_ASM("st/no4/nonmatchings/unk_55678", func_us_801D8D44);
+// #ifndef NON_MATCHING
```
sozud, line 30: Could we clean up the commented-out parts? also the #endif below

src/st/no4/e_init.c
```diff
@@ -207,7 +207,7 @@ EInit D_us_80180C04 = {ANIMSET_OVL(0x03), 0x01, 0x4B, 0x204, 0x06A};
 EInit D_us_80180C10 = {ANIMSET_OVL(0x03), 0x01, 0x4B, 0x204, 0x06B};
 EInit D_us_80180C1C = {ANIMSET_OVL(0x04), 0x01, 0x4A, 0x206, 0x115};
 EInit D_us_80180C28 = {ANIMSET_OVL(0x04), 0x13, 0x4A, 0x206, 0x005};
-EInit D_us_80180C34 = {ANIMSET_OVL(0x04), 0x00, 0x4A, 0x206, 0x116};
+EInit g_EInitSkeleton = {ANIMSET_OVL(0x04), 0x00, 0x4A, 0x206, 0x116};
```
sozud, line 210: How did you come up with this rename?

amelenty: (TL;DR looks like I was wrong and I don't have enough information to figure out which entity the code refers to.)

Now that I'm thinking about this, I'm not sure this is correct, but this was my original thought process:
- `func_us_801D8D44` is a non-duplicate function in `no4`, so this must be something unique to this stage.
- the function's logic is related to explosions in some way (because of `self->pfnUpdate = EntityExplosion`)
- the one thing I could find that is related to explosions and unique to Underground Caverns (using [this ](https://castlevania.fandom.com/wiki/Underground_Caverns) as my reference) is Skeleton Apes throwing exploding barrels.
- the code of this function is similar to other code related to initializing enemy entities (I was looking at `EntityAxeKnightThrowingAxe` from `entity_axe_knight.h` for comparison).
- when I was trying to figure out what `D_us_80182780`, a different reference present in the function, meant, I ran `grep -r D_us_80182780` (couldn't find anything in VSCode), and among other things I got this:
```
asm/us/st/no1/matchings/e_skeleton/func_us_801CD670.s:    /* 4DA6C 801CDA6C 1880013C */  lui        $at, %hi(D_us_80182780)
asm/us/st/no1/matchings/e_skeleton/func_us_801CD670.s:    /* 4DA74 801CDA74 80272290 */  lbu        $v0, %lo(D_us_80182780)($at)
asm/us/st/no1/data/e_skeleton.data.s:glabel D_us_80182780
asm/us/st/no1/data/e_skeleton.data.s:.size D_us_80182780, . - D_us_80182780
```
Which made me think this was related to skeletons, which is why I used `g_EInitSkeleton `.

Now that I'm looking at this again, I'm coming to a different conclusion:
- the reference code I was looking at was not for the knights, but specifically for their throwing axes, and it uses `g_EInitAxeKnightAxe` from `no0/e_init.c`.
- the code of the function I'm decompiling would make sense for the exploding barrels that are thrown by skeleton apes. It doesn't handle explosions directly, but it sets `self->pfnUpdate = EntityExplosion;` when the entity is destroyed, so presumably the explosion is handled at a later time (`func_8011A870` from `src/dra/7879C.c` looks like it could be relevant here).
- it's plausible that thrown barrel animation data would be stored together with the skeleton apes, so the reference to `e_skeleton.data.s` still makes sense. `no1/e_init.c` contains functions for both `EntitySkeletonApe` and `EntitySkeletonApeBarrel`, so it makes sense that data from a different stage is reused.
- I'm now realizing that skeleton apes throwing barrels isn't unique to `no4` (as there is an `e_skeleton_ape.c` in `no1`). It doesn't seem very plausible that skeleton ape / exploding barrel implementation would be put in `no4` given it's already implemented in `no1`.

My conclusion here is that I don't have enough data yet to identify which entity this code is referring to. I suggest changing `g_EInitSkeleton` to `g_EInitUnkId<X>` (given there exists a `g_EInitUnkId13`). Does this make sense to you? Is there a convention for what number I should pick for X?

ProjectOblivion: The number used for naming entity ID's is related to their offset in the data structure.  Looking at `EntityUpdates` (line 116, e_init.c) `g_EInitUnkId13` actually seems to be off by one and should be `g_EInitUnkId12`.  Though this is only a naming convenience until somebody figures out what it actually is, so the off by one naming is largely irrelevant.

Going by the memory location (`80180C34`), it doesn't look like this particular symbol should be in `EntityUpdates`, though.  Looking at the code, this looks like it is more likely something called by it's "main" entity, which would be in `EntityUpdates`

_My ever present caveat of "I have no idea what I'm talking about compared the other reviewers" and while I do my best, anything I say about decompiled code could be incorrect or inaccurate._

gamezter: The last element of an EInit struct is an index into `g_EnemyDefs` (src/config_us.h). This is most likely related to the Fishhead enemy : 
image

amelenty: Thanks a lot for the context!

Based on this discussion, I have renamed the following:
- `func_us_801D8D44` -> `EntityUnkId3F` (given that it's at offset `0x3F` in `no4/e_init.c`
- `D_us_80180C34`/`g_EInitSkeleton` -> `g_EInitUnkId3F`

I'd rather not increase the scope of this PR; I'll try to investigate the relation to "Fishhead" next and hopefully submit my findings in a separate PR.

Please let me know if there are any further issues I should address!

PR #2399, opened 2025-04-03
Import some `lib` assets

gamezter:
Import `sprite_banks`, `entity_layouts` and `rooms`. I had issues importing the other asset categories, but this is a good start


Xeeynamo (approved):
Looks good, cheers

PR #2402, opened 2025-04-04
Fix decompile.py failing on duplicate C filenames

amelenty:
Check that the C path is taken from the relevant overlay.

Before:
```
$ dec func_us_801C9764
Traceback (most recent call last):
  File "/<repo_path>/sotn-decomp/./tools/decompile.py", line 334, in <module>
    decompile(args.function, args.number_occurrence, args.force)
  File "/<repo_path>/sotn-decomp/./tools/decompile.py", line 269, in decompile
    funcs = get_nonmatching_functions(asm_dir, func_name)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/<repo_path>/sotn-decomp/./tools/decompile.py", line 67, in get_nonmatching_functions
    function = NonMatchingFunc(full_path)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/<repo_path>/sotn-decomp/./tools/decompile.py", line 56, in __init__
    assert len(c_paths) == 1
           ^^^^^^^^^^^^^^^^^
AssertionError
```

After:
```
$ dec func_us_801C9764
In file included from include/stage.h:5,
                 from src/st/no4/no4.h:2,
                 from src/st/no4/first_c_file.c:2:
include/common.h:23:2: warning: #warning "Version not specified. Falling back to the US version." [-Wcpp]
   23 | #warning "Version not specified. Falling back to the US version."
      |  ^~~~~~~
function 'func_us_801C9764' decompiled but cannot be compiled
python3 tools/asm-differ/diff.py -mwo --overlay st/no4 func_us_801C9764
```

### DIFF
--- a/tools/decompile.py
+++ b/tools/decompile.py
@@ -52,7 +52,7 @@ class NonMatchingFunc(object):
         self.text_offset = split[split.index("nonmatchings") + 1]
 
         assumed_path = "/".join(split[split.index("nonmatchings") + 1 : -1]) + ".c"
-        c_paths = [src for src in src_files if src.endswith(assumed_path)]
+        c_paths = [src for src in src_files if src.endswith(assumed_path) and self.overlay_name in src.split("/")]
         assert len(c_paths) == 1
         self.src_path = c_paths[0]
 
--- a/tools/decompile.py
+++ b/tools/decompile.py
@@ -52,7 +52,11 @@ class NonMatchingFunc(object):
         self.text_offset = split[split.index("nonmatchings") + 1]
 
         assumed_path = "/".join(split[split.index("nonmatchings") + 1 : -1]) + ".c"
-        c_paths = [src for src in src_files if src.endswith(assumed_path) and self.overlay_name in src.split("/")]
+        c_paths = [
+            src
+            for src in src_files
+            if src.endswith(assumed_path) and self.overlay_name in src.split("/")
+        ]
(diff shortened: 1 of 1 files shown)


ProjectOblivion:
Make sure that you run make format before every PR, it will save you a little bit of hassle.

tools/decompile.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, nonmatching_path):
         self.text_offset = split[split.index("nonmatchings") + 1]
 
         assumed_path = "/".join(split[split.index("nonmatchings") + 1 : -1]) + ".c"
-        c_paths = [src for src in src_files if src.endswith(assumed_path)]
+        c_paths = [src for src in src_files if src.endswith(assumed_path) and self.overlay_name in src.split("/")]
```
ProjectOblivion, line 55: While this seems on the surface as though it would fix the issue, I'm not certain that it will behave in this manner in all circumstances.  My primary concern is that this could change an explicit error into a silently skipped condition and lead to ambiguous behavior.

I can't point to a specific instance where this would be the case, so I will leave it up to the judgement of the maintainers.

Xeeynamo (approved):
Thanks for taking the initiative and fix this. There are a lot of assumptions made with this tool anyway. I am happy to keep iterating on it as we find new issues along the way.

PR #2403, opened 2025-04-05
Decompile several functions of PSP DRA menu

bismurphy:
This file is truly gigantic, definitely not going to try to do it all in one PR.

I started with just doing MenuHandle, but then there were a whole bunch of other functions that I had to get working first before I could get that one to come in, due to weird stuff with data.

But now they're all matching, and we have a big batch of new symbols to use which should be helpful all around.

I might continue with this or move onto something else for a bit, these functions are really annoying with the O4.


bismurphy:
Not sure what I'm doing wrong where the linter wants to remove some symbols from NO4, been needing to fix this manually for each of my PRs recently. Kind of weird.

Xeeynamo (approved):
Cheers!

PR #2404, opened 2025-04-05
Dedupe `no4` e_spear_guard

gamezter:
Slightly different from spear guard in no1. `func_us_801C82B8` in the middle of `first_c_file` seems to be similar to `EntitySpearGuard` and spawns a `EntitySpearGuardBlock`


src/st/e_spear_guard.h
```diff
@@ -0,0 +1,454 @@
...
+    self->hitboxOffX = guardHitboxOffsets[animCurFrame].x;
+    self->hitboxOffY = guardHitboxOffsets[animCurFrame].y;
+    self->hitboxWidth = guardHitboxSizes[animCurFrame].x;
+    self->hitboxHeight = guardHitboxSizes[animCurFrame].y;
+    self->facingLeft = parent->facingLeft;
+    self->posX.i.hi = parent->posX.i.hi;
+    self->posY.i.hi = parent->posY.i.hi;
+#ifdef STAGE_IS_NO4
+    if (parent->entityId != E_SPEAR_GUARD &&
+        parent->entityId != E_UNK_NO4_SPEAR_GUARD) {
```
gamezter, line 404: This is the only difference with no1, This entityID is temporary, I thought it would be better than `E_ID_3A`

JoshSchreuder: Looking at the entity func (`func_us_801C82B8`) it seems like this is an alternate spear guard entity (has the same entity ID in EInit) that has some alternate behaviour for `g_CastleFlags[NO4_WATER_BLOCKED]`

JoshSchreuder: Looking at some playthrough my guess is it's the drowning version of the Spear Guard, see https://youtu.be/mQtTt4pNho0?t=290 where the water gets blocked

sozud (approved):

PR #2405, opened 2025-04-05
Clean up and trim undefined symbols file

JoshSchreuder:
Went through and cleaned up the `undefined_syms` files for PSP, US and HD, removing a bunch which were no longer required, and moving others into (mostly) the `symbols.[version].txt` file

As part of this I also discovered a few PSP symbols which I've mapped, and also added auto-sorting to the undefined syms file as part of the normal `make format`


src/dra/4A538.c
```diff
@@ -615,13 +615,14 @@ void LoadEquipIcon(s32 equipIcon, s32 palette, s32 index) {
 
     if (D_80137538[index] != palette) {
         for (i = 0; i < 16; i++) {
-            if (D_800705CC) { // FAKE
+            if (g_Clut[0x1D00]) { // FAKE
```
Xeeynamo, line 618:
```suggestion
            if (g_Clut[0x1D00]) {
```
let's remove the `FAKE` label

src/dra/loading.c
```diff
@@ -735,16 +734,16 @@ void HandleEnding(void) {
...
+        LoadClut((u_long*)g_Clut, 0x200, 0xF0);
         g_GameStep++;
         break;
     case 2:
         if (g_UseDisk) {
             g_CdStep = CdStep_LoadInit;
             g_LoadFile = CdFile_StagePrg;
         }
-        g_GameStep = 3;
+        g_GameStep++;
```
Xeeynamo, line 746: This is an interesting discovery! We did not know GCC would optimise the `++` in such way

Xeeynamo (approved):
This is a great clean-up! Thank you for doing so. There is one single `// FAKE` to remove as I believe it might cause confusion if we keep it there. But I'll be very happy to merge this right after

src/dra/loading.c
JoshSchreuder: Yep, got this from [the PSP side](https://decomp.me/scratch/9QZjQ) which I was using to try to work out some mismatches (though I did not fully solve it there)

src/dra/4A538.c
JoshSchreuder: This was bugging me too actually, so I was looking into the PSP side about how to solve. I have worked out how to get rid of this FAKE and have updated the PR

PR #2407, opened 2025-04-05
RIC PSP: match some subweapon functions

Xeeynamo:
This took me a while. It matches around 25% of the whole RIC overlay on PSP and it is merged with PSX. There are few differences between PSP and PSX, but matching between the two required a lot of unexpected code massaging.

The biggest news is that `u16 hitboxOffX;` is now gone from `struct Entity`.


src/ric/319C4.c
```diff
@@ -1208,55 +1236,56 @@ void RicEntityAguneaLightning(Entity* self) {
...
         }
         prim = &g_PrimBuf[self->primIndex];
-        temp_s0 = self->params / 0x100 * 0x200;
-        sp20 = (temp_s0)-0x100 + rand() % 0x200;
-        randomSeed = rand();
+        sp20 = ((self->params & 0xFF00) >> 8) * 0x200;
+        sp20 += rand() % 0x200 - 0x100;
+        randomSeed = rand()
+#if defined(VERSION_PSP)
+                     & 0x7FFF
```
sozud, line 1261: There's an existing workaround for this https://github.com/Xeeynamo/sotn-decomp/blob/8913913c89006f4d091d1691b6b7fbd77e2bc518/include/game.h#L27

sozud (approved):
Can't tell what the diff is for src/ric/2C4C4.c but seems good

Xeeynamo:
> Can't tell what the diff is for src/ric/2C4C4.c

Unfortunately from https://github.com/Xeeynamo/sotn-decomp/pull/2407/files would be impossible to tell. I have one commit for every matched PSP function, and I always port it back to PSX. To check the differences with `2C4C4.c` you would need to check the commits from `RicEntitySubwpnReboundStone` to `RicEntityCrashReboundStoneParticles`. Let me know if you find anything odd, otherwise I am happy for you to merge it!

PR #2410, opened 2025-04-05
Rename animFrameIdx to pose and animFrameDuration to pl_ptimer

Xeeynamo:
I found multiple `FntPrint("pl_pose:%02x\n", PLAYER.animFrameIdx);` scattered around the various overlays, and we decided to strip the `pl_` prefix as it is probably only intended to be used for `g_Entities[0]`. Same for `animFrameDuration`, where `ptimer` was found from `FntPrint("pl_ptimer = %d\n", PLAYER.animFrameDuration)` on PSP. I assume it is a short for `poseTimer`.

I think `pose` and `ptimer` make perfect sense in this context. In an entity animation, each frame represents a pose and each pose has a time to live.

I decided to not rename anything else that uses `pose` and `ptimer` as I did not want the PR to be a scope-creep and hard to review.


sozud:
In my opinion our names are more clear but if others like the original names that's fine. A lot of the original names are unclear abbreviations or nonsensical in English like `hkyori_search` vs. `GetDistanceToPlayerX`

Xeeynamo:
> In my opinion our names are more clear but if others like the original names that's fine. A lot of the original names are unclear abbreviations or nonsensical in English like `hkyori_search` vs. `GetDistanceToPlayerX`

I'd leave `pose`, but I am open to rename `ptimer` to `poseTimer`. `hkyori_search` is a bit of an extreme scenario as it uses a mix of japanese and english

JoshSchreuder:
+1 for `poseTimer` if we're going with `pose`. 

Regarding `pose` I'm also in soft agreement that `animFrameIdx` is clearer, but looking at the code it might be possible that the numerical `pose` values could be mapped to enum values? Like `PLAYER.pose == POSE_CROUCH` or something like that? In which case `pose` would become quite clear I think.

Xeeynamo:
I changed it to `poseTimer`. I agree in using an enum of POSEs, we have a bunch of hard-coded numbers at the moment

Xeeynamo:
I think we gathered enough feedback. @sozud @bismurphy I am seeking an approval from either of you. We should decide fast because this PR is bound to have more conflicts with the incoming contributions.

sozud (approved):
If others like the new names that's fine, I'm not blocking this

Xeeynamo:
No worries, I just did not want to override you. I'll look into merging this one after #2426 :crossed_fingers:

PR #2411, opened 2025-04-06
Permuter loader: fix function names terminating with 's'

Xeeynamo:
I found `permute ric 182C8 RicSetSubweaponParams` strips the `s` character at the end, resulting the permuter trying to work in `sotn-decomp/tools/sotn_permuter/permute.me/RicSetSubweaponParam/`.

### DIFF
--- a/tools/sotn_permuter/permuter_loader.py
+++ b/tools/sotn_permuter/permuter_loader.py
@@ -255,8 +255,10 @@ if __name__ == "__main__":
                 for a in sys.argv
                 if not a.startswith("--settings") and not a.endswith(".c")
             ]
-            sys.argv[1] = sys.argv[1].split("/")[-1].rstrip(".s")
-            args.function = args.function.split("/")[-1].rstrip(".s")
+            args.function = args.function.split("/")[-1]
+            if args.function.endswith(".s"):
+                args.function = args.function[-2]
+            sys.argv[1] = args.function
 
     if args.action == "permute" or (
         args.action == "import" and args.permute and args.overlay


tools/sotn_permuter/permuter_loader.py
```diff
@@ -255,8 +255,10 @@ def sotn_finalize_compile_command(cmdline: importer.List[str]) -> str:
                 for a in sys.argv
                 if not a.startswith("--settings") and not a.endswith(".c")
             ]
-            sys.argv[1] = sys.argv[1].split("/")[-1].rstrip(".s")
-            args.function = args.function.split("/")[-1].rstrip(".s")
```
ProjectOblivion, line 259: I keep doing this no matter how many times I tell myself it's stupid.  rstrip removes the characters from the right side of the string until it none of the characters match the last character.  It should be .replace(".s")

tools/sotn_permuter/permuter_loader.py
```diff
@@ -255,8 +255,10 @@ def sotn_finalize_compile_command(cmdline: importer.List[str]) -> str:
...
+            args.function = args.function.split("/")[-1]
+            if args.function.endswith(".s"):
+                args.function = args.function[-2]
+            sys.argv[1] = args.function
```
ProjectOblivion, line 261: I can't think of any issues that this adjustment would cause.

tools/sotn_permuter/permuter_loader.py
Xeeynamo: I was a bit confused too. I thought `rstrip` stood for "reverse strip". I ended up using `endswith(".s")` with `[-2]` to reduce potential side effects, as we know `.s` is only found at the very end of the path

ProjectOblivion (approved):

PR #2412, opened 2025-04-06
Rename symbol for `g_UserLanguage` on PSP side

JoshSchreuder:
Added a symbol / renamed this so it's clearer in code / assemblies what it's used for.

~~Also did a slight update of the LANG enum to use more "typical" 2 character ISO codes for German (DE) and Spanish (ES)~~


Xeeynamo:
While I'd normally use the ISO standard as well, `sp` and `gr` are based to `disks/pspeu/PSP_GAME/USRDIR/res/ps/ms/`. It is the only reference I found so far, but IMHO we should stick with the game nomenclature as much as possible.

JoshSchreuder:
Makes sense! I've reverted the enum change

Xeeynamo (approved):

PR #2413, opened 2025-04-06
match func_060668D4

sozud:


src/saturn/game.c
```diff
@@ -13,7 +13,29 @@ INCLUDE_ASM("asm/saturn/game/f_nonmat", f60664E0, func_060664E0);
...
+// func_060668D4
+// similar to func_8011A9D8
+void func_8011A9D8(void) {
+    Entity* entity;
+    s32 i;
+
+    func_06064684();
+
+    entity = &g_Entities[0x40];
+    for (i = 0x40; i <= 255; i++, entity++) {
```
Xeeynamo, line 30:
```suggestion
    entity = &g_Entities[STAGE_ENTITY_START];
    for (i = STAGE_ENTITY_START; i <= 255; i++, entity++) {
```
I know that `func_8011A9D8` is not yet taking advantage of that define, but let's try using it for new code. `256` is `TOTAL_ENTITY_COUNT`, but I am not sure how to fit it to `i <= 255` if it can't be changed to `i < 256`

Xeeynamo (approved):
Left a non-blocker comment. Other than that, good stuff! Saturn really needs some love from all of us.

PR #2414, opened 2025-04-06
Decompile PSP `lib` init and stage name functions

JoshSchreuder:
Decompiled some PSP exclusive LIB functions


src/st/lib_psp/e_stage_name.c
```diff
@@ -1,6 +1,22 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "../lib/lib.h"
 
-INCLUDE_ASM("st/lib_psp/psp/lib_psp/e_stage_name", func_psp_0923C0C0);
+// these look to be indexed off language,
+// but they don't appear to be valid text?
+extern char D_psp_0926E170[]; // eng
```
JoshSchreuder, line 6: @gamezter since you're the string king, maybe you recognise this data format?
But it doesn't appear like text to me, despite it being array indexed off the user language.

src/st/lib_psp/st_init_lib.c
```diff
@@ -1,7 +1,30 @@
...
+
+// entity table to allocate new entities
+extern PfnEntityUpdate OVL_EXPORT(EntityUpdates)[];
+extern Overlay OVL_EXPORT(Overlay);
+
+extern char D_psp_09276330[];
+extern char D_psp_09276520[];
+extern char D_psp_092766E8[];
+extern char D_psp_092768E0[];
+extern char D_psp_09276AC8[];
```
JoshSchreuder, line 20: Have held off importing this data for now, since it's related to the cutscene between Alucard and Libarian, and I think it likely has some state machine / control characters baked in to control switching between who's talking and what-not.

It will probably make more sense once `LIB_EntityCutscene` is decompiled

image

src/st/lib_psp/e_stage_name.c
gamezter: Could these be localized cutscene data ?

JoshSchreuder: I don't think so but not certain - this function is called from e_stage_name, but the address it stores these pointers to (`D_psp_0926EDE4`) doesn't appear to get referenced anywhere else, so I'm not clear on what it could be used for.

And this is LIB specific code also, not present in the other PSP stages we have so far

src/st/lib_psp/st_init_lib.c
Xeeynamo: Did you try specifying the offset of the cutscene at `config/assets.us.yaml`? It should decompile the state machine for you

src/st/lib_psp/st_init_lib.c
```diff
@@ -1,7 +1,30 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "../lib/lib.h"
 
-INCLUDE_ASM("st/lib_psp/psp/lib_psp/1B7F0", func_psp_09253D58);
+#include "../pfn_entity_update.h"
+
+// entity table to allocate new entities
+extern PfnEntityUpdate* PfnEntityUpdates;
```
Xeeynamo, line 7: This and `EntityUpdates` are already defined in `pfn_entity_update`

src/st/lib_psp/st_init_lib.c
JoshSchreuder: This is for PSP side, I don't think we support that for the asset extraction tool do we?

Xeeynamo: The tool should be platform-agnostic. So as long as we create a new asset file for PSP it _should_ work. I am not sure if you want to do it on a separate PR. Maybe it's a good idea?

JoshSchreuder: I will try it on a separate PR if it works

src/st/lib_psp/st_init_lib.c
JoshSchreuder: You are spot on, I have updated this (and also cleaned up similar code in ST0 and WRP PSP)

PR #2422, opened 2025-04-08
Import `psp` `lib` cutscene data

JoshSchreuder:
This required some changes to the cutscene handler to keep accented characters in the STRING section and not convert to BYTE data.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -637,9 +637,3 @@ get-phony:
 get-silent:
 	echo ".SILENT:" > make.silent.targets
 	$(foreach target,$(MUFFLED_TARGETS),echo $(target) >> make.silent.targets;)
-
-extract_assets: $(SOTNASSETS)
-	cd tools/sotn-assets; $(GO) install
-	$(SOTNASSETS) extract config/assets.$(VERSION).yaml
-build_assets: $(SOTNASSETS)
-	$(SOTNASSETS) build config/assets.$(VERSION).yaml
--- a/Makefile
+++ b/Makefile
@@ -637,3 +637,9 @@ get-phony:
 get-silent:
 	echo ".SILENT:" > make.silent.targets
 	$(foreach target,$(MUFFLED_TARGETS),echo $(target) >> make.silent.targets;)
+
+extract_assets: $(SOTNASSETS)
+	cd tools/sotn-assets; $(GO) install
+	$(SOTNASSETS) extract config/assets.$(VERSION).yaml
+build_assets: $(SOTNASSETS)
+	$(SOTNASSETS) build config/assets.$(VERSION).yaml
--- a/Makefile.psp.mk
+++ b/Makefile.psp.mk
@@ -13,6 +13,11 @@ $(BUILD_DIR)/%.ld: $(CONFIG_DIR)/splat.$(VERSION).%.yaml $(BASE_SYMBOLS) $(CONFI
 	$(muffle)$(SPLAT) $<
 
 # Step 2/2 of extract
+extract_assets: $(SOTNASSETS)
+	cd tools/sotn-assets; $(GO) install
+	$(SOTNASSETS) extract config/assets.$(VERSION).yaml
+build_assets: $(SOTNASSETS)
+	$(SOTNASSETS) build config/assets.$(VERSION).yaml
 extract_pspeu: $(addprefix $(BUILD_DIR)/,$(addsuffix .ld,$(call get_targets,prefixed)))
 	make extract_assets
 	make build_assets
--- a/Makefile.psp.mk
+++ b/Makefile.psp.mk
@@ -14,6 +14,8 @@ $(BUILD_DIR)/%.ld: $(CONFIG_DIR)/splat.$(VERSION).%.yaml $(BASE_SYMBOLS) $(CONFI
 
 # Step 2/2 of extract
 extract_pspeu: $(addprefix $(BUILD_DIR)/,$(addsuffix .ld,$(call get_targets,prefixed)))
+	make extract_assets
+	make build_assets
 
 # Step 1/5 of build
 $(BUILD_DIR)/%.s.o: %.s $(AS)
--- a/include/cutscene.h
+++ b/include/cutscene.h
@@ -15,7 +15,7 @@ typedef enum {
     CSOP_CLOSE_DIALOG,
     CSOP_PLAY_SOUND,
     CSOP_WAIT_FOR_SOUND,
-    CSOP_UNK_11,
+    CSOP_SCRIPT_UNKNOWN_11,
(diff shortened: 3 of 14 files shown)


Makefile
```diff
@@ -637,3 +637,9 @@ get-phony:
 get-silent:
 	echo ".SILENT:" > make.silent.targets
 	$(foreach target,$(MUFFLED_TARGETS),echo $(target) >> make.silent.targets;)
+
+extract_assets: $(SOTNASSETS)
+	cd tools/sotn-assets; $(GO) install
+	$(SOTNASSETS) extract config/assets.$(VERSION).yaml
+build_assets: $(SOTNASSETS)
+	$(SOTNASSETS) build config/assets.$(VERSION).yaml
```
JoshSchreuder, line 645: @ProjectOblivion can you please have a look and make sure these Makefile changes fit with your stuff? I moved this out of `Makefile.psx.mk` and just put it at the bottom of the file but I can move it elsewhere if you let me know where it should go.

Also see changes in the PSP Makefile to add assets steps

include/cutscene.h
```diff
@@ -15,7 +15,7 @@ typedef enum {
     CSOP_CLOSE_DIALOG,
     CSOP_PLAY_SOUND,
     CSOP_WAIT_FOR_SOUND,
-    CSOP_UNK_11,
+    CSOP_SCRIPT_UNKNOWN_11,
```
JoshSchreuder, line 18: The inconsistency was bugging me so I renamed this to match the other unknowns

tools/sotn-assets/assets/cutscene/handler.go
```diff
@@ -70,11 +70,22 @@ func (h *handler) Build(e assets.BuildArgs) error {
...
 				} else {
-					sb.WriteString(fmt.Sprintf("'%c',", text[i]))
+					byteValue := byte(r)
+
+					// If the byte is in the PSP's upper language character position,
+					// output the hex literal
+					if os.Getenv("VERSION") == "pspeu" && byteValue >= 0xA0 && byteValue <= 0xDF {
+						// TODO: This can probably output something like _SE() which
+						// can be handled by sotn_str
+						sb.WriteString(fmt.Sprintf("'\\x%02X',", byteValue))
```
JoshSchreuder, line 84: I think in future this can probably be updated to output and pass through sotn_str, though I did not get it working in the quick attempt I tried.

For example:
image

`0xD0` is the `è` mapping from [sotn_str](https://github.com/Xeeynamo/sotn-decomp/blob/master/tools/sotn_str/src/main.rs#L455)

I have left this as a future exercise if someone wants to take a look at it

tools/sotn-assets/assets/cutscene/handler.go
```diff
@@ -212,6 +223,9 @@ func parseScript(r io.ReadSeeker, baseAddr, addr psx.Addr, length int) ([][]stri
 			script = append(script, command)
 		} else if op < 0x7F {
 			text += string([]byte{byte(op)})
+		} else if os.Getenv("VERSION") == "pspeu" && op >= 0xA0 && op <= 0xDF {
```
JoshSchreuder, line 226: I am not sure if it is necessary but in the interest of defensive coding I only made this happen for PSP, since PSX does not use multi-lang characters

Makefile.psp.mk
```diff
@@ -14,6 +14,8 @@ $(BUILD_DIR)/%.ld: $(CONFIG_DIR)/splat.$(VERSION).%.yaml $(BASE_SYMBOLS) $(CONFI
 
 # Step 2/2 of extract
```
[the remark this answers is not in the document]
JoshSchreuder: Done, thanks for taking a look 👍

Makefile
JoshSchreuder: Done 👍

tools/sotn-assets/assets/cutscene/handler.go
sozud: What was preventing it from working?

JoshSchreuder: I am not certain, I didn't spend too much time on it, but from what I could tell `_SE` as a define was not available, and after moving it outside `#ifndef SOTN_STR` to make it defined, it didn't appear to be running through `sotn_str` processing.

I'm not sure if the auto-generated headers don't go through the string pipeline or whether it's something else.

JoshSchreuder: I don't think it would process data imported via headers by the looks?

```
❯ tools/sotn_str/target/release/sotn_str process -f src/st/lib_psp/cutscene_data.c -p
#include "../lib/lib.h"
#include <cutscene.h>

u8 cutscene_data_it[] = {
#include "cutscene_data_it.h"
};

u8 cutscene_data_sp[] = {
#include "cutscene_data_sp.h"
};

u8 cutscene_data_fr[] = {
#include "cutscene_data_fr.h"
};

u8 cutscene_data_ge[] = {
#include "cutscene_data_ge.h"
};

u8 cutscene_data_en[] = {
#include "cutscene_data_en.h"
};
```

sozud (approved):
Xeeynamo should take a look at the `go` part but otherwise looks good

tools/sotn-assets/assets/cutscene/handler.go
Xeeynamo: It's fine. The various `cutscene_data.h` files are generated and needs to be treated as intermediary data. Outputting a byte literal from the build step of the asset tool is fine. Realistically, almost everyone will look/modify the asset file in YAML instead.

tools/sotn-assets/assets/cutscene/handler.go
Xeeynamo: Here the same previous comment applies, where we could just always dump the hex literal if it's easier. If you wish to still output a character and let `sotn_str` handle the rest, we should move `os.Getenv` out of each loop. The main reason being the use of a mutex in the `os.Getenv` to ensure multi-thread synchronization, which will impact the performance of building cutscenes on a CPU that allows great concurrency.

JoshSchreuder: I will move the version check outside the loop, thanks for the hint, I didn't know that about the performance characteristics.

There are still some BYTE data which I don't think are actually part of strings which I think we would probably like to preserve, rather than always treating these higher order characters as string values, so I think this check probably makes sense (at least for now)

eg.
image

JoshSchreuder: Done 👍

Xeeynamo (approved):
Great stuff, thanks a lot!

PR #2426, opened 2025-04-08
Several More `BO4` and `RBO5` Functions

hohle:
Matching 11 additional functions across both `BO4` and `RBO5`.

A small change to `ET_DisableAfterImage` which appears to be used for Doppleganger as well. The names and comments did not match the offsets.

Some minor cleanup for things encountered while looking for similar or matching code.

### DIFF
--- a/src/boss/rbo5/unk_44954.c
+++ b/src/boss/rbo5/unk_44954.c
@@ -5,7 +5,32 @@ extern PlayerState g_Dop;
 
 // n.b.! this file is the same as bo4/unk_45354.c
 
-INCLUDE_ASM("boss/rbo5/nonmatchings/unk_44954", func_us_801C4954);
+extern s16 D_us_801D433E;
+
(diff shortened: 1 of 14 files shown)


src/boss/rbo5/unk_44954.c
```diff
@@ -165,11 +205,74 @@ void func_us_801C4FDC(void) {
     g_Dop.unk44 = 0x10;
 }
 
-INCLUDE_ASM("boss/rbo5/nonmatchings/unk_44954", func_us_801C504C);
+void func_us_801C504C(void) {
+    if (CheckMoveDirection() != 0) {
+        SetSpeedX(0x30000);
```
gamezter, line 210: FIX(3.0)

hohle: Thanks for catching that. I had it correct in RBO5 somehow.

sozud (approved):

Xeeynamo (approved):

PR #2427, opened 2025-04-08
Decompile func_us_801C542C

amelenty:
Thanks to @mkst for the help!


src/st/no4/first_c_file.c
```diff
@@ -57,7 +57,28 @@ INCLUDE_ASM("st/no4/nonmatchings/first_c_file", func_us_801C5134);
...
+    if (self->step == 0) {
+        InitializeEntity(g_EInitCommon);
+        self->animSet = ANIMSET_OVL(1);
+        self->animCurFrame = D_us_801815DC[params];
+        self->facingLeft = D_us_801815EC[params];
+        self->velocityY = D_us_8018160C[params];
+        self->drawFlags = FLAG_DRAW_ROTZ;
+        self->rotZ = false;
+    }
+    if (HIH(self->velocityY) < 8) {
```
sozud, line 76: I think this is supposed to be like
```
    if(self->velocityY.i.hi < 8)
    {
        self->velocityY.val += FIX(0.25);
    }
```
but that requires a giant refactor

Xeeynamo: Both @gamezter and @bismurphy suggested a `F(self->velocityY).i.hi`. It should be a bit more accurate to how the original source code might have looked like.

PR #2428, opened 2025-04-08
Cleaned up decompile.py and added decomp.me upload

ProjectOblivion:


tools/decompile.py
```diff
@@ -1,338 +1,446 @@
...
+                print(
+                    f"Received an error when attempting to upload to decomp.me: {response_json.get("error", response.text)}"
+                )
+        except Exception as e:
+            print(
+                f"An unhandled error occurred while attempting to upload to decomp.me: {e}"
+            )
 
+    def resolve_jumptables(self) -> None:
+        """Adds any jump tables to the function asm file if they aren't already there"""
```
ProjectOblivion, line 132: I didn't do much to this function since it is mostly deprecated by splat migrating rodata.  Just flattened it down and cleaned it up for readability.

tools/decompile.py
```diff
@@ -1,338 +1,446 @@
...
+    parser.add_argument(
+        "-j",
+        "--resolve-jumptables",
+        help="Attempt to add any jump tables for the function to the asm file, if they're not already there (uncommon)",
         action="store_true",
+        required=False,
+    )
+    parser.add_argument(
+        "-u",
+        "--upload",
```
amelenty, line 431: I suggest calling this argument `-d/--decompme` to match the existing argument of `permute`.

tools/decompile.py
```diff
@@ -1,338 +1,446 @@
...
+            status = inject_decompiled_function(repo_root, sotn_func)
+
+        if args.resolve_jumptables:
+            sotn_func.resolve_jumptables()
+        if args.upload:
+            sotn_func.upload_to_decompme()
+
+    match status:
+        case Status.NOT_FOUND:
+            message = f"Could not find function {args.function}, are you sure it exists in version {args.version}?"
```
amelenty, line 373: Since the message is tightly coupled to the Status value, it would make sense to store the format string within `Status`. Here are some possible approaches:
1. Move the `match` statement to a `__str__()` method of `Status` and then call `print(status)` here.
2. Use a [multi-value enum](https://docs.python.org/3/howto/enum.html#multivalueenum) and store the format string as the second field.
3. Use a [dataclass enum](https://docs.python.org/3/howto/enum.html#dataclass-support) to store multiple fields without needing a complicated `new()` method.

tools/decompile.py
```diff
@@ -1,338 +1,446 @@
...
+    """Steps backward from the file location to infer the root of the repo"""
+    if next(current_dir.glob("src"), None) or current_dir.name == "/":
+        return current_dir
+    else:
+        return get_repo_root(current_dir.parent)
+
+
+def get_function_path(
+    asm_dir: Path, args: argparse.Namespace
+) -> tuple[Status | None, tuple[Path, ...] | None]:
```
amelenty, line 260: I suggest using [Optional](https://docs.python.org/3/library/typing.html#typing.Optional) for clearer typing.

tools/decompile.py
```diff
@@ -1,338 +1,446 @@
...
+        """Leverages m2c to decompile the function and stores it so that decompilation only needs to run once"""
+        if not self.c_code:
+            self.c_code = self._guess_unknown_type(self._run_m2c())
+        return self.c_code
+
+    def upload_to_decompme(self) -> None:
+        """Uploads the decompiled code to decomp.me using the extracted context and SotN decomp.me presets"""
+        print("Loading up the catapult...")
+        # No current expectation of these changing, so hardcoding is probably fine
+        presets = {"us": 14, "pspeu": 132}
```
amelenty, line 98: Are we sure that this number will not change on decomp.me? It would be nice to have some validation that this is the correct preset in the code.

tools/decompile.py
```diff
@@ -1,338 +1,446 @@
 #!/usr/bin/env python3
+"""This script is intended to make it easier for someone to decompile a function within the sotn-decomp project
+and optionally upload that function to decomp.me.
+Any questions or issues related to running it should be directed to the sotn-decomp discord #tooling channel:
+https://discord.com/channels/1079389589950705684/1135205782703570994
```
amelenty, line 5: IMO it's better to link to https://sotn-discord.xee.dev/ here and on line 393: the link won't work for someone who hasn't joined the discord yet, and someone who already joined can easily find the channel in the Discord app.

tools/decompile.py
ProjectOblivion: No, they should not change.  Those numbers are the id's pulled from the API, the handling code just isn't necessary here so I didn't add it.
We also know it's correct because it works.

tools/decompile.py
ProjectOblivion: I prefer keeping arguments separate from tools and aligning our arguments with third party tools isn't something I think we should be trying to do.
That being said, I'm not hard against it if that's what we're looking for in the project overall.

tools/decompile.py
ProjectOblivion: That's a fair point.  I was thinking toward people already in the Discord when I wrote the blurb, but ultimately I was wanting some sort of standardized blurb like this that we can stick at the top of "sotn-decomp" tools.

tools/decompile.py
ProjectOblivion: I'm not an expert on type hints, but I looked at it like this:
It's not really optional, the code will terminate earlier if these values are None, but None is a valid type for these vars.

I'm happy to adjust it if I'm in the collective minority.

tools/decompile.py
ProjectOblivion: I appreciate this feedback, but it feels like I'm getting it from an LLM.  Are you a robot?
Kidding, of course, I appreciate the feedback regardless of the original source.

This is something I had debated and kind of just didn't feel like making a decision on to be honest, so I went with the lazy solution.  Since the messages use variables that could encounter scoping issues, it isn't as straightforward as a pure string would be.  I was looking at revisiting the approach and cleaning it up if possible as I add in proper pspeu support, so I'm definitely taking note of the suggestions to look at.

tools/decompile.py
Xeeynamo: It is fine to keep it `-u/--upload`. It is concise enough.

tools/decompile.py
Xeeynamo: I agree with the comment from @amelenty . Although I am proposing in another comment a way where we could trim down almost all messages, eliminating the problem of choosing how the implementation should look like.

Xeeynamo (commented):
This is a much better than the existing tool. I have no strong opinion about the code and the implementation itself, but more about the functionality.

* We can remove the injection of the jump table(s).
* We can avoid checking if the decompiled function actually compiles or matches. Unless the function is microscopic, it will almost always fail to compile. Let's just print on screen the asm-differ command after injecting the m2c output, then exit.
* This adds a second way of uploading a function to decompme. What was the rationale of including it here? Don't get me wrong, it's something I use very often. But I wonder which one will be the intended one when this gets in master.

As always I aim to merge in code that is simple and straightforward. And I prefer to reduce options when possible to reduce the maintainability burden and to not overwhelm who consume our code base & tool interface. Let's reduce the scope of the tool. I'll do a quick second round of review afterwards :crossed_fingers: 

Also, a bit of a side-line comment @amelenty I appreciate you did a code review. I also noticed @ProjectOblivion you did some code review on other people's PRs. I encourage contributors to keep doing it. That releases some burden from the project maintainers (me, Sozud, Bismurphy).

tools/decompile.py
```diff
@@ -1,338 +1,445 @@
...
+            )
+            self._overlay = (
+                self.relpath.parts[nonmatching_index - 1] if nonmatching_index else None
+            )
+        return self._overlay
+
+    @property
+    def asm_differ_command(self) -> str:
+        """Constructs the string to run asm-differ if the decompiled function code can be compiled"""
+        return f"{Path(sys.executable).name} {self.root.joinpath("tools/asm-differ/diff.py").relative_to(Path.cwd())} -mwo --overlay {self.overlay} {self.name}"
```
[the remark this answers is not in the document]
ProjectOblivion: This is simply a shortcut for constructing the string.  Functionality remains as it was before, just giving the command as console feedback.

tools/decompile.py
ProjectOblivion: https://www.decomp.me/api/compilers

It was saturn that was exactly why I went with a dict, but your suggestion doesn't work as probably expected.  Building the payload uses the version to get the preset from the dict.
I can drop the dict, but the code would just need to be adjusted for uploading saturn to decomp.me in the future.

tools/decompile.py
```diff
@@ -1,338 +1,445 @@
...
-    elif inject_res == InjectRes.NOT_INJECTED:
-        print(f"function '{func.name}' might already be decompiled")
-    else:
-        print("unhandled error!")
+        return Status.DOES_NOT_COMPILE
+
+
+def main(args: argparse.Namespace) -> None:
+    """Executes the decompilation workflow and gives console feedback based on the results"""
+    if args.version != "us":
```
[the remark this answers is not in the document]
ProjectOblivion: I wasn't aware that it worked for anything but `us` previously.  That logic is pretty easy to adjust to consider architecture instead of version assuming `saturn`, `psp`, and `psx` sufficiently handles it.

tools/decompile.py
ProjectOblivion: I've been looking at solutions for this particular problem and they all amount to a sliding puzzle, simply shifting the complexity around.  There is no way that I can see to make the current situation work cleanly, so I'm not going to continue trying to do it.
I'm just going to remove the enum message situation that was really just a carryover from previously, anyway.  Necessary feedback messages can just be printed at the time of the situation.

ProjectOblivion:
> This adds a second way of uploading a function to decompme. What was the rationale of including it here? Don't get me wrong, it's something I use very often. But I wonder which one will be the intended one when this gets in master.

Two reasons:
- This creates one tool for the two possible ways to decompile a function, local and decomp.me.  Permuter is an optional tool so having what I consider to be required functionality tied to it doesn't make any sense to me.
- Adding functionality to permuter requires far more hurdles than I feel like dealing with and it doesn't have the ability to specify preset, which I think is an important part to smooth accessibility for newer folks.

tools/decompile.py
ProjectOblivion: I discovered that we already had a saturn preset, so I added that in and enabled uploading to decomp.me for all versions.

tools/decompile.py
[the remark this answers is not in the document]
ProjectOblivion: I enabled decompilation of all versions except saturn.

tools/decompile.py
```diff
@@ -1,338 +1,324 @@
...
+        if self.version != "saturn" and not self.c_code:
+            self.c_code = self._guess_unknown_type(self._run_m2c())
+        return self.c_code
+
+    def upload_to_decompme(self) -> None:
+        """Uploads the decompiled code to decomp.me using the extracted context and SotN decomp.me presets"""
+        print("Loading up the catapult...")
+        # No current expectation of these changing, so hardcoding instead of retrieving them from the API programmatically is probably fine
+        # The preset id's can be found at https://www.decomp.me/api/compilers, searching the page for Castlevania: Symphony of the Night
+        presets = {"us": 14, "pspeu": 132, "saturn": 21}
```
Xeeynamo, line 99:
```suggestion
        presets = {"us": 14, "hd": 14, "pspeu": 132, "saturn": 21}
```
If we want to go ahead with using a map of presets, we need to support the HD version

ProjectOblivion: This was my oversight.  I intended to shift this to go by platform, rather than version.

ProjectOblivion: It also ended up being cleaner to go with your original suggestion, so I went with that.

Xeeynamo (approved):
Let's have it merged and collect feedback. Great job in improving this tool!

PR #2429, opened 2025-04-10
Decompile `no4` func_us_801D96FC

gamezter:
PSX: https://decomp.me/scratch/1JmyJ
PSP: https://decomp.me/scratch/JQPXI

I created a new scratchpad macro that *could* replace `SPAD`. Let me know if it's a good idea or if I should revert to using `SPAD`.

This was the largest non-decompiled NO4 function, and was pretty tough:

- `points` and `sxy` were originally 4 separate pointer variables while I was decompiling on PSP, but PSX requires them to be arrays. The stack variable offsets *grow* downwards on PSX, except for variables in arrays, while they *grow* upwards on PSP even in arrays.
- `&g_Entities[self->ext.et_801D96FC.unkA8[j]]` looked like `&g_Entities[(self + j)->unkA8]` until I understood there was an array in the extension struct.
- `func_us_801D9560` / `func_pspeu_09255F98` had to be declared static in the same file to get a match on PSP
- Figuring out scratchpad variables is always a brainteaser but so rewarding when it all makes sense.


src/st/no4/unk_55678.c
```diff
@@ -42,11 +40,748 @@ INCLUDE_ASM("st/no4/nonmatchings/unk_55678", func_us_801D8FE0);
...
+        }
+        MoveEntity();
+        break;
+
+    case 4:
+        switch (self->step_s) {
+        case 0:
+            self->animCurFrame = 9;
+            self->ext.et_801D96FC.unk84 = 0x10;
+            g_api.PlaySfx(0x733);
```
sozud, line 527: SFX_FROZEN_SHADE_SCREAM

src/st/no4_psp/unk_1E818.c
```diff
@@ -1,11 +1,749 @@
...
+        }
+        MoveEntity();
+        break;
+
+    case 4:
+        switch (self->step_s) {
+        case 0:
+            self->animCurFrame = 9;
+            self->ext.et_801D96FC.unk84 = 0x10;
+            g_api.PlaySfx(0x733);
```
sozud, line 489: SFX_FROZEN_SHADE_SCREAM

sozud (approved):

include/game.h
```diff
@@ -31,6 +31,7 @@
 #endif
 
 #define SPAD(x) ((s32*)SP(x * sizeof(s32)))
+#define SP2(x, type) ((type*)SP(x * sizeof(void*)))
```
Xeeynamo, line 34: I am not a fan of the introduction of this helper. The file `src/st/st0/3C5C0.c` has an interesting interpretation of how a structured stackpad could be read. Essentially you define a struct and then you leverage `offsetof` to accurately get back the offset of the field you're interested on. This comes at the sacrifice that you would have to explicitly cast the various `(Point16*)`, `(SVECTOR*)` and so on, but it guarantees the code is moddable and portable.

src/st/no4/unk_55678.c
```diff
@@ -42,11 +40,748 @@ INCLUDE_ASM("st/no4/nonmatchings/unk_55678", func_us_801D8FE0);
...
+                    }
+                    prim->p2 = 0;
+                    if (prim->p1 == (self->ext.et_801D96FC.unk9E & 3) &&
+                        prim->u3 < 0x30) {
+                        tempEntity =
+                            &g_Entities[self->ext.et_801D96FC.unkA8[j]];
+                        tempEntity->posX.i.hi =
+                            LOH(prim->clut) - tilemap->scrollX.i.hi;
+                        tempEntity->posY.i.hi =
+                            LOH(prim->tpage) - tilemap->scrollY.i.hi;
```
JoshSchreuder, line 155: Not a comment on PR quality, but these two seem rather suspicious to me... perhaps this is actually a Primitive variant and this is not actually using `clut` / `tpage`

Xeeynamo (commented):
Left a comment for the introduction of `SP2`

src/st/no4/unk_55678.c
Xeeynamo: > Not a comment on PR quality, but these two seem rather suspicious to me... perhaps this is actually a Primitive variant and this is not actually using `clut` / `tpage`

`Primitive` sometimes seems to be used as a blob of data. This primitive is a `PRIM_G4`, an untextured coloured quad that does not make use of `clut`, `tpage`, `u0`, `v0`, and similar. @gamezter it's probably worth copy&paste the existing `Primitive` on top of this function and change name+type for these.

PR #2434, opened 2025-04-12
Improve Docker support

Xeeynamo:
Enhance the SOTN Decomp experience with Docker. This is a full break-down of the changes:
* Enhance the instruction steps to build the image.
* Use default `ubuntu:ubuntu` user. It should map directly to 1000:1000 from the host, without messing with new users and permissions. There was a problem where the container used a mix of two users, causing various "permission denied".
* The installation of dosemu2 was broken, it is now fixed.
* The Python set-up step is now removed from the build image step and moved in a later stage. This is now done in the container itself with `make update-dependencies`.
* Added a `/sotn/.venv` overlay, so it will not conflict with the native `.venv` folder (if this is present). This allows me to use concurrently both .venv from Arch Linux and the containerised one from Ubuntu without them to collide.
* Added a `/sotn/build` overlay to avoid back and forth between the host and the container. This improves the build speed on macOS and Windows, as Docker Desktop uses a VM under the hood.
* Make clear that a container can be stopped and resumed any time, without having to re-configure it again from scratch.

### DIFF
--- a/.github/workflows/dockerfile.yaml
+++ b/.github/workflows/dockerfile.yaml
@@ -1,0 +1,18 @@
+name: Build Docker image
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'Dockerfile'
+      - 'tools/**'
+      - 'Makefile*'
+  workflow_dispatch:
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


sozud (approved):
Seems good, I think having a docker-based build in the CI would be good to make sure this is maintained

PR #2435, opened 2025-04-12
Change PSP and US splits to align first PSP C file

bismurphy:
Decided to start things from the beginning and go to the first C file in PSP. Decompiled one function, then realized the files don't properly align between the two versions.

Now both match up, and US also ends up making more sense too - the file splits are at much more natural locations.

Next will attempt to enable code-sharing on this file, of course :)


sozud (approved):
Is the maspsx change intentional? other than that seems good

bismurphy:
> Is the maspsx change intentional? other than that seems good

That's just pulling the latest maspsx, where mkst made a readme adjustment right after his latest push of maspsx into sotn-decomp. I figured it would be fine to include; when I made this PR I was still on the old maspsx and when I pulled the newest, it was one commit ahead of sotn-decomp's version, but given that it's a meaningless commit it seems harmless.

Xeeynamo (approved):

PR #2438, opened 2025-04-14
Fix decompile script for PSP overlays

Xeeynamo:
This fixes a small bug on tools/decompile.py due to the false assumption that the `nonmatchings` part of the path is different between PSX and PSP. All the PSP overlays produced by `make-config.py` use `nonmatchings` in their path.

There is also a fix when assuming the name of the overlay on PSP. I found it was returning `ric_psp` as the overlay name for me. Doing a quick replace from `_psp` to an empty string fixed it. 

```
# before the fix
$ dec RicEntityMariaPowers
Could not find function RicEntityMariaPowers, are you sure it exists in version pspeu?

# after the fix
RicEntityMariaPowers was successfully decompiled, but likely will not compile without adjustments.
When it successfully compiles, the following command can be used to look for the differences:
	python3 tools/asm-differ/diff.py -mwo --overlay ric RicEntityMariaPowers   
```

### DIFF
--- a/tools/decompile.py
+++ b/tools/decompile.py
@@ -52,23 +52,19 @@ class SotnFunction(object):
     def overlay(self) -> str:
         """Attempts to infer the overlay from the asm file path or returns None if one is not found.
         The '_psp' extension will be included if the version is psp"""
-        nonmatchings_dir = (
-            "f_nonmat"
-            if args.version == "saturn"
-            else "psp" if "psp" in args.version else "nonmatchings"
-        )
+        nonmatchings_dir = "f_nonmat" if args.version == "saturn" else "nonmatchings"
         if not self._overlay:
             nonmatching_index = next(
                 (
                     i
                     for i, part in enumerate(self.relpath.parts)
-                    if part == nonmatchings_dir
+                    if nonmatchings_dir in part
                 ),
                 None,
             )
             self._overlay = (
                 self.relpath.parts[nonmatching_index - 1] if nonmatching_index else None
-            )
+            ).replace("_psp", "")
         return self._overlay
 
     @property
@@ -172,14 +168,14 @@ class SotnFunction(object):
 
     def _infer_src_path(self) -> Optional[Path]:
         inferred_c_name = f"{self.abspath.parent.parent.name if self.version == "saturn" else self.abspath.parent.name}.c"
-        inferred_c_files = (
-            file
-            for file in self.src_dir.rglob(inferred_c_name)
-            if self.overlay in file.parts or self.overlay in file.name
-        )
-        return next(
-            (path for path in inferred_c_files if self.name in path.read_text()), None
-        )
+        files = [file for file in self.src_dir.rglob(inferred_c_name)]
+        if len(files) > 1 and self.overlay:
+            files = [
+                file
+                for file in files
+                if self.overlay in file.parts or self.overlay in file.name
+            ]
+        return files[0]
 
 
 def get_repo_root(current_dir: Path = Path(__file__).resolve().parent) -> Path:
@@ -194,11 +190,7 @@ def get_function_path(asm_dir: Path, args: argparse.Namespace) -> Optional[Path]
     """Uses the version asm directory and the passed args to find any files matching the function name."""
     file_name = f"{args.function.replace("func_0", "f") if args.version == "saturn" else args.function}.s"
     matchings_dir = "f_match" if args.version == "saturn" else "matchings"
-    nonmatchings_dir = (
-        "f_nonmat"
-        if args.version == "saturn"
-        else "psp" if "psp" in args.version else "nonmatchings"
-    )
+    nonmatchings_dir = "f_nonmat" if args.version == "saturn" else "nonmatchings"
 
     candidates = tuple(file for file in asm_dir.rglob(file_name))
     matching = tuple(c for c in candidates if matchings_dir in c.parts)


tools/decompile.py
```diff
@@ -172,14 +168,14 @@ def _guess_unknown_type(self, decompiled_code: str) -> str:
...
-            (path for path in inferred_c_files if self.name in path.read_text()), None
-        )
+        files = [file for file in self.src_dir.rglob(inferred_c_name)]
+        if len(files) > 1 and self.overlay:
+            files = [
+                file
+                for file in files
+                if self.overlay in file.parts or self.overlay in file.name
+            ]
+        return files[0]
```
ProjectOblivion, line 178: What is the purpose of completely changing this around?
It takes away the function being able to guarantee that it has returned the correct file, as well as adding unnecessary iteration.

tools/decompile.py
```diff
@@ -52,23 +52,19 @@ def __init__(self, func_root: Path, function_path: Path, args: argparse.Namespac
     def overlay(self) -> str:
         """Attempts to infer the overlay from the asm file path or returns None if one is not found.
         The '_psp' extension will be included if the version is psp"""
-        nonmatchings_dir = (
-            "f_nonmat"
-            if args.version == "saturn"
-            else "psp" if "psp" in args.version else "nonmatchings"
-        )
+        nonmatchings_dir = "f_nonmat" if args.version == "saturn" else "nonmatchings"
```
ProjectOblivion, line 55: Previously extracted overlays still use psp as the nonmatchings directory, this will break the tool for them.

tools/decompile.py
```diff
@@ -52,23 +52,19 @@ def __init__(self, func_root: Path, function_path: Path, args: argparse.Namespac
...
         if not self._overlay:
             nonmatching_index = next(
                 (
                     i
                     for i, part in enumerate(self.relpath.parts)
-                    if part == nonmatchings_dir
+                    if nonmatchings_dir in part
```
ProjectOblivion, line 61: Why change this?

tools/decompile.py
```diff
@@ -52,23 +52,19 @@ def __init__(self, func_root: Path, function_path: Path, args: argparse.Namespac
...
                 ),
                 None,
             )
             self._overlay = (
                 self.relpath.parts[nonmatching_index - 1] if nonmatching_index else None
-            )
+            ).replace("_psp", "")
```
ProjectOblivion, line 67: What is the issue with _psp being included in the overlay name?

ProjectOblivion:
The only things that need to be done to fix it is changing line 65 in master to `if part == nonmatchings_dir or part == "nonmatchings"` and lin 208 in master to `if (nonmatchings_dir in c.parts or "nonmatchings" in c.parts) and (not args.overlay or args.overlay in c.parts)`.

ProjectOblivion (changes requested):
The only things that need to be done to fix it is changing line 65 in master to `if part == nonmatchings_dir or part == "nonmatchings"` and line 208 in master to `if (nonmatchings_dir in c.parts or "nonmatchings" in c.parts) and (not args.overlay or args.overlay in c.parts)`

```
@@ -64,3 +64,3 @@ class SotnFunction(object):
                     for i, part in enumerate(self.relpath.parts)
-                    if part == nonmatchings_dir
+                    if part == nonmatchings_dir or part == "nonmatchings"
                 ),
@@ -207,3 +207,3 @@ def get_function_path(asm_dir: Path, args: argparse.Namespace) -> Optional[Path]
         for c in candidates
-        if nonmatchings_dir in c.parts and (not args.overlay or args.overlay in c.parts)
+        if (nonmatchings_dir in c.parts or "nonmatchings" in c.parts) and (not args.overlay or args.overlay in c.parts)
```

tools/decompile.py
Xeeynamo: When `self.overlay` is nil, this logic just crashes. It would be nicer to receive a suggestion how to keep the logic while avoiding unnecessary iterations, as you said.

tools/decompile.py
Xeeynamo: > All the PSP overlays produced by make-config.py use nonmatchings in their path.

From my PR description.

tools/decompile.py
Xeeynamo: Because:

1. Other places uses it. For example `nonmatchings_dir in c.parts`.
2. If the overlay is `ric` and one of the path parts is `ric_psp`, this will get me the wrong `nonmatching_index` and just produce the wrong output.

tools/decompile.py
Xeeynamo: It's either here or asm_differ. My logic is that there are no two overlays `ric` and `ric_psp`. There's only one `ric` with multiple versions. It has to be fixed here.

tools/decompile.py
ProjectOblivion: I would have given a suggestion previously, but didn't understand the goal.
Based on how the code was written, there will not be a circumstance where it can get to this point in the code without something being extracted from the path as an overlay value.  If there is an instance where some value wouldn't be extracted as the overlay, then it would throw this message:
`Could not find function RicEntityMariaPowers, are you sure it exists in version pspeu?`

tools/decompile.py
ProjectOblivion: I'm confused, are you saying that we don't need to care about being able to use `decompile.py` on overlays that use `psp` as `nonmatchings_dir`?

tools/decompile.py
ProjectOblivion: 1. Two different purposes.  Your example simply needs to know if the value is in the list, this is specifically looking for where the value is in the list.
2.  That's literally not possible.  For `part == nonmatchings_dir` to return an index, it has to match exactly.  Meaning it would only return an index if `nonmatchings_dir == "ric_psp"` and `part == "ric_psp"`.  Your change is actually want makes it so this occurs.  If `nonmatchings_dir == "psp"` and `"psp"` is in part, such as `"psp" in "ric_psp"`, then it will give the wrong index.

tools/decompile.py
ProjectOblivion: In the `asm` directory, the `ric` overlay is identified by `ric_psp`, there is no `ric` directory.
By your logic, shouldn't that be fixed?

tools/decompile.py
Xeeynamo: I am not sure why some of the existing overlays use a different nonmatchings directory. These overlays need to be adjusted accordingly. We have tools that relies on that being part of the path.

ProjectOblivion: This I definitely agree on, I actually raised this question in discord a week ago: https://discord.com/channels/1079389589950705684/1087866009983139870/1358808728257953967

PR #2440, opened 2025-04-14
Decompile rbo5 func_us_801B5004

bryanr1992:
Thanks to everyone for the help

### DIFF
--- a/src/boss/rbo5/doors.c
+++ b/src/boss/rbo5/doors.c
@@ -527,12 +527,16 @@ void func_us_801B4A30(Entity* self) {
 extern s16 D_us_80180680[];
 extern s16 D_us_80180694[];
 void func_us_801B5004(Tilemap* map, s32 arg1) {
-    Tilemap* tmap = &g_Tilemap;
-
-    s16 tilePos = D_us_80180680[arg1 >> 1];
-    u16* tileData = &D_us_80180694[arg1 << 2];
-
+    Tilemap* tmap;
+    s16 tilePos;
+    u16* tileData;
     s32 i;
+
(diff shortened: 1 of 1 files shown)


sozud:
Run `make format` to fix the linter

bryanr1992:
Will do

src/boss/rbo5/doors.c
```diff
@@ -524,6 +524,19 @@ void func_us_801B4A30(Entity* self) {
...
-INCLUDE_ASM("boss/rbo5/nonmatchings/doors", func_us_801B5004);
+extern s16 D_us_80180680[];
+extern s16 D_us_80180694[];
+void func_us_801B5004(Tilemap* map, s32 arg1) {
+    Tilemap* tmap = &g_Tilemap;
+
+    s16 tilePos = D_us_80180680[arg1 >> 1];
+    u16* tileData = &D_us_80180694[arg1 << 2];
+
+    s32 i;
```
Xeeynamo, line 535: While the order of variables do not matter on PSX, they do for PSP and Saturn. We found often we had to separate the variable declaration from the actual assignment to be able to move them around. Let's do that here as well to align it with the other functions' style.

Xeeynamo (commented):
Minor nitpick, otherwise it looks good. Congrats on your first function match!

src/boss/rbo5/doors.c
bryanr1992: Understood. I will declare all the variables first before assignment from now on! Thanks for your input.

sozud (approved):
Thanks for contributing!

PR #2443, opened 2025-04-16
More BO4 & RBO5 functions

hohle:
Another round of duplictes between `BO4` and `RBO5`.

I found that `BO4` and `RBO5` needed `g_pfn_800EA5AC` to have a default signature. Instead of casting, I updated the struct to match. This works on PS and PSP, however, may not be desirable for portability later.


hohle:
Needed to learn how to use `sotn-assets`.

hohle:
Taking down while I figure this out.

PR #2444, opened 2025-04-16
Better match for `dra` ExecSoundCommands

gamezter:
I was testing merging dra `91EBC.c`, `92F60.c`, `93290.c`, `93BDC.c`, `94F50.c` and `953A0.c` (since `ExecCdSoundCommands` in `91EBC.c` needs `AdvanceCdSoundCommandQueue` in `953A0.c` to be static for PSP) but ran into an issue where `ExecSoundCommands` would mismatch when `PlaySeq` would be in the same file because it has the wrong signature (`PlaySeq(u8)` instead of `PlaySeq(s16)`). 
I got a better match for `ExecSoundCommands` by working on PSP https://decomp.me/scratch/CirCq (not 100% match, I suspect there are other missing static functions). This new PSX `ExecSoundCommands` works with `PlaySeq(s16)`.


Xeeynamo (approved):
An improvement is an improvement. Thanks for the submission!

PR #2445, opened 2025-04-16
PSP RIC: pl_blueprints

Xeeynamo:
A monster of PR. Matches several PSP functions and merge them with the PSX overlay. You can see how PSX functions changed by exploring the individual commits. I did not want to duplicate functions and merge them separately, therefore I am submitting everything at once.

The name `pl_blueprints.c` comes from the fact this function mostly creates entities from the blueprints. I am unsure if the name is really that accurate with the content, but it's the best I could find. I am surprised there are so many functions in one file, but cross-referencing PSX and PSP suggests that being the case. I could mark different functions as `static` that were not previously possible. Not all the self-contained functions can be `static` as the PSP code will start to not match. All the weird fake-ry code was removed too.

This decompiles around 24% of the RIC overlay. We're almost there!


src/ric/pl_blueprints.c
```diff
@@ -0,0 +1,3100 @@
...
+        }
+    }
+    if (D_80174FAC) {
+        D_80174FAC--;
+        if (D_80174FAC & 1) {
+            g_api.g_pfn_800EA5AC(1, D_80174FB0, D_80174FB4, D_80174FB8);
+        }
+    }
+    D_80174F80[1] = D_80174F80[2] = 0;
+    if (g_Entities[16].enemyId == 1) {
```
sozud, line 492: E_WEAPON? could fill out other e_Entities magic nums with enum
https://github.com/Xeeynamo/sotn-decomp/blob/7ae20702f5dac93037f8c1334cf83eec5d05588d/include/game.h#L2019

src/ric/pl_blueprints.c
```diff
@@ -0,0 +1,3100 @@
...
+    if (PLAYER.animCurFrame == 140) {
+        self->hitboxOffY = 0;
+    }
+
+    if (PLAYER.animCurFrame == 141) {
+        self->hitboxOffY = 12;
+    }
+
+    if (self->hitFlags) {
+        g_Player.unk44 |= 0x80;
```
sozud, line 841: might be good to start an enum for unk44 values

src/ric/pl_blueprints.c
```diff
@@ -0,0 +1,3100 @@
...
+        self->step++;
+    }
+}
+
+// Entity ID #2. Created by 6 blueprints:
+// 0, 1, 24, 74, 75, 76.
+// Matches DRA func_8011B5A4
+static s16 pos_x_80154C50[] = {0, -4, -8, -12, -16, -20};
+static s32 velocity_x_80154C5C[] = {
+    -0x3000, -0x4000, -0x6000, -0x8000, -0xA000, -0xC000};
```
sozud, line 899: I think these are FIX?

src/ric/pl_blueprints.c
```diff
@@ -0,0 +1,3100 @@
...
+                    tilePrim->velocityY.val =
+                        (i - 10) * (i - 10) * FIX(0.09375) + FIX(0.125);
+                }
+                tilePrim->delay = 0x3F;
+                break;
+            case 8:
+                tilePrim->posX.i.hi = x;
+                tilePrim->posY.i.hi = y;
+#if defined(VERSION_PSP)
+                tilePrim->velocityX.val = ((rand() & 0x7FFF) - FIX(0.25)) >> 1;
```
sozud, line 1183: could use randmask

sozud (approved):
This is huge so I just read through the first half, seems good

src/ric/pl_blueprints.c
Xeeynamo: Great catch! I did not notice we had an enum for that. I'll make a commit

src/ric/pl_blueprints.c
Xeeynamo: I will make a note for a separate PR. `unk44`, together with some other flags in `g_Player` are very complicated to understand. I will most likely integrate whatever enum I come up with the linter too.

src/ric/pl_blueprints.c
Xeeynamo: This unfortunately wouldn't work. Using `& 0xFFFF` on PSX gives a mismatch. If you notice, I used `PSP_RANDMASK` just a few lines before. But this here I couldn't make it work unfortunately.

Xeeynamo: image

I do not have a scratch at hand, but this is what I get locally

PR #2449, opened 2025-04-16
Decompile first PSP DRA file (80.c), except one func

bismurphy:
Frustrating to need to stop it here, but I figured this was at least good progress, enough to be worth contributing.

Last function is a monster, scratch here for the brave: https://decomp.me/scratch/yNOg6


Xeeynamo (approved):
I'll be merging this as soon as the CI queue is free. Great stuff as always

PR #2454, opened 2025-04-18
PSP RIC: pl_whip

Xeeynamo:
plus a small patch for that undefined behaviour, which is now better documented in the code base


src/ric/pl_whip.c
```diff
@@ -0,0 +1,548 @@
...
+        self->ext.whip.unk8C = 0x600;
+    }
+    if (self->step) {
+        diffX = self->posX.val - self->ext.whip.curX;
+        diffY = self->posY.val - self->ext.whip.curY;
+        self->ext.whip.prevX += diffX;
+        self->ext.whip.prevY += diffY;
+    }
+    aimPath = g_Player.padPressed & (PAD_UP | PAD_RIGHT | PAD_DOWN | PAD_LEFT);
+    if (aimPath == PAD_UP) {
```
sozud, line 107: I think when porting people might get bitten by this, what about 
```
#ifdef VERSION_PC
// you need to sanitize your inputs
assert(aimPath != PAD_LEFT + PAD_RIGHT);
assert(aimPath != PAD_LEFT + PAD_LEFT);
#endif
```
if we are not going to initialize psp_s0?

sozud (approved):

src/ric/pl_whip.c
Xeeynamo: Good point. I just added that. Better safe than sorry

PR #2457, opened 2025-04-18
Align `dra` sound functions + decompile most `dra_psp` sound functions

gamezter:


config/splat.us.dra.yaml
```diff
@@ -131,7 +131,7 @@ segments:
       - [0x1D1E0, .data, 91EBC] # g_SeqInfo
       - [0x1D224, assets, music_xa, g_XaMusicConfigs]
       - [0x1F554, assets, sfx, g_SfxData]
-      - [0x2097C, .data, 953A0]
+      - [0x2097C, .data, sfx_scripts]
```
gamezter, line 134: I had to move this data to a separate c file because 91EBC would otherwise be split by `g_XaMusicConfigs` and `g_SfxData` that are not inlined in the c file. Let me know if there is a better way to do this

Xeeynamo:
You need to add `sfx_scripts` in CMakeLists to fix the build error

config/splat.us.dra.yaml
sozud: I'm thinking there's two options better than what we currently have with the spat extensions:

Just move this data to C (this is probably what it was originally).
Add support to the assets tool and #include it like the other assets.

Either of those ways avoid the problem

@Xeeynamo What's your opinion?

sozud: I should add that you don't have to fix it for this PR, just saying in general

Xeeynamo: I agree with @sozud POV. Have everything as C data in its own file makes sense for this PR. I am happy to wrap that data with either macros or from structured data generated by the asset tool at some point.

> better than what we currently have with the spat extensions

Splat extensions are cool to kick-start a new decomp. But as this decomp is becoming more mature, extensions should be slowly phased-out.

sozud (approved):

PR #2461, opened 2025-04-20
Import `dra_psp` 63C08.c data

gamezter:


src/dra/91EBC.c
```diff
@@ -290,73 +350,14 @@ void KeyOnRange(s32 minVoice, s32 maxVoice, s16 vabId, s16 prog, s16 tone,
...
-
 void func_80132A04(s16 voice, s16 vabId, s16 prog, s16 tone, s16 note,
                    s16 volume, s16 distance) {
     if (!distance) {
         g_VolL = volume;
         g_VolR = volume;
     } else {
-        g_VolR = (volume * g_CdVolumeTable[distance * 2 + 144]) >> 7;
-        g_VolL = (volume * g_CdVolumeTable[distance * 2 + 145]) >> 7;
+        g_VolR = (volume * (D_800BD17C + 8)[distance][0]) >> 7;
```
sozud, line 359: What happens if you add a symbol at D_800BD17C + 8 and leave the earlier part unused? I'm guessing it would get stripped on PSP?

gamezter: It does get stripped, I think distance is supposed to go from [-8, +8]. I tried `D_800BD17C[distance + 8][0]` which matches on PSP but doesnt match on PSX

PR #2465, opened 2025-04-20
Complete code-share for first C file in DRA PSP

bismurphy:
Took me a while since this is all O4, but nice to see another one done!


src/dra/62EB8.c
```diff
@@ -56,72 +66,117 @@ void func_801030B4(s32 arg0, POLY_GT4* poly, s32 arg2) {
...
+#elif defined(VERSION_PSP)
+#define CONFIRM (D_psp_08B42050 | PAD_START | PAD_SQUARE)
+#define EXIT D_psp_08B42054
 #endif
-    Primitive* prim1;
+
+// Somewhat clumsy way to avoid ifdef throughout the code.
+// V86P70 means "Value is 86, except on PSP it's 70"
+#ifdef VERSION_PSP
+#define V86P70 70
```
sozud, line 141: I think these should be renamed like `CONFIRM_UPPER_LEFT_POS` or whatever to have actual meaning but it can be documented later

sozud (approved):

PR #2466, opened 2025-04-21
Do some misc PSP tidy ups

JoshSchreuder:
A few miscellaneous things I noticed while working on CHI PSP that I wanted to fix up before I forgot.


include/entity.h
```diff
@@ -2067,7 +2067,7 @@ typedef struct {
 // Elevator at the top of CEN.
 // Exists in both CEN and NO0 (lowers you into CEN)
 typedef struct {
-    /* 0x7C */ struct Primitve* prim;
+    /* 0x7C */ struct Primitive* prim;
```
JoshSchreuder, line 2070: Spelling mistake I noticed in decomp.me (forward declaration warning)

I'm surprised this wasn't causing any issues?

sozud: I think what's happening is that if you don't do prim->whatever it's just an opaque extern pointer. `cenElevator` refer dereferences anything inside

sozud (approved):

PR #2463, opened 2025-04-20
Import `dra_psp` sfx_scripts

gamezter:
Leaving as draft because I can't get it to code-share. `sfx_scripts` is identical for PSX and PSP but using `dra/sfx_scripts` instead of `dra_psp/sfx_scripts` in the splat config does not work

PR #2469, opened 2025-04-22
PSP RIC: decompile init function + data

Xeeynamo:
A very small function that initializes the overlay on PSP. I figured it was a good time to import data for some quick wins.


Makefile
```diff
@@ -129,7 +129,7 @@ define get_conf_merged
 	import os;\
 	yaml_file=open("$(CONFIG_DIR)/splat.$(VERSION).$(2)$(1).yaml");\
 	config = yaml.safe_load(yaml_file); yaml_file.close();\
-	c_subsegments = [x for x in config["segments"][1]["subsegments"] if type(x) == list and x[1] == "c"];\
+	c_subsegments = [x for x in config["segments"][1]["subsegments"] if type(x) == list and (x[1] == "c" or x[1] == ".data")];\
```
sozud, line 132: Is this related to this problem? https://github.com/Xeeynamo/sotn-decomp/pull/2463#issuecomment-2819394851

src/ric/pl_header.c
```diff
@@ -11,8 +11,15 @@ PlayerOvl RIC_player = {
     RicInit,
     RicUpdatePlayerEntities,
     RicGetPlayerSensor,
+#if defined(VERSION_PSP)
+    (u8**)0x092703C0,
```
sozud, line 15: It's not possible to use a symbol here?

sozud (approved):

Makefile
JoshSchreuder: Actually I had this problem too a couple of days ago trying to codeshare [d_prize_drops.c](https://github.com/Xeeynamo/sotn-decomp/pull/2466/files#diff-63acc9dff5f75687f1a214d7eca5711d6dc78f9f0b4e5e121197832f54cad811), wasn't possible since it was in data. I didn't look into what would be required for a fix

src/ric/pl_header.c
Xeeynamo: I didn't think of it. Good point. I submitted a commit.

Makefile
Xeeynamo: I'll be merging this now. Hopefully it will unblock both of you

JoshSchreuder: Actually my use case was merging a PSP BSS -> PSX Data
I think it's probably technically possible but needs further changes to this Makefile script. It's not that important I think, so I'll just leave it as a separate source file on PSP for now

PR #2472, opened 2025-04-23
Import `dra_psp` 63C08.c bss

gamezter:


src/dra_psp/63C08.c
```diff
@@ -1,20 +1,108 @@
...
-extern s32 D_psp_09236880;
-extern char D_psp_09236888[];
-extern s32 D_psp_09236910;
-extern s32 D_psp_09236920;
-extern s32 D_psp_09236928;
-extern s16 D_psp_092374B0;
-extern s16 D_psp_092374B8;
-extern SpuVoiceAttr* D_psp_09237578;
+
+s32 D_8013B61C;
```
sozud, line 9: Note that I don't think we can combine this with PS1 due to psy-q bss hashtable thing

sozud (approved):

PR #2475, opened 2025-04-23
Align `dra_psp` 61E38.c to `dra` 66590.c

gamezter:
I couldnt get a full PSP match for `DrawEntitiesHitbox` because it uses `g_CurrentBuffer->tiles` which is one of the last elements in the struct and depends on the sizes of all the elements before it, which are currently unknown on PSP. Here's my current scratch : https://decomp.me/scratch/tjxvc

edit: I found this PSP scratch https://decomp.me/scratch/qY61I that touches nearly all members of the `GpuBuffer` struct. It turns out most of the element counts are the same for PSX / PSP apart from `ot`


Xeeynamo (approved):
Looks good. But I wonder if it's possible to merge dra/66590.c with dra_psp/61E38.c before merging?

I find great to see a joint effort on tackling DRA PSP. Thanks for your contribution :pray:

PR #2479, opened 2025-04-24
Full code-share for Primitive handling functions (including RenderPrimitives!)

bismurphy:
A fun and useful one - lots of functions looking more readable now. Took a whole lot of commits, but very satisfying to see it come together.


sozud (approved):

bismurphy:
> I think it would be better to `#undef` the various macros in `libgpu.h` and re-define them differently in `game.h` (at least for now) whenever VERSION_PSP is defined. We should try to minimize as much as possible anything in `include/psxsdk` and preserve the original headers.
> 
> Structs such as `P_TAG` are more complicated and it is fine to leave it like that for now.

I played with this a bit and couldn't get them to match. For example the first mismatch, DR_ENV_CODE_SIZE, I put into game.h a block of:
```
#if defined(VERSION_PSP)
#undef DR_ENV_CODE_SIZE
#define DR_ENV_CODE_SIZE 8
#endif
```

but that doesn't match, I think because of the setting that the variable holds at the time it's being used in DR_ENV. By the time you undef in game.h, it's too late.

Perhaps we want to make a totally separate libgpu.h for PSP, and put the conditional on the include statement?

PR #2484, opened 2025-04-24
Fixed file listing for fully merged psp overlays

ProjectOblivion:
I also took the opportunity to pull the python out of Makefile and put it in it's own .py file.


ProjectOblivion:
The issue is not with `get_o_files`, though that is where it manifests.  The issue with maria is that it doesn't follow the structure of psp overlays with `asm_path: asm/pspeu/ovl_psp`.  Fix that and those files will be found.

PR #2486, opened 2025-04-24
Saturn RIC: decompile some small functions

Xeeynamo:
I am kick-starting the decompilation for the Richter overlay on Saturn. The process is quite tedious as I need to constantly look at the data section at the bottom of the function to understand values and offsets.

The content in `src/saturn` is not picked up by make, so I am running `rm build/saturn/richter.*; rm build/saturn/RICHTER.PRG; cp src/saturn/* build/saturn/; make` to check if the overlay matches locally after making a change.

Functions are barely different from PSX and PSP, it's mostly copy&paste on decomp.me and fill the gaps with enums or missing struct fields. `RicSetSpeedX` is inlined, which is an interesting optimization compared to PSX and PSP.

The various `FIX` values that works horizontally are all different, I assume to accommodate the change in the horizontal resolution from 256 to 320. I suspect we could define an additional `FIX` with a fixed multiplier baked in to adjust all the values seamlessly (hopefully the OG devs used something similar). For now I am leaving the raw un-FIX-ed values as I find a bit hard to work them out without an utility.


sozud (approved):
Great work, good to see saturn progress

PR #2487, opened 2025-04-25
Clean up `RenderTilemap`

gamezter:


include/psxsdk/libgpu.h
```diff
@@ -1,27 +1,28 @@
...
 #define LINEG2_CODE 0x9
 #define TILE_CODE 0xB
+#else
+#define SPRT_CODE 0x64
+#define SPRT16_CODE 0x7C
+#define POLYGT3_CODE 0x34
+#define POLYG4_CODE 0x38
+#define POLYGT4_CODE 0x3C
+#define LINEG2_CODE 0x50
+#define TILE_CODE 0x60
```
gamezter, line 25: I can revert these if it's an issue. IMO `#ifdef ...` is much easier to parse than `#ifndef ...`

PR #2488, opened 2025-04-25
Saturn RIC: Add symbols and split hints

Xeeynamo:
Document almost all functions from the overlay. Almost all functions have been thoughtfully verified manually. The objective is to allow multiple contributors to work on the same file more easily.

I only verified a fraction of the entities from the Richter entity list. I found the list is 1:1 with PSX/PSP and that was quite straightforward.

Some undocumented functions naturally follow the same function order from PSX, it was a bit redundant to rename them. I named most of them to give guidance on the order and the suggested file split.

All the functions at the bottom of the overlay are exclusive to Saturn, and I have no idea what their purpose is.


sozud:
Yeah sorry it's not set up to handle renames before decompilation at the moment, functions can be renamed once they are decompiled

sozud (approved):

PR #2493, opened 2025-04-26
PSP: Add maria overlay

Xeeynamo:
As decompiling the Richter overlay for both PSX and PSP gave us precious hints about differences, more accurate matches and file splits, I am now kick-starting this exclusive PSP overlay with a good amount of functions and splits already documented. Hopefully having the file pre-split will stimulate collaboration.


sozud (approved):
Let's add it to `dups`, otherwise looks good

sozud:
You're right, I forgot we weren't doing PSP. I think this overlay would be a good reason to get dups running on PSP though.

PR #2500, opened 2025-04-28
Decompile `no4` `EntityBat`

JoshSchreuder:
PSX: https://decomp.me/scratch/F8yPX
PSP: https://decomp.me/scratch/fqCHE


src/st/no4/e_bat.c
```diff
@@ -0,0 +1,107 @@
...
+#include "no4.h"
+
+static u8 bat_anim_fly[] = {4, 21, 1, 22, 1, 23, 1, 30, 1, 24, 1, 25, 4, 26,
+                            2, 27, 2, 28, 2, 29, 1, 30, 2, 23, 2, 22, 0, 0};
+static u8 bat_anim_drop[] = {5, 31, 5, 32, 5, 31, 5, 32, 5,  31, 5,   32, 4, 31,
+                             4, 32, 3, 31, 3, 32, 2, 31, 12, 32, 255, 0,  0, 0};
+
+// This is a slightly modified version of the Bat enemy
+// present in other overlays. Has a different SFX and some
+// slightly enhanced movement behaviour.
```
JoshSchreuder, line 11: This is mostly pretty similar to the existing shared header for Bat enemy used in NO3 / NP3, however there were enough differences (IMO) that putting conditionals into that header just to share code made it harder to read.

Here's an example scratch of how that might look in situ:
https://decomp.me/scratch/pG66O

If it is preferably I can update the shared header and use this approach instead - all data is identical.

sozud: I could go either way, it's not essential to dedupe it for this PR IMO

PR #2503, opened 2025-04-28
RIC+MARIA blueprints as assets

Xeeynamo:
Sample in `assets/ric/blueprints_def.yaml`:

```yaml
BP_SUBWPN_CROSS_PARTICLES:
  entity: E_80169C10
  amount: 3
  nPerCycle: 1
  isNonCritical: true
  incParamsKind: true
  timerCycle: 2
  kind: B_DECORATION
  origin: B_ORIGIN_DEFAULT
  timerDelay: 0
```

compiled in `src/ric/blueprints_def.h` as:

```c
B_MAKE(E_80169C10,3,1,true,true,2,B_DECORATION,B_ORIGIN_DEFAULT,0),
```

Long-term, this will help a lot in documenting blueprints and entities. The exported asset is synchronized with the enums, making everything more seamless and connected.


src/ric/pl_blueprints.c
```diff
@@ -561,103 +561,21 @@ Entity* RicCreateEntFactoryFromEntity(
...
 }
 
-#if defined(VERSION_US)
-#define NON_CRITICAL false
-#else
-#define NON_CRITICAL true
-#endif
 static FactoryBlueprint blueprints[] = {
-    // clang-format off
-    B_MAKE(E_SMOKE_PUFF, 5, 1, true, true, 2, B_DECOR, 0, 0),
```
sozud, line 571: I think this isn't as clean as some of the other yaml or json based things since there's a dependency on the entity function pointer array and enum but it's not necessarily an issue

sozud (approved):

src/ric/pl_blueprints.c
sozud: My understanding is to add an effect that uses a new entity I'd have to:

1. Add the entity function to the array of function pointers
2. Create an enum value at the same position
3. Use that in the yaml here

I don't think anything needs to be changed, was just pointing out it's not as "pure data" as some other examples.

PR #2505, opened 2025-04-28
Align blueprints across overlays

Xeeynamo:
First of all, I am sorry for the big PR.

I am now importing as an asset the blueprints from RIC, DRA and MARIA across all versions in the decomp. I took this opportunity to align enums, entity names and documentation. DRA never had a blueprint enum; most of the changes in this PR comes from its introduction.

The most important change was to **duplicate** the four enums `Entities`, `Blueprints`, `BlueprintKinds` and `BlueprintOrigins` across the three overlays. While DRA and RIC share mostly the same structure when spawning entities using a Factory Pattern, the actual implementation and the meaning of each value changes between them. Even the macro `B_MAKE` is different between RIC and DRA. The MARIA overlay is mostly identical to RIC, so enums+docs are copy&pasted.

I noticed BO3 and RBO3 might need the same treatment I did to DRA (cc @hohle) since entities, blueprints and factory are duplicated there as well for the Doppelganger bosses.


src/maria/maria.h
```diff
@@ -218,6 +218,91 @@ enum MarBlueprints {
     NUM_BLUEPRINTS,
 };
 
+// NOTE: B_MAKE is different between DRA and RIC!
```
sozud, line 221: maybe a comment explaining if it's the same for ric and maria?

sozud (approved):

PR #2506, opened 2025-04-29
Code-Share `dra` 4AEA4.c

gamezter:


src/dra/4AEA4.c
```diff
@@ -116,8 +175,12 @@ static s32 DecompressData(u8* dst, u8* src) {
...
         case 0:
             ch = DecompressReadNibble();
-            op = DecompressReadNibble();
-            count = (ch << 4) + op + 0x13;
+            ch2 = DecompressReadNibble();
+#ifdef VERSION_PSP
+            count = ch2 + (ch << 4) + 0x13;
+#else
+            count = (ch << 4) + ch2 + 0x13;
+#endif
```
gamezter, line 183: Could not figure out a way to get this to match on both platforms

JoshSchreuder: Does `len = (len * 16) + op + 0x13;` work?

sozud (approved):

PR #2508, opened 2025-04-29
Refactor/get lang

gamezter:
Depends on https://github.com/Xeeynamo/sotn-decomp/pull/2506

`GetLang` seems to be a static O0 function included into multiple files. We can tell because some overlays have multiple GetLang-like functions and they are always O0.
This PR updates `get_lang.h` so that `GetLang` is static and 'merges' the multiple `lang.c` files into the following c file. The only exception seems to be `st0` where `GetLang` is used 4 files after it's definition, but maybe that's a clue that all of the files are actually merged together.


sozud:
Rebase

PR #2518, opened 2025-05-02
saturn ric func_8015CC28

sozud:


sozud:
Yeah I think once we confirm a few more offsets it makes sense to use regular entity.h

PR #2521, opened 2025-05-02
Decompile `dra_psp` 1DDC0.c

gamezter:
There's still work to do on this file, but I think this is a good place to make a PR


src/dra_psp/1DDC0.c
```diff
@@ -1,8 +1,975 @@
...
+                    SIM_PTR,
+                    0x00002000,
+                    SIM_17,
+                },
+                {
+                    "hdbin/C_GOB_EU.BIN;1",
+                    SIM_PTR,
+                    0x00002000,
+                    SIM_17,
+                }};
```
gamezter, line 936: @JoshSchreuder You were right that this is inline data ;)

PR #2522, opened 2025-05-03
Dedupe `no4` `st_init`

JoshSchreuder:
~~Also cleaned up some of the function naming I noticed~~


src/st/st_debug.h
```diff
@@ -18,9 +18,11 @@ void EntityBackgroundBlock(Entity* self) {
         if (objInit->flags) {
             self->flags = objInit->flags;
         }
+#ifndef STAGE_IS_NO4
```
JoshSchreuder, line 21: This code should be shareable with this change in more places (BO4, etc.) but I have not done so for now

PR #2525, opened 2025-05-03
Match NO4 func_us_801c86f4 (aka EntityBreakableCrystalFloor)

SestrenExsis:
In room ID 1 of Underground Caverns (which I call Hidden Crystal Entrance), there is a breakable floor at the bottom made of crystal (or maybe it's ice?). After striking it three times, the tiles completely break away, leaving a hole that the player can fall through to enter the room below.

The function references an array of data that stores the tilemap IDs to use when updating after each break attempt. This array should look like the following:
```c
static s16 D_us_80181880[] = {
    // Phase 0
    0x054C, 0x054D, 0x054E,
    0x0180, 0x0551, 0x0552,
    // Phase 1
    0x054C, 0x0740, 0x0748,
    0x0180, 0x0551, 0x0552,
    // Phase 2
    0x054C, 0x074D, 0x074E,
    0x0180, 0x0551, 0x0552,
    // Phase 3
    0x055E, 0x0000, 0x0000,
    0x055F, 0x0000, 0x0000
};
```
I have not tried to add this array as part of the PR, but am identifying it in case it proves helpful for someone else later.

Special thanks to @JoshSchreuder for helping me figure out the final bit of the match on this function.


src/st/no4/no4.h
```diff
@@ -136,3 +136,14 @@ extern s16 D_us_801DF788;
...
 extern s8 D_us_801DF78F;
+
+// *** EntityBreakableCrystalFloor properties START ***
+
+typedef enum {
+    BREAKABLE_CRYSTAL_FLOOR_INIT,
+    BREAKABLE_CRYSTAL_FLOOR_IDLE,
+    BREAKABLE_CRYSTAL_FLOOR_BREAK,
+    BREAKABLE_CRYSTAL_FLOOR_CHECK
+} BREAKABLE_CRYSTAL_FLOOR_STEPS;
```
bismurphy, line 147: These kinds of enum, which are only used for a single entity in its own file, should be in that file, rather than in the shared header for every entity in the stage.

include/entity.h
```diff
@@ -1841,6 +1841,13 @@ typedef struct {
     /* 0x9C */ s16 rotSpeed;
 } ET_BreakableWallDebris;
 
+// NO4 Room 1, Bottom, Breakable Crystal Floor
+typedef struct {
+    /* 0x7C */ char pad_7C[0x4];
```
bismurphy, line 1846: Take a look at the other Ext entries for how they write unused pad members; let's use the `    s32 : 32;` syntax here rather than creating named entries.

bismurphy (approved):
Nice job, seems good to me.

PR #2526, opened 2025-05-04
Import data for `EntityBreakableCrystalFloor`

JoshSchreuder:
For @SestrenExsis to show an example

I also split the file up - it appears that the 2 functions after this are related to another secret wall in `no4` based on PSP splits.


config/splat.us.stno4.yaml
```diff
@@ -43,6 +43,8 @@ segments:
       - [0xCF4, data]
       - [0xD5C, .data, st_debug]
       - [0xDDC, data]
+      - [0x1880, .data, e_secrets]
+      - [0x18B0, data, e_secrets]
```
JoshSchreuder, line 47: `.data` marks this section as imported into C code. This is for `D_us_80181880`

`data` section (without the .) is not yet imported and relates to `D_us_801818B0` which is used by `func_us_801C89F4` in the same `e_secrets` file

sozud (approved):

PR #2528, opened 2025-05-05
Next PSP DRA code share file (mostly related to game map)

bismurphy:
Yet another file code shared. PSX seems to have some nice improvements and de-faking as well.

The RunMainEngine function is complicated and has a lot of little differences. I started to ifdef it, but it got ugly fast, so I decided to pull it out to a separate .h file. Seems like the cleanest solution. The alternatives would be to have 50 different spots of ifdef, or to copy both of the functions in full with an ifdef between them, so I figured a dedicated file would minimize clutter.


bismurphy:
Oh man I don't know what I did but I rebased and now there are over 300 new files pulled into this thing???

sozud:
The gitignore is messed up or something, I think those are all generated files

bismurphy:
Ah, I think the issue is that the gitignore was modified between when I originally made this branch and when I rebased, so I had all the new files and the rebasing didn't respect the not-yet-updated gitignore file.

I think I'll start over, this is becoming a mess.

PR #2533, opened 2025-05-07
Add NO3 overlay to PSP version

bismurphy:
Took a bit of fighting to get this to work, but now we have a working NO3 overlay for PSP.

I intend to chip away at this slowly when I get time to do decomp stuff. Should be a nice quiet corner :)


bismurphy:
> Should be added to progress.py otherwise seems good

I'm not seeing PSP overlays referenced in progress.py, am I missing something?

sozud:
Nevermind, I forgot that the version is implicit

PR #2539, opened 2025-05-07
Decompile `dra_psp` play.c

gamezter:


src/dra_psp/play.c
```diff
@@ -1,16 +1,679 @@
...
+extern s32 D_8C630D4;
+extern s32 D_8C630D8;
+extern u32 D_8D35C40;
+extern u32 D_8D45C40;
+extern u32 D_8D47C40;
+extern u8 D_psp_0914C318[];
+extern RECT D_psp_0914C328[];
+extern char D_psp_0914C3B8[];
+extern char D_psp_0914C3C8[];
+extern char* D_psp_0914C3D8[];
```
gamezter, line 14: These are compiler pooled strings for `"alucard "`, `"richter "` and `"maria   "` that `loading.c` uses, indicating that play.c and loading.c are the same file

PR #2550, opened 2025-05-09
Misc signature and future-proof fixes

Xeeynamo:
I added to the CI the compilation using std c99 and c17 (which also includes c11). This revealed a cascade of problems that were ignored when using GNU extensions. Apparently we cannot use -std=c23 (default on GCC 15) as `-Wno-implicit-function-declaration` is no longer effective due to to stricter requirements from the recent C standard, I couldn't include that very easily.

This is important as GCC 15 uses c23 as default. It's not yet the default compiler for the most stable distros, but it is good to start looking ahead and make the code more portable. At the same time we want to ensure the code gets compiled on the old GCC 2.6.x and more compilers, so this PR makes a few changes to enable this.

I am not sure if we should use `-std=gnu90` for non-PSP/PSX code. To enable it, the only change would be to adjust the various `for (int i = ` to avoid a loop initial declaration. 

The different C standards are only tested on Ubuntu. I think that covers enough scenarios.


sozud (approved):
> I am not sure if we should use -std=gnu90 for non-PSP/PSX code. To enable it, the only change would be to adjust the various for (int i = to avoid a loop initial declaration.

IMO this restriction is just annoying so unless there's a compiler that requires it I don't think it's necessary

PR #2551, opened 2025-05-09
Import `dra_psp` sfx_data

gamezter:
I started working on this thinking I could merge `g_SeqInfo`, `g_XaMusicConfigs`, `g_SfxData` and `g_SfxScripts` under the same file , but quickly realized that `g_SfxData` is different for each platform. Here I get rid of the splat_ext import of `g_SfxData` and manually import the data for PSX, HD and PSP, but ideally this would be imported by an asset importer and included into the same file. Leaving as Draft to get some feedback


gamezter:
I agree that they should be handled by an asset importer, I'm not very familiar with Go so I think it would be better if you handled it. As for the 'json loading stuff' I'm not exactly sure what that code is for and  how to test it.

src/pc/sotn.c
```diff
@@ -245,15 +245,15 @@ bool InitGame(struct InitGameParams* params) {
     FileOpenRead(InitPalEquipIcons, "assets/dra/g_PalEquipIcon.bin", NULL);
     InitVbVh();
 
-    if (!FileAsString(InitSfxData, "assets/dra/sfx.json", NULL)) {
+    /*if (!FileAsString(InitSfxData, "assets/dra/sfx.json", NULL)) {
```
sozud, line 248: This can just be removed rather than commented out

sozud (approved):

PR #2554, opened 2025-05-10
Change path for generated code from gen_ to gen/

Xeeynamo:
Suggested by @gamezter, I am changing the path for the generated source code files to isolate them in the directory `gen/` instead of adding the prefix `gen_`. This will help to avoid having loads of unwanted files between the code we want to maintain.

All the changes from `os.WriteFile` to `util.WriteFile` in Go is to use a custom wrapper that creates the targeted directory if it does not exist.

Some Makefile adjustments had to be made to evaluate the source code generated in `gen/`


fivemeyestoreserver:
Interesting solution to this problem! I've been working on similar challenges recently.

sozud (approved):

sozud:
This is large so let's get this in before more conflicts show up

PR #2556, opened 2025-05-11
Begin NO3 PSP with creation of st_init_no3.c

bismurphy:
Planning to do NO3 PSP in linear order. This represents the PSP initialization C file that appears to be in every PSP stage. It does not appear to always be the first file in the stage, but in this one it is.

Naturally this had some other consequences on the rest of the codebase.

EntityBackgroundLightning is in a weird place, so that has its own .h file now.

Perhaps it would make sense to have a `src/st/entities/` directory to hold all the individual .h files (and in the future, .c files) for various entities? Not sure if it makes sense to have overlays like `src/st/no3` at the same level as entities. But that's a different PR :)


src/st/e_background_lightning.h
```diff
@@ -0,0 +1,237 @@
...
+    }
+#if defined(STAGE_IS_NO3)
+    if (self->params == 1) {
+        switch (self->ext.backgroundLightning.unk84) {
+        case 0:
+            g_api.PlaySfx(SET_UNK_90);
+            self->ext.backgroundLightning.unk84++;
+            D_80097928 = 1;
+#ifdef VERSION_PSP
+            D_80097910 = MU_METAMORPHOSIS_III;
```
sozud, line 205: The SFX offset is an ongoing issue, it doesn't have to be in this PR but we ought to create an enum and add or remove a frew entries to get rid of this ifdef

sozud (approved):

src/st/e_background_lightning.h
bismurphy: Ah okay so it's not a different track, just an offset - I'll change the PSP version to use a hex constant for now instead of a false ID

src/st/no3_psp/cutscene_data.c
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "../no3/no3.h"
+#include <cutscene.h>
+
+u8 cutscene_data_it[] = {
+#include "gen_cutscene_data_it.h"
```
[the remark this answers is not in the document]
bismurphy: Ah didn't see that had been merged, nice improvement! Will find-and-replace. Thanks!

PR #2557, opened 2025-05-12
Replace Makefile with Ninja

Xeeynamo:
## Why Ninja

[Ninja](https://ninja-build.org/) is a fast build system originally developed for Chromium. It’s a bare-bones assembler rather than a scripting language like Make [phylosophical overview](https://ninja-build.org/manual.html#_philosophical_overview). I’ve long considered a similar approach. While I initially planned to create a custom solution, I was inspired by the work in #2499 and realized that it would have been simple enough to build a proof of concept.

It consists of a simple `build.ninja` file containing basic rules to generate outputs from inputs under defined conditions. Since the build "script" is minimal, the file is not committed to version control but instead generated by `tools/build-system/gen.py`. This approach gives us much more flexibility in specifying what files to build and how to build them.

As the project is expected to grow significantly, we need a build system that can handle rapid and reliable iteration. Having the generator script written in Python will also simplify debugging and troubleshooting.

## The desire of sunsetting Makefile

The current build system has many issues. Running `make clean && make -j extract && make -j build` is a brute-force workaround for a difficult problem. Because of it, the build speed currently has unacceptably long build times, discouraging contributors. The Makefile has grown to an unmanageable level of complexity, making reviews difficult and limiting understanding to just a couple of contributors. Additionally, Make 4.4 (my current version) is known to be problematic, causing targets to rebuild unnecessarily or relying on hardcoded paths that are difficult to change without breaking multiple dependencies.

We should still retain a very minimal Makefile to handle basic setup tasks, such as generating the Ninja build file and invoking it.

## Development

This is very much a work in progress. I will continue updating this PR. Once every version reliably gets an :ok:, I’ll replace the underlying Makefile logic and mark the PR as ready. Much of the current challenge lies in porting the splat assets, which will eventually be phased out in favor of the asset tool.

One of the biggest pain points in the existing build system is that asset extraction/build only starts after all splat extractions are complete, and code compilation only begins after the assets are built. This severely limits parallelism on faster CPUs. The new build process allows splat extraction, asset extraction, asset building, and code compilation to happen in parallel.

* [X] c and asm from splat fil
* [X] link elf and produce final bin
* [X] check sum1
* [X] asset tool
* [x] `memcard_*.png` (wondering if I should just port it to the asset tool)
* [X] dra (dirt patches not applied)
* [x] servants (padding not added)
* [x] weapon (linker script very different)
* [X] st/mad (beta symbol list)
* [X] st/sel (handwritten assembly)
[...]

PR #2562, opened 2025-05-13
Decomp/codeshare PSP NO3 Entrance Weights

bismurphy:
This is the pair of weights in the upper right next to Cube of Zoe which opens the shortcut to Marble Gallery. Not too terribly many surprises in the process of solving this, but some small code details cleared up by PSP.

Too bad the built in diffing can't identify the old versus new code when it gets pulled into a .h file.


bismurphy:
Hmmm actually the file got committed to master in my last PR, so I think this PR is deleting the file, it just shows up weird in the diff viewer since it's a deletion of an empty file.

sozud (approved):

bismurphy:
> > Too bad the built in diffing can't identify the old versus new code when it gets pulled into a .h file.
> 
> This is a very annoying issue I noticed since we started sharing the code, forcing us to just hit a CTRL+S on the C file for differ to pick up the changes.
> 
> I am working on #2557, which also aims to fix the issue

Oh no sorry, when I said "the built in diffing", I meant github's "Files changed" tab, not anything about the diff tools.

When code moves from a given .c file to a .h file, and, for example, I pull in an enum value for a hard-coded hex value, you can't see that I made that change, because the whole .h file shows up as green "new code". Even in this PR - you can't see what I did to modify the code to make it PSP-compatible. I had to add a few temp variables here and there. In a normal PR review you can compare old versus new code to confirm it's better, but when the whole file is shown as changed, you can't see that.

PR #2565, opened 2025-05-13
Decompile/Codeshare NO3 Cavern Door

bismurphy:
Another relatively straightforward one. One thing I'm not sure of with our infrastructure, will leave a PR comment.


src/st/e_cavern_door.h
```diff
@@ -0,0 +1,250 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#ifdef VERSION_PSP
+extern s32 E_ID(CAVERN_DOOR_PLATFORM);
```
bismurphy, line 3: Is this how I should deal with this? The E_ID syntax is clever but we need to make sure we don't declare an s32 of E_ID when we're on PSX, since PSX needs it to be in an enum, so it can't be a variable.

JoshSchreuder: Yeah we have just been #ifdefing this for PSP as you did, eg.
https://github.com/Xeeynamo/sotn-decomp/blob/1ec982a822f412c5cef66315e5f7dc2e89ba2d84/src/st/chi/en_demon_switch_wall.c#L11-L13

bismurphy: Okay perfect, thanks!

sozud (approved):

PR #2566, opened 2025-05-14
Import `dra_psp` d_E294.c

gamezter:


src/dra/d_E294.c
```diff
@@ -1,1317 +1,1478 @@
...
+static AnimationFrame D_800AE2DC[] = {
+    POSE(255, 0x4F, 1),
+    POSE_LOOP(0),
 };
 
-static u32 D_800AE2DC[] = {
-    0x024F00FF,
-    0x00000000,
+static AnimationFrame D_800AE2E4[] = {
+    POSE(-3, 0x10, 0), POSE(-3, 0x11, 0), POSE(6, 0x13, 1),
```
gamezter, line 25: `POSE(-3....` seems to be a special case that probably deserves a separate macro: https://github.com/Xeeynamo/sotn-decomp/blob/1b9424aa6cdb605abc7908b08ae86f4cccf03f94/src/dra/alu_anim.c#L159

gamezter: Maybe `SUBPOSE()` ?

sozud: Makes sense to me, either as part of this PR or as a followup

sozud (approved):

PR #2577, opened 2025-05-18
Renaming/Updating/Improving for NO3

bismurphy:
- Added git clean for the new `gen/` directories. Not cleaning these was giving me weird issues with stale data.
- Split out a new file "sky entities", representing the Death sky-swirl (and its helper), as well as the lightning thunder and cloud entities. This was done on both versions.
- Filled out the full EntityIds enum, making the names of the enum members match the names of the functions that they correspond to. This also makes it much easier to read the e_layout/e_laydef files since they sample from the enum.
- Updated some function names to be more descriptive - for example, what was "Explosion 2" is now the merman's fire spit.
- Updated mwccgap to latest. Not necessary for any particular features, but I was getting weird stuff with shift-jis which turned out to be related to sotn-str parsing and was resulting in a mwccgap error.


bismurphy:
Weird that the PR run is failing due to creation of src/ric/gen/richter.h, not sure what I changed that would have caused this. All the cleaning should be happening before that file would be created.

bismurphy:
Oh wait I think I get it, the Gen files have to be extracted from the Playstation game and then are used in the Linux build, but part of the Linux build is to clean, so by cleaning, the Gen files are lost. Need a solution, maybe we have a "soft clean" that doesn't do Gen wiping, and use that for Linux and other builds.

bismurphy:
I'll revert the clean changes for now, I think it's a known thing now so it doesn't matter if it's in this PR or a little later on

PR #2581, opened 2025-05-19
NO3 PSP EntityExplosionPuffOpaque

bismurphy:
This one cleaned up nice, huge reduction of temp variables.

As usual, prepped for next file decomp by getting the splits lined up.


src/st/np3/e_explosion_puff_opaque.c
```diff
@@ -1,4 +1,21 @@
...
+static u8 D_80180FC0[] = {
+    0x02, 0x01, 0x02, 0x02, 0x02, 0x03, 0x02, 0x04, 0x02, 0x05,
+    0x02, 0x06, 0x02, 0x07, 0x02, 0x08, 0x02, 0x09, 0x02, 0x0A,
+    0x02, 0x0B, 0x02, 0x0C, 0x02, 0x0D, 0x02, 0x0E, 0xFF, 0x00,
+};
+static Unkstruct_80180FE0 D_80180FE0[] = {
+    {0x000E, 0x0079, 0x0000, 0x30, D_80180FA4},
+    {0x8006, 0x0053, 0x0000, 0x00, D_80180FC0},
+    {0x8006, 0x0053, 0x0003, 0x00, D_80180FC0},
+    {0x000E, 0x0079, 0x0000, 0x30, D_80180FA4}};
```
JoshSchreuder, line 19: Can we move data into the header if it's imported (and especially if it's the same) for all overlays?

Not sure if there's any thoughts on this from the wider group but I think having the data defined in with the code is preferable personally

bismurphy: That's the thing - it's not the same across all overlays.

The animation arrays above are the same, but the four lines at the bottom have enough variation that #ifdef for all of them would be clumsy.

JoshSchreuder: Ah ok, no worries then!

sozud (approved):

PR #2586, opened 2025-05-21
Several More `BO4` and `RBO5` Functions

hohle:
`OVL_EXPORT(CheckFloor)` and `OVL_EXPORT(CheckCeiling)` are Doppleganger versions of the same functions in `DRA`. Like most of these parallel functions, points to different entity, "player" state, and global data. There is some difference in logic since Doppleganger cannot transform into wolf, among other things. These rely on known global data and were able to be shared trivially.

Several other matching functions. Some are the same with different global references. Others are slight variations. When data is imported, these can be merged.


src/boss/bo4/unk_365FC.c
```diff
@@ -1,5 +1,181 @@
...
+        MoveEntity();
+        self->velocityY += FIX(0.125);
+
+        blockIndex = self->params & 0xFF;
+        blockYPos = ((3 - blockIndex) * 16) + 96 + g_Tilemap.scrollY.i.hi;
+
+        if (blockYPos < self->posY.i.hi) {
+            if (blockIndex != 4) {
+                next = self + 1;
+                next->ext.ILLEGAL.u8[8] = 1;
```
sozud, line 131: use the correct ext?

hohle: Eek. Missed that.

src/boss/bo4/unk_365FC.c
```diff
@@ -1,5 +1,181 @@
...
+        self->ext.dopBGLight.unk84 = 0;
+
+        if (!self->params) {
+            self->params += leftBlockFlag;
+            next = self + 1;
+            for (i = 1; i < 4; i++, next++) {
+
+#ifdef VERSION_PSP
+                // no declaration
+                CreateEntityFromCurrentEntity(D_pspeu_0926BCA8, next);
```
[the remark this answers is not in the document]
hohle: Definitely. Good catch.

hohle:
Added the `CheckStageCollision` implementation as well. This helped identify enough symbols to share `CheckWallLeft` and `CheckWallRight` implementations.

PR #2584, opened 2025-05-20
Import `dra_psp` d_6B420.c

gamezter:
I modified sotn-assets to accept a png with data larger than width * height which is needed for `D_psp_09149E90` - `D_psp_0914A248`

PR #2587, opened 2025-05-21
 Decompile multiple `dra_psp` E588.c functions 

gamezter:
Depends on https://github.com/Xeeynamo/sotn-decomp/pull/2584


src/dra/menu.c
```diff
@@ -529,17 +529,21 @@ u8 func_800F548C(u8 arg0) {
     return temp << 4;
 }
 
-bool IsOutsideDrawArea(s32 x0, s32 x1, s32 y0, s32 y1, MenuContext* context) {
-    if (!(x1 > context->unk1.x)) {
+static u32 IsOutsideDrawArea(
```
gamezter, line 532: PSP requires the return value to be unsigned so I can't use `bool` here. Is there a way to force the underlying type of the bool enum to be unsigned ?

sozud (approved):

PR #2592, opened 2025-05-24
Avoid Touching Symbols During `make format` & Exclude `gen`

hohle:
Now that we have correctly modeled symbol table dependencies an overlay is automatically rebuilt whenever a symbol table file is touched. Two utility functions in `symbols.py`, however, update these files regardless of whether changes are being made to them. This change updates the `sort` and `remove-orphans` to only modify the file if a change to the contents is necessary. This allows subsequent builds to be no-ops when no changes are necessary as well as only requiring rebuilds to overlays that are modified.

`remove-orphans` was updated to no longer remove symbols with the `allow-duplicates:true` attribute.

The Makefile targets for these have been updated to derive from the list of splat files for each version. This eliminates the need to update this target when new overlays are added and allows these targets to be run in parallel (with `-j` or the `ff` target).

The next longest running task in was `clang-format`. Excluding `gen` saves around 30% of the total runtime for `clang-format`.

Together these changes allow `ff` to run 25-30% faster.


hohle:
What is purpose of the symbols I removed in `symbols.us.weapon.txt.in`?

PR #2594, opened 2025-05-24
Major improvements to outdoor entities

bismurphy:
- Decompilation/Codeshare with PSP
- Pull in all data for all functions, on both versions
- Rename all data and functions to be meaningful - no `D_8675309` style placeholders or unknown entities
- Whole lot of general improvements to functions making code cleaner

Still work to do as always but I think this is in a nice place now.


config/symbols.pspeu.stno3.txt
```diff
@@ -27,9 +27,9 @@ EntityExplosionPuffOpaque = 0x0923D428;
 EntityPushAlucard = 0x0923D998;
 EntityCastleDoorTransition = 0x0923DD00;
 EntityForegroundTree = 0x0923DEE8;
-EntityUnkId50 = 0x0923E340;
+EntityFoliageMaker = 0x0923E340;
 EntityBackgroundPineTrees = 0x0923E588;
-EntityUnkId52 = 0x0923F0F0;
+EntityDistantTreeMaker = 0x0923F0F0;
```
JoshSchreuder, line 32: Not a blocker but perhaps we should standardise on terminology for stuff like this?

Unless I'm misunderstanding the functionality on these, we sometimes call these EntityXFactory, EntityXSpawner, and now EntityXMaker?

bismurphy: I don't think this should be considered a spawner, since I think those should be entities that actively spawn new entities as the game runs - biggest example is the Zombie spawners in the long hallways of NO3. It could be a factory, but I'm not sure if we want that either. Not a huge deal either way though, I don't think standardizing on one particular naming has a ton of benefit as things stand right now.

PR #2595, opened 2025-05-24
CI: Merge generate-progress-report

Xeeynamo:
I do not remember why `generate-progress-report` was done, but we do not need to keep it separate. This should save around 1m30s in the CI, plus the time is needed for GH Actions to spin up the agent


.github/workflows/validate-and-report.yml
```diff
@@ -83,8 +86,6 @@ jobs:
         run: make -j build
       - name: Check if they match
         run: make check
-      - name: progress dry run
```
sozud, line 86: We're still going to have a dry run as part of the PR CI?

PR #2614, opened 2025-05-27
Migrate cmpgfx from splat to asset tool

Xeeynamo:
Loads of changes to make the code more portable. This also eliminates the old `splat_ext/cmpgfx.py`.

With the suggestion from @sozud , I tried using `incbin` but I ran into a few issues:
```c
#define ADD_DATA(sym, path)                                                    \
    __asm__(".pushsection .data\n"                                             \
            ".globl " #sym "\n" #sym ":\n"                                     \
            ".incbin \"" path "\"\n"                                           \
            ".popsection\n");
```
The macro above works pretty well using the MIPS compiler. Unfortunately it does not make the code portable enough. I remember I got suggested graphitemaster/incbin, but it has some limitations with MSVC. I also tried making a macro to use `incbin` when on PSX and `#include` in any other situation, but `#include` cannot be used in a macro due to a C pre-processor limitation. I thought the effort was not justified. 

This solution increases the build time by around 10%, but it is noticeable only when doing a `make clean`, which is currently only required when switching game builds. If the increase of build time represents a problem, I'll think of another solution.


sozud:
It's not available everywhere but #embed from C23 should help with the issue on modern compilers https://gcc.gnu.org/gcc-15/changes.html#cxx

```
Some more C23 features have been implemented:
    #embed preprocessing directive support.
```

PR #2622, opened 2025-05-30
Give "opacity" name to Entity.unk6C

bismurphy:
Turns out this was already discovered in a comment in `ghost.c`. Named now, both the entity member and the draw flag.

Mostly a find and replace, but not a fully automatic one; I checked as I went to make sure the member didn't get renamed in any other structs, so this should just be entities.

### DIFF
--- a/src/dra/7879C.c
+++ b/src/dra/7879C.c
@@ -3349,8 +3349,9 @@ void EntityWingSmashTrail(Entity* entity) {
         entity->animSet = PLAYER.animSet;
         entity->animCurFrame = PLAYER.animCurFrame | ANIM_FRAME_LOAD;
         entity->zPriority = PLAYER.zPriority - 2;
-        entity->drawFlags = PLAYER.drawFlags |
-                            (FLAG_DRAW_OPACITY | FLAG_DRAW_ROTY | FLAG_DRAW_ROTX);
+        entity->drawFlags =
+            PLAYER.drawFlags |
+            (FLAG_DRAW_OPACITY | FLAG_DRAW_ROTY | FLAG_DRAW_ROTX);
         entity->opacity = 0x80; // a lifetime counter
         entity->drawMode = DRAW_TPAGE2 | DRAW_TPAGE;
         entity->rotZ = PLAYER.rotZ;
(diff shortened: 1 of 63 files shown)


src/dra/7879C.c
```diff
@@ -3349,9 +3349,10 @@ void EntityWingSmashTrail(Entity* entity) {
         entity->animSet = PLAYER.animSet;
         entity->animCurFrame = PLAYER.animCurFrame | ANIM_FRAME_LOAD;
         entity->zPriority = PLAYER.zPriority - 2;
-        entity->drawFlags = PLAYER.drawFlags |
-                            (FLAG_DRAW_UNK8 | FLAG_DRAW_ROTY | FLAG_DRAW_ROTX);
-        entity->unk6C = 0x80; // a lifetime counter
+        entity->drawFlags =
+            PLAYER.drawFlags |
+            (FLAG_DRAW_OPACITY | FLAG_DRAW_ROTY | FLAG_DRAW_ROTX);
+        entity->opacity = 0x80; // a lifetime counter
```
sozud, line 3355: Is the comment wrong now? There's a bunch of places with this same comment so applies to all of them

src/servant/tt_001/ghost.c
```diff
@@ -644,10 +644,10 @@ void UpdateServantDefault(Entity* self) {
         self->ext.ghost.opacityAdjustmentAmount *= -1;
     }
 
-    // unk6C is used in conjunction with self->drawFlags = FLAG_DRAW_UNK8 to set
-    // the alpha/saturation.  When it's zero, it's invisible, when it's 255,
-    // it's at the opacity of the sprite itself and is "bright".
-    self->unk6C = self->ext.ghost.opacity;
+    // opacity is used in conjunction with self->drawFlags = FLAG_DRAW_OPACITY
+    // to set the alpha/saturation.  When it's zero, it's invisible, when it's
+    // 255, it's at the opacity of the sprite itself and is "bright".
```
sozud, line 649: Not your comment but opaque is probably more precise than bright

sozud (approved):

src/servant/tt_001/ghost.c
bismurphy: Pushed updated version of this comment that I think states what we know explicitly.

PR #2627, opened 2025-06-01
Decompile no4 water effects functions

JoshSchreuder:
Sorry for the massive PR. My original plan was to do this on PSP and PSX side and have splits and data imported.

As I went along I kept expecting a PSP file split that never came. It looks like most if not all of the functions in first_c_file are consistent with what's on the PSP side, so this may just be one big file (or researching the functions may give some more clear context wrt splits).

All of these functions should be matching on PSP side also so should be pretty trivial to codeshare later on. Main reason I'm putting this up now is I realised some of what I was working on was starting to overlap with @cram0's research here and I want to get out of their way as they seem more on top of what the functions are doing and documenting them better 😄


src/st/no4_psp/unk_5EE8.c
```diff
@@ -0,0 +1,68 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "common.h"
+
+INCLUDE_ASM("st/no4_psp/psp/no4_psp/unk_5EE8", func_pspeu_0923D568);
```
JoshSchreuder, line 4: This is the start of first_c_file on PSX side. I'm pretty sure most if not all of these functions are consistent with that file and likely can be codeshared directly once it's all done

src/st/no4/first_c_file.c
```diff
@@ -1,39 +1,1812 @@
...
+#ifdef VERSION_PSP
+                var_s5 = (D_us_8018152C[self->ext.et_801C12B0.unk80] * 4) +
+                         0xBB - scrollX - 9;
+                prim->x2 = prim->x0 = var_s5 + (rand() & 3);
+#else
+                var_s5 = (rand() & 3) - 9;
+                prim->x2 = prim->x0 =
+                    (D_us_8018152C[self->ext.et_801C12B0.unk80] * 4) + 0xBB -
+                    scrollX + var_s5;
+#endif
```
Xeeynamo, line 1116: This looks a bit weird, including the `u32 scrollX`. Can you link the scratch on Discord? I'll have a look. I am happy to get this merged for the time being

JoshSchreuder: @Xeeynamo agreed, I had a bit of trouble matching this one

PSX: https://decomp.me/scratch/4YxCX
PSP: https://decomp.me/scratch/SgBVI

Happy to update if you can work it out 😄

PR #2632, opened 2025-06-02
Added DAI for us and pspeu

ProjectOblivion:
All symbols, text segments, and rodata segments were automatically generated.


src/st/dai/blit_char.c
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "dai.h"
+
+INCLUDE_ASM("st/dai/nonmatchings/blit_char", BlitChar);
+
+INCLUDE_RODATA("st/dai/nonmatchings/blit_char", D_us_801C0080);
```
ProjectOblivion, line 6: This rodata item is only referenced by a data item that is only referenced by EntityRelicOrb.  EntityRelicOrb immediately follows BlitChar for both rodata and text segments.

Given that EntityRelicOrb has to be part of e_collect for pspeu and e_collect, blit_char, and e_relic_orb are contiguous, this strongly suggests that both BlitChar and EntityRelicOrb should be a part of e_collect.

ProjectOblivion: @Xeeynamo @sozud @bismurphy 
Do you have an opinion 👍 or 👎 on this?  If it makes sense to everyone else, I'd like to adjust it for this PR.

sozud: Bismurphy just did BlitChar for no3 a couple of days ago, let's do what he thinks? https://github.com/Xeeynamo/sotn-decomp/blob/8b15d522b79a834889d3997bb3480563a5a830ea/src/st/no3/blit_char.c#L5

bismurphy: If DAI can match, with these files being separated they are in the existing overlays, then let's have it match the other overlays. If you want to have a followup PR to pull these files together you can do that, but for now let's keep the structure unchanged and just add the new overlay, and have it be like the existing ones.

ProjectOblivion: Not arguing the response, but asking for clarity.
There are several different implementations between different overlays and us vs pspeu.  Even the same overlays don't have the same implementation between us and pspeu.  What would I match?
Also to note is the fact that the segments currently in this PR also don't match the other overlays.

My original question was an attempt to identify what the implementation should be to be consistent across the board (as consistent as is reasonable, anyway)

bismurphy: What are you asking? The question at the top of this comment chain is about unifying the files of relic orb, blit char, and e_collect. What are you saying is different about what implementations?

src/st/dai/st_debug.c
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "dai.h"
+
+extern ObjInit2 OVL_EXPORT(BackgroundBlockInit)[];
+extern u16 g_EInitCommon[];
+
+// params: Index of ObjInit to use
+//         (== 1) Use a different hardcoded scaleY and scaleX value
```
Xeeynamo: This comment is not valid for this function

src/st/dai/e_init.c
```diff
@@ -0,0 +1,150 @@
...
+    /* 0x08 */ EntityPrizeDrop,
+    /* 0x0C */ EntityDamageDisplay,
+    /* 0x10 */ DAI_EntityRedDoor,
+    /* 0x14 */ EntityIntenseExplosion,
+    /* 0x18 */ EntitySoulStealOrb,
+    /* 0x1C */ EntityRoomForeground,
+    /* 0x20 */ EntityStageNamePopup,
+    /* 0x24 */ EntityEquipItemDrop,
+    /* 0x28 */ EntityRelicOrb,
+    /* 0x2C */ EntityHeartDrop,
```
Xeeynamo: We used to have the entity ID as a comment, so `/* 0x01 */`, `/* 0x02 */` and so on. We are not using it anymore though as it's better to have the Entity ID to be used as a look-up table. See here for example: https://github.com/Xeeynamo/sotn-decomp/blob/master/src/maria/maria.h#L111

We should get rid of those comments representing the offset of the array.

src/st/dai/dai.h
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "stage.h"
+
+#define OVL_EXPORT(x) DAI_##x
+
+typedef enum EntityIDs {
+    /* 0x00 */ E_NONE,
+} EntityIDs;
```
Xeeynamo: It would save a lot of time to have all the enums as `E_1`, `E_2` and so on with the corresponding function from `PfnEntityUpdate OVL_EXPORT(EntityUpdates)[]` as a comment. See https://github.com/Xeeynamo/sotn-decomp/blob/master/src/maria/maria.h#L111

src/st/dai/e_init.c
```diff
@@ -0,0 +1,150 @@
...
+void func_us_801C5920(Entity* self);
+void func_us_801C5B88(Entity* self);
+void EntityMistDoor(Entity* self);
+
+PfnEntityUpdate OVL_EXPORT(EntityUpdates)[] = {
+    /* 0x00 */ EntityUnkBreakable,
+    /* 0x04 */ EntityExplosion,
+    /* 0x08 */ EntityPrizeDrop,
+    /* 0x0C */ EntityDamageDisplay,
+    /* 0x10 */ DAI_EntityRedDoor,
```
Xeeynamo:
```suggestion
    /* 0x10 */ OVL_EXPORT(EntityRedDoor),
```
and similar?

src/st/dai/blit_char.c
ProjectOblivion: While the question here was directly about those, the issue is related to how e_collect, blit_char, e_misc and e_misc_2 are handled in all of the overlays across the two versions.
I looked back and leaving blit_char as is doesn't create any major divergence and you're right, I can better present my suggested solution in a separate PR.

config/symbols.us.stdai.txt
```diff
@@ -0,0 +1,155 @@
...
+g_EInitObtainable = 0x80180944;
+g_EInitParticle = 0x80180950;
+g_EInitSpawner = 0x8018095C;
+g_EInitInteractable = 0x80180968;
+g_EInitUnkId13 = 0x80180974;
+g_EInitLockCamera = 0x80180980;
+g_EInitCommon = 0x8018098C;
+g_EInitDamageNum = 0x80180998;
+DAI_BackgroundBlockInit = 0x80180AC8;
+DAI_Invincibility0 = 0x80181070;
```
ProjectOblivion: I just noticed that this got renamed from UNK to DAI.  I had addressed it already, but hadn't pushed the changes to main yet and missed it.

ProjectOblivion:
You can do this for yourself by cloning https://github.com/ProjectOblivion/decomp_utils into the tools directory and running `python3 tools/decomp_utils/generate_sotn_config.py dai` for each version.  You'll want to use the dev branch because I found a couple issues while doing this that I resolved but haven't pushed to main.

I manually added the clang-format lines.  I'd like to add this to the tool output, though I'm not sure of the specific criteria for when it is needed.

I did some things manually that I've been working on, but I don't have them incorporated functionally into the tool yet.
- e_init.c generation
- header.c generation
- Added DAI_EInitUnkBreakable, g_EInitObtainable, g_EInitParticle, g_EInitSpawner, g_EInitUnkId13, g_EInitLockCamera, g_EInitDamageNum, and DAI_BackgroundBlockInit to the symbols file
- Added the splat suggestion comments

I've been looking at replacing the INCLUDE_ASM with actual code or include automatically, but I don't think it is a capability that is right for the initial generation.  The places where that was done were done manually, which includes the data and bss segments related to them.

I added the isolated comments manually when I validated the match for ones that didn't have enough context just from the .c file.

I also had to do a few additional things manually for pspeu:
- Added EntityMagicallySealedDoor, DAI_Load, and DAI_Overlay to the symbols file, working on doing it automatically
- Added the 3 GetLang functions to the symbols file and named them manually.  I'm thinking on how to handle the naming automatically in a way that also correctly preserves their association with their respective segments
- DAI is the first with EntityMagicallySealedDoor so I added that manually
- Since the GetLang functions aren't named automatically, I renamed them manually.

One known issue is that 0x801C58C4 is misattributed as func_8018F890 instead of func_801961DC.  I have resolution for this, but it is not incorporated yet.

Just noticed that I also added in StepTowards and func_801961DC code manually.

src/st/dai/dai.h
ProjectOblivion: I think I did what you were getting at, let me know if I missed the mark.

src/st/dai/e_init.c
ProjectOblivion: I think I did what you were getting at, let me know if I missed the mark.

PR #2633, opened 2025-06-03
Decompile `no4` cutscene functions

JoshSchreuder:
Having trouble finishing the one mismatch on the main cutscene function - https://decomp.me/scratch/WoDh2

Putting this up for now which includes the NO4 dialogue struct, and the simpler cutscene functions decompiled.


src/st/no4/dialogue.h
```diff
@@ -0,0 +1,18 @@
...
+    /* 0x0A */ u8 nextCharTimer;
+    /* 0x0B */ u8 unkB;
+    // Of course, offsets beyond here won't be right on PSP
+#if defined(VERSION_PSP)
+    /* 0x0C */ Primitive* prim[4]; // for dialogue graphics rendering
+#else
+    /* 0x0C */ Primitive* prim[6]; // for dialogue graphics rendering
+#endif
+    /* 0x1C */ s32 primIndex[3]; // primIndices: unk, actorName, unk
+} OVL_EXPORT(Dialogue);          // size = 0x28
```
JoshSchreuder, line 18: Not sure if using an overlay named struct here is a good idea or not.

This struct on NO4 is much simplified compared to usual as the cutscene only has one actor (Ferryman), so much of the code is simplified in the handler (fewer prims, loops unrolled, etc.)

src/st/no4/cutscene.c
```diff
@@ -1,23 +1,200 @@
...
+    g_Dialogue.nextCharTimer = 0;
+    g_Dialogue.unkB = 8;
 }
 
-INCLUDE_ASM("st/no4/nonmatchings/cutscene", func_us_801C9568);
+static u8 SetCutsceneScript(u8* script) {
+    Primitive* prim;
+
+    g_Dialogue.primIndex[2] = g_api.AllocPrimitives(PRIM_SPRT,
+#if defined(VERSION_PSP)
```
sozud: A `#define` is probably cleaner here

src/st/no4/dialogue.h
sozud: Seems fine to me

sozud (approved):

src/st/no4/cutscene.c
Xeeynamo:
```
#if defined(VERSION_PSP)
#define NUM_CUTSCENE_PRIM 7
#else
#define NUM_CUTSCENE_PRIM 5
#endif

...

g_Dialogue.primIndex[2] = g_api.AllocPrimitives(PRIM_SPRT, NUM_CUTSCENE_PRIM);
```

or something along those lines, as Sozud suggested

src/st/no4/dialogue.h
Xeeynamo: I think it's fine even without. This struct is isolated in `no4.h` and no other overlays can access it. It was most likely copy&pasted and changed ad-hoc by the developers. Up to you, it's not a big deal either way.

src/st/no4/cutscene.c
JoshSchreuder: Updated

PR #2636, opened 2025-06-04
Create new tool for viewing entity animations

bismurphy:
As I've been pulling in lots more data for NO3, I've found that there are some animations which were unused - the PSP dead-stripping makes it very obvious when and where these exist. I've been making an effort to give them names to highlight the fact that they are unused.

Naturally, it is interesting to try to dig into what these animations might have looked like. To that end, I have extended the capabilities of the Python tools I've made in the past, so that they can now show these animations.

As specified in the comment, if you run PCSX-Redux with its HTTP debugging server, you can then run `python3 tools/play_animation.py LIVE no3 src/st/no3/e_warg.c` and you will be prompted with an interactive system which will ask you what animation you want to view. It automatically scrubs through the specified file for any arrays declared as "static u8", and gives you the opportunity to select an animation to view. If there is an entity that calls that animation, it will then auto-detect that entity's InitializeEntity initializer (which specifies the graphics tpage, clut, etc) and use that to render the animation. If no entity calls the animation (as in an unused animation), it will list all initializers used in the file, and prompt the user to select one.

Some very light testing on the Warg indicates that this is operational, but as always there could be bugs. But hopefully this is a helpful tool for future SOTN Archaeology :)

### DIFF
--- a/tools/play_animation.py
+++ b/tools/play_animation.py
@@ -117,6 +117,9 @@ def get_initializer_for_ent(anim_name, src_file, overlay):
             if "ANIMSET_OVL" in initSet[0]:
                 arg = re.findall(r"(?<=\()[^\)]*", initSet[0])[0]
                 initSet[0] = hex(0x8000 | int(arg, 0))
+            if "ANIMSET_DRA" in initSet[0]:
+                arg = re.findall(r"(?<=\()[^\)]*", initSet[0])[0]
+                initSet[0] = arg
             return [int(x, 0) for x in initSet]
 
 


tools/play_animation.py
```diff
@@ -0,0 +1,289 @@
...
+        initName = options[initializer_idx]
+    # Now we know the name of the entity initializer.
+    # Fetch the value of it from the overlay e_init file.
+    with open("src/st/" + overlay + "/e_init.c") as f:
+        initlines = f.readlines()
+    for line in initlines:
+        if initName in line:
+            initSet = re.findall(r"(?<={)[^}]*", line)[0].split(",")
+            print(initSet)
+            return [int(x, 0) for x in initSet]
```
JoshSchreuder, line 123: This fails for initializers containing `ANIMSET_OVL`

```
  File "/home/josh/dev/sotn-decomp/tools/play_animation.py", line 117, in get_initializer_for_ent
    return [int(x, 0) for x in initSet]
            ^^^^^^^^^
ValueError: invalid literal for int() with base 0: 'ANIMSET_OVL(0x05)'
```

eg. `EInit D_us_80180AAC = {ANIMSET_OVL(0x05), 0x00, 0x24, 0x000, 0x003};`

Can be worked around easily enough by replacing manually in code, eg. with `0x8005` but would be nice if it can be handled gracefully 😄

bismurphy: Good catch! I think I got ANIMSET_OVL working. However, NO0 will, for the time being, not be usable with this tool because it does not yet use the assets tool to extract its sprite_banks members. We need those in order for this tool to work.

JoshSchreuder: Yep I was playing around with sprite extraction here and got something "working" but it appeared to be wrong or not working with this tool so I'll need to dig into it further

JoshSchreuder: Btw there is also ANIMSET_DRA(x), I am not sure if it is used in any st_init off hand but might need support too

bismurphy: Added ANIMSET_DRA!

sozud (approved):

Xeeynamo:
Hey @bismurphy sorry for not seeing this merged yet. I did not have time to test the tool unfortunately. Both Sozud and JoshSchreuder reviewed, I am happy to blindly merge this!

PR #2638, opened 2025-06-04
Add playable Maria on PSX

Xeeynamo:
image

The mod replaces Richter with Maria as a playable character.

I created the folder `mods/`, where I put `maria-psx` and planning to move the debug module in as well. I am not sure if the path is what we can all agree on, but let's decide before merging this in.

I re-compiled the PSP code by rebasing the address to `0x8013C000` with the simple `player.ld` in the repo root. I plan on migrating `RIC.BIN` to use it too and allow relocating the overlay. `stage.ld` and `servant.ld` will follow. This follows the `weapon0.ld` pattern found in the repo root. I am happy to move those new linker scripts anywhere you suggest me to.

PSP Maria loads most of the graphics on the same Video RAM address that is used on PS1 by the second screen buffer. In `ovl_load.c` you will find some ifdef to keep the PSP match while relocating the Maria graphics to the Video RAM section that were used for Richter exclusive content (e.g. Kid Maria and the four beasts during the prologue). The graphics need to be strategically located to minimize the code changes on entities that use UV coordinates. The only exception so far is the `u -= 0x80;` you will find on the Dragon Crash.

The palette also had to be relocated from `0x114` (PSP Maria) to `0x120` (PSX RIC). The reason is the 0x114 palette row is replaced at runtime by the NO1 and RNO1  overlay, resulting on a Maria all black and all her sub-weapons with the incorrect colours. Luckily the `enum MarPalette` made it very easy to re-locate everything.

The HUD was the main source of issue, as it is found in DRA and not in the player overlay. I created a `PatchHudSubweaponUV` to patch DRA on the fly. Alucard and Richter have separate UV coordinates to show their respective subweapon graphics in the HUD. Since the mod replaces Richter, I am just patching the Richter UVs with Maria.

Another issue is where the sub-weapon IDs need to be re-mapped. Maria uses the subweapon DOLL when subweapon == 0. As PS1 DRA will not show any icon when the subweapon is zero, I am pulling and pushing the correct ID right before and right after the player loop. I needed to do the same task for the other subweapons to reflect the PSP behaviour, where the Doll replaced Bible, Dragon replaces Vithubi and so on. It is all documented in the respective functions. Like the PSP, picking up a subweapon that does not exist for maria (e.g. Cross), will not dismiss the current subweapon and instead it will increase the heart count by 1.

Maria on PSP allows to hold two sub-weapons. The active subweapon can be swapped with Triangle. While the swapping logic is present in the Maria overlay, the HUD cannot be patched. There is a way to make the HUD working nicely, but we need to agree to what extend this mod can go.

All Maria grunts are absent: we need to find a way to convert PSP sound effects to PS1, with the appropriate VH/VB container. But I think it is out of scope for this PR.

The graphics of the subweapon drops from candles will incorrectly show the Richter subweapons. I can think of two options to fix this:
1. Replace the UV in DRA. For example `D_800D1A28` contains the UVs of the Axe subweapon. If they get replaced by the correspondent Maria subweapon, the correct graphics will always be displayed. The downside is soft-resetting / gameover to play with Alucard will result in the wrong graphics.
2. We could patch the graphics file by blitting the Maria subweapons over the Richter ones. This has the advantage of being more self-contained. But the Bird icon cannot replace the Knife as they're of different size. EDIT: the palette will be incorrect, so this is not possible.

---

To all the non-contributors of the project: PLEASE BE PATIENT. PLEASE DO NOT REDISTRIBUTE WITHOUT PERMISSION. KINDLY, COMPLY WITH THE LICENSE. When this is ready, it will be properly announced and distributed in an official way. Until then, please ask permission before taking any action.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ __pycache__
 .venv/
 asm/
 build/
+release/
 expected/
 disks/
 nonmatchings/
--- a/mods/mariapsp-forpsx/Makefile
+++ b/mods/mariapsp-forpsx/Makefile
@@ -17,7 +17,7 @@ OBJS = $(BUILD_DIR)/maria.o $(patsubst $(SRC_DIR)/%.c, $(BUILD_DIR)/%.o, $(SRCS)
 ASSETS = $(BUILD_DIR)/.assets_done
 
 
-.PHONY: all build clean disk
+.PHONY: all build clean release disk
 
 all: build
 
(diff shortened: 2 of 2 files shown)


mods/maria-psx/Makefile
```diff
@@ -0,0 +1,41 @@
+ROOT_DIR = ../..
```
sozud: I think we should have this in the CI to make sure it's always building

mods/maria-psx/maria.c
```diff
@@ -0,0 +1,209 @@
...
+    for (ptr = (s32*)0x800A0000; (u32)ptr < 0x8013C000U; ptr++) {
+        if (DoesRichterHudMatch(ptr)) {
+            return (SubwpnIconParams*)ptr;
+        }
+    }
+    return NULL;
+}
+static void PatchHudSubweaponUV() {
+    // offset of g_ricSubwpnIcons, from build/us/dra.map
+    s32* DefaultOffset = (s32*)0x800A3014;
```
sozud: IMO we should just have alterations to DRA as part of this mod so that this would be compatible with other platforms for example. Doesn't have to be part of this PR though

hohle:
In the future there will possibly be two Marias that can be integrated into any game version. Should the mod name reflect which Maria this is?

mods/maria-psx/Makefile
Xeeynamo: Good point thanks. Done

Xeeynamo:
@sozud I would like a second review if possible.

Note that the `sotn-maria` image will not yet run since this branch is not yet merged in `master`, and we do not yet have the commands `sotn-disk inject` implemented. But all the infrastructure should be ready to allow users to easily generate a modded disk

sozud (approved):
Great work

sozud:
I think since this makes CI changes just merge it when you're ready to monitor the CI and fix if necessary

PR #2641, opened 2025-06-05
Fix funcs report not generating correctly

JoshSchreuder:
Duplicate check was skipped locally so you can ignore that part
https://gist.github.com/JoshSchreuder/e141eab0d4e2813d93d3213c1bf2ba6e

Now correctly includes `no4` / `rbo` etc. functions again

### DIFF
--- a/.github/workflows/validate-and-report.yml
+++ b/.github/workflows/validate-and-report.yml
@@ -18,7 +18,7 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
-  build-and-test: 
+  build-and-test:
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/validate-and-report.yml
```diff
@@ -152,8 +152,8 @@ jobs:
             find ./build -type f -name "*.ld" -delete
       - name: Generate duplicates report
         run: |
-          make force_symbols
           make build || true
+          make force_symbols
```
JoshSchreuder, line 156: This relies on the `.elf` files being built already, otherwise it will strip all symbols from `symbols.*.txt`

PR #2642, opened 2025-06-05
Merge e_collect to a single C file

Xeeynamo:
Instead of having multiple `e_collect.c` that simply do `#include "../e_collect.h`, use a single C file that gets compiled once and linked to every overlay.

This is to help PSP, where `sotn_str` cannot reach the strings in `e_collect.h`. The pull request do not yet tackle all the `e_collect.c` we have. I am creating it as a proof of concept to understand if we can continue with this approach and apply to other cases too, while also unblocking @bismurphy .

### DIFF
--- a/src/st/chi_psp/e_collect.c
+++ b/src/st/chi_psp/e_collect.c
@@ -127,7 +127,7 @@ static u8* g_ExplosionAnimations[] = {
 
 #include "../collect_life_vessel.h"
 
-#include "../collect_dummy.h"
+static void CollectDummy(u16 id) { DestroyEntity(g_CurrentEntity); }
 
 #include "../entity_prize_drop.h"
 
--- a/src/st/lib_psp/e_collect.c
+++ b/src/st/lib_psp/e_collect.c
@@ -126,7 +126,7 @@ static u8* g_ExplosionAnimations[] = {
 
 #include "../collect_life_vessel.h"
 
-#include "../collect_dummy.h"
+static void CollectDummy(u16 id) { DestroyEntity(g_CurrentEntity); }
 
 #include "../entity_prize_drop.h"
 
--- a/src/st/st0/e_collect.c
+++ b/src/st/st0/e_collect.c
@@ -108,6 +108,7 @@ void CollectDummy2(void) { DestroyEntity(g_CurrentEntity); }
 
 #include "../collect_subweapon.h"
 
+#if !defined(VERSION_PSP)
 static void CollectDummy(void) { DestroyEntity(g_CurrentEntity); }
 // Extra unused function, putting it in this same if-block.
 Entity* func_801939C4(void) {
@@ -115,6 +116,11 @@ Entity* func_801939C4(void) {
     g_CurrentEntity->params = 4;
     return g_CurrentEntity;
 }
+#else
+
+// PSP is more in-line with the classic e_collect.c from other stages
+static void CollectDummy(u16 id) { DestroyEntity(g_CurrentEntity); }
+#endif
 
 #include "../entity_prize_drop.h"
 
--- a/src/st/wrp_psp/e_collect.c
+++ b/src/st/wrp_psp/e_collect.c
@@ -123,7 +123,7 @@ static u8* g_ExplosionAnimations[] = {
 
 #include "../collect_life_vessel.h"
 
-#include "../collect_dummy.h"
+static void CollectDummy(u16 id) { DestroyEntity(g_CurrentEntity); }
 
 #include "../entity_prize_drop.h"
 #include "../entity_explosion.h"


CMakeLists.txt
```diff
@@ -305,6 +305,10 @@ set(SOURCE_FILES_STAGE_SEL
     src/st/sel/cutscene.c
 )
 
+set(SOURCE_FILES_STAGE_SHARED
+    src/st/e_collect.c
```
sozud, line 309: I don't think this is going to work because there's multiple versions of e_collect.c for different overlays. If you only compile it once you're not going to get those differences

Xeeynamo: e_collect seem to be different only for ST0 and MAD. On PSX, now `build/us/src/st/e_collect.c.o` is used when linking the various stage overlays, and we still get a :ok: . For ST0 and MAD obviously the approach needs to change and we might probably just do `OVL_EXPORT(x)`

config/splat.hd.stwrp.yaml
```diff
@@ -50,7 +50,7 @@ segments:
       - [0x6BC, .data, collision]
       - [0xA7C, .data, e_red_door]
       - [0xA94, .data, st_common]
-      - [0xC94, .data, e_collect]
+      - [0xC94, .data, ../e_collect]
```
sozud, line 53: Didn't we already try this and have to give up because it messes up the paths in the asm folder? https://github.com/ethteck/splat/issues/266

CMakeLists.txt
bismurphy: I think if it's almost all the same between versions, perhaps the thing to do is to have e_collect and e_collect_alt (for ST0/MAD).

src/st/cen/e_collect.c
```diff
@@ -3,7 +3,8 @@
 
 #if !defined(VERSION_HD)
 
-#include "../e_collect.h"
+// TODO: fix?
+#include "../e_collect.c"
```
bismurphy: Heads up in my branch I have HD CEN resolved so there's no ifdef, so no problem there.

Xeeynamo: Does this PR fix your problem with PSP? You could either merge your PR here or I merge mine to yours

config/splat.hd.stwrp.yaml
Xeeynamo: That is correct. The original problem were our tools, yes? This PR is to unblock @bismurphy and the fact we cannot easily inject sotn_str after the pre-processor. We lose either one or the other. And I think at this point it is easier if we adapt the dups tool to handle this specific scenario than keep circumnavigating a problem we are bound to face at some point anyway.

CMakeLists.txt
Xeeynamo: ST0 and MAD have slightly different variations between each other, unfortunately that is not possible.

The best way going forward is to just do a `diff` between the three different variants if someone wants to know how the file evolved over time. Overall, I strongly believe we should move away from shared code with slight variations based on the STAGE macro.

sozud: IMO this is a step in the wrong direction because how are we going to keep track of all the slightly modified versions of files? Right now it's enforced because the PSX build would become nonmatching. This is decoupling it from that so I think we're going to end up with subtle entity differences.

config/splat.hd.stwrp.yaml
sozud: If we move everything up a level that might work better? e.g. like st_common becomes wrp/st_common if you don't want to share and just st_common if you do. My impression is that you're not supposed to use `../` with splat since it messes up the directories.

CMakeLists.txt
Xeeynamo: We are already doing it with `EntityBackgroundBlock` and the various `EntityCutscene`. I do not see why the various functions inside `e_collect.c` should be any different. They contain many more differences than other cases. And the only differences are only for ST0 and MAD, as I mentioned earlier. Three variants of the same file is much better than what we have in other places.

config/splat.hd.stwrp.yaml
Xeeynamo: Moving everything up to one level will only work with what we have in `src/st/`, not in `src/boss`. We would have to move up two levels. Unless we return back to the idea of having the boss stages in `src/st/`.

bismurphy:
I don't think this PR does what it says it does. It doesn't merge to a single C file, it merges to different C files for the different versions (main, ST0, MAD). I personally don't think this is the right solution. When two files are substantially similar, we should endeavour to use only one file in one place. Splitting things up means things will inevitably get de-synchronized. For example, all the data in this file, with D_8675309 name placeholders. At some point, someone will research what each of them is for, and they will give those a real name. But will they do it across all three files? With the exception of highly diverging files/functions, we should try to keep a spirit of "everything only exists in one place" to the extent that we can. This project is ultimately a quest of knowledge gathering. Trying to find what every function does and why. We figure out the C code for a function, we figure out how data is encoded, we figure out the purpose of every struct member. And when a new piece of knowledge is discovered, we should try to make it so that the knowledge can be incorporated into the codebase as smoothly as possible. Having the same functions in three places does not do that.

My initial goal with my branch I'm working on was to be able to flatten e_collect. If we continue having three files, then that would be counterproductive because we're increasing the amount of duplicated code in the codebase. We've put a lot of work into reducing duplicated code and I think this PR, as it stands, increases duplication.

I think this PR is really close to being a huge step forward - sharing C files would be fantastic compared to H files. But that only applies if the C files are shared - not if we have to fracture it into three C files that are scattered in different places.

config/splat.hd.stwrp.yaml
bismurphy: I think I like the idea of every overlay having its splat root just at src, and creating a `src/common` directory to hold all shared files. The splat would become slightly more verbose, but I don't see that as a problem.

Xeeynamo:
With two maintainers believing this is not the right approach, I am happy to close this proposal. IMHO it is good we keep exploring different approaches. Some will not make the cut and it is perfectly fine.

Closing until we figure out something better.

CMakeLists.txt
ProjectOblivion: EntityBackgroundBlock was recently specifically moved away from a common include file, by your direction, to avoid having subtle differences handled by a single file.

PR #2647, opened 2025-06-09
Adjusted pspeu nonmatchings_dir to be nonmatchings for all overlays

ProjectOblivion:


sozud:
What's the goal of this change?

ProjectOblivion:
To unify the nonmatchings directories across the board.  The primary problem is that some pspeu overlays use psp, some use nonmatchings.

ProjectOblivion:
Sure, that's no problem.  I'll do that at the same time as I resolve the conflicts later today.

PR #2648, opened 2025-06-09
Adjusted decompile.py for nonmatchings as psp nonmatchings_dir

ProjectOblivion:
Depends on #2647 to avoid breaking decompilation for overlays that currently use `psp` nonmatchings_dir.


tools/decompile.py
```diff
@@ -74,7 +71,7 @@ def overlay(self) -> str:
     @property
     def asm_differ_command(self) -> str:
         """Constructs the command string the user can enter to run asm-differ against the overlay this function resides in"""
-        return f"{Path(sys.executable).name} {self.root.joinpath("tools/asm-differ/diff.py").relative_to(Path.cwd())} -mwo --overlay {self.overlay} {self.name}"
+        return f"{Path(sys.executable).name} {self.root.joinpath("tools/asm-differ/diff.py").relative_to(Path.cwd())} -mwo --overlay {self.overlay.replace("_psp", "")} {self.name}"
```
ProjectOblivion, line 74: @Xeeynamo I looked at doing this replace at assignment time like you had done previously, but that value is used to identify the correct overlay asm directory.  Since psp uses the ovl_psp directory name, doing that would break the script without further adjustments.

PR #2650, opened 2025-06-09
Split `st_debug` into `e_room_bg` and `e_lock_camera`

ProjectOblivion:
This splits the text segments and renames the data segments, but the data segments are not fully converted.
Since for whatever reason, this happened to be a place that creates matching binaries, I'm using it to break up the PR.
I will be reviewing all of the associated data segments and making sure that they follow the same pattern across all overlays for a follow-up PR.


sozud:
Conflicted

Issue #2616, opened 2025-05-28
Integrate assets in the code base using `incbin`


Xeeynamo:
Follow-up to: https://github.com/Xeeynamo/sotn-decomp/pull/2614#issuecomment-2914666799

Currently, we use the sotn-assets tool to extract data from the binaries in disks/ into the assets/ folder, either as readable files or raw binary. We then run sotn-assets build to convert those assets into header files, which are included in the C source like this:

```c
u8 text_item_up[] = {
#include "gen/text_item_up.h"
};
```

This works well for portability. Since the data is part of the C source, it builds cleanly across compilers and platforms. However, it has a downside: longer build times. As the project grows, this is becoming more noticeable.

One potential improvement is using .incbin to include raw binary data directly via assembly. This was likely how the original code handled assets, and switching to it could reduce build times by ~20%.

The main challenge is making this approach portable. `.incbin` doesn’t always play nice with all compilers or targets, for example, it currently fails when I last tried building with GCC 15 targeting x86_64.

This issue is about exploring how we might modify the build system to support .incbin in a clean, portable way. If we can find a reliable method to integrate .incbin without breaking cross-compiler compatibility (Clang, MSVC, MWCC, etc.), it could be a solid upgrade.

We’re assuming a portable implementation is possible. We just need to figure out what changes are needed in the build chain to make it work smoothly, then port the current assets to be used as blob files.

PR #2653, opened 2025-06-10
Import `dra_psp` d_92E80.c data

gamezter:
I had to do a 'hack' because some of these `rawgfx` textures have 4 bytes at the start describing the dimensions of the texture. If I give `rawgfx` the correct offset, 4 bytes will be truncated off the end. Here I give an offset 4 bytes from the start and hard-code that data.


config/assets.hd.yaml
```diff
@@ -28,6 +28,28 @@ files:
           - [0x24908, cmpgfx, load_room_tile, 80, 16, 4]
           - [0x24B34, cmpgfx, save_room_tile, 80, 16, 4]
           - [0x24D90, skip]
+          - [0x24D94, rawgfx, D_800C4CEC, 16, 16, 4]
```
sozud, line 31: it seems like we probably want a rawgfx_header type or similar that takes the header into account?

Xeeynamo: I agree with this, it would make things a tad bit simpler. Especially if we plan to solve #2616 or if we want to encourage moddability.

@gamezter would a new asset type that processes the header work for you? We could merge as it is as it looks really good, then I could iterate on it to make the header part automated.

gamezter: All good !

PR #2656, opened 2025-06-11
Cleanup former st debug data

ProjectOblivion:
I went through and added or adjusted the data segments for e_room_bg and e_red_door.  I moved BackgroundBlockInit to e_room_bg and moved RedDoorTiles to e_red_door, importing the data when necessary.  In order for this to work since g_eRedDoorUV is not in a contiguous location, I moved it out to it's own file.

The end result of doing all of that allows e_room_bg, e_red_door, and e_init to be aligned between us and pspeu and merged.


src/st/dre/e_room_bg.c
```diff
@@ -1,11 +1,38 @@
...
-extern ObjInit2 OVL_EXPORT(BackgroundBlockInit)[];
 extern u16 g_EInitCommon[];
 
-// params: Index of ObjInit to use
-//         (== 1) Use a different hardcoded scaleY and scaleX value
+static u8 AnimFrames_80180518[] = {0x40, 0x01, 0xFF, 0x00};
+static u8 AnimFrames_8018051C[] = {
+    0x02, 0x25, 0x02, 0x26, 0x02, 0x27, 0x02, 0x26};
+
+static u8 pad[] = {0, 0, 0, 0};
```
ProjectOblivion: I'm not sure if this is the right thing to do, but the data was shifted without it.

ProjectOblivion:
> Maybe I am out of the loop, but could you provide me an example where this would be beneficial?

Without changes (whether these or others) you cannot share e_init files between us and pspeu without ifdefs.  There are also other places that would need special handling.  With these changes, e_init.c, e_red_door, e_room_bg, and e_red_door_uv segments will be the same across all overlays and both versions.  The only difference between overlays/versions would be the definition within the config.

The primary issue is that in us the order goes e_init > BackgroundBlockInit > RedDoorTiles with that data being contiguous, but pspeu goes RedDoorTiles > e_init with BackgroundBlockInit being in a very different location.

This means that the only way to have them be unified between both platforms is to use ifdefs.  This in and of itself may not be a major issue, but the data is not always in that structure.  NO1 for example goes e_init > e_room_bg > e_red_door_uv > e_red_door.

> The .data you defined in US BO4 (as an example) is e_init, e_room_bg, e_red_door, e_lock_camera and so on. But the .text order is e_room_bg, e_lock_camera

What you're not seeing is that RedDoorTiles was previously lumped in with e_init or not imported, so while you don't see that e_red_door segment previously, it has always been the place that RedDoorTiles has been.  e_red_door has two pieces of data, RedDoorTiles and g_eRedDoorUV.  Previously RedDoorTiles was externed, but now g_eRedDoorUV is externed instead.

The bottom line is that the previous way could not use the same pattern across the board, this can.  I looked at multiple different ways and compared multiple overlays for commonalities and this is the only way I could get everything to agree.  Do I think having the e_red_door_uv oddball segment is the final "right" answer?  No, but it is better than before and will be easier to adjust in the future, as well as having the precedent set by e_init itself.

> The reason why you're getting a match with the data order as you expect is due to splat generating a linker script with the exact order the sections are defined in the splat config. And why I do not think this approach will work long-term is because we will
migrate the source code to use a much simpler linker script, similar to the weapon0.ld located in the root of our repo.

I don't think that is the case, but that is largely an academic argument because this approach works in the same way as e_init itself does.  If it doesn't work in the future, then this isn't the only thing.

Xeeynamo:
> The primary issue is that in us the order goes e_init > BackgroundBlockInit > RedDoorTiles with that data being contiguous, but pspeu goes RedDoorTiles > e_init with BackgroundBlockInit being in a very different location.

Couldn't we leave `g_eRedDoorUV` in `e_red_door.c` and have `OVL_EXPORT(RedDoorTiles)` in a separate C file? That way `.text`+`.data` will still respect the same order, and both PSX and PSP can keep using the same files without ifdeffing.

ProjectOblivion:
You mean having what is in the PR as e_red_door_uv be e_red_door and what is currently e_red_door be e_red_door_tiles?
It seems like that would work fine, I can't think of a reason off hand why it would be an issue.

ProjectOblivion:
Swapping UV and tiles worked without issue.

src/st/no1/e_room_bg.c
```diff
@@ -14,8 +14,7 @@ u8 D_us_80180B88[] = {0x40, 0x58, 0xFF, 0x00};
 u8 D_us_80180B8C[] = {0x40, 0x59, 0xFF, 0x00};
 u8 D_us_80180B90[] = {0x04, 0x1E, 0xFF, 0x00};
 
-static ObjInit envTileInit[] = {
-
+static ObjInit2 OVL_EXPORT(BackgroundBlockInit)[] = {
```
Xeeynamo: Should we do the same for `OVL_EXPORT(BackgroundBlockInit)` too? I noticed it is often declared before `OVL_EXPORT(RedDoorTiles)` (see `src/st/wrp/e_init.c`) but not always (see PSP no4). I do not feel confident enough to bundle it with `e_room_bg.c` for now.

ProjectOblivion: This is exactly what I was talking about with RedDoorTiles.  us data has e_init > BackgroundBlockInit > RedDoorTiles, pspeu has RedDoorTiles > e_init and BackgroundBlockInit somewhere else.

Where this becomes an issue is the fact that RedDoorTiles has two pieces of data associated with it that are in two totally separate places.

BackgroundBlockInit was included with e_room_bg (st_debug) before I started on this.  While there are a couple slight deviations for BackgroundBlockInit, the location is fairly consistent within the version.  Obviously, the location will be different between us and pspeu.  e_room_bg is the only function within the decompiled code that uses BackgroundBlockInit.

Where is it that you are concerned with combining them as you would if it were a newly decompiled/imported function?

Xeeynamo: > Where is it that you are concerned with combining them as you would if it were a newly decompiled/imported function?

that data is not contiguous for the same reason I explained in the other comments.

ProjectOblivion: Assuming you're referring to:
> The .data you defined in US BO4 (as an example) is e_init, e_room_bg, e_red_door, e_lock_camera and so on. But the .text order is e_room_bg, e_lock_camera. There is no e_red_door between. Then further down I see .data on st_update, collision, e_red_door_uv, st_common and so on, with .text ordered as st_update, collision, create_entity (this one does not have any data), e_red_door, st_common and so on. This suggests the data was correct as before, and it should not be split in the separate file e_red_door_uv.

Is it simply that you're looking for this same type of order alignment between .data and .text for e_room_bg?

ProjectOblivion:
I moved BackgroundBlockInit to it's own file, but I feel like it's moving in the wrong direction.  I can't say that keeping it with e_room_bg is definitely the right answer, but I think it's a better answer for the knowledge we currently have.

PR #2657, opened 2025-06-11
Merged `pspeu` `e_init` and matched `e_init.c` symbols between `us` and `pspeu`

ProjectOblivion:


src/st/dai/e_init.c
```diff
@@ -153,12 +153,16 @@ EInit OVL_EXPORT(EInitBreakable) = {ANIMSET_DRA(3), 0, 0, PAL_DRA(0), 0};
 
 #include "../e_init_common.h"
 
-// Looks to be unused
+// Unused, not in psp
+#ifndef VERSION_PSP
```
JoshSchreuder: Are these ifndefs required or does PSP automatically deadstrip these?

ProjectOblivion: Not necessary, apparently.

JoshSchreuder: Unrelated to this PR but there is a bunch of these left around the place from before Xee got deadstripping working which can likely be removed

src/st/dai_psp/e_red_door.c
```diff
@@ -1,6 +1,12 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "../dai/dai.h"
 
+u16 OVL_EXPORT(RedDoorTiles)[][8] = {
+    {0x04B3, 0x04B3, 0x04B3, 0x04B3, 0x0000, 0x0000, 0x0000, 0x0000},
+    {0x000D, 0x000E, 0x000F, 0x000E, 0x01F0, 0x01F6, 0x020E, 0x01F6},
+    {0x000A, 0x000B, 0x000C, 0x000B, 0x0086, 0x0088, 0x00A1, 0x0088},
+};
```
Xeeynamo: we would need to separate this as well

PR #2658, opened 2025-06-11
Refactor `dra_psp` menu.c

gamezter:
Here I:

1. Rename `dra_psp/E588` to `dra_psp/menu`
2. Merge `dra_psp/d_6B420` into `dra_psp/menu` to match `dra/menu` closer
3. Import `dra_psp/menu` bss
4. Convert `MenuData g_MenuData;` to `MenuContext g_Menus[NUM_MENU]` and remove `MenuData` struct

For some reason, bss data for `dra/menu` is in the same order as `dra_psp/menu` except for `g_PrevEquippedWeapons`.


src/dra/bss.c
```diff
@@ -92,10 +92,8 @@ s32 g_IsSelectingEquipment;
 s32 g_EquipmentCursor;
 s32 D_80137614;
 s32 g_EquipOrderType;
-MenuData g_MenuData;
-s32 D_801377FC[NUM_MENU];
-s32 D_8013783C;
-s32 D_80137840;
+MenuContext g_Menus[NUM_MENU];
```
Xeeynamo, line 95: I am positively surprised this worked! I remember we had to containerise this field in a struct due to the `MenuContext` struct not being 4-byte aligned. I am glad you found a way to simplify it.

gamezter:
I think the two could be merged eventually. There are still 3 functions that I havent managed to get a match on PSP, and a few places where the code 'should' be the same but I havent found a way to get it to match on all platforms

PR #2666, opened 2025-06-15
Partial configuration for BO4 assets

hohle:
BO4 is the first of several overlays which use yet another `Overlay` struct. This follows the same layout as `Overlay` and `AbbreviatedOverlay`, but contains only two sprite banks and no cutscene handler. There are a few other overlays that use this structure that haven't been started yet.

`sotn-assets` is having trouble with the sprite banks (possibly because of this new overlay struct), as well as the palette definitions, so those are imported in code. Everything else is extracted as assets.

### DIFF
--- a/config/assets.us.yaml
+++ b/config/assets.us.yaml
@@ -573,6 +573,22 @@ files:
           - [0x1308, skip]
           - [0x1424, cutscene, cutscene_data]
           - [0x15DC, skip]
+  - target: disks/us/BOSS/BO4/BO4.BIN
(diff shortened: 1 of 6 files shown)


hohle:
@ProjectOblivion found that there are 3 header types (4 if you count SEL) after looking at the remaining overlays. I'm not sure if we'll need to add any more unique headers.

ProjectOblivion:
There are only 3 structures, SEL uses the Overlay struct with different values.

PR #2667, opened 2025-06-16
Begin cleaning up data for NO3 PSP

bismurphy:
With NO3 PSP fully decompiled, I'm now beginning to clean up the data and most importantly, get the two versions sharing everything.

I've done what I can here; most (if not all) of the remaining material is handled by the asset manager, which needs some slight tweaking, as mentioned in Discord. Once the asset tool is adjusted, I will follow up with another PR.


src/st/no3/d_water_data.c
```diff
@@ -1,7 +1,9 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-#include "../no3/no3.h"
+#include "no3.h"
+
+// 7C3 needs a name!
+u16 g_WaterSounds[] = {NA_SE_EV_WATER_SPLASH, 0x7C3};
```
sozud: Could we give it a placeholder name in the enum to track it?

bismurphy: Sure, I'll put something together.

PR #2668, opened 2025-06-16
Import `dra` D_801375BC

gamezter:
`D_801375BC` takes 4 bytes on PSP indicating the following data is part of the same array / struct. To fix this I put `D_801375C0` inside the same struct. 
I also made all 'moved' .bss data so far static since those variables are only used in the same file. 
The next variable to be imported is `g_canRevealMap ` according to PSP, but on PSX that variable is in the `.comm` segment which I have yet to figure out how to use.


src/pc/sotn.c
```diff
@@ -14,7 +14,7 @@
 
 u16 g_RawVram[VRAM_W * VRAM_H];
 GameApi g_ApiInit = {0};
-extern u8* g_DemoPtr;
+u8* g_DemoPtr;
```
Xeeynamo: Let's remove this one and the line `g_DemoPtr = g_DemoRecordingBuffer;`. We could restore the pc behaviour through `DEMO_KEY_PTR` later on

PR #2671, opened 2025-06-16
Decompiled, imported, and merged `dai` `e_bone_pillar`

ProjectOblivion:
EntityBonePillarHead
PSX: https://decomp.me/scratch/Ta5tY
PSP: https://decomp.me/scratch/v9dIV

EntityBonePillarSpikeBall
PSX: https://decomp.me/scratch/R6Nbi
PSP: https://decomp.me/scratch/8a5E9

EntityBonePIllarFireBreath
PSX: https://decomp.me/scratch/fC1P3
PSP: https://decomp.me/scratch/jvQxB

EntityBonePillarDeathParts
PSX: https://decomp.me/scratch/DORoo
PSP: https://decomp.me/scratch/nBGrs

### DIFF
--- a/src/st/dai/e_bone_pillar.c
+++ b/src/st/dai/e_bone_pillar.c
@@ -284,7 +284,7 @@ void EntityBonePillarSpikeBall(Entity* self) {
                 entity->nFramesInvincibility = 0x10;
                 entity->stunFrames = 4;
                 entity->hitEffect = 1;
-                entity->ext.et_bonePillar.unkB0.i.hi = 0;
+                entity->ext.et_bonePillar.unkB2 = 0;
                 entity->flags =
                     FLAG_POS_CAMERA_LOCKED | FLAG_KEEP_ALIVE_OFFCAMERA;
                 g_api.func_80118894(entity);
--- a/include/entity.h
+++ b/include/entity.h
@@ -3182,23 +3182,24 @@ typedef struct {
 } ET_801C12B0;
 
 typedef struct {
-    struct Primitive* prim;
-    s16 unk80;
-    s16 : 16;
-    u8 unk84;
-    u8 unk85;
-    u8 unk86;
-    struct Entity* unkEntity;
-    u8 : 8;
-    s32 : 32;
-    s32 : 32;
-    s32 : 32;
-    s32 : 32;
-    s32 : 32;
-    s32 : 32;
-    s32 : 32;
-    s32 : 32;
-    f32 unkB0;
+    /* 0x7C */ struct Primitive* prim;
+    /* 0x80 */ s16 unk80;
+    /* 0x82 */ s16 : 16;
+    /* 0x84 */ u8 unk84;
+    /* 0x85 */ u8 unk85;
+    /* 0x86 */ u8 unk86;
+    /* 0x88 */ struct Entity* unkEntity;
+    /* 0x89 */ u8 : 8;
+    /* 0x90 */ s32 : 32;
+    /* 0x94 */ s32 : 32;
+    /* 0x98 */ s32 : 32;
+    /* 0x9C */ s32 : 32;
+    /* 0xA0 */ s32 : 32;
+    /* 0xA4 */ s32 : 32;
+    /* 0xA8 */ s32 : 32;
+    /* 0xAC */ s32 : 32;
+    /* 0xB0 */ s16 : 16;
+    /* 0xB2 */ s16 unkB2;
 } ET_BonePillar;
 
 typedef struct {


src/st/dai/e_bone_pillar.c
```diff
@@ -0,0 +1,489 @@
...
+            self->drawFlags = FLAG_DRAW_ROTATE;
+            self->hitPoints = 0x7FFF;
+            self->hitboxState = 1;
+            self->velocityX = FIX(-0.75);
+            self->velocityY = FIX(-1.0);
+            self->step++;
+            entity = AllocEntity(&g_Entities[32], &g_Entities[47]);
+            if (entity != NULL) {
+                DestroyEntity(entity);
+                entity->entityId = E_SEALED_DOOR;
```
ProjectOblivion, line 277: I feel like this should instead be a second spiked ball E_ID, rather than E_SEALED_DOOR.  Seems like something may be slightly off with the E_IDs in dai.h

src/st/dai/e_bone_pillar.c
```diff
@@ -0,0 +1,489 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "dai.h"
+
+#ifdef VERSION_PSP
+s32 D_us_80181C54[] = {0x00200000, 0x00040000, 0xFFFC0004, 0x0000FFF8};
```
sozud: These are arguments to UnkCollisionFunc3 so they ought to be s16
```
s32 UnkCollisionFunc3(s16* sensors) {
``

File: src/st/dai/e_bone_pillar.c [hunk 3]
```diff
@@ -0,0 +1,489 @@
...
+            entity = AllocEntity(&g_Entities[32], &g_Entities[47]);
+            if (entity != NULL) {
+                DestroyEntity(entity);
+                entity->entityId = E_SEALED_DOOR;
+                entity->pfnUpdate = EntityBonePillarSpikeBall;
+                entity->attack = self->attack;
+                entity->hitboxWidth = self->hitboxWidth;
+                entity->hitboxHeight = self->hitboxHeight;
+                entity->hitboxState = 2;
+                entity->attackElement = 0x8000;
```
sozud: ELEMENT_FIRE

src/st/dai/e_bone_pillar.c
JoshSchreuder: Additionally I would guess the PSX exclusive part is probably a separate array and deadstripped on PSP side.
So you could pull it out into a separate import and remove the #ifdef I think

This is fairly common where we have unused arrays on the PSX side which are not present in PSP data

JoshSchreuder: Also minor note, but good practice to make all of this data `static` if possible, assuming it's only used in this file

src/st/dai/e_bone_pillar.c
```diff
@@ -0,0 +1,489 @@
...
+s32 D_us_80181C84[] = {0x000E0000, 0x00000000};
+s16 D_us_80181C8C[] = {
+    0x0002, 0xFFF1, 0x0002, 0xFFEF, 0xFFFC, 0x000A, 0xFFFC, 0x0009};
+s16 D_us_80181C9C[] = {0x0026, 0x0026, 0x000E, 0x000A};
+s16 D_us_80181CA4[] = {0x02C0, 0x0540, 0x0AC0, 0x0D40};
+s32 D_us_80181CAC[] = {
+    0x02020130, 0x02020102, 0x02020102, 0x02020102, 0x02020102,
+    0x04020302, 0x04050502, 0x04020602, 0x07020602, 0x07020602,
+    0x07020802, 0x07020802, 0x04110602, 0x000000FF};
+s32 D_us_80181CE4[] = {
```
JoshSchreuder: Similarly to sozud's comment above, these are `AnimateEntity` arguments so should be `u8` arrays

src/st/dai/e_bone_pillar.c
```diff
@@ -0,0 +1,489 @@
...
+            }
+        }
+        entity = AllocEntity(&g_Entities[224], &g_Entities[256]);
+        if (entity != NULL) {
+            CreateEntityFromEntity(E_EXPLOSION, self, entity);
+            entity->params = 2;
+        }
+        if ((!self->params) &&
+            ((entity = self + 1, (entity->entityId) == E_BONE_PILLAR_HEAD) ||
+             ((entity->entityId) == E_BONE_PILLAR_SPIKE_BALL))) {
```
JoshSchreuder: Do you have a scratch for this? The assign and condition within an `if`, while valid, is pretty uncommon within the codebase, and more likely to be a more simple assignment enclosed within the `if`. Should be easier to read that way also.

eg. something like

```c
if (!self.params) {
   entity = self + 1;
   if (entity->entityId == E_BONE_PILLAR_HEAD || entity->entityId == E_BONE_PILLAR_SPIKE_BALL) {
     entity->ext.et_bonePillar.unk85 = 1;
   }
}
```

src/st/dai/e_bone_pillar.c
```diff
@@ -0,0 +1,489 @@
...
+        if ((self->pose == 0x11) && (unkVar & 0x80)) {
+            entity = AllocEntity(&g_Entities[160], &g_Entities[192]);
+            if (entity != NULL) {
+                CreateEntityFromEntity(E_BONE_PILLAR_FIRE, self, entity);
+                entity->facingLeft = self->params;
+                if (self->params) {
+                    entity->posY.i.hi -= 0xE;
+                    break;
+                }
+                entity->posY.i.hi += 8;
```
gamezter: This is probably an `if(){}else{}`

src/st/dai/e_bone_pillar.c
```diff
@@ -0,0 +1,489 @@
...
+        InitializeEntity(g_EInitBonePillarPieces);
+        self->hitboxState = 0;
+        self->animCurFrame = (self->params & 0xF) + 0x12;
+        self->zPriority += self->params & 0xF;
+        if (self->params & 1) {
+            self->ext.et_bonePillar.unk80 = 0x18;
+            break;
+        }
+        self->ext.et_bonePillar.unk80 = 0xC;
+        break;
```
gamezter: This is probably an `if(){}else{}`

src/st/dai/e_bone_pillar.c
ProjectOblivion: This one was a little difficult for me.  I felt like it was overly complex, but I couldn't figure out a better way.  Since it is written this way in e_fishhead, that's what I went with.  Your suggestion worked fine, though, and I'll be pushing it up with the other review adjustments.

I also updated the description of this PR with the scratches.

src/st/dai/e_bone_pillar.c
ProjectOblivion: @JoshSchreuder How would I split it if there is only one call using the symbol?

JoshSchreuder: @ProjectOblivion would need to test on both PSX and PSP but it should work if you take the common part as the current symbol, and then define the PSX only data as a separate array:

s32 D_FOO[] = { ... };
s32 unused[] = { ... };

The variable with the symbol is used in code and will be kept on PSX and PSP, the unused data will get stripped on PSP side so you don't need ifdef

ProjectOblivion: Oh, I see what you mean now.  Essentially defining the data for psx, but since it isn't used it isn't relevant to be in the variable itself.  Thanks!

JoshSchreuder: Example here as well if you need https://github.com/Xeeynamo/sotn-decomp/blob/master/src/st/st0/2A218.c

src/st/dai/e_bone_pillar.c
Xeeynamo: How did this get renamed as such? 
We should probably revert it to func_80xxxxxx if it is wrong. This can be done in a separate PR.

src/st/dai/e_bone_pillar.c
```diff
@@ -0,0 +1,495 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "dai.h"
+
+static s16 sensors_unk1[] = {
```
Xeeynamo: Nit:
```suggestion
static s16 sensors_bone_pillar[] = {
```

src/st/dai/e_bone_pillar.c
```diff
@@ -0,0 +1,495 @@
...
+    0x0000, 0x0020, 0x0000, 0x0004, 0x0004, 0xFFFC, 0xFFF8, 0x0000};
+// This appears to be unused and deadstripped on PSP
+static s16 unused[] = {
+    0x0000, 0x0008, 0x0000, 0x0004, 0x0002, 0xFFFC, 0xFFFC, 0x0000,
+    0x0000, 0x0024, 0x0000, 0x0004, 0x0002, 0xFFFC, 0xFFFC, 0x0000};
+static s16 sensors_unk2[] = {0x0000, 0x000E, 0x0000, 0x0000};
```
Xeeynamo: Nit:
```suggestion
static s16 sensors_spike_ball[] = {0x0000, 0x000E, 0x0000, 0x0000};
```
Since you were asking how to name the sensors

src/st/dai/e_bone_pillar.c
```diff
@@ -0,0 +1,495 @@
...
+        if (primIndex == -1) {
+            DestroyEntity(self);
+            return;
+        }
+        self->flags |= FLAG_HAS_PRIMS;
+        self->primIndex = primIndex;
+        prim = &g_PrimBuf[primIndex];
+        self->ext.et_bonePillar.prim = prim;
+        prim->tpage = 0x13;
+        prim->clut = 0x221;
```
Xeeynamo, line 370: It would be a good idea to start creating an enum for this. I did this for the first time in our code base for Maria (see `MarPalette` in `src/maria/maria.h`) and it was quite useful.

src/st/dai/e_init.c
```diff
@@ -162,10 +162,10 @@ EInit g_EInit_801809C8 = {ANIMSET_OVL(2), 0, 72, PAL_DRA(202), 5};
 EInit D_us_801809D4 = {ANIMSET_OVL(2), 37, 72, PAL_DRA(201), 101};
 EInit D_us_801809E0 = {ANIMSET_OVL(4), 0, 72, PAL_DRA(533), 180};
 EInit D_us_801809EC = {ANIMSET_OVL(0), 0, 0, PAL_DRA(0), 181};
-EInit D_us_801809F8 = {ANIMSET_OVL(5), 0, 73, PAL_DRA(534), 183};
-EInit D_us_80180A04 = {ANIMSET_OVL(5), 0, 73, PAL_DRA(534), 2};
-EInit D_us_80180A10 = {ANIMSET_OVL(5), 0, 73, PAL_DRA(534), 184};
-EInit D_us_80180A1C = {ANIMSET_OVL(5), 0, 73, PAL_DRA(534), 185};
+EInit g_EInitBonePillarHead = {ANIMSET_OVL(5), 0, 73, PAL_DRA(534), 183};
```
Xeeynamo, line 165:
```suggestion
EInit g_EInitBonePillarHead = {ANIMSET_OVL(5), 0, 73, PAL_DRA(0x216), 183};
```
not for this PR, but we should use hex numbers as it is easier to match them with the decompiled code. The number used here is the same used in `prim->clut` but without the `0x8000` bit.

src/st/dai/e_bone_pillar.c
ProjectOblivion: It didn't get renamed, it was set as this initially by mapping it to EntityUpdates.  I raised this point so that I can understand if and how I should have done it differently.

Link: https://github.com/Xeeynamo/sotn-decomp/pull/2632/files#diff-acf8b11fdc7196b7e1df5ca1c522f45bac9da708303fc4f9e8e1f0a2aa15416a

src/st/dai/e_bone_pillar.c
ProjectOblivion: Thanks for this. I didn't like the name either, but I wasn't sure how to name it.  I asked the question in Discord naming if you wouldn't mind giving some more universal guidance on it.
Nevermind, I see that you already did in the next comment.

PR #2674, opened 2025-06-17
Added entity tables for all overlays

ProjectOblivion:


sozud (approved):
Haven't checked any of the values but looks reasonable to me

PR #2676, opened 2025-06-17
Cleanup NO3

bismurphy:
This finishes pulling in all data for PSP and US NO3. It also removes all unneeded entries from no3.h, and completely removes everything in symbols.pspeu.stno3.txt.

With this, I think my work on NO3 is done. There is always more to do (could replace every D_812345678 variable name), but for now, I think this overlay is in good shape.


bismurphy:
Ah - building locally I was doing US first, and then PSPEU would use US's assets. Looks like CI runs them separately so they can't cross-connect the resources. I guess I'll make PSP definitions for the asset tool then.

bismurphy:
@Xeeynamo I don't think the assets tool is capable of handling PSP at this point? I'm getting errors due to the hard-coded values of `RamStageBegin` and similar variables. I will mark this PR as draft, until the tool is capable of working.

bismurphy:
Reopening due to accidental closing.

bismurphy:
For now I have commented out the references to the asset tool. This should make this PR work for now.

PR #2677, opened 2025-06-18
Failing Lints & AttackElement Lint

hohle:
Adds two new "failing" lints which identify problems but do not automatically fix them. The first identifies automatic symbols with addresses that are within the `Entity` table. The second finds RODATA strings being passed as global variables to `FntPrint`. These will cause `make format` to fail with file and line number information.

Adds an `AttackElement` transformer for finding `attackElement` field values.


tools/lints/sotn-lint/src/linter.rs
```diff
@@ -0,0 +1,79 @@
...
+        let Some(captures) = SYMBOL_PATTERN.captures(&stripped.as_str()) else {
+            return Ok(());
+        };
+
+        let addr_str = captures.get(1).map(|m| m.as_str().to_string());
+        let addr = u32::from_str_radix(&addr_str.clone().unwrap(), 16).unwrap();
+
+        if (addr >= 0x800733D8 && addr < 0x8007EF1C) ||
+            (addr >= 0x091e1680 && addr < 0x91ED1C4) {
+            return Err("global should be entity".to_string())
```
sozud: I'm thinking people are going to be confused about this message. Maybe like "Variable X should index into g_Entities" or whatever?

tools/lints/sotn-lint/src/linter.rs
```diff
@@ -0,0 +1,79 @@
...
+            regex: Regex::new(expression).unwrap()
+        }
+    }
+}
+
+impl Linter for RegexLinter {
+    fn check_line(&self, line: &str) -> Result<(), String> {
+        let stripped = line.to_string().strip_line_comment();
+        if self.regex.is_match(&stripped) {
+            Err(format!("{}: {}", self.name, self.regex))
```
sozud, line 75: What does this error message look like?

tools/lints/sotn-lint/src/linter.rs
Xeeynamo: Agree here. I'd like to see a hint of what the actual variable should be

hohle: The line content is displayed by `main.rs`, but I've updated the message to include additional context:

Untitled

tools/lints/sotn-lint/src/linter.rs
hohle: See comment above.

hohle:
Another easy lint to added would be checking SFX ID constants:

```bash
$> grep -r 'Play.*Sfx.*(0x' src/ | wc -l 
117
```

Possibly increment operator, though that's not universal (still investigating):

```bash
$> grep -r ' += 1;' src | wc -l
218
```

PR #2678, opened 2025-06-18
Pulled in code and merged known segments for DAI

ProjectOblivion:
I have also decompiled e_cutscene, but I'm leaving that for a separate PR since e_cutscene has a lot going on.

I want to say that I did very little work for this comparatively.  This is 99% just pulling in what has already been done, so thank you to anybody who decompiled any of the code in these .h files.


src/st/dai/dai.h
```diff
@@ -1,8 +1,15 @@
...
+#ifndef DAI_H
+#define DAI_H
+
 #include "stage.h"
 
+#define STAGE_IS_DAI
 #define OVL_EXPORT(x) DAI_##x
 
+#define STAGE_FLAG OVL_EXPORT(STAGE_FLAG)
+#define SEALED_DOOR_PALETTE 0x245
```
JoshSchreuder, line 11: `DaiPalette` for this one perhaps?

src/st/dai/e_init.c
```diff
@@ -172,11 +172,11 @@ EInit D_us_80180A28 = {ANIMSET_OVL(7), 1, 72, PAL_UNK22A, 118};
...
 EInit D_us_80180A40 = {ANIMSET_OVL(8), 0, 75, PAL_UNK22B, 128};
 EInit D_us_80180A4C = {ANIMSET_OVL(8), 0, 75, PAL_UNK22B, 3};
-EInit D_us_80180A58 = {ANIMSET_OVL(9), 0, 74, PAL_BAT, 64};
+EInit g_EInitBat = {ANIMSET_OVL(9), 0, 74, PAL_BAT, 64};
 EInit D_us_80180A64 = {ANIMSET_OVL(10), 0, 79, PAL_UNK22F, 114};
 EInit D_us_80180A70 = {ANIMSET_OVL(10), 0, 79, PAL_UNK22F, 115};
-EInit D_us_80180A7C = {ANIMSET_OVL(11), 1, 72, PAL_UNK200, 11};
-EInit D_us_80180A88 = {ANIMSET_OVL(11), 43, 72, PAL_UNK200, 12};
+EInit g_EInitSkelerang = {ANIMSET_OVL(11), 1, 72, PAL_UNK200, 11};
+EInit g_EInitSkelerangBoomerang = {ANIMSET_OVL(11), 43, 72, PAL_UNK200, 12};
```
JoshSchreuder: Can rename these palette enum values with known names

src/st/dai/header.c
```diff
@@ -2,12 +2,16 @@
 #include "dai.h"
 
 extern RoomHeader OVL_EXPORT(rooms)[];
-extern s16** OVL_EXPORT(spriteBanks)[];
+extern SpriteParts* OVL_EXPORT(spriteBanks)[];
 extern u_long* OVL_EXPORT(cluts)[];
+#ifdef VERSION_PSP
+extern LayoutEntity* g_pStObjLayoutHorizontal[];
+#else
 extern LayoutEntity* OVL_EXPORT(pStObjLayoutHorizontal)[];
```
JoshSchreuder, line 12: Does including `pfn_entity_update.h` here allow you to remove the ifdefs?

src/st/dai/dai.h
ProjectOblivion: Yeah, good call.

ProjectOblivion: I can't get the logic to work with an enum, so unless somebody knows a way to do it that I'm missing, it would need to be a separate PR converting EntitySealedDoor to use an enum instead of the macro.

src/st/dai/header.c
ProjectOblivion: You mean including `pfn_entity_update.h` here instead of `st_init`?
I'll look into it.

JoshSchreuder: I think it's just externs so it could be in both

It should handle the conditional define here for you I think

ProjectOblivion: It is just externs, but I don't see any sense in having it be in multiple places.
How about including it in `dai.h`, is that a dumb idea?

ProjectOblivion: So, it looks like `Overlay` wants `g_pStObjLayoutHorizontal` as a constant, but it is currently defined in `pfn_entity_update.h` as a pointer for `psp` and required as such in the code. `pfn_entity_update.h` can probably be adjusted to work for both the header and existing code, but that would be outside of the scope of this PR.  For now, I've replicated the relevant logic (adjusted for Overlay) from `pfn_entity_update.h` in this file.

ProjectOblivion:
I added st_init, but I didn't realize it had already been approved.  Just didn't want it to seem like I was trying to sneak something in.

src/st/dai/e_cutscene_script.c
```diff
@@ -0,0 +1,29 @@
...
+};
+
+u8 OVL_EXPORT(cutscene_script_en)[] = {
+#include "../dai_psp/gen/cutscene_script_en.h"
+};*/
+#else
+// u8 OVL_EXPORT(cutscene_script)[] = {
+// #include "gen/cutscene_script.h"
+// };
+#endif
```
Xeeynamo, line 29: unused?

src/st/dai/st_init_psp.c
```diff
@@ -0,0 +1,157 @@
...
+extern s32 D_pspeu_09297270;
+extern s32 D_pspeu_09297278;
+extern s32 D_pspeu_09297280;
+
+void InitEntityIds(void) {
+    D_pspeu_09297280 = 0x11;
+    D_pspeu_09297278 = 0x12;
+    D_pspeu_09297270 = 0x13;
+    D_pspeu_09297268 = 0x14;
+    D_pspeu_09297260 = 0x15;
```
Xeeynamo: These need to be initialized according to the various `DaiEntities`

src/st/dai/st_init_psp.c
```diff
@@ -0,0 +1,157 @@
...
+        (u8*)OVL_EXPORT(cutscene_script_fr),
+        (u8*)OVL_EXPORT(cutscene_script_sp),
+        (u8*)OVL_EXPORT(cutscene_script_ge),
+        (u8*)OVL_EXPORT(cutscene_script_it));
+    InitEntityIds();
+    PfnEntityUpdates = OVL_EXPORT(EntityUpdates);
+    g_pStObjLayoutHorizontal = &D_8D2E2B0;
+    g_pStObjLayoutVertical = &D_8D2E384;
+    func_892A018();
+    memcpy(&g_api, &OVL_EXPORT(Overlay), 0x40U);
```
Xeeynamo:
```suggestion
    memcpy(&g_api, &OVL_EXPORT(Overlay), sizeof(Overlay));
```

src/st/e_clouds.h
```diff
@@ -0,0 +1,232 @@
...
+    {.unkPtr = D_us_80181334, .unk4 = 0xFF20, .priority = 0x1C},
+    {.unkPtr = D_us_80181374, .unk4 = 0xFF00, .priority = 0x18},
+    {.unkPtr = D_us_801813B4, .unk4 = 0x0060, .priority = 0x1C}};
+
+static SVECTOR empty = {0};
+
+void EntityClouds(Entity* self) {
+#ifdef VERSION_PC
+    u8 sp[SP_LEN];
+#endif
```
Xeeynamo: is this needed? I noticed it is unused

ProjectOblivion: I have no idea, this was just a copy paste from RBO3 e_clouds.c

src/st/dai/e_cutscene_script.c
ProjectOblivion: This is ultimately because of the issue with the asset tool parsing the bytes.  Since it is referenced in the config, I included this as a stub.

src/st/e_clouds.h
Xeeynamo: let's delete it then. If something breaks, we will know

PR #2679, opened 2025-06-18
Added palette enum and change e_init palettes to hex

ProjectOblivion:


sozud:
Needs a rebase

src/st/dai/dai.h
```diff
@@ -3,6 +3,22 @@
 
 #define OVL_EXPORT(x) DAI_##x
 
+enum DaiPalette {
+    PAL_NONE = 0,
+    PAL_UNKC9 = 0xC9,
+    PAL_UNK200 = 0x200,
+    PAL_UNK215 = 0x215,
+    PAL_BONE_PILLAR = 0x216,
```
sozud, line 12: Is the plan to actually use the enum in a later PR?

ProjectOblivion: I have no plan, this was something suggested by @Xeeynamo .
It does make things a lot easier to correlate since this value is used in a lot of circumstances which relate to each other and is a value within the EInits.

sozud: OK I would think that Xeeynamo meant for the enum to be used, otherwise this is just dead code

Xeeynamo:
```
EInit g_EInitBonePillarHead = {ANIMSET_OVL(5), 0, 73, PAL_DRA(0x216), 183};
EInit g_EInitBonePillarPieces = {ANIMSET_OVL(5), 0, 73, PAL_DRA(0x216), 2};
EInit g_EInitBonePillarFireBreath = {
    ANIMSET_OVL(5), 0, 73, PAL_DRA(0x216), 184};
EInit g_EInitBonePillarSpikeBall = {ANIMSET_OVL(5), 0, 73, PAL_DRA(0x216), 185};
```
I think `PAL_BONE_PILLAR` was meant to be used here

src/st/dai/background_block_init.c
```diff
@@ -8,7 +8,7 @@ ObjInit2 OVL_EXPORT(BackgroundBlockInit)[] = {{
     .zPriority = 0x1FA,
     .facingLeft = 0,
     .unk5A = 0,
-    .palette = PAL_DRA(0),
+    .palette = PAL_NONE,
```
ProjectOblivion, line 11: It makes sense to me to use the enum here also, but I'm not certain about it.  I can change it back if it isn't right.

src/st/dai/dai.h
ProjectOblivion: It was.

ProjectOblivion:
I think I got all of the enum places now.

PR #2680, opened 2025-06-18
Import `no0` header and spritebanks

JoshSchreuder:


tools/sotn-assets/assets/spritebanks/spritebanks.go
```diff
@@ -96,8 +96,8 @@ func buildSprites(fileName string, outputDir string) error {
 	if err := json.Unmarshal(data, &spritesBanks); err != nil {
 		return err
 	}
-	if len(spritesBanks.Indices) != banksCount {
-		return fmt.Errorf("the number of banks must be exactly %d, got %d", banksCount, len(spritesBanks.Banks))
+	if len(spritesBanks.Indices) != banksCount && len(spritesBanks.Indices) != banksCount + 1 {
+		return fmt.Errorf("the number of banks must be exactly %d, got %d", banksCount, len(spritesBanks.Indices))
```
JoshSchreuder, line 100: @Xeeynamo I'm guessing the message here was a mistake? Updated it to match the condition check.

tools/sotn-assets/assets/spritebanks/spritebanks.go
```diff
@@ -96,8 +96,8 @@ func buildSprites(fileName string, outputDir string) error {
 	if err := json.Unmarshal(data, &spritesBanks); err != nil {
 		return err
 	}
-	if len(spritesBanks.Indices) != banksCount {
-		return fmt.Errorf("the number of banks must be exactly %d, got %d", banksCount, len(spritesBanks.Banks))
+	if len(spritesBanks.Indices) != banksCount && len(spritesBanks.Indices) != banksCount + 1 {
```
JoshSchreuder, line 99: This is very much a hack around.
NO0 has 25 sprite banks, vs. the expected 24 `const banksCount = 24 // the number seems to be fixed`

I don't know if you have any good ideas for this @Xeeynamo?

src/st/entity_axe_knight.h
```diff
@@ -124,7 +124,7 @@ static void AxeKnightUnkFunc1(AxePrim* prim) {
 }
 
 // Called by EntityAxeKnight
-extern s16* sprites_nz0_3[];
+extern s16* SPRITE_BANK[];
```
JoshSchreuder, line 127: This code is shared between `nz0` and `no0` so we shouldn't reference a specific spritebank by name. I've made this a define so the overlay can specify it.

Potentially this define should be indexing into `OVL_EXPORT(spriteBanks)`, eg. `OVL_EXPORT(spriteBanks)[8]` rather than also by name? I did not yet test this matches, but assuming it does would this be preferable?

src/st/no0/e_marionette.c
```diff
@@ -36,7 +36,7 @@ static u8 D_us_8018214C[] = {8, 23, 6, 22, 6, 24, 6, 21, 6, 20, 6, 23, 4,  19,
 static u8* D_us_80182178[] = {D_us_801820FC, D_us_80182124, D_us_8018214C};
 static u8 D_us_80182184[] = {6, 1, 4, 1, 4, 2, 6, 3, 5, 4, 5, 5, 0, 0};
 
-extern s16* D_us_801BA12C[];
+extern s16* sprites_no0_8[];
```
JoshSchreuder, line 39: Marionette is only in NO0 so I think this is fine hardcoded

tools/sotn-assets/assets/spritebanks/spritebanks.go
Xeeynamo: Odd but we can live with that. I think too much stuff is hardcoded in the sotn-asset tool for those asset types that are related to the stage overlay header. I'll make a note to fix that in the future.

PR #2683, opened 2025-06-19
Share `dra_psp` data

gamezter:


sozud:
Fixing the build error will involve removing this and changing the loader code https://github.com/Xeeynamo/sotn-decomp/blob/da125aba37b0021ad60ef62b7db47685116d7232/src/pc/sotn.c#L55

gamezter:
We no longer generate `assets/dra/g_GfxEquipIcon.bin`, how should I handle the loader code ?

sozud:
It doesn't need to be loaded because the data is already present in g_GfxEquipIcon due to the header file. You can remove the loader code

PR #2685, opened 2025-06-19
Push sotn-build image to Docker Hub

Xeeynamo:
Depending to #2684 . The goal here is to have an image ready to be pulled that can be used to build SOTN with the minimum amount of steps possible. This will also be used to build #2638 in a 5yo step-by-step guide.

I am not sure if we can use this image for our GitHub CI. I plan to experiment on it on a different pull request.

I decided to use Docker Hub as a registry as it is the default one Docker and Docker Desktop use. If we ever want to use a Docker image on GHA, I will also upload it to GitHub Registry to prevent inbound traffic and the Docker Hub pull limit.

### DIFF
--- a/.github/workflows/dockerfile.yaml
+++ b/.github/workflows/dockerfile.yaml
@@ -6,8 +6,8 @@ on:
       - master
     paths:
       - 'Dockerfile'
-      - 'tools/**'
-      - 'Makefile*'
+      - 'tools/requirements-debian.txt'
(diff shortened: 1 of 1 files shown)


sozud (approved):
From what I see none of the real part of the repo is baked into the container so I think this should be fine:

https://github.com/Xeeynamo/sotn-decomp/blob/b50ca8a33a15ea88e7875da990815f9fdd8ecfc2/Dockerfile

PR #2686, opened 2025-06-19
Several More Doppleganger Functions

hohle:
`ControlBatForm` and `EntityDopplegangerBlinkWhite` are analogs to Alucard's functions in DRA. There continues to be enough difference between the two being just `g_Player`/`g_Dop` and `PLAYER`/`DOPPLEGANGER` that its hard to suggest combining the implementations as shared code. All state enums are shifted by one, there are frequently differences in sound effects, entire sections related to forms Doppleganger can't take, pallet values, etc.

BO4's `func_us_801B5774` and RBO5's `func_us_801B4210` are shared closely with the warp room's lighting effect. AFAICT this is never used for RBO5 and suggests it was just copied from BO4.

`F_BO4.BIN` and `F_RBO5.BIN` were missing from the checksum file and have been added.


hohle:
Failure seems to be caused by a transient go download failure.

src/boss/bo4/unk_46E7C.c
```diff
@@ -486,21 +486,329 @@ void func_8011690C(s16 arg0) {
...
+            }
+        } else {
+            if (g_Dop.unk66 == 0) {
+                newEntity = CreateEntFactoryFromEntity(
+                    g_CurrentEntity, FACTORY(BP_BLINK_WHITE, 0x20), 0);
+
+                func_8010FAF4();
+                g_Dop.unk66++;
+            }
+            DecelerateX(0x480);
```
sozud: Does FIX work here? Same for all other DecelerateX/Y

hohle: Yes. Of course!

PR #2688, opened 2025-06-19
Cleanup `dra` data types

gamezter:


src/dra/d_37d8.c
```diff
@@ -340,21 +340,22 @@ GfxBank** g_GfxSharedBank[] = {
...
-extern s16* D_800D34B8[];
-extern s16* D_800D372C[];
-extern s16* D_800D3B88[];
-extern s16* D_800D5628[];
-extern s16* D_800D327C[];
-extern s16* D_800D56AC[];
-extern s16* D_800D5868[];
-extern s16* D_800D684C[];
+extern s16* D_800CF324[]; // AluFrame
+extern s16* D_800CFF10[]; // SpritePart
```
gamezter: BTW the struct is called `SpriteParts`, but the asset importer calls it `spriteset`. It would be nice to use the same names

sozud (approved):
Looks good, @Xeeynamo should check the comment about `spriteset`

PR #2693, opened 2025-06-21
Decompiled `dai` `e_cutscene`

ProjectOblivion:
I did what I could to bring pspeu and us together, but I wasn't able to get it all the way.  I think there is the possibility of merging them with a couple ifdefs for additional psp code, but it seems to involve larger scale code adjustments that are beyond my current abilities.

There are two segments associated with the cutscene that seem like they should be a part of the segment.  I tried everything I could think of, but psx wouldn't accept it.  It seems that the cutscene section of code was likely heavily modified for pspeu, so it is difficult to draw correlations.  This additional segment appears to be code needed to "tare" the environment for the cutscene, for example changing Alucard to human form, centering the screen, positioning Alucard, the elevator in cen, etc.

That being said, I found that psx had a consistent order of:
```
1. CutsceneUnk1 -> SetCutsceneScript
2. CutsceneUnk3 -> CutsceneUnk4
3. DrawCutsceneActorName
4. SetCutsceneEnd -> CutsceneRun -> CutsceneSkip -> ScaleCutsceneAvatar
5. EntityCutScene
6. cutscene_init (three function segment, same order psx and psp)
```
And pspeu had a consistent order of (numbered order psx, visual order psp):
```
6. cutscene_init (three function segment, same order psx and psp)
0. GetLang
3. DrawCutsceneActorName
2. CutsceneUnk3 -> CutsceneUnk4
1. CutsceneUnk1 -> SetCutsceneScript
4. SetCutsceneEnd -> CutsceneRun -> CutsceneSkip -> ScaleCutsceneAvatar
5. EntityCutScene
```

- I added enum items to PlayerPad for the forms to be used instead of ifdefs
- I added a small ifdef at the beginning of cutscene_actor_name.h so that psx and psp can reference the same .h file, even though the code is different.  I intend to add get_lang.h to the psp version in a future PR.
- I created three new .h files consolidating functions using the consistent subordering
  - cutscene_script_text.h, which combines cutscene_unk1.h and set_cutscene_script.h
  - cutscene_script_box.h, which combines cutscene_unk3.h and cutscene_unk4.h
  - cutscene_exec.h (I intend to convert cutscene_run.h to this in a future PR), which combines set_cutscene_end.h, cutscene_run.h, and cutscene_scale_avatar.h
- Added cutscene asset extraction for both psx and psp.  pspeu required added the trailing data manually.
- Updated CSOP_SCRIPT_UNKNOWN_4 to be CSOP_HIDE_DIALOG
  - Added CSOP_HIDE_DIALOG  and CSOP_SCRIPT_UNKNOWN_18 macros

I did what made sense to do to begin cleaning up cutscene related code, but there is still a lot that can be done.


sozud:
This needs a rebase

src/st/cutscene_exec.h
```diff
@@ -0,0 +1,114 @@
...
+            entity->entityId = *g_Dialogue.scriptEnd++;
+            entity->pfnUpdate = OVL_EXPORT(EntityUpdates)[entity->entityId - 1];
+            entity->posX.i.hi = *g_Dialogue.scriptEnd++ * 0x10;
+            entity->posX.i.hi |= *g_Dialogue.scriptEnd++;
+            entity->posY.i.hi = *g_Dialogue.scriptEnd++ * 0x10;
+            entity->posY.i.hi |= *g_Dialogue.scriptEnd++;
+#ifdef STAGE_IS_DAI
+            entity->posX.i.hi -= g_Tilemap.scrollX.i.hi;
+            entity->posY.i.hi -= g_Tilemap.scrollY.i.hi;
+#endif
```
Xeeynamo: We want to move away from ifdeffing based on the stage the header is imported from. Different overlays will always have minor variants of certain functions. And we cannot keep enlarging those included files with those small variations.

Instead, it is better to keep those functions duplicates.

src/st/dai/cutscene_init.c
```diff
@@ -0,0 +1,233 @@
...
+    gTilemap = &g_Tilemap;
+    gEntities = g_Entities;
+
+    posScrollX = self->posX.i.hi + gTilemap->scrollX.i.hi;
+    posScrollY = self->posY.i.hi + gTilemap->scrollY.i.hi;
+
+    switch (self->step) {
+    case 0:
+        InitializeEntity(g_EInitInteractable);
+        self->animSet = -0x7FF1;
```
Xeeynamo: this is an ANIMSET_OVL

src/st/dai/cutscene_init.c
```diff
@@ -0,0 +1,233 @@
...
+        self->unk5A = 0x48;
+        self->palette = PAL_UNK251;
+        break;
+    case 1:
+        if (g_CutsceneFlags & 4) {
+            self->step++;
+        }
+        break;
+    case 2:
+        if (!AnimateEntity(&D_us_80181024, self)) {
```
Xeeynamo:
```suggestion
        if (!AnimateEntity(D_us_80181024, self)) {
```

src/st/dai/cutscene_init.c
```diff
@@ -0,0 +1,233 @@
...
+            SetStep(3);
+            self->velocityX = FIX(1.5);
+        }
+        break;
+    case 3:
+        pan = AnimateEntity(&D_us_80181030, self);
```
Xeeynamo:
```suggestion
        pan = AnimateEntity(D_us_80181030, self);
```
and similar

src/st/dai/cutscene_init.c
```diff
@@ -0,0 +1,233 @@
...
+        if (pan & 0x80 && (self->pose == 3 || self->pose == 7)) {
+            pan = (self->posX.i.hi - 120) / 16;
+            if (pan < -8) {
+                pan = -8;
+            }
+            if (pan > 8) {
+                pan = 8;
+            }
+            g_api.PlaySfxVolPan(0x64C, 0x50, pan);
```
Xeeynamo:
```suggestion
            g_api.PlaySfxVolPan(SFX_STOMP_SOFT_B, 0x50, pan);
```
and similar

src/st/dai/e_cutscene.c
```diff
@@ -1,24 +1,506 @@
...
+s32 g_IsCutsceneDone;
 
-INCLUDE_ASM("st/dai/nonmatchings/e_cutscene", DrawCutsceneActorName);
+// These three are reversed in order from psp
+#include "../cutscene_script_text.h"
+#include "../cutscene_script_box.h"
+#include "../cutscene_actor_name.h"
 
-INCLUDE_ASM("st/dai/nonmatchings/e_cutscene", SetCutsceneEnd);
+#include "../cutscene_exec.h"
```
Xeeynamo: Let's inline functions that are different, instead of ifdeffing from the stage name

src/st/cutscene_exec.h
ProjectOblivion: So the goal is to never ifdef if there is a difference between overlays, no matter how minor?
Are we still looking at using macros if values are different, but code itself is the same?

Xeeynamo: Yes.

Values are a bit of a grey area. I'd say it's fine for now and evaluate it case by case. But I have a few examples where they're causing problems too. They're easier to untangle than ifdeffed blocks of code, so I'm okay to green light them for now

sozud:
I merged master since the changes were marked resolved, but they aren't actually present on the branch yet

ProjectOblivion:
That's really weird.  I had to completely delete my local branch, repull from origin, and recommit the changes for it to actually push them to the branch.

PR #2695, opened 2025-06-22
Import `no0` stage data

JoshSchreuder:


config/assets.us.yaml
```diff
@@ -248,7 +248,37 @@ files:
...
-          - [0x90, skip]
+          - [0x90, paldef, palette_def]
+          - [0x150, layers, layers]
+          - [0x5B8, gfx_banks, graphics_banks]
+          #- [0x778, layout, entity_layouts]
+          - [0x778, skip]
+          - [0x48A0, cmpgfx, stage_title_no0, 128, 128, 4]
+          - [0x4CF4, cmpgfx, e_diplocephalus_1, 128, 128, 4, 0x16EBC]
+          - [0x5954, cmpgfx, e_diplocephalus_2, 128, 128, 4, 0x16EBC]
+          - [0x6A48, cmpgfx, e_diplocephalus_3, 128, 128, 4, 0x16EBC]
```
JoshSchreuder, line 281: @Xeeynamo the part I'm not clear on - how do you go about matching the last value here (palette), back to the actual palette address

This would help to name them when importing the palettes later on eg. `- [0x16EBC, palette, e_diplocephalus_pal, 16]`

config/assets.us.yaml
```diff
@@ -248,7 +248,37 @@ files:
         vram: 0x80180000
         assets:
           - [0x2C, sprite_banks, sprite_banks]
-          - [0x90, skip]
+          - [0x90, paldef, palette_def]
+          - [0x150, layers, layers]
+          - [0x5B8, gfx_banks, graphics_banks]
+          #- [0x778, layout, entity_layouts]
```
JoshSchreuder: Getting  `unable to extract asset "entity_layouts" in "assets/st/no0": gap between data detected: 0x801839C2 != 0x80183A64` when trying to import this

Xeeynamo: Let's just ignore it for now. Unfortunately the tool is very picky about unused data between arrays

config/assets.us.yaml
Xeeynamo: I am not sure to understand the question. You mean doing something like ` [0x6A48, cmpgfx, e_diplocephalus_3, 128, 128, 4, e_diplocephalus_pal]`?

JoshSchreuder: @Xeeynamo so I have the palette definitions and can import them like
`- [0x16EBC, palette, e_diplocephalus_pal, 16]`

But the piece I am missing is how to know which palette corresponds with each sprite gfx.
The only way I could match `- [0x4CF4, cmpgfx, e_diplocephalus_1, 128, 128, 4, 0x16EBC]` was to go through each palette definition manually and check the output png to see if the colours looked right.

Do you have any tips about how to match the two? I could not find how the `clut` value in `EInit` of the entity corresponds with a paticular palette definition address in memory

Makefile
```diff
@@ -159,7 +159,7 @@ clean: ##@ clean extracted files, assets, and build artifacts
 	git clean -fdx assets/
 	git clean -fdx asm/$(VERSION)/
 	git clean -fdx build/$(VERSION)/
-	git clean -fdx src/**/gen
+	git clean -fdx src/**/gen/
```
JoshSchreuder, line 162: @Xeeynamo I assume this was the intention - previous clean was not cleening up the old `gen` files for me

JoshSchreuder:
Thanks for your help @Xeeynamo - I was able to get the palettes mostly nicely sorted and named.

I still had a few outstanding questions about some which I've posted in Discord, but this should be good to review and hopefully merge now.

PR #2701, opened 2025-06-24
Added more g_Entities addresses to symbols

ProjectOblivion:
I found that pspeu was missing some of the useful g_Entities symbols entries that us had, so I added them to pspeu as well as adding a couple extras to both.


config/symbols.pspeu.txt
```diff
@@ -184,7 +184,7 @@ D_8003C708 = 0x091e1308; // size:0x8
 g_BgLayers = 0x091E1328; // size:0x300
 g_Tilemap = 0x091e1628; // size:0x50
 g_CurrentEntity = 0x091E1678;
-g_Entities = 0x091e1680;
+g_Entities = 0x091E1680; // size:0xBC
```
JoshSchreuder: I may be wrong, but I think I tried this before and splat considers this to be a duplicate range with the `PLAYER_*` symbols below? I don't think it really supports having both at the moment from memory.

ProjectOblivion: It didn't complain when I ran it locally and the tests are good, so I dunno.

config/symbols.pspeu.txt
```diff
@@ -194,7 +194,14 @@ PLAYER_zPriority = 0x091E16A4;
 PLAYER_step = 0x91E16AC;
 PLAYER_step_s = 0x91E16AE;
 D_80073510 = 0x091E17B8;
+g_Entities_32 = 0x091E2E00; // size:0xBC
```
sozud: At some point there was a discussion that we were going to remove these "helper" entries? There were various people objecting to them

ProjectOblivion: I thought the discussion was around the ones that said "symbol+0xdead" kind of thing.  For these, the size hint isn't terribly concerning, but getting the symbols renamed to g_Entities_160 and g_Entities_224 has recently been very helpful.

config/symbols.pspeu.txt
JoshSchreuder: Fair enough, it used to be when extracting splat would give "Duplicate range" warning, maybe that got changed in a newer splat since we've updated a bit recently.

The other thing to check is whether it still continues to pick up stuff like `PLAYER_step` in the output assembly, or if it will now use `g_Entities + 0xFOO`.

Ideally in my mind it should use the named symbol offsets where possible but fall back to `+ offset` in the assembly if one does not exist, but I don't think it was doing that previously

hohle: Splat will print a warning about the range, but it still works. It would be nice if it handled "overlay" symbols, though.

config/symbols.pspeu.txt
hohle: Agreed. Some of the well-known indexes are very helpful.

config/symbols.pspeu.txt
ProjectOblivion: It sounds to me like it would be better to drop the size hint in a general sense, but leave the symbol name.
Are there any of the g_Entities that benefit from having the size hint?

Xeeynamo: If it complains about the range, we could just remove the various `PLAYER_` symbols and have the linter to check for `g_Entities->` and `g_Entities[0]`. Let's do it as part of another PR.

PR #2704, opened 2025-06-24
Add new overlay PSP NP3

bismurphy:
After finishing NO3, I figured it makes sense for me to move onto NP3 and do the same stuff. This adds the new overlay, and I believe it does all the standard stuff for a new one, but if I missed something, please let me know.

I have already opened an Issue to update the steps for a new overlay, so hopefully this can go more smoothly in the future.


bismurphy:
Thanks, will do - i estimate 7 hours from now.

bismurphy:
Wait sorry @sozud what needs to be added to progress.py? From what I can see there's nothing in there about US vs PSPEU so I don't see what I would add for this, given that we already have NP3 for US.

sozud:
Nevermind I keep forgetting that that there's nothing to add in this case

PR #2708, opened 2025-06-25
Imported `dai` `e_stage_name` data

ProjectOblivion:
I realized that I missed importing e_stage_name data.
I also moved pspeu EntityStageNamePopup to e_stage_name.c since it had ovl based ifdefs in it.


config/assets.pspeu.yaml
```diff
@@ -218,6 +218,11 @@ files:
         vram: 0x09237700
         assets:
           - [0x80, skip]
+          #- [0x23A18, cmpgfx, stage_name_en, 128, 128, 4]
+          #- [0x23C38, cmpgfx, stage_name_sp, 128, 128, 4]
+          #- [0x23E88, cmpgfx, stage_name_it, 128, 128, 4]
+          #- [0x24090, cmpgfx, stage_name_ge, 128, 128, 4]
+          #- [0x24360, cmpgfx, stage_name_fr, 128, 128, 4]
```
Xeeynamo:
```suggestion
          - [0x23A18, cmpgfx, stage_name_en, 128, 128, 4]
          - [0x23C38, cmpgfx, stage_name_sp, 128, 128, 4]
          - [0x23E88, cmpgfx, stage_name_it, 128, 128, 4]
          - [0x24090, cmpgfx, stage_name_ge, 128, 128, 4]
          - [0x24360, cmpgfx, stage_name_fr, 128, 128, 4]
```

cmpgfx works on PSP: https://github.com/Xeeynamo/sotn-decomp/blob/master/config/assets.pspeu.yaml#L12

please let's not import giant blobs of copyrighted data in the repo.

ProjectOblivion: First of all, I didn't realize it was any different than the other data we import, my mistake.

Second of all, it doesn't work.  This is the precise issue I was asking about in Discord, https://discord.com/channels/1079389589950705684/1135205782703570994/1387414430664364052, and was told to assume that the asset tool doesn't work for pspeu.

Xeeynamo: I will have a look once I have access to my PC. But `data length (8328) < width*height (16384)` should definitively not happen. 128*128/2 is 8192. I am surprised the tool is complaining by expecting the image to be 16384 long.

ProjectOblivion: It says it right in the message, length < width*height

```
if len(data) < width*height {
		return fmt.Errorf("data length (%d) < width*height (%d)", len(data), width*height)
	}
```

Xeeynamo: I get it. But `data length (8328) < width*height (16384)` hints that you initially chosen `256, 128, 4`, `128, 256, 4` or `128, 128, 8`.

ProjectOblivion: 128*128=16384
I don't fully understand why you're expecting the /2
But the code specifically uses `width*height` for the error.

Xeeynamo: /2 is to compensate the 4bpp format. I will debug this once I can and come back to you.

ProjectOblivion: The obvious answer is that the error message is accurate and I have the size wrong, which I do.  It is not a square, it is a rectangle.  I don't know what the actual dimensions are, though, but the GFXENTRY says `128*128`.
`128*64` is accurate visually, but it doesn't match.  I thought 8328 was an odd size, so it could be that there is trailing data.

ProjectOblivion: And I got it figured out.  Looks like there is some trailing data after fr that the asset tool was picking up, but isn't used.

src/st/dai/e_stage_name.c
```diff
@@ -2,17 +2,407 @@
...
+// sp and it are swapped in relation to maria in order to match the order in
+// GetLang
+static u8* get_lang[] = {
+    stage_name_en, stage_name_fr, stage_name_sp, stage_name_ge, stage_name_it};
+static void func_psp_0923C0C0(void) {
+    stage_name[3] = (u_long*)get_lang[g_UserLanguage - 1];
+    func_91040A0(stage_name);
 }
 
-#include "../entity_stage_name_popup_jp.h"
```
sozud, line 15: Are we not de-duplicating anymore? I know Xeeynamo has said he wants to change to copy-pasting

src/st/entity_stage_name_popup_jp.h
```diff
@@ -23,7 +23,7 @@ void EntityStageNamePopup(Entity* self) {
     s16 var_s3;
     s16 var_s2;
 
-#if defined(STAGE_IS_LIB) || defined(STAGE_IS_DAI)
+#if defined(STAGE_IS_LIB)
```
ProjectOblivion, line 26: @sozud This conditional applies to dai as well.  Since @Xeeynamo said we're moving away from ovl based ifdefs in favor of multiple versions, that is why I moved to code directly to dai/e_stage_name.c

src/st/dai/e_stage_name.c
ProjectOblivion: I'm not quite understanding what you mean.  It seems like you answered your own question.

sozud:
I think while the dedupe thing is being debated we should just merge this rather than hold it in limbo

PR #2710, opened 2025-06-25
Decompile PSP NP3 st_init_np3.c

bismurphy:
Good start - no big surprises compared to the existing overlays.

Only hangup was that there was a fake function identified in the middle of InitEntityIds; after a while of staring at the compiled code I figured out what was going wrong and made it work the normal way.


sozud (approved):

src/st/np3_psp/st_init_np3.c
```diff
@@ -0,0 +1,172 @@
...
+extern s32 E_ID(HAMMER);
+extern s32 E_ID(GURKHA_BODY_PARTS);
+extern s32 E_ID(HAMMER_WEAPON);
+extern s32 E_ID(GURKHA);
+extern s32 E_ID(GURKHA_SWORD);
+extern s32 E_ID(BLADE);
+extern s32 E_ID(BLADE_SWORD);
+
+void InitEntityIds(void) {
+    E_ID(ID_11) = 0x11;
```
Xeeynamo:
```suggestion
    E_ID(ID_11) = E_ID_11;
```
and so on

bismurphy: Thanks, done - on NP3 and also on NO3.

PR #2712, opened 2025-06-26
Ignore `matchings` When Finding Orphaned Symbols

hohle:
I'm not sure if it always worked this way, but `matchings` expands symbols configured in each overlay's symbol tables. Because these files are included in the set of files evaluated by `remove-orphans`, very few unneeded symbols were ever removed. This excludes `matchings` from the set of files used for "live" symbols.

`g_MainGame`, which is never defined in code, is moved to `undefined_syms`.

On the PSP `g_DebugPlayer` has been moved to the system symbol table.

There are several static symbols in `tt_000` on PSP which are not defined but which also are not referenced by unmatched code. These have been given a `used:true` attribute to indicate they are still necessary.


hohle:
Instead of `used:true`, something like `unallocated_bss:true` may be more appropriate here since it describes why these aren't found.

PR #2717, opened 2025-06-29
Import Some Doppleganger Data

hohle:
Partial data import for `BO4` and `RBO5`. There is still asset data that needs to be mapped.

Some preliminary work has been done to share data & code, but those changes will come in a later commit.


sozud (approved):

sozud:
Needs to be linted

hohle:
Had some cruft in my local branch that was preventing `clang-format` from running. Fixed.

PR #2721, opened 2025-07-01
Lint for Global Definitions Declared `extern`

hohle:
Adds a new lint rule to find initialized global variables that are declared with `extern` linkage. Since `extern` values cannot be initialized, this is undefined behavior.

These won't always be `static`, but in this case all of the found cases were.


hohle:
Screenshot 2025-06-30 at 9 09 25 PM

https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Storage-Class-Specifiers

PR #2722, opened 2025-07-01
Decompile Sword Familiar's `func_us_80176F28`

hohle:
I had thought `D_us_80178670` was a struct of some kind, but addressing only worked as an array of arrays. Unfortunately, there are still other functions which require symbols that cut into these arrays.

The tail of BSS has been moved back into a `bss` segment to support the type declarations the decompiled function want as well as all of the "fake" symbols that the remaining ASM functions want.


sozud (approved):

sozud:
Actually this needs to be linted

PR #2725, opened 2025-07-02
Import `NO2` e_owl_knight.c

gamezter:
Depends on https://github.com/Xeeynamo/sotn-decomp/pull/2724


src/st/no2/e_owl_knight.c
```diff
@@ -0,0 +1,1100 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "no2.h"
```
sozud, line 2: Is the plan to dedupe this later? I think this is one where the "I don't want to modify multiple files to change owl knight in my mod" guidance applies

gamezter: Yes. It is partially decompiled in NP3, but it would have been a lot of work to get that version up to date so I opted to do it in a separate PR.

sozud (approved):

PR #2730, opened 2025-07-03
Document Jewel Sword drops

SestrenExsis:
Note that the item IDs used in the table for MAD are different. They appear to be off by 4 entries. I don't know of a way to verify the Jewel Sword behavior in MAD since there are no enemies in the stage, but I believe the new item IDs correspond to Lapis Lazuli, Ring of Ares, Gold Ring, and Silver Ring.

I also removed an unused variable in `HitDetection` that I noticed while renaming `randCompare`.


include/game.h
```diff
@@ -974,6 +974,18 @@ typedef enum EquipKind {
     NUM_EQUIP_KINDS,
 } EquipKind;
 
+// Note that these item IDs are not the same as the item IDs in the BodyItems
+// enum
+typedef enum {
+    DROPPED_ITEM_ZIRCON = 360,     // 0x0168
```
ProjectOblivion: I'm not sure it really matters, but this name is a little lengthy.  Would DROPPED_XXX be enough?

src/st/collision.h
```diff
@@ -198,7 +229,7 @@ void HitDetection(void) {
     s32* scratchpad_2;
     Entity* iterEnt2;
     s32* scratchpad_1;
-    u16* randCompare;
+    u16* dropThresholdAndItem;
```
ProjectOblivion: Perhaps just `dropTable`?
That would maintain the naming association with `g_jewelSwordDropTable` and is a little less clunky.

src/st/mad/collision.c
```diff
@@ -70,9 +70,19 @@ static u16 g_testCollPrizeTable[] = {
...
+
+static u16 g_jewelSwordDropTable[] = {
+    // clang-format off
+    0x0C00, DROPPED_ITEM_GARNET,
+    0x0F00, DROPPED_ITEM_OPAL,
+    0x0FD0, DROPPED_ITEM_DIAMOND,
+    0x0FF0, 0x016F, // Lapis Lazuli?
+    0x0FF8, 0x0170, // Ring of Ares?
+    0x0FFD, 0x0171, // Gold Ring?
+    0x0FFF, 0x0172, // Silver Ring?
```
ProjectOblivion, line 84: Are you sure that those items aren't the same as the other table, just with different IDs for `mad`?

include/game.h
SestrenExsis: I think `DROP_XXX` reads a little better and is even shorter.

src/st/mad/collision.c
SestrenExsis: I'm not sure how to test that in `MAD` since there are no enemies. However, I'll update the file to reflect the lack of certainty rather than assuming the item IDs are equivalent.

ProjectOblivion: If you look at the mad IDs vs the non-mad IDs, the mad IDs both start and end by an offset of +4.  I'm not certain how you could validate this, but given that the drop rates seem to be the same I think this is the likely scenario.

sozud: It's likely that they removed a few items from the enum after compiling MAD, you should be able to  just do like:
```
#ifdef VERSION_BETA
 DROP_ZIRCON = 0x016C, 
#else
 DROP_ZIRCON = 0x0168, 
#endif
```
The SFX IDs in MAD are also shifted vs. the retail release

SestrenExsis: I find that very plausible, but is it confirmed somehow that that is the case? I'd rather leave it as question marks unless it is somehow confirmed what item those IDs equate to.

PR #2731, opened 2025-07-03
Code-share multiple `no2_psp` files

gamezter:


gamezter:
Did not realize @ProjectOblivion was working on this, I'll try and fix the conflicts.

ProjectOblivion:
I'm not working on it, I just noticed that a lot was missing from your initial PR and added it in.

src/st/no2_psp/e_init.c
```diff
@@ -0,0 +1,167 @@
...
+    /* 0x3A */ func_pspeu_092529B0,
+    /* 0x3B */ func_pspeu_0924F990,
+    /* 0x3C */ func_pspeu_09250ED0,
+    /* 0x3D */ func_pspeu_0924AB98,
+    /* 0x3E */ func_pspeu_0924BE50,
+    /* 0x3F */ func_pspeu_0924BF68,
+    /* 0x40 */ func_pspeu_092571F8,
+};
+
+EInit D_pspeu_09266618 = {ANIMSET_DRA(0x03), 0x00, 0x00, 0x000, 0x000};
```
ProjectOblivion: You'll see it when you deconflict, but there is a new enum for palettes.

src/st/no2_psp/st_init_no2.c
ProjectOblivion, line 1: This should probably be named `st_init_psp.c`.  Since it will be in the no2 folder, naming it `_no2` doesn't add anything helpful whereas `_psp` allows it to be put in the no2/ folder and still be identified as the psp st_init.

src/st/no2_psp/e_init.c
```diff
@@ -0,0 +1,167 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "../no2/no2.h"
+
+void func_pspeu_09256D08(Entity* self);
```
ProjectOblivion: If you map this name to the us name as I did with DAI it makes segment matching really easy, as well as allowing immediate merging of e_init.c

src/st/no2_psp/st_init_no2.c
```diff
@@ -0,0 +1,120 @@
...
+#include "../no2/no2.h"
+
+#include "../pfn_entity_update.h"
+
+extern LayoutEntity* D_8D2E1B4;
+extern LayoutEntity* D_8D2E288;
+
+extern Overlay OVL_EXPORT(Overlay);
+
+s32 E_ID(ID_11);
```
ProjectOblivion, line 11: I opted for E_UNK_11 (for example) instead of E_ID_11.  It better communicates that we don't know what it is for yet.

src/st/no2_psp/e_init.c
```diff
@@ -65,74 +59,106 @@ void func_pspeu_092529B0(Entity* self);
...
-    func_pspeu_09254520,
-    func_pspeu_092515D0,
-    func_pspeu_092529B0,
-    func_pspeu_0924F990,
-    func_pspeu_09250ED0,
-    func_pspeu_0924AB98,
-    func_us_801D6FC8,
-    func_pspeu_0924BF68,
-    func_801A8328,
+    /* 0x00 */ EntityUnkBreakable,
```
ProjectOblivion: We are no longer mapping entity IDs in this way, the function name is instead added as a comment in no2.h.

ProjectOblivion:
Were the symbol removals just because they are no longer needed, or are there other reasons?

gamezter:
`make format` will remove useless symbols

ProjectOblivion:
Yeah, I was just asking if it was because they were unused or if you removed them intentionally for some reason.  Like if you found that there was something erroneous.  I'm really only concerned about the ones I added, just making sure I didn't add something incorrectly.

PR #2734, opened 2025-07-06
Lots of codeshares for NP3

bismurphy:
Many files merged in, but very minimal changes to C code. Mostly just symbols/splat file changes to use existing code.

Some flattening on e_particles.h.

And a tiny bit of data pulled in for PSX NP3.

Next function in the sequence is stuff for Gurkha and Hammer, and that's currently a bit of a mess in `src/np3` so that will be a separate PR since I expect it to take a fair amount of modification.


sozud (approved):

bismurphy:
Managed to find where hammer/gurkha/blade ends. That's its own file, and now all of NP3 PSP is decompiled except those entities. So next PR will just handle cleaning those up in both versions. Then I'll do a final pass to pull in data.

PR #2740, opened 2025-07-12
NP3 PSP Blade enemy

bismurphy:
Mostly just taking US stuff and porting it over with tweaks to be compatible with PSP, but also pulling in data, and cleaning up some struct stuff.

Still lots to learn here, but once we have the code cleaned up the research will become easier.

Prepping for next PR which should be Gurkha.


src/st/np3/blade.c
```diff
@@ -102,31 +117,179 @@ static s32 func_801D0B78(Entity* unused) {
...
-        (((self + 13)->ext.GH_Props.unk88 != 0) ||
-         ((self + 10)->ext.GH_Props.unk88 != 0))) {
-        PlaySfxPositional(0x750);
-        func_801CE1E8(0x18);
+    if ((self->flags & FLAG_DEAD) && (self->step < 0x18)) {
+        ent_s0 = self + 13;
+        ent_s4 = self + 10;
+
+        if (ent_s0->ext.GH_Props.unk88 || ent_s4->ext.GH_Props.unk88) {
+            PlaySfxPositional(0x750);
```
sozud, line 279: sfx enum value?

sozud (approved):

PR #2742, opened 2025-07-13
Decompiled st/no2 EntityFrozenShadeCrystal

TomoshibiAkira:
First PR here!
The asm has matched, but I'm sure about the name. It's different from the function in ``st/no4``.

### DIFF
--- a/src/st/no2/unk_33A94.c
+++ b/src/st/no2/unk_33A94.c
@@ -7,7 +7,14 @@ INCLUDE_ASM("st/no2/nonmatchings/unk_33A94", func_us_801B3D8C);
 
 INCLUDE_ASM("st/no2/nonmatchings/unk_33A94", func_us_801B3F30);
 
-INCLUDE_ASM("st/no2/nonmatchings/unk_33A94", EntityFrozenShadeCrystal);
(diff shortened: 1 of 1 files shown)


src/st/no2/unk_33A94.c
```diff
@@ -7,7 +7,14 @@ INCLUDE_ASM("st/no2/nonmatchings/unk_33A94", func_us_801B3D8C);
 
 INCLUDE_ASM("st/no2/nonmatchings/unk_33A94", func_us_801B3F30);
 
-INCLUDE_ASM("st/no2/nonmatchings/unk_33A94", EntityFrozenShadeCrystal);
+void EntityFrozenShadeCrystal(Entity* self) {
+    if (!self->step) {
+        InitializeEntity(g_EInitCommon);
+        self->animSet = -0x7FFE;
```
sozud: decomp.me is down so I can't check but this is probalby ANIMSET_OVL(1) or similar

TomoshibiAkira: Yes it is! Would use the macro instead.
Also decomp.me is back: https://decomp.me/scratch/vOrQ8

sozud (approved):
Thanks!

PR #2743, opened 2025-07-14
Decompile `lib` `LIB_EntityCutscene` for PSP

JoshSchreuder:
Another big func falls. Thanks to @bismurphy's work on NO3 PSP here which finally unlocked this one for me.


src/st/lib_psp/st_init_lib.c
```diff
@@ -152,19 +152,19 @@ void InitEntityIds(void) {
...
         0, (u8*)cutscene_data_en, (u8*)cutscene_data_fr, (u8*)cutscene_data_sp,
         (u8*)cutscene_data_ge, (u8*)cutscene_data_it);
 
-    D_psp_092A54E8 = GetLangAt(
+    OVL_EXPORT(cutscene_data_copy) = GetLangAt(
         0, (u8*)cutscene_data_en, (u8*)cutscene_data_fr, (u8*)cutscene_data_sp,
         (u8*)cutscene_data_ge, (u8*)cutscene_data_it);
 
-    D_psp_092A54E0 = GetLangAt(
+    OVL_EXPORT(cutscene_data_offset_four) = GetLangAt(
```
JoshSchreuder, line 167: It looks like these BSS pointers were used to set the 0th, 4th, 8th bit of the cutscene script, eg. in `e_shop.c` and once you've already met the librarian.

I don't really have a good idea for how to name these but open to suggestions.

src/st/no3_psp/cutscene.c
```diff
@@ -124,17 +124,7 @@ void CutsceneUnk1(void) {
...
-void SetCutsceneEnd(u8* ptr) {
-#if !defined(VERSION_PSP)
-    g_Dialogue.scriptEnd = ptr + 0x100000;
-#else
-    g_Dialogue.scriptEnd = ptr;
-#endif
-    g_Dialogue.timer = 0;
-    g_Dialogue.unk3C = 1;
-}
+#include "../set_cutscene_end.h"
```
JoshSchreuder, line 127: Since I hit the same issue as bismurphy here, and unk1 also needed to have static removed, I have done this in the header, and moved this back to being shared. It has no impact on other code so it's likely these were never static.

PR #2747, opened 2025-07-15
PSP NP3 Gurkha

bismurphy:
Planning to do some cleanup on both Gurkha and Blade, after finishing the full trio, hoping to get all the data pulled in and managed before doing that, just in case there are surprises.


sozud:
Rebase

src/st/no2/no2.h
```diff
@@ -131,3 +131,13 @@ typedef struct {
     s16 unk0;
     s16* unk4;
 } unkStr_801CDD80;
+
+// note on member types: length and zOffset definitely look signed
+// in the data, but psp uses lhu instead of lh.
+typedef struct {
```
sozud, line 137: I'm thinking this should be in giantbros.h or whatever instead of including it in the stage header 2x?

sozud (approved):
The header can be adjusted as a follow-up

PR #2750, opened 2025-07-15
Code-Share e_background_house

gamezter:
This is the 3D background houses seen in this video : https://youtu.be/6g6t9klC4Eg?t=257


src/st/e_background_house.h
```diff
@@ -0,0 +1,361 @@
...
+        }
+        iPtr = sideIndices;
+        for (i = 0; i < 2; i++) {
+            prim = DrawSides(prim, iPtr, (u16*)modelData);
+            iPtr += 4;
+        }
+        modelData += 3;
+        iPtr = roofIndices;
+        for (i = 0; i < 2; i++) {
+            prim = DrawRoof(prim, iPtr, (u16*)modelData);
```
gamezter, line 352: These don't actually draw the primitives, but they set up the primitives clip space vertices, uvs, colors, etc. I wouldn't mind renaming these if someone can come up with a better name.

sozud (approved):

PR #2751, opened 2025-07-16
Import `no2` e_sealed_door

gamezter:
Unfortunately `no2` `EntitySealedDoor` is different enough from e_sealed_door.h that it can't really be shared


src/st/no2/33D8C.c
```diff
@@ -50,18 +50,18 @@ void func_us_801B3D8C(Entity* self) {
 
 INCLUDE_ASM("st/no2/nonmatchings/33D8C", func_us_801B3F30);
 
-void EntityFrozenShadeCrystal(Entity* self) {
-    if (!self->step) {
+void func_us_801B4148(Entity* self) {
+    if (self->step == 0) {
```
gamezter, line 54: This was wrongly named

sozud (approved):

PR #2752, opened 2025-07-16
Decompiled func_us_801B3F30 / func_pspeu_09257810

TomoshibiAkira:
PSX: https://decomp.me/scratch/0Tvva
PSP: https://decomp.me/scratch/iedH8

It looks like a damage animation handling routine TBH.
I'm not sure about what ``colorLo`` is about. Is there a more detailed explanation (or code) about the contents in the CLUT?

Also, where should I put the the data stored in ``.s``?
Hmm... Binary is not matching even asm is matched?


sozud:
This needs to be linted and the psp build isn't matching

src/st/no2/33D8C.c
```diff
@@ -48,7 +48,71 @@ void func_us_801B3D8C(Entity* self) {
...
+        self->animSet = ANIMSET_OVL(2);
+        self->animCurFrame = 3;
+        self->ext.et_801B3F30.unk7C = 2;
+        self->ext.et_801B3F30.unk80 = 0x10;
+        self->zPriority = 0x80;
+        self->drawMode = 0x70;
+        break;
+    case 1:
+        if (g_Tilemap.scrollY.i.hi >= 0x304) {
+            deltaPosXHi = self->posX.i.hi - PLAYER_posX_i_hi;
```
JoshSchreuder: `PLAYER.posX.i.hi`

src/st/no2/33D8C.c
```diff
@@ -48,7 +48,71 @@ void func_us_801B3D8C(Entity* self) {
     }
 }
 
-INCLUDE_ASM("st/no2/nonmatchings/33D8C", func_us_801B3F30);
+/* st/no2/data/B7C.data.s */
+extern u16 D_us_80180B7C[];
+extern u32 D_us_80180B8C[];
+
+extern s16 PLAYER_posX_i_hi;
```
JoshSchreuder: This can be removed, see below

include/entity.h
```diff
@@ -3275,6 +3275,12 @@ typedef struct {
     /* 0x80 */ struct Entity* unk80;
 } ET_801CEB08;
 
+typedef struct {
+    /* 0x7C */ u16 unk7C;
+    /* 0x7E */ u16 unk7E;
+    /* 0x80 */ u16 unk80;
+} ET_801B3F30;
```
JoshSchreuder, line 3282: Believe this PR isn't matching because these are `s16` in your scratch but you made them `u16` here.
image

src/st/no2/33D8C.c
```diff
@@ -49,8 +49,8 @@ void func_us_801B3D8C(Entity* self) {
 }
 
 /* st/no2/data/B7C.data.s */
-u16 D_us_80180B7C[] = {0x8044, 0x8048, 0x8049, 0x804A, 0x804B, 0x804C, 0x804D, 0x0000};
-u32 D_us_80180B8C[] = {0x4, 0x8, 0x9, 0xA, 0xB, 0xC, 0xD};
```
JoshSchreuder: We can go through it more in Discord if you like, but if you want to import data, you need to exclude the segment from the splat config `/config/splat.us.stno2.yaml`, otherwise the data effectively gets included in the bin twice which won't match.

You would do something to the effect of
```
      - [0xB7C, .data, 33D8C]
      - [0xBA8, data]
```

Which has the effect of:

image

But we don't require importing data right away and this can be done later on so making these extern is completely fine

include/entity.h
TomoshibiAkira: Thanks! I probably copied the wrong struct lol

src/st/no2/33D8C.c
TomoshibiAkira: Got it! Just saw gametzer's PR too, I'll try it out!

PR #2757, opened 2025-07-18
Decompile `no2` e_secrets

gamezter:


src/st/no2/e_secrets.c
```diff
@@ -0,0 +1,581 @@
...
+    x = (prim->x0 << 0x10) + prim->x1;
+#endif
+    x += LOW(prim->x2);
+    prim->x0 = HIHU(x);
+    prim->x1 = LOHU(x);
+#ifdef VERSION_US
+    y = (prim->y0 << 0x10) + (u16)prim->y1;
+#else
+    y = (prim->y0 << 0x10) + prim->y1;
+#endif
```
gamezter, line 417: This sucks, but it's the only way I found to get both platforms to match

sozud (approved):

PR #2759, opened 2025-07-18
Decompile `no2_psp` e_stage_name

gamezter:


src/st/entity_stage_name_popup_jp.h
```diff
@@ -381,51 +381,18 @@ void EntityStageNamePopup(Entity* self) {
...
-        prim = self->ext.stpopupj.unk84;
-        prim->x1 -= 4;
-        var_s2 = prim->x3 = prim->x1;
-        prim = self->ext.stpopupj.prim;
-#ifdef VERSION_PSP
-        self->ext.stpopupj.unk80 += 4;
-#endif
-        for (i = 0; i < 2; i++) {
+        switch (self->step_s) {
+        case 0:
```
gamezter, line 385: This is actually a switch with a single case. Thanks @JoshSchreuder !

sozud (approved):

config/assets.pspeu.yaml
```diff
@@ -250,6 +250,21 @@ files:
...
+    segments:
+      - start: 0x80
+        vram: 0x09237700
+        assets:
+          - [0x80, skip]
+          - [0x24FF0, cmpgfx, D_pspeu_0925C670, 128, 32, 4]
+          - [0x25220, cmpgfx, D_pspeu_0925C8A0, 128, 32, 4]
+          - [0x25478, cmpgfx, D_pspeu_0925CAF8, 128, 32, 4]
+          - [0x25708, cmpgfx, D_pspeu_0925CD88, 128, 32, 4]
+          - [0x25918, cmpgfx, D_pspeu_0925CF98, 128, 32, 4]
```
gamezter, line 266: For some reason these are 128x32 textures (the left half of the localized stage name graphic), but the GFXBank says they are 128x128. Where is the right half of these graphics ?

gamezter: Ex: 
D_pspeu_0925C8A0
D_pspeu_0925C670
D_pspeu_0925CAF8
D_pspeu_0925CD88
D_pspeu_0925CF98

sozud: I don't really know the answer to this, we can do another pass of fixing to figure out what's going on here

JoshSchreuder: I think maybe it's an issue with the assets tool, if you try increasing the width to 256 for example you can see the rest of the text is actually below this
image

PR #2766, opened 2025-07-22
Codeshare `no2` gurkha

gamezter:
I renamed:
EntityGurkhaSword -> EntityGurkhaWeapon
EntityBladeSword -> EntityBladeWeapon
and their related `EInit`s and `E_ID`s since the next giantbrother entity to be code-shared is hammer, whos secondary entity is `EntityHammerWeapon`. I figured we should keep the naming consistent across all giantbrothers.

I also tried to format the data in `e_blade.h` and the new `e_gurkha.h` to make it more readable and consistent. 
I know @bismurphy is working on related code right now so this PR can wait while we get their OK


sozud (approved):

sozud:
I don't think there's going to be any objections to merging this, it's pretty straightforward.

PR #2770, opened 2025-07-26
Decompile and import no4 PSP stage name

JoshSchreuder:
Mostly this is focused on importing the (existing) stage name code for NO4 PSP.

While I was here I standardised and cleaned up the data imports a bit for NO2, LIB and CHI PSP to match renames here, and a very small codeshare for NO4  `popup`


config/assets.pspeu.yaml
```diff
@@ -259,11 +280,11 @@ files:
...
-          - [0x24FF0, cmpgfx, D_pspeu_0925C670, 128, 32, 4]
-          - [0x25220, cmpgfx, D_pspeu_0925C8A0, 128, 32, 4]
-          - [0x25478, cmpgfx, D_pspeu_0925CAF8, 128, 32, 4]
-          - [0x25708, cmpgfx, D_pspeu_0925CD88, 128, 32, 4]
-          - [0x25918, cmpgfx, D_pspeu_0925CF98, 128, 32, 4]
+          - [0x24FF0, cmpgfx, stage_name_en, 128, 64, 4]
+          - [0x25220, cmpgfx, stage_name_it, 128, 64, 4]
+          - [0x25478, cmpgfx, stage_name_sp, 128, 64, 4]
+          - [0x25708, cmpgfx, stage_name_ge, 128, 64, 4]
+          - [0x25918, cmpgfx, stage_name_fr, 128, 64, 4]
```
JoshSchreuder, line 287: @gamezter I think this is still not quite right but it is slightly closer, you can see the full text at least. 

Still think maybe this needs tweaks in the assets tool, I tried playing around with 128x128 2bpp and some other ratios but couldn't quite work it out.

src/st/entity_stage_name_popup_jp.h
```diff
@@ -326,7 +331,8 @@ void EntityStageNamePopup(Entity* self) {
             }
             break;
         case 3:
-            PrimDecreaseBrightness(self->ext.stpopupj.unk88, 4);
+            prim = self->ext.stpopupj.unk88;
+            PrimDecreaseBrightness(prim, 4);
```
JoshSchreuder, line 335: This change was required on NO4 but does not break existing PSP usages of the header, so I suspect it is probably the canonical implementation.

sozud (approved):

PR #2775, opened 2025-07-30
Codeshare `no2` hammer

gamezter:
Once hammer in np3 is ready, this file can be merged.

And with that all functions are decompiled in `no2` / `no2_psp` ! All that is left are `no2_psp` assets that currently can't be imported because of issues with sotn-assets.


src/st/e_gurkha.h
```diff
@@ -165,43 +165,41 @@ static s32 func_801CF7A0(Entity* ent) {
...
+        if (dx < 64) {
             g_CurrentEntity->ext.GH_Props.unk8C = 0;
         }
     }
 
     if (step != g_CurrentEntity->step) {
-        do {
-            func_801CE1E8(step);
-        } while (0); // no idea why, found by permuter
+        func_801CE1E8(step);
```
gamezter, line 199: IDK why but making this function look like its counterpart in `hammer` made this unnecesary

PR #2776, opened 2025-07-30
General code clean up

gamezter:


src/dra/5087C.c
```diff
@@ -1830,7 +1830,7 @@ void RunMainEngine(void) {
             }
             D_801375A4 = D_8013759C - PLAYER.posX.val;
             D_801375A8 = D_801375A0 - PLAYER.posY.val;
-            for (i = 0, ent = &g_Entities[0]; i < LEN(g_Entities); i++, ent++) {
+            for (i = 0, ent = g_Entities; i < TOTAL_ENTITY_COUNT; i++, ent++) {
```
JoshSchreuder, line 1833: PLAYER?

gamezter: I opted to leave it as `g_Entities` since the intention is to get a pointer to the first element and iterate over all of the entities.

gamezter: Maybe `&g_Entities[0]` would be clearer, but we use `g_Entities` elsewhere in similar code.

src/boss/bo4/doors.c
```diff
@@ -122,8 +122,7 @@ void EntityUnkId18(Entity* self) {
         offsetX = 0x100 - g_Tilemap.scrollX.i.hi;
         offsetY = 0x80 - g_Tilemap.scrollY.i.hi;
         for (i = 0; i < 2; i++) {
-            dop =
-                AllocEntity(g_Entities + 224, g_Entities + TOTAL_ENTITY_COUNT);
+            dop = AllocEntity(&g_Entities[224], &g_Entities[256]);
```
sozud, line 125: If the point is making the number of entities modifiable it seems like we want to keep TOTAL_ENTITY_COUNT? same for several other spots in this PR

gamezter: This line is so common in the project, I figured it would be simpler to eventually do that modification if they all look the same

PR #2778, opened 2025-07-31
Decompile `dra_psp` DrawConsumableCount

gamezter:
Thanks to @JoshSchreuder for help getting this function matching


src/dra_psp/menu.c
```diff
@@ -1834,7 +1834,41 @@ void MenuDrawLine(s32 x0, s32 y0, s32 x1, s32 y1, s32 isColorStatic) {
...
+        *ptr++ = (displayCount / 10) + CH('0');
+        *ptr++ = (displayCount % 10) + CH('0');
+    }
+    // Finish off with a ) and string terminator
+    *ptr++ = CH(')');
+#ifdef VERSION_PSP
+    *((s8*)&(*ptr++)) = 0xFF;
+#else
+    *ptr++ = 0xFF;
+#endif
```
gamezter, line 1867: This sucks, on PSP we store a `-1`, on PSX we store a `0xFF`

JoshSchreuder: I think this can be changed with compiler flags for unsigned char, but I'm not sure where we decided that originally / what other impact it might have on other PSP

JoshSchreuder: `-char signed`
https://decomp.me/scratch/QTBQm

It looks like the default for MWCC if you don't specify one is signed? So maybe we just had it wrong given how many of the compiler flags seem to be left default.

gamezter: `-char signed` gives me a ton of mismatches, we probably need to leave them unsigned : 
image

sozud: We'd have to look into it deeper but it probably makes sense to make a github issue for this change

PR #2779, opened 2025-07-31
Decompile `main` FntOpen

gamezter:
Mostly based off @Xeeynamo's scratch : https://decomp.me/scratch/6MCvv

I couldnt find a way around using `(*&Font[D_8002B810].tile)` when accessing the tile

### DIFF
--- a/src/main/psxsdk/libgpu/font.c
+++ b/src/main/psxsdk/libgpu/font.c
@@ -1,8 +1,6 @@
 #include "common.h"
 #include <libgpu.h>
 
-const char a0123456789abcd[] = "0123456789ABCDEF";
-
 typedef struct {
     /* 0x00 */ TILE tile;
     /* 0x10 */ DR_MODE draw_mode;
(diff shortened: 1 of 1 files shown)


src/main/psxsdk/libgpu/font.c
```diff
@@ -128,4 +170,128 @@ u_long* FntFlush(s32 id) {
...
+    if (id < 0 || id >= D_8002B810) {
+        f = (char*)id;
+        id = D_8002B814;
+        if (Font[id].buffer == NULL) {
+            return -1;
+        }
+    } else {
+        args = &id + 2;
+        f = (char*)*(&id + 1);
+    }
```
gamezter, line 203: This is weird, the top side of the branch acts as if the signature is `s32 FntPrint(const char* fmt, ...)` and the bottom side acts as if it looks like `s32 FntPrint(s32 id, const char* fmt, ...)`.

src/main/psxsdk/libgpu/font.c
```diff
@@ -128,4 +170,128 @@ u_long* FntFlush(s32 id) {
...
+                sign = '-';
+            }
+            len = 0;
+            do {
+                do {
+                    *--bufPtr = (num % 10) + '0';
+                    num /= 10;
+                    len++;
+                } while (len == 0);
+            } while (num != 0);
```
gamezter, line 248: This looks wrong, any ideas ?

src/main/psxsdk/libgpu/font.c
Xeeynamo: This looks a very weird optimization, but it makes sense.

The [official signature](file:///run/media/xeeynamo/wd_black/SDK/sony_ps1/psyq_sdk/DOCS/LIBREF46.PDF) is `long FntPrint(long id, char *format)`, but according to our code base and various samples from the SDK, it is possible to omit the `ID` parameter and just pass the string.

src/main/psxsdk/libgpu/font.c
Xeeynamo: I tried a few things, but I was not able to remove the unnecessary `while (len == 0)`. It's probably real and an artifact of a previous code iteration. Let's just keep it as it is.

PR #2783, opened 2025-08-04
Dedupe `cat` funcs

JoshSchreuder:
Deduped a bunch across PSX + PSP


src/st/st_update.h
```diff
@@ -1,9 +1,11 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include <game.h>
 
+#ifndef STAGE_IS_CAT
 static u16 unused[] = {
     0x0000, 0x0000, 0x0000, 0x0000, 0x0001, 0x0000, 0x0001, 0x0000,
     0x0001, 0x0001, 0x0001, 0x0001, 0x0001, 0x0001, 0x0001, 0x0101};
+#endif
```
JoshSchreuder, line 8: Only one notable difference for CAT, I think this is ok.
Possible this belongs to a prior segment on other overlays

PR #2784, opened 2025-08-06
More `cat` dedupes

JoshSchreuder:
Mostly standard, a couple of functions in `e_misc` on PSP side which had been deadstripped out of earlier overlays but are used in CAT, so needed minor updates.


src/st/e_misc.h
```diff
@@ -317,31 +317,34 @@ bool UnkCollisionFunc5(s16* pointXY) {
 }
 
 #if !defined(STAGE_IS_NO0)
-u8 UnkCollisionFunc4(s32 arg0) {
+u8 UnkCollisionFunc4(u8 arg0) {
```
JoshSchreuder, line 320: This had not been used on PSP up til now so needed tweaks

src/st/e_misc.h
```diff
@@ -616,8 +624,8 @@ void MakeEntityFromId(u16 entityId, Entity* src, Entity* dst) {
 }
 
 void MakeExplosions(void) {
-    s32 temp_s3;
-    s8 temp_s4;
+    u8 temp_s4;
+    s16 temp_s3;
```
JoshSchreuder, line 628: This had not been used on PSP up til now so needed tweaks

sozud (approved):

PR #2786, opened 2025-08-08
Dedupe `cat` giantbros helpers for PSX

JoshSchreuder:
These appear to go unused in CAT - perhaps Hammer or Gurkha was originally intended to appear in this stage but was removed at some point.

Since it's unused on PSX I am pretty sure these are deadstripped out of PSP as I didn't notice them there.


src/st/cat/cat.h
```diff
@@ -108,4 +108,9 @@ extern EInit g_EInitObtainable;
 extern EInit g_EInitWereskeleton;
 extern EInit D_us_80180A88;
 
+typedef struct {
```
sozud, line 111: Is there a reason to not put this in a shared header?

JoshSchreuder: I had a similar thought but it's used in the helpers and also in Hammer/Gurka files

Might make sense moving elsewhere like entity.h but I wasn't sure where best to put it

sozud: giantbro_defs.h or similar?

sozud: This is a small thing and doesn't need to block the PR, we can adjust later

PR #2791, opened 2025-08-12
Decompile `main` _exeque & _addque2

gamezter:
`_exeque`: https://decomp.me/scratch/ij8G6
`_addque2`: https://decomp.me/scratch/iSyoQ

### DIFF
--- a/src/main/psxsdk/libgpu/sys.c
+++ b/src/main/psxsdk/libgpu/sys.c
@@ -21,10 +21,13 @@ typedef struct {
 } gpu;                                          // size = 0x40
 
 struct QueueItem {
-    s32 unk0;
-    s32 unk[0x4A / 4];
+    void (*unk0)(s32*, s32);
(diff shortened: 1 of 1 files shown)


src/main/psxsdk/libgpu/sys.c
```diff
@@ -789,23 +792,135 @@ s32 _param(s32 arg0) {
...
+
+    DMACallback(2, _exeque);
+    ret = 1;
+    goto temp0;
+
+temp2:
+
+    SetIntrMask(intrMask);
+    return ret;
+}
```
gamezter, line 911: These `goto`s aren't ideal. Hopefully we can figure this out later using different versions of `_exeque`

sozud: `goto` the end of a function for cleanup is a typical C pattern so at least one of these is probably real

PR #2792, opened 2025-08-13
Better Build Dependencies for `us` and `hd`

hohle:
Adds options to `cpp` to emit dependency files which contain all `#include` dependencies. This allows ninja to track changes across all headers, recursively, included by a file and rebuild all objects which rely on those files. This was originally mentioned by @sozud in #2557

This only affects `us` and `hd` now. `mwccpsp` supports creating dependency files, but the way that `mwccgap` works causes them to be discarded. Once that's fixed PSP should be able to build based on dynamic dependencies as well.

In practice this means that changing a shared entity like `src/st/e_red_door.h` will cause all `e_red_door.c` that include it to be rebuilt. It also means explicit dependencies on high level headers an be removed and only files that depend on less used headers like `sfx.h` require rebuilding when those files change. Some of the latter work will require additional refactoring to minimize unnecessary headers.

One limitation here is that `.include`s for asm files are not handled by `cpp`. There is no change in behavior from before, but worth pointing out. When `splat` regenerates these files the overlay's objects are already recompiled.


hohle:
Before:
```bash
$> make clean ; make
⋮
[1949/1949] check config/check.us.sha
⋮
$> touch src/st/e_flea_man.h
$> make
⋮
[1/1] check config/check.us.sha
⋮
$> touch include/stage.h
$> make
⋮
[1/1] check config/check.us.sha
⋮
$> touch include/game.h
$> make
⋮
[1324/1324] check config/check.us.sha
⋮
```

Only the hardcoded headers trigger rebuilds, and when they rebuild, they build more than necessary.

After:
```bash
$> make clean ; make
⋮
[1949/1949] check config/check.us.sha
⋮
$> touch src/st/e_flea_man.h
$> make
⋮
[7/7] check config/check.us.sha
⋮
$> touch include/stage.h
$> make
⋮ 
[788/788] check config/check.us.sha
⋮
$> touch include/game.h
$> make
⋮
[1113/1113] check config/check.us.sha
⋮
```

Builds rebuild the things that need to be rebuilt and not the things that don't.

PR #2793, opened 2025-08-14
Decompile `main` sprintf

gamezter:
https://decomp.me/scratch/ML8eW

Thanks to @roeming for help fixing mismatches

Unfortunately this function has a ton of `do {} while (0)`. I suspect some of them are real if I compare it to glibc printf code I found online (ex: https://codebrowser.dev/glibc/glibc/stdio-common/vfprintf-internal.c.html#89)

Edit: Thanks to @hohle for the idea to use the macros from `stdarg.h`


src/main/psxsdk/libc/sprintf.c
```diff
@@ -1,6 +1,304 @@
...
+                    info.precision = (info.precision * 10) + (ch - '0');
+                    ch = *++f;
+                }
+            }
+            if (info.precision >= 0) {
+                info.usePrecision = true;
+            }
+        }
+
+        bufPtr = &args - 4;
```
gamezter: bufPtr actually points to `buf[0x200]` (1 after the end of buf) but I had to do this to force `args` to be a stack variable.

src/main/psxsdk/libc/sprintf.c
```diff
@@ -1,6 +1,304 @@
...
+
+    va_start(args, f);
+    ch = *f;
+    written = 0;
+    for (; ch = *f, ch != 0; ++f) {
+        if (ch != '%') {
+            out[written++] = ch;
+            continue;
+        }
+        info = D_8002D3A8;
```
gamezter, line 39: I'd love this line to be `info = (printf_info){0};` but that forces `D_8002D3A8` into RO_DATA.

src/main/psxsdk/libc/sprintf.c
hohle: I added some comments to this in my scratch. It’s one word before the end of the buf array so that things like `’n’` work.

src/main/psxsdk/libc/sprintf.c
hohle: I have this in the scratch as well. It’s a non-const global initialized to `{0}` and then assigned.

PR #2794, opened 2025-08-14
Decompile `main` CdControl functions

gamezter:
I was stuck on `CdControl` until I found this scratch for a different game : https://decomp.me/scratch/eDL7i

After doing the same inline function trick for all 3 CdControl functions, I figured out that we could just reuse the same inline function and pass the slight differences through arguments.

CdControl : https://decomp.me/scratch/SRLVh
CdControlF : https://decomp.me/scratch/Frkos
CdControlB: https://decomp.me/scratch/9W6Gu


Mc-muffin:
As extra info the inline is real, it was called `cd_cw` and can be seen in earlier SDK 3.0 https://decomp.me/scratch/gHiMe

PR #2795, opened 2025-08-16
Decompile `main` CD_cw

gamezter:
All credit goes to @devperson7 : https://decomp.me/scratch/wQ7Dv
I just changed small things to make it fit this project.

Something neat is that `callback` was already decompiled and put at the end of the file, but making it static inline and placing it above `CD_cw` does not cause a mismatch ?? I figure the compiler is generating a non-inline version at the end of the file since it is passed as a function pointer in other functions. Maybe this is happening elsewhere in the project ?

### DIFF
--- a/src/main/psxsdk/libcd/bios.c
+++ b/src/main/psxsdk/libcd/bios.c
@@ -2,28 +2,155 @@
 #include "../libspu/libspu_internal.h"
 #include <common.h>
 
+typedef struct {
+    unsigned char sync;  // sync state
+    unsigned char ready; // ready state
+    unsigned char c;
+} CD_flush_struct;
+
+typedef struct Alarm_t {
+    int unk0;
+    int unk4;
+    char* unk8;
+} Alarm_t;
+
+extern volatile unsigned char* D_80032D68;
+extern volatile unsigned char* D_80032D6C;
+extern volatile unsigned char* D_80032D70;
+extern volatile unsigned char* D_80032D74;
+extern volatile int* D_80032D78;
+extern volatile CD_flush_struct D_80032D80;
+extern union SpuUnion* D_80032D7C;
+// TODO: CD_status is a word here, but a byte in sys.c
+extern int CD_status;
+extern int CD_status1;
+extern volatile unsigned char D_80032D84;
+extern int* D_80032D9C;
+extern int* D_80032DA0;
+extern int* D_80032DA4;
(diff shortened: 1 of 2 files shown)


src/main/psxsdk/libcd/bios.c
```diff
@@ -2,29 +2,156 @@
...
+        if (((Alarm_t*)&Alarm)->unk0 < VSync(-1) ||
+            ((Alarm_t*)&Alarm)->unk4++ > 0x3C0000) {
+            puts(&aCdTimeout);
+            printf(D_800106B4, ((Alarm_t*)&Alarm)->unk8, D_80032AC8[CD_com],
+                   D_80032B48[D_80032D80.sync], D_80032B48[D_80032D80.ready]);
+            CD_flush();
+            var_v0_2 = -1;
+        } else {
+            var_v0_2 = 0;
+        }
```
gamezter: This looks like an inline function that returns -1 / 0. Also I know the string `D_800106B4` is reused multiple times in other non-decompiled functions. We could probably pull it out later when those functions are decompiled

Mc-muffin: You are right on the money, see this same function from SDK 4.0 with the inlines ported from SDK 3.0 https://decomp.me/scratch/dlFwm

src/main/psxsdk/libcd/bios.c
```diff
@@ -2,29 +2,164 @@
...
+    if (((Alarm_t*)&Alarm)->unk0 < VSync(-1) ||
+        ((Alarm_t*)&Alarm)->unk4++ > 0x3C0000) {
+        puts(&aCdTimeout);
+        printf(D_800106B4, ((Alarm_t*)&Alarm)->unk8, D_80032AC8[CD_com],
+               D_80032B48[D_80032D80.sync], D_80032B48[D_80032D80.ready]);
+        CD_flush();
+        return -1;
+    }
+    return 0;
+}
```
gamezter, line 107: These look similar to get_alarm / set_alarm in libgpu : https://github.com/Xeeynamo/sotn-decomp/blob/e8e2ffd9bbfebb658af3385fa3e1a281b90cb63c/src/main/psxsdk/libgpu/sys.c#L985

devperson7:
Just want to add @Mc-muffin helped me with this function/was the one to match it in the end. I did spend many hours on it, but could not have matched it without help.

PR #2812, opened 2025-08-20
Decompiled, merged, and imported dai/e_priest

ProjectOblivion:
EntityPriest
PSX: https://decomp.me/scratch/5WDKF
PSP: https://decomp.me/scratch/W2iyf

EntityPriestBlades
PSX: https://decomp.me/scratch/wIkeG
PSP: https://decomp.me/scratch/5hPEW

EntityPriestBladeDebris
PSX: https://decomp.me/scratch/XMLyD
PSP: https://decomp.me/scratch/INjY9


src/st/dai/e_priest.c
```diff
@@ -2,10 +2,434 @@
...
+#ifdef VERSION_PSP
+        self->ext.priest.unk88 = 0;
+        if (self->params & 1) {
+            func_892A620(0, 1);
+        }
+        break;
+    case 1:
+        self->ext.priest.unk88++;
+        if ((self->ext.priest.unk88 == 1) && (self->params & 1)) {
+            g_api.PlaySfx(MU_CURSE_ZONE_LOOP_POINT);
```
sozud: The SFX ids are different on psp so this is not the right track

src/st/dai/e_priest.c
```diff
@@ -2,10 +2,434 @@
...
+                if (params == 4) {
+                    entity->params = 1;
+                }
+            }
+        }
+        g_api.PlaySfx(SFX_EXPLODE_B);
+        break;
+    case 1:
+        params = self->params;
+        self->posX.val += 0xFFFD8000;
```
sozud: FIX

src/st/dai/e_priest.c
```diff
@@ -2,10 +2,434 @@
...
+            SetStep(4);
+        }
+        break;
+    case 4:
+        if (self->ext.priest.unk80) {
+            self->ext.priest.unk80--;
+            return;
+        }
+        params = self->params;
+        self->posX.val += 0xFFFC8000;
```
sozud: FIX

src/st/dai/e_priest.c
ProjectOblivion: Even though it is the same ID for us as well?
Nevermind, I got mixed up on where this was.

ProjectOblivion: The enum says it applies to us and psp.

sozud: Curse Zone is this track, does the priest cause this to play on PSP? https://www.youtube.com/watch?v=B8TR_ZsDdBc

We have existing changes to align on PSP https://github.com/Xeeynamo/sotn-decomp/blob/87d87716af955dc0be577333dfa5543273da261d/include/sfx.h#L163

src/st/dai/e_priest.c
ProjectOblivion: What value are you expecting to be used here with FIX?
I tried both 65533.5 (0xFFFD8000/65536) and -2.4999847412109375 (0x27FFF/65526).  Neither worked, nor made sense to me.

src/st/dai/e_priest.c
ProjectOblivion: So, are you saying that this is incorrect?
```
// us: includes US build
    // jp: includes HD and PSP builds
    /* us:0x301, jp:0x301 */ MU_LOST_PAINTING = 0x301,
    /* us:0x302, jp:0x302 */ MU_LOST_PAINTING_LOOP_POINT,
    /* us:0x303, jp:0x303 */ MU_CURSE_ZONE,
    /* us:0x304, jp:0x304 */ MU_CURSE_ZONE_LOOP_POINT,
```

src/st/dai/e_priest.c
sozud: 65532.5 works fine?

image

sozud: Actually `self->posX.val += FIX(-3.5);` is better

src/st/dai/e_priest.c
ProjectOblivion: FIX(65532.5) works here, but I think there is more at play here and I don't think that's the right thing.  65532.5 is a wildly unusual number compared to the existing codebase and since these two places are doing the same thing, it seems that they should both work using the same method.

src/st/dai/e_priest.c
ProjectOblivion: FIX(65532.5) does not work here.

src/st/dai/e_priest.c
ProjectOblivion: And FIX(-3.5) here.  Looks like these two numbers being so close messed us both up.  Got it now.

src/st/dai/e_priest.c
ProjectOblivion: FIX(-2.5) here.

src/st/dai/e_priest.c
ProjectOblivion: I looked into it and I don't think it plays that music.  I think psp 0x304 is equivalent to us 0x204 and probably the priest's VO.  I took a stab at adjusting the enum, please let me know if I should have done it differently.

sozud: Ok how about just do `#define SFX_UNK_PSP_304 0x304` and take out the edit to the sfx enum until more research is done?

sozud (approved):

PR #2817, opened 2025-08-21
Mapping BO4 Data

hohle:
Maps most of BO4 assets and data.

The `blueprintsdef` assets handler now filters headers through `cpp` to support includes and other macros. This was allows bo4 and rbo5 to continue to share enums and use assets.

The current implementation of `sprite_banks` and `paldef` do not work for this overlay, so those are excluded for now.

This nearly passes a picci build, but there are a number of duplicate functions symbols from DRA that need to be namespaced first.


tools/sotn-assets/assets/blueprintsdef/handler.go
```diff
@@ -162,12 +163,23 @@ func sourcePath(dir, name string) string {
 }
 
 func fetchEnum(srcDir, ovlName, enumName string) (map[int]string, error) {
-	f, err := os.Open(fmt.Sprintf("%s/%s.h", srcDir, ovlName))
+    cpp, err := exec.LookPath("cpp")
```
sozud, line 166: @Xeeynamo I don't know enough about your tool to review this change

sozud:
If the C preprocessor part becomes a problem we can revert

PR #2820, opened 2025-08-22
Decompile 2 `libcd` functions

gamezter:


src/main/psxsdk/libcd/iso9660.c
```diff
@@ -2,10 +2,18 @@
 #include "libcd_internal.h"
 #include "common.h"
 
+typedef struct {
+    s32 unk0;
+    s32 unk4;
+    s32 unk8;
+    char unkC[0x20];
+} CdlDIR;
```
gamezter, line 10: This might not be an appropriate name, I named it this because `D_80039884` is size `CdlMAXDIR` and it looks like the `CdlFILE` counterpart: `extern CdlFILE file[CdlMAXFILE];`

PR #2824, opened 2025-08-23
Fix Duplicates Report, Analyze Calls, & Function Finder

hohle:
The duplicates report is no longer picking up dynamic symbols, which represent an increasing percentage of all function symbols. Previously generating the report would also overwrite files in the work tree which made it difficult to run ad-hoc. `make duplicates-report` should work in any working tree without modifying any files outside of `asm` or `build`.

`symbols.py` now has a `dynamic` command that will dump all of the symbols in an elf file that are not defined in symbol tables for that overlay. This uses the same filtering rules as the `elf` command. Any symbols with the same name or address of one of the static symbols will be excluded from the output.

The `force_symbols` make target emits a table of dynamic symbols for each of the `us` overlays and stores them in `$BUILD_DIR/config`. This "dynamic" symbol file is appended to the static symbol file list of each overlay at extract time. This makes it possible to build the duplicates report without moving `src` or overwriting anything in `config`.

`analyze_calls.py` has been updated with current rules based on new patterns and supports mapping matching functions.

Reports are now written to the build directory to keep the working directory clean. The github workflow moves them to the expected place for the `gh-duplicates` branch.


hohle:
[duplicates.txt](https://gist.github.com/hohle/d78ef21645186d610885aa419c91fbef)

.github/workflows/validate-and-report.yml
```diff
@@ -150,16 +150,18 @@ jobs:
...
         run: |
             find ./build -type f -name "*.ld" -delete
+      - name: Create reports directory
+        run: |
+          mkdir -p gh-duplicates
       - name: Generate duplicates report
         run: |
           make build || true
-          make force_symbols
-          .venv/bin/python3 tools/function_finder/fix_matchings.py
```
JoshSchreuder, line 157: Guess this can be deleted now?

hohle: Yeah, it’s not necessary.

hohle:
After spending a little more time with this this morning I’m going to move most of this to ninja. Thr core concepts will be the same, but generating the symbol tables and splat config is straight forward and can leverage the ninja dag.

hohle:
Removed the make targets so now the dynamic symbol tables as well as the injected splat config to use those tables are part of the ninja build graph. This removes the list of targets and makes the dynamic symbol logic portable to other versions as well.

This will run quickly from a clean build, but still requires a full rebuild after running. This isn't something most people will be affected by and the rebuild happens automatically the next time `make` is run.

hohle:
New failure for PSP. Will look in the morning.

hohle:
The PSP issue was caused by undefined symbols handling in `symbols.py`. That's been fixed.

The tooling test issue is a little more fundamental and I'm still working through it. Previously, the `force_symbols` target didn't use the build graph so didn't have to worry about new overlays. I'm working through different scenarios (clean repo, partially built, etc.) that work for both reports and make-config

hohle:
Converting to draft while I work out the ninja DAG along with best-effort symbol table generation.

hohle:
I think I found a reasonable pattern that avoids circular dependencies, fails gracefully when there are no built overlays, and works when there are.

`force_symbols` should now handle running without built overlays better than before. if there are no elf files, no dynamic symbols will be added, but none of the static symbols will be removed, either. The `config` directory isn't touched so `force_symbols` is safe to run in a dirty working tree. After symbols are forced, ninja takes care of identifying the "splat" rule as dirty and will rebuild everything during the next build.

sozud:
A review from @Xeeynamo would be helpful since he wrote most of the modified code

PR #2831, opened 2025-08-25
Clean Up Build Warnings

hohle:
Make the `FntPrint` function signature consistent between header and impl and and put guard around duplicate `CLAMP` macro in `libgpu/sys.c`.


sozud:
Needs build fixes

sozud:
Could you post the warnings you're trying to avoid with these changes?

hohle:
Both of these currently affect `us`. The `CLAMP` redeclaration also affects the portable build.

```
[551/1984] psx cc src/main/psxsdk/libgpu/font.c
src/main/psxsdk/libgpu/font.c:181: conflicting types for `FntPrint'
include/psxsdk/libgpu.h:603: previous declaration of `FntPrint'
```

and
```
[567/1984] psx cc src/main/psxsdk/libgpu/sys.c
src/main/psxsdk/libgpu/sys.c:68: warning: "CLAMP" redefined
   68 | #define CLAMP(value, low, high)                                                \
      | 
In file included from src/main/psxsdk/libgpu/sys.c:2:
include/common.h:146: note: this is the location of the previous definition
  146 | #define CLAMP(x, min, max) x < min ? min : (x > max ? max : x)
      | 
```

sozud:
The original signature of FntPrint is `extern int FntPrint(...);` in C++ mode or `extern int FntPrint();` in C mode. I don't have an answer to this problem but figured I should point that out

https://github.com/sozud/psy-q/blob/2266d676f8de8a7e7a96d9cbb92854f0a55a86a1/3.5/PSX/INCLUDE/LIBGPU.H#L694

hohle:
Unfortunately, the C++ signature, which is listed in the PDF docs isn't valid C, though it best represents the function. I'll try the empty arg version again and see if I can make progress there.

JoshSchreuder:
@Xeeynamo I tried these changes locally and it seemed all good to me, warnings are gone.

JoshSchreuder (approved):

PR #2832, opened 2025-08-26
Decompile `libspu` func_800286E0 + code cleanup

gamezter:
https://decomp.me/scratch/8SkhQ


Mc-muffin:
This function is actually called `_spu_gcSPU` and it's a static function, worth mentioning that all but 2 of the `libspu` functions listed in the [functions list](https://github.com/Xeeynamo/sotn-decomp/blob/gh-duplicates/functions.md) for this game are solved, though they are using other conventions different from this project :/
Here's my take of this function for instance https://decomp.me/scratch/EOb4q

sozud:
You need to remove this to fix the build error https://github.com/Xeeynamo/sotn-decomp/blob/14222ece8bfa286c9f097c29863bc4b65f441cf7/src/pc/psxsdk/libsnd.c#L595

PR #2834, opened 2025-08-27
Separate Assets that Differ Between Versions

hohle:
Splits sound effects and graphics that differ between versions. Adds macros for choosing the appropriate gen files at build time.

Building different versions of the game without cleaning caused build errors due to assets writing to the same files. These conflicting and differing assets have now been moved to `gen/$version`. This makes it easy to find which specific assets differ and keep the code unaware when they do.

`sotn-assets`'s `spritesheet` handler now treats the `name` field as a relative path like other handlers. The created variable name is based on the last component of that path.

This does not make it possible to build different game versions in parallel. `build.ninja`, several generated C files, and assets that do not differ between versions are still overwritten during each build. Even if they don't conflict, reading a file as it is being written by another build may cause transative failures. So for now different builds need to be done serially, but no longer require a clean between them.


hohle:
Pending PC build check

PR #2837, opened 2025-08-27
Decompile `_spu_pitch2note`

gamezter:
Copied from @Mc-muffin's scratch : https://decomp.me/scratch/k3ClY

EDIT: also decompiled `_spu_t` based off @Mc-muffin's scratch : https://decomp.me/scratch/EfAgN + code cleanup


src/main/psxsdk/libspu/s_n2p.c
```diff
@@ -30,4 +30,44 @@ s32 _spu_note2pitch(s32 center_note, u16 center_note_frac, s32 offset_note,
...
+    // Find least significant bit of arg2
+    lsb = 0;
+    for (i = 0; i < 16; i++) {
+        if ((arg2 >> i) & 1) {
+            lsb = i;
+        }
+    }
+
+    // Equivalent to `lsb_mask = 1 << lsb;`
+    lsb_mask = (1 << (lsb + 1)) - (1 << lsb);
```
gamezter, line 51: Why would they do this ? Maybe it comes from a macro ?

gamezter: Proof : https://roeming.github.io/C-Math-Evaluator/?seq=%281+%3C%3C+%28x+%2B+1%29%29+-+%281+%3C%3C+x%29

sozud: I'm not sure but it may be some boilerplate thing that gets copied around since I see it exists elsewhere https://github.com/liupengsay/PyIsTheBestLang/blob/265fb4516a5d37db7745b85e098a0c40bbec4a03/src/math/bit_operation/problem.py#L1265

Mc-muffin: Looks like a "flat" version of a "create mask" macro, something like `MASK(_lsb, _msb) ((1 << (_msb + 1)) - (1 << _lsb))` with both parameters being the same (perhaps another macro? `BIT(_bit) (MASK(_bit, _bit))`?), so you can create a mask for a given bit range

PR #2840, opened 2025-08-29
Add BO4 to Portable Builds

hohle:
Namespaces several BO4 symbols that were duplicates of those from DRA and RBO5 to allow the BO4 code to be built in the portable build.

Stubs out a few globals which are not yet supported by `sotn-assets`.


hohle:
I'm not sure if I'm building locally with clang or gcc, but will rerun with both.

hohle:
`clang` wasn't happy with using an implicit declaration and then later providing a definition in the same file. gcc produced the same diagnostic, but as a warning.

I added the `unused` attribute and then went back to replace it with a conditional `UNUSED` macro, but missed one case. mcvc doesn't understand `__attribute__`s, so it was failing as well. Replacing with the macro resolved the issue.

hohle:
Another msvc failure due to zero length arrays in the overlay loader to stub out some missing data. This is a language extension not supported by msvc. My hope was they would cause early errors (as well as errors in valgrind and other tools) by being zero length

hohle:
Now running into link errors as well on windows. I'll namespace these later today.

hohle:
Fixed with a combination of namespacing and excluding unused functions from portable builds.

PR #2841, opened 2025-09-01
BO6 Richter Import

hohle:
Import BO6 and dedup most of the shared code.

There are a lot of implementation similiarities between boss Richter and player Richter, but also both Dopplegangers.


sozud:
Needs to be added to progress.py I think

PR #2843, opened 2025-09-03
Added accommodation of unused gfx banks to sotn-assets

ProjectOblivion:
Found that DAI has an unused gfx bank, so I adjusted sotn-assets to iterate through the data for the banks instead of using gfxBanks.
This only affects what ovl_gfxbank_X symbols are added to the .h file and maintains proper construction of OVL_gfxBanks itself.

I also removed some minor duplication and moved handling of null addresses to util.SortUniqueOffsets.


sozud:
@Xeeynamo could you take a look since you wrote this tool

tools/sotn-assets/util/utils.go
```diff
@@ -123,6 +123,9 @@ func SortUniqueOffsets(slice []psx.Addr) []psx.Addr {
 		}
 		return 0
 	})
+	if newSlice[0] == psx.RamNull { // ignore null address
+		newSlice = newSlice[1:]
+	}
```
Xeeynamo: It's dangerous to add this fix for `graphics` to a shared utility. `SortUniqueOffsets` is also used by `layout` and I am not sure if the change could cause problems on newer overlays.

My suggestion is to create another helper that wraps `SortUniqueOffsets` and specifically removes the first null pointer from the sorted list. This new helper could be located in either `assets/gfxbanks` or could still live in `util`

tools/sotn-assets/assets/gfxbanks/graphics.go
```diff
@@ -87,10 +87,10 @@ func ReadGraphics(r io.ReadSeeker, ramBase, addr psx.Addr, symbol func(addr psx.
...
 	var banks []GfxBank
 	var ranges []datarange.DataRange
-	for _, addrGfxBank := range util.SortUniqueOffsets(addrGfxBanks) {
-		if addrGfxBank == psx.RamNull { // exception for ST0
-			continue
-		}
+	var newRange datarange.DataRange
+	sortedBanks := util.SortUniqueOffsets(addrGfxBanks)
+	lastAddr := sortedBanks[len(sortedBanks)-1]
+	for addrGfxBank := sortedBanks[0]; addrGfxBank <= lastAddr; {
```
Xeeynamo:
```suggestion
	for addrGfxBank := sortedBanks[0]; addrGfxBank <= lastAddr; addrGfxBank = newRange.End() {
```
to make clearer `addrGfxBank` is the "index" of the loop.

tools/sotn-assets/util/utils.go
ProjectOblivion: While I don't necessarily disagree with that, since we don't know what the future overlays might hold, I do offer the counter question of whether null should be considered a valid address for offsets.
The reason I opted for this approach is because I don't consider null to be a valid offset personally.  I'm not against your approach, though, if you still think that's the way to go.

It seems that it would make the most sense in util with the main portion, to keep it together.

Xeeynamo: I prefer to fail explicitly (e.g. null pointer dereference) and get a stack trace for future overlays where the current code needs attention, instead of silently ignoring data that might lead to generating unmatching data. Other than that, it is also better to not couple too much logic into an utility function.

hohle:
I don't see any issues with the change, but does this make DAI's bank gen usable? Can the asset config be added as well?

ProjectOblivion:
Yes, I've fully completed us DAI assets.  There are some other issues with sotn-asset compatibility for pspeu.
If you meant adding the config to this PR, I don't think that is in scope, but I could if needed.

PR #2845, opened 2025-09-03
Fix Matchings & Reporting Fixes

hohle:
Runs `fix_matchings.py` prior to generating reports like was previously done.

Maps `0x0($v0)` to `"UnknownEntityFunction"` in `analyze_calls.py`. This appears to be new in `bo6`.

`function_finder.py` now looks for the duplicates report in `build/us/reports`.


sozud:
linter

hohle:
Yes, but it's also not referenced anywhere. I've changed the `duplicates-report` target to use this definition and made threshold a var (`make DUPS_THRESHOLD=0.8 duplicates-report`).

PR #2849, opened 2025-09-05
Organized structs

ProjectOblivion:
I went through and collected all of the various structs into more central locations in order to facilitate easier reuse of structs:
- several files that were only used to contain a small number of structs were removed and their structs moved to relevant locations
- primitive related structs were move to primitive.h
- dra structs were mostly collected in dra.h, with some that made sense to be in game.h or were related to structs already in game.h being moved there
- stage structs were move to stage.h, with the exception of lib/e_shop structs, which were moved to lib.h since it is clear that they are very specific to lib and almost certainly not used anywhere else
- servant structs remain in their respective .h files with the exception of ItemPrimitiveParams, which was move to primitive.h
- weapon structs were collected in weapon.h
- pc structs were untouched
- main/psxsdk structs were untouched
- saturn structs were untouched
- sel structs were untouched

I moved the two scratchpad structs (no4 and st0) to stage.h even though they are likely not used anywhere else.  The reason for this is to raise some awareness of the scratchpad struct concept for future decompiling.  I suspect it is used in dai somewhere (I haven't looked for where yet) and seems to be a concept that is probably used in other overlays as well.


ProjectOblivion:
> Most structs are meant to be self-contained to the C file that use them. For example struct TilemapRenderer is only used in src/dra/4CE2C.c by RenderTilemap and it is not meant to be shared. This applies to many other structs being moved in your PR.

This is fair, I debated a bit on how to handle this.  Since my two primary objectives were consolidating structs for reuse and eliminating the extra files, I'll review the dra changes to see which ones actually need to be moved and revert the remaining.

include/game.h
```diff
@@ -1564,6 +1595,55 @@ typedef struct {
     /* 0x40 */ u8* scriptEnd;    // pointer to the end of the script
 } Dialogue;                      // size = 0x44
 
+// st0_psp/3101C, st0/bss.c, st0/prologue_scroll.c, st0/3101C
+typedef struct {
```
ProjectOblivion, line 1599: Ultimately, the set of Dialogue structs should probably be in cutscene.h. Currently Dialogue and DIalogue2 are used in st0/bss so collecting them here made more sense to me for now.

include/game.h
```diff
@@ -1965,6 +2096,37 @@ typedef struct {
...
+    /* 0x80097414 */ s32 BottomCornerTextPrims;
+    /* 0x80097418 */ s32 unk18;
+    /* 0x8009741C */ s32 unk1C;
+    /* 0x80097420 */ s32 unk20;
+    /* 0x80097424 */ s32 unk24;
+
+    // size must be 8 for the loop in RunMainEngine, while
+    // PreventEntityFromRespawning suggests it has a size of 32
+    /* 0x80097428 */ s32 D_80097428[8];
+} unkGraphicsStruct;
```
ProjectOblivion, line 2092: This and the next one were previously in unkstruct.h.

include/primitive.h
ProjectOblivion, line 1: I moved the prim related structs here because it seems to me like there is a high chance of them being reused.

include/sfx.h
```diff
@@ -711,4 +711,11 @@ enum Sfx {
 
 #endif
 
+// related to SFX
+typedef struct {
+    /* 0x00 */ s16 sndId;
+    /* 0x02 */ u16 sndVol;
+    /* 0x04 */ s16 sndPan;
+} SfxRingBufferItem;
```
ProjectOblivion, line 719: This was previously in unkstruct.h.

include/stage.h
```diff
@@ -8,6 +8,11 @@
 #include "objects.h"
 #include "sfx.h"
 
+// used by InitializeEntity to pre-fill certain entity fields
+// cannot be declared as a struct, field order:
+// animSet, animCurFrame, unk5A, palette, enemyID
+typedef u16 EInit[5];
```
ProjectOblivion, line 14: This was moved to the top to allow it to be used in stage.h structs.

src/dra/dra.h
ProjectOblivion, line 1: Only structs that are used in multiple segments were moved here.

include/weapon.h
```diff
@@ -24,6 +24,40 @@ typedef struct {
...
+    s16 unk8;
+    u16 unkA;
+    s16 unkC;
+    s16 unkE;
+    s16 unk10;
+    s16 spawnDelay;
+    s16 unk14;
+    s16 state;
+} FireShieldDragon;
+
```
hohle: If these are just used in those weapons, can they just be declared in those weapon's header files?

include/game.h
```diff
@@ -1676,6 +1756,57 @@ typedef struct {
...
+    f32 posX;
+    f32 posY;
+    s16 angle1;
+    s16 angle2;
+    s16 size;
+    s16 xOffset;
+    s16 yOffset;
+    s16 pad;
+} mistStruct; // size = 0x14
+
```
hohle: This is used in multiple overlays, but it's not really a "game" wide structure. Putting it here makes it a dependency of nearly every compiled file.

Could this be moved to something like `include/alucard.h` or something broader than `mist` but more granular than `game`?

hohle:
> Most structs are meant to be self-contained to the C file that use them.

I'd like to go through this a bit deeper, but I want to agree with this as well.

The US and HD versions of the build will now rebuild all header dependents when the header changes (there's no specific checks to rebuild everything when particular header files change). I'm trying to get `mwccgap` to support the same thing for the psp builds. The less the header include graph looks like a straight line from every file to `game.h` the faster local builds and iteration are. (e.g. do something like `touch src/st/e_collect.h; make` and compare to `touch include/game.h; make`).

My preference would be to keep headers as granular as possible (single struct is overkill), but think of them like modules, packages, or whatever in any other language. My preference would be that refactors move things out of `game.h` and not into it. A global header that every file depends on should be stable and rarely change.

include/weapon.h
ProjectOblivion: Did you mean the c files? I didn't see any additional weapon specific header files.
Both of them seem to me like they could be reused somewhere else, but I have no major objections if leaving those two in their respective .c files is the overall preference.

include/game.h
ProjectOblivion: Thinking about this and your other comment, these structs make much more sense being in a maria/ric/alucard include file, rather than game.h.
I'll wait for feedback from @Xeeynamo before I make any more changes.

ProjectOblivion:
> > Most structs are meant to be self-contained to the C file that use them.
> 
> I'd like to go through this a bit deeper, but I want to agree with this as well.
> 

I agree with this for structs that we have fully identified and are confident that they aren't used elsewhere.  I found quite a few structs that were duplicated in multiple places because of that practice.

> The US and HD versions of the build will now rebuild all header dependents when the header changes (there's no specific checks to rebuild everything when particular header files change). I'm trying to get `mwccgap` to support the same thing for the psp builds. The less the header include graph looks like a straight line from every file to `game.h` the faster local builds and iteration are. (e.g. do something like `touch src/st/e_collect.h; make` and compare to `touch include/game.h; make`).
> 
> My preference would be to keep headers as granular as possible (single struct is overkill), but think of them like modules, packages, or whatever in any other language. My preference would be that refactors move things out of `game.h` and not into it. A global header that every file depends on should be stable and rarely change.

I agree with this.  The structs added to game.h in this PR make more sense in a separate header file, though I think moving Dialogue out of game.h, rather than moving DIalogue2/3 into game.h, for this PR would be a bit more troublesome than collecting them and moving them out later.

include/weapon.h
Xeeynamo: > weapon specific header files

`weapon_private.h`

include/game.h
Xeeynamo: Yes, eventually we'd like to isolate all the Alucard-related code. Maybe it's a good opportunity to re-use `player_steps.h` as `alucard.h` and put `mistStruct` there?

include/primitive.h
Xeeynamo: I think it is fine for now. But let's keep an eye on those "primitive" structs. I am very suspicious the original struct have been copy&pasted then modified ad-hoc for multiple entities.

include/game.h
ProjectOblivion: Any objection to adding Props_80161FF0, FactoryBlueprint, unkStr_8011E4BC, and mistStruct to player_steps.h and renaming it to simply player.h?
I'm not sure if there is anything else that make sense in that file, but that would minimize how much is being added to game.h.

Xeeynamo: * `Props_80161FF0` it's better if it remains self-contained to ric and maria. We do not want to couple structs across overlays that could otherwise remain private to the C file they've been used.
* I am conflicted about `FactoryBlueprint` for the same reason as above. But since we have one blueprint handler in SOTN Asset that is used on three different overlays, I think it's better if we keep it shared for now
* We do not want `mistStruct` to be included in ric and maria. So maybe we could have the two separate headers `player.h` and `alucard.h`?

ProjectOblivion: - I moved Props_80161FF0 back to the original files
- I left FactoryBlueprint and unkStr_8011E4BC in game.h. When I found this also touched pc, I realized that it was better to leave this PR at the zero sum operation of simply moving it from unkstruct.h, which was previously included in game.h, directly into game.h.
- I moved mistStruct back to the locations it was in originally, but just inlined it to still remove dra/mist.h. This made much more sense than creating a new header file

include/weapon.h
ProjectOblivion: I moved these structs to weapon_private.h

PR #2852, opened 2025-09-08
Decompile `SpuSetCommonAttr` & `_SpuSetVoiceAttr`

gamezter:
`SpuSetCommonAttr` mostly copied from @Mc-muffin 's scratch : https://decomp.me/scratch/M3gZL
`_SpuSetVoiceAttr` looked similar but no one had decompiled it yet. I managed to get a match with the permuter's help : https://decomp.me/scratch/cWIg3

### DIFF
--- a/src/main/psxsdk/libspu/s_sva.c
+++ b/src/main/psxsdk/libspu/s_sva.c
@@ -9,6 +9,7 @@ void SpuSetVoiceAttr(SpuVoiceAttr* arg) {
 
 static inline u16 inline_fn(u16 arg0, u16 arg1) { return arg0 & arg1; }
 
+#ifndef VERSION_PC
 s32 _SpuSetVoiceAttr(SpuVoiceAttr* attr, s32 min, s32 max, s32 arg3) {
     volatile s32 sp10;
     volatile s32 sp14;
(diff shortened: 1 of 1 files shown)


src/main/psxsdk/libspu/s_sva.c
```diff
@@ -3,6 +3,284 @@
...
 
-void SpuSetVoiceAttr(SpuVoiceAttr* arg) { _SpuSetVoiceAttr(arg, 0, 0x17, 0); }
+void SpuSetVoiceAttr(SpuVoiceAttr* arg) {
+    _SpuSetVoiceAttr(arg, 0, NUM_SPU_CHANNELS - 1, 0);
+}
 
-INCLUDE_ASM("main/nonmatchings/psxsdk/libspu/s_sva", _SpuSetVoiceAttr);
+static inline u16 inline_fn(u16 arg0, u16 arg1) {
+    return arg0 & arg1;
+}
```
gamezter: This might actually be an inline function that does something like `return (arg0 & arg1) | arg2` that could be used in multiple functions in `libspu` but I havent tried getting it to work in other places. For now it is the only way I managed to get a match in `_SpuSetVoiceAttr`

PR #2854, opened 2025-09-09
More `CutsceneRun` Sharing

hohle:
`CutsceneRun` has a common difference in many stages where the entity is moved in relation to the tile map's scroll position. Instead of continuing to add stages to this difference, I've added a new macro that stages can opt into to get the same behavior which decouples the cutscene code a bit. This allows `dai` and `bo6` to use the shared `CutsceneRun`.

On the other hand, I introduced a new check for `DRE` specifically (since that's the only stage that behaves this way) for how the `scriptEnd` and `timer` are decremented. This allows `dre` to use the shared `CutsceneRun`.

### DIFF
--- a/src/boss/bo6/cutscene.c
+++ b/src/boss/bo6/cutscene.c
@@ -20,7 +20,8 @@ extern const char* actor_names[];
 
 #include "../../st/set_cutscene_end.h"
 
-INCLUDE_ASM("boss/bo6/nonmatchings/cutscene", CutsceneRun);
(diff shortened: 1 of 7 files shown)


ProjectOblivion:
I think more can actually be done with this since the sequence involving `CutsceneRun` has consistency across the overlays for both pspeu and us.
SetCutsceneEnd -> CutsceneRun -> CutsceneSkip -> ScaleCutsceneAvatar

I think with the changes from this PR, these could all be merged into one file, though I haven't confirmed it.

PR #2855, opened 2025-09-10
Decompile `SpuVmSetVol`

gamezter:
Thanks to @andre_cabral who helped me with a final mismatch : https://discord.com/channels/1079389589950705684/1079395108501331990/1415085514788438138

### DIFF
--- a/src/main/psxsdk/libsnd/vmanager.c
+++ b/src/main/psxsdk/libsnd/vmanager.c
@@ -945,7 +945,74 @@ s32 SpuVmGetProgPan(s16 arg0, s16 arg1) {
     }
 }
 
-INCLUDE_ASM("main/nonmatchings/psxsdk/libsnd/vmanager", SpuVmSetVol);
+s32 SpuVmSetVol(s16 seq_sep_no, s16 vabId, s16 prog, u16 arg3, u16 arg4) {
+    s32 var_s2;
+    u8 voice;
+    s32 new_var;
+    u32 volr;
+    u32 voll;
+    u32 volr_t;
+    u32 voll_t;
+    struct SeqStruct* temp_s1;
+    s32 mvol_scaled;
+    s16 new_var2;
+    register u16 pan_arg asm("s3") = arg4;
+
+    var_s2 = 0;
+    temp_s1 = &_ss_score[seq_sep_no & 0xFF][(seq_sep_no & 0xFF00) >> 8];
+    SpuVmVSetUp(vabId, prog);
+    _svm_cur.field_16_vag_idx = seq_sep_no;
+    for (voice = 0; voice < spuVmMaxVoice; voice++) {
+        new_var2 = prog;
+        if ((_svm_voice[voice].unke == seq_sep_no) && (_svm_voice[voice].prog == new_var2) && (_svm_voice[voice].vabId == vabId)) {
+            if (temp_s1->vol[temp_s1->channel] != arg3) {
+                if (temp_s1->vol[temp_s1->channel] == 0) {
+                    temp_s1->vol[temp_s1->channel] = 1;
+                }
+            }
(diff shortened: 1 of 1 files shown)


src/main/psxsdk/libsnd/vmanager.c
```diff
@@ -945,7 +945,80 @@ s32 SpuVmGetProgPan(s16 arg0, s16 arg1) {
...
+                volr = (volr * _svm_pg[_svm_voice[voice].unk10].mpan) / 0x3F;
+            } else {
+                voll = (voll * (0x7F - _svm_pg[_svm_voice[voice].unk10].mpan)) /
+                       0x3F;
+            }
+            if ((pan_arg & 0xFF) < 0x40) {
+                volr = (volr * (pan_arg & 0xFF)) / 0x3F;
+            } else {
+                voll = (voll * (0x7F - (pan_arg & 0xFF))) / 0x3F;
+            }
```
gamezter: This pattern is repeated many times in this file. I tried converting it to an inline function but that requires passing `voll` and `volr` by reference which causes a mismatch. Maybe it's a macro ?

src/main/psxsdk/libsnd/vmanager.c
```diff
@@ -945,7 +945,82 @@ s32 SpuVmGetProgPan(s16 arg0, s16 arg1) {
...
+    u8 voice;
+    s32 new_var;
+    u32 volr;
+    u32 voll;
+    u32 volr_t;
+    u32 voll_t;
+    struct SeqStruct* temp_s1;
+    s32 mvol_scaled;
+    s16 new_var2;
+    register u16 pan_arg asm("s3") = arg4;
```
sozud: Have you tried the permuter? Using `register` isn't really a match

gamezter: I have, it gets stuck with this last mismatch : https://decomp.me/scratch/JSt2f  . I'm not a fan of this solution either, I can leave it as draft until we find a better solution ?

hohle: @sozud while not optimal, would having a macro like `REGISTER(type, var, reg)` that expands to `register type var asm(#reg)` and  `type var` everywhere else make sense?

sozud: I think if we want to have a non-matching version we could do `#ifdef VERSION_PC` or similar. What does the `REGISTER` macro gain us? To me if we use `register` it takes the function out of this list, and removes a INCLUDE_ASM, so it's harder to tell it's nonmatching https://github.com/Xeeynamo/sotn-decomp/blob/gh-duplicates/functions.md

hohle: Yeah, it would just be a way to get a version working for both `VERSION_PC` and `us` without redefining the entire function. I get the point about removing it from non matching, though.

gamezter: Found a match without using `register` :)

src/main/psxsdk/libsnd/vs_vh.c
```diff
@@ -66,13 +65,13 @@ s16 SsVabOpenHeadWithMode(u8* addr, s16 vabid, s16 arg2, u32 sbaddr) {
...
     magic = vab_hdr_2->form;
-    if ((magic >> 8) != 0x564142) {
+    if ((magic >> 8) != ('V' << 0x10 | 'A' << 0x8 | 'B')) {
         _svm_vab_used[vabId_2] = 0;
         _spu_setInTransfer(0);
         _svm_vab_count -= 1;
         return -1;
     }
-    if ((magic & 0xFF) == 0x70) {
+    if ((magic & 0xFF) == 'p') {
```
gamezter, line 74: VAB header starts with string "VABp" : https://psx-spx.consoledev.net/cdromfileformats/#vab-header-vh

PR #2856, opened 2025-09-10
Various `main` code cleanups

gamezter:


src/main/psxsdk/libcd/sys.c
```diff
@@ -84,11 +84,10 @@ CdlCB CdReadyCallback(CdlCB func) {
 }
 
 static inline cd_cw(u8 com, u8* param, Result_t* result, s32 arg3) {
-    s32 i;
-
     CdlCB old = CD_cbsync;
+    int count = 4;
 
-    for (i = 3; i != -1; i--) {
+    while (count--) {
```
gamezter, line 90: This pattern was pretty common in `main`, I realized that it could be converted to a simpler while loop. Also the 'count' / 'size' variable doesn't work with `s32` ? I was sure `s32` and `int` were the same type.

PR #2861, opened 2025-09-15
Decompile `SpuVmKeyOn`

gamezter:
https://decomp.me/scratch/WOS1y

The final clue that unblocked me was figuring out that this was using an inlined function `_SsVmSelectToneAndVag` : https://github.com/ps2dev/ps2sdk/blob/566ed82da2c57ed0e86dfbec5b69a8f1d0ecb76b/iop/sound/libsnd2/src/vm/vm_key.c#L54

I placed it below `SePitchBend` according to this list of symbols : https://github.com/psyqz/psyz/blob/3409b8f8d7cb14f737985506eef6e9506c5e2493/decomp/symbols.470.txt#L2326


src/main/psxsdk/libsnd/vmanager.c
```diff
@@ -307,6 +307,24 @@ void SePitchBend(u8 voice, s16 arg1) {
     }
 }
 
+static inline u8 _SsVmSelectToneAndVag(u8* toneIdx, u8* vagIdx) {
```
gamezter, line 310: It might actually be called `SpuVmSelectToneAndVag` : https://github.com/abelbriggs1/tm1_decomp/blob/a289574f0218fb1bdacd3b9fe744bb2abeb39142/config/JP/symbol_addrs.txt#L1776 

It seems a lot of these functions are named `_Ss...` in other versions, and renamed `Spu...` in this version.

sozud: This may become clear once we add newer versions to psy-q-decomp since the newer ones have more symbols

sozud (approved):

PR #2867, opened 2025-09-17
Decompile `SpuSetReverbModeParam` + cleanup

gamezter:
https://decomp.me/scratch/AZYa5

### DIFF
--- a/src/main/psxsdk/libsnd/vmanager.c
+++ b/src/main/psxsdk/libsnd/vmanager.c
@@ -105,24 +105,23 @@ u8 SpuVmAlloc(s32 arg0) {
 
 #ifndef VERSION_PC
 void SpuVmKeyOnNow(u8 vagCount, u16 arg2) {
+    struct SeqStruct* score =
+        &_ss_score[_svm_cur.field_16_vag_idx & 0xFF]
+                  [(_svm_cur.field_16_vag_idx & 0xFF00) >> 8];
     u16 pos;
     u16 bitsUpper;
     u16 bitsLower;
-    u32 volr;
-    u32 voll;
-    u32 volr_t;
-    u32 voll_t;
-    struct SeqStruct* score;
+    u32 voll, volr;
+    u32 voll_t, volr_t;
     s32 mvol_scaled;
 
     pos = _svm_cur.field_0x1a * 8;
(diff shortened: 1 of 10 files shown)


src/main/psxsdk/libsnd/vmanager.c
```diff
@@ -466,18 +462,21 @@ void SetAutoPan(s16 voice) {
...
-    temp_hi = (temp_hi * _svm_cur.field_D_vol) / 0x3F01U;
-    temp_v1_4 = temp_hi;
+
+    voll_t = ((_svm_cur.field_4_voll * mvol_scaled) / 0x7F) / 0x7F;
+    volr_t = ((_svm_cur.field_4_voll * mvol_scaled) / 0x7F) / 0x7F;
+
+    voll_t =
+        ((voll_t * _svm_cur.field_A_mvol * _svm_cur.field_D_vol) / 0x7F) / 0x7F;
+    volr_t =
+        ((volr_t * _svm_cur.field_A_mvol * _svm_cur.field_D_vol) / 0x7F) / 0x7F;
```
gamezter, line 472: Turns out these cursed `/ 0x3F01` are actually `/ (0x7F * 0x7F)`. Based off this in-progress scratch : https://decomp.me/scratch/v82al I think these lines could even look like
`voll_t = (voll_t * _svm_cur.field_A_mvol) / 0x7f;`
`voll_t = (voll_t * _svm_cur.field_D_vol) / 0x7f;` 
but that doesn't match. Maybe these come from inlined functions ?

gamezter: When functions get inlined there *seems* to be an optimisation pass to remove dead code. I'm thinking maybe that pass would rearrange the operations?

sozud: Here's a strings dump of the 4.0 libsnd ecoff.  https://gist.github.com/sozud/3eeb33337ff0e121378ecbc4a0e8113f
So I believe the way inlines work is that they are still in the symbols for later psy-q versions. You can see that `_SsVmSelectToneAndVag` is there for example. If you look for `vmNoiseOn` I don't see a function that would be an obvious inline name near it. These could be macros though, or it may be an inline and I'm just misunderstanding how it works

PR #2869, opened 2025-09-22
Decompile `CD_newmedia`

gamezter:
https://decomp.me/scratch/0c0QA

### DIFF
--- a/src/main/psxsdk/libcd/iso9660.c
+++ b/src/main/psxsdk/libcd/iso9660.c
@@ -2,18 +2,30 @@
 #include "libcd_internal.h"
 #include "common.h"
 
+typedef union {
+    s32 addr;
+    struct {
+        u8 unk0;
+        u8 unk1;
+        u8 unk2;
+        u8 unk3;
+    } i;
+} LBA;
+
 typedef struct {
     s32 unk0;
     s32 unk4;
-    s32 unk8;
+    LBA unk8;
     char unkC[0x20];
 } CdlDIR;
 
 extern int CD_nopen;
 extern int D_80032AB0;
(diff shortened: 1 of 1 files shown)


src/main/psxsdk/libcd/iso9660.c
```diff
@@ -98,7 +110,68 @@ int _cmp(const char* str1, const char* str2) {
...
+        return 0;
+    }
+    if (strncmp(&load_buf[1], "CD001", 5) != 0) {
+        if (D_80032AB0 > 0) {
+            printf("CD_newmedia: Disc format error in cd_read(PVD)\n");
+        }
+        return 0;
+    }
+
+    (&pathTableLBA)->i = ((LBA*)(&load_buf[0x8C]))->i;
```
gamezter, line 132: I'm not sure I understand this fully. Is loading LBA->i inverting the byte order when we later access LBA->addr ?

sozud: As far as I understand it's the same byte order
```
#include <stdio.h>


typedef union {
    int addr;
    struct {
        char unk0;
        char unk1;
        char unk2;
        char unk3;
    } i;
} LBA;

int main() {
    LBA a;
    char test[4] = {0,1,2,3};
    a.i = ((LBA*)test)->i;
    printf("%08X\n", a.addr);
    printf("%08X\n", a.i);
    return 0;
}
```
```
03020100
03020100
```

gamezter: What happens if you do `printf("%08X\n", LOW(test[0]));` ?

PR #2873, opened 2025-09-23
Build All Versions Together

hohle:
Changes the default target of `make` and `tools/builds/gen.py` to build all decomped versions of the game together in one bulid graph.

There are two parts to this change: making assets writes atomic to allowing assets to be shared across versions, and updating the build graph to not duplicate shared targets.

Different versions of the game are currently configured to write to the same asset paths. I think this is a good thing™ since it makes differing assets obvious. The work to separate these assets was done in cdda8fcd87e20ee88edbecea069cbd1108f07883 but running multiple versions of `sotn-assets` in parallel resulted in race conditions that prevented multiple versions of the game from building together (whether in the same graph or not). This change updates `sotn-assets` to write files atomically which avoid partial reads while allowing multiple instances of `sotn-assets` to run concurrently. An alternative to this approach would be to put `sotn-assets` rules in the same ninja pool, but that would force them to be serialized, which seems unnecessary, but possibly simpler.

The second area is merging all three build graphs into a single monolithic graph. This is primarily meant eliminating duplicate asset steps for shared assets.

`make` will now build `us`, `hd`, and `pspeu`. using the `VERSION` env var still works and will only build that version. `make clean` cleans all versions.

picci builds are out of scope for this change.

### DIFF
--- a/.github/workflows/validate-and-report.yml
+++ b/.github/workflows/validate-and-report.yml
@@ -97,6 +97,46 @@ jobs:
         if: github.ref == 'refs/heads/master' && github.event_name == 'push'
         run: .venv/bin/python3 tools/progress.py --version ${{ matrix.version }}
 
+  # unlike the matrix build above, this task builds all versions of the
+  # game from the same build graph
+  build-all:
+    # Building and testing cannot work if the repository owner is not Xeeynamo
+    # due to the missing secrets to clone the CI dependencies
+    if: github.repository == 'Xeeynamo/sotn-decomp'
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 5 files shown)


hohle:
It would probably be a good idea to add the combined build to the build matrix as well.

hohle:
This has been working well for me for several weeks and I've used it as the base for several development branches. Marking as ready for review.

hohle:
Care should be taken with the workflow change. I don't have a good way to run or test this.

Xeeynamo (approved):
> Care should be taken with the workflow change. I don't have a good way to run or test this.

I'd say to merge it, and if it has problems we will fix them along the way. Let's move fast, this is a very good addition to the project!!

Xeeynamo (approved):
.

Xeeynamo:
I temporarily reverted it to fix the build on master

PR #2875, opened 2025-09-23
Tooling impovements

misterjdrg:
1. Cleaner code
2. Less `.clone()`ing
3. macro for defining flag and enum transformer types in sotn-lint
4.  interface change `fn transform_line(&self, line: &str) -> String;` -> `fn transform_line(&self, line: &str) -> Option<String>;` to allow implementations that don't allocate String when nothing changes
5. interface change `fn check_line(&self, line: &str) -> Result<(), String>;` -> `fn check_line(&self, line: &str) -> Vec<String>;` to allow linters return multiple errors on line
6. `EntityRangeLinter` -> `ObjectRangeLinker::<Gentries>` to allow other object impls
7. External behavior after applying this pr should be same


Xeeynamo:
Hi, next time please open an issue before refactoring our code base. Or at least create a discussion to suggest a linter+formatter to cover these changes automatically.

PR #2877, opened 2025-09-26
Added `dai` asset extraction

ProjectOblivion:


src/st/dai/gfx_data.c
```diff
@@ -0,0 +1,166 @@
...
+
+u8 gfx_spectral_sword[] = {
+#include "gen/gfx_spectral_sword.h"
+};
+
+#ifndef VERSION_PSP
+u8 gfx_sealed_door[] = {
+#include "gen/gfx_sealed_door.h"
+};
+#endif
```
Xeeynamo, line 70: There is no need to do `#ifndef VERSION_PSP` for unused data. Unused symbols get stripped at link time `--gc-sections`.

src/st/dai/gfx_data.c
```diff
@@ -0,0 +1,166 @@
...
+
+u8 gfx_cutscene_1[] = {
+#include "gen/gfx_cutscene_1.h"
+};
+
+u8 gfx_cutscene_2[] = {
+#include "gen/gfx_cutscene_2.h"
+};
+
+#ifndef VERSION_PSP
```
Xeeynamo, line 92: Same here

src/st/dai/gfx_data.c
```diff
@@ -0,0 +1,166 @@
...
+u8 gfx_sealed_door[] = {
+#include "gen/gfx_sealed_door.h"
+};
+#endif
+
+u16 pal_skelerang_spittle_bone[] = {
+#include "gen/pal_skelerang_spittle_bone.h"
+};
+
+#ifndef VERSION_PSP
```
Xeeynamo, line 106: same here

src/st/dai/gfx_data.c
ProjectOblivion: This one is in different places for psp an psx, so there would be a conflict without the ifndef.

src/st/dai/gfx_data.c
ProjectOblivion: gfx_stage_name_en.h doesn't exist for psp, so it throws an error without the ifndef.

src/st/dai/gfx_data.c
ProjectOblivion: This also doesn't exist for psp.

src/st/dai/gfx_data.c
Xeeynamo: Oh you're right. I just noticed that when building only the PSP version from a clean state :sweat_smile: . I will take that in mind for future reviews.

PR #2879, opened 2025-09-27
BO6 Cutscene Init & Symbols

hohle:
Decompilation of the Richter cutscene init functions.

Also includes room assets and additional symbols.


src/boss/bo6/e_cutscene_actors.c
ProjectOblivion, line 1: I found that this segment is better identified as e_cutscene_actors.  It does things throughout the cutscene lifecycle, rather than just init, and it handles entities like other e_* segments.

src/boss/bo6/cutscene_init.c
```diff
@@ -0,0 +1,315 @@
...
+                DRAW_TPAGE2 | DRAW_TPAGE | DRAW_COLORS | DRAW_TRANSP;
+        }
+        break;
+    case 2:
+        if (g_CutsceneFlags & 0x100) {
+            g_unkGraphicsStruct.unk20 = 0;
+            self->step++;
+            self->ext.utimer.t = 0;
+            prim = &g_PrimBuf[self->primIndex];
+            prim->r0 = prim->r1 = prim->r2 = prim->r3 = prim->g0 = prim->g1 =
```
ProjectOblivion: You should be able to use `PCOL(prim)` to clean up this line and many other similar lines.

PR #2880, opened 2025-09-27
Identified explosion anims and added ExplosionTypes enum

ProjectOblivion:
Enum for use with E_EXPLOSION entity instantiations.
Random paren removal is because of my VSCode save formatting that I didn't catch before the commit.


include/stage.h
```diff
@@ -8,6 +8,14 @@
 #include "objects.h"
 #include "sfx.h"
 
+typedef enum {
+    EXPLOSION_SMALL,
+    EXPLOSION_FIREBALL,
+    EXPLOSION_SMALL_MULTIPLE,
+    EXPLOSION_BIG,
+    EXPLOSION_UNK_4
+} ExplosionTypes;
```
Xeeynamo, line 17: I am suspicious the various explosion types might differ between overlays. Maybe @gamezter or @JoshSchreuder can confirm?

ProjectOblivion: They are tied to the actual animations in e_collect.h, so if the overlay uses that file then they should be the same.

JoshSchreuder: @ProjectOblivion can you give an example usage of this enum?
Is it for eg. this params val?
https://github.com/Xeeynamo/sotn-decomp/blob/997fb62fb47f5021c67b4e10b0e90eb8729dab5e/src/st/no4/e_fishhead.c#L278-L279

ProjectOblivion: Yeah.  I don't remember if there are any other ways it is handled, but the enum is specifically for the `EntityExplosion` params value, like your reference.

JoshSchreuder: @Xeeynamo I think this seems ok.

NZ0/NP3 Gaibon small projectile uses params = 0 which is EXPLOSION_SMALL here
NZ0/NP3 Gaibon large projectile uses params = 1 which is EXPLOSION_FIREBALL here
ST0 Dracula fireball uses params = 2 which is EXPLOSION_SMALL_MULTIPLE here
ST0 Dracula meteor ball uses params = 2 which is EXPLOSION_FIREBALL here
CAT Lossoth fireball attack uses params = 0 which is EXPLOSION_SMALL here
NO4 Fishhead death parts uses params = 2 whcih is EXPLOSION_SMALL_MULTIPLE here

All of these seem to be reasonable I think. Didn't see any existing usages of 3 or 4 in the decompiled code

JoshSchreuder (approved):

include/stage.h
ProjectOblivion: There are several that use params =3, EXPLOSION_BIG, though I also couldn't find any that use params = 4.  Spectral Sword uses it in it's death animation (the main sword, not the floating weapons) and Bone Pillar uses it in the Spike Ball death animation.

PR #2882, opened 2025-09-27
Added AnimateEntityFrame struct

ProjectOblivion:
For using with u9 animations sent to AnimateEntity.

### DIFF
--- a/src/st/dai/background_block_init.c
+++ b/src/st/dai/background_block_init.c
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "dai.h"
 
-static u8 anim_frames[] = {64, 1, -1, 0};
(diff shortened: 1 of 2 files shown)


src/st/dai/background_block_init.c
```diff
@@ -11,5 +11,5 @@ ObjInit2 OVL_EXPORT(BackgroundBlockInit)[] = {{
     .palette = PAL_NONE,
     .drawFlags = DRAW_DEFAULT,
     .drawMode = DRAW_TPAGE,
-    .animFrames = anim_frames,
+    .animFrames = (u8*)anim_frames,
```
ProjectOblivion, line 14: The animFrames element in BackgroundBlockInit requires a cast, but the cast is not needed when sending data to AnimateEntity.

PR #2894, opened 2025-10-01
Documented `dai` `e_room_fg_2`

ProjectOblivion:


src/st/dai/e_room_fg_2.c
```diff
@@ -1,16 +1,24 @@
...
     Primitive* prim;
     s32 primIndex;
     s32 posX;
     s32 count;
 
+    // Nop for when the tiles have been initialized
     if (self->step != 0) {
-        s32 nop; // This does nothing, but
-        nop;     // pspeu seems to require it to match
+        (void)1;
```
JoshSchreuder, line 21: Probably not that important, but maybe we should add a NOP macro for this

src/st/dai/e_room_fg_2.c
```diff
@@ -1,16 +1,24 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "dai.h"
+// Used ext.prim because it didn't seem worthwhile to create an entire ext
+// simply for a prim
+
+/*
+There are two different entities used for the castle wall foreground tiles.
+EntityCastleWall1: Most castle wasll foregruond tiles
```
JoshSchreuder: Nit: misspelt foreground

ProjectOblivion: I don't consider that a nit, lol. Thank you for pointing it out!

src/st/dai/e_room_fg_2.c
ProjectOblivion: I had the same thought, but ultimately decided that it didn't really improve the readability or understandability of the code.

JoshSchreuder (approved):

src/st/dai/e_room_fg_2.c
Xeeynamo: It is fine to keep it like this IMHO. Devs who will stumble on this and remove it, they will see it's no longer matching, then understand it is required to be there.

Xeeynamo:
There's a conflict between the two `ET_Bell et_bell;` and `ET_Bell bell;`

PR #2895, opened 2025-10-01
Documented and coalesced `e_chair`

ProjectOblivion:


src/st/dai/dai.h
```diff
@@ -42,6 +42,13 @@
 #define NUM_CUTSCENE_PRIM 7
 #endif
 
+// These are used by both e_chair and e_confessional
+#define STAGE_HAS_CONFESSIONAL
+#define CONFESSIONAL_GHOST_PRIEST 0
+#define CONFESSIONAL_GHOST_PARISHIONER 1
+#define CONFESSIONAL_GHOST_BAD 0x100
+#define CONFESSIONAL_GHOST_READY 1
```
JoshSchreuder, line 49: I think typically we would put very specific defines like this in the c file (eg. DAI's `e_chair.c`) to keep the defines closer to the usage?

Eg. I am thinking https://github.com/Xeeynamo/sotn-decomp/blob/be8a585c9c6f69a051ec84cb833027d09052ed93/src/st/no1/e_collect.c#L4

src/st/e_chair.h
```diff
@@ -54,23 +106,33 @@ void EntityChair(Entity* self) {
...
+            self->step = CHAIR_CHECK_SIT;
         }
         break;
-
-    case 4:
-        g_Player.unk14 = self->params;
-        if (PLAYER.step != 0 || PLAYER.step_s != 4) {
-            self->step = 1;
+    case CHAIR_PLAYER_SITTING:
+        g_Player.unk14 = self->PARAMS;
```
JoshSchreuder: Random but did `PARAMS` get uppercased here? I'm surprised this isn't a compilation error or warning at least?

src/st/dai/dai.h
ProjectOblivion: I would have done that, but they are used by two different segments.

src/st/e_chair.h
ProjectOblivion: This is a macro defined at the beginning of this file.  The confessional uses `self->params & 0xFF` where the other ovls use just `self->params`.

JoshSchreuder: I don't have a lot of familiarity with C and macros to know the idiomatic way to do this, but this seems a bit confusing to read to me, and personally I think would be clearer expressed as something like

```c
#ifdef STAGE_HAS_CONFESSIONAL
#define PARAMS self->params & 0xFF
#else
#define PARAMS self->params
#endif

...

case CHAIR_PLAYER_SITTING:
        g_Player.unk14 = PARAMS;
```

To distinguish it from a struct member

ProjectOblivion: I also don't know the idiomatic way of handling it.  My concern with your approach was that it loses the association with self when reading the section.  While I personally don't find what I did confusing, since I always read all caps names as macros or enum items, I also tend to interpret things quite differently than most.  I'm sure there is a good answer that satisfies both points.
I'm leaning toward SELF_PARAMS, though I'll defer to whatever @Xeeynamo wants.

Xeeynamo: I think the `STAGE_HAS_CONFESSIONAL` and the `PARAMS` macro are stretching too much to the point where readability hurts. Most entities in this game are modified copies from other entities. If they are meant to be separated then so be it, and we should not try to converge everything with crazy ifdefs or macros.

PR #2899, opened 2025-10-03
Decompiled, imported, and merged `pspeu` `chi` `e_stage_name`

ProjectOblivion:


src/st/e_stage_name_alternate.h
```diff
@@ -0,0 +1,345 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
```
JoshSchreuder: I would be tempted to keep this isolated to CHI's en_stage_name.c for now until we know it's used elsewhere as there is already the PSX and PSP versions of this shared header, and I think 'alternate' muddies the waters a little.

ProjectOblivion: `pspeu` `nz0` also uses this code, though I do agree that the naming could probably be improved.  That is the best name I could think of.

PR #2903, opened 2025-10-04
Documented `dai` `e_corner_guard`

ProjectOblivion:


src/st/dai/e_corner_guard.c
```diff
@@ -318,21 +359,24 @@ void EntityCornerGuard(Entity* self) {
             entity = AllocEntity(&g_Entities[224], &g_Entities[256]);
             if (entity != NULL) {
                 CreateEntityFromEntity(E_EXPLOSION, self, entity);
-                entity->params = 2;
+                entity->params = EXPLOSION_SMALL_MULTIPLE;
             }
-            self->animCurFrame = 0;
+            self->animCurFrame = NULL;
```
sozud: animCurFrame is an index, not a pointer so not sure NULL makes sense

ProjectOblivion: You're right, I'm not sure it makes sense for an index either.  Thanks for pointing it out.

PR #2886, opened 2025-09-27
Documented `dai` `e_bell`

ProjectOblivion:
Also added `VramFlags` enum for use with `g_Player.vram_flag` checks.

PR #2905, opened 2025-10-04
Add linter for `g_Player.vram_flag`

JoshSchreuder:
Should cover off part of @ProjectOblivion's findings in #2886 as well.


sozud (approved):
Looks good but needs a rebase

PR #2906, opened 2025-10-05
Move dialogue to cutscene.h

ProjectOblivion:
I also removed any need for Dialogue in the overlay .h files and eliminated bss files which only contained cutscene variables.


ProjectOblivion:
To be clear, the overall readability and organization of this file is terrible.  I don't think these changes are what we want as final changes for the codebase ultimately, but I think it's a necessary evil to move toward better.
This PR is likely already painful enough to review, trying to fix everything in one go would be horrendous.

PR #2907, opened 2025-10-06
Decompile multiple `sel_psp` functions

gamezter:
This is still a work in progress. So far PSP shows that the PSX file splits are wrong, hopefully adjusting the PSX file splits should help with the werid RODATA issues we've met in that overlay

### DIFF
--- a/src/st/sel_psp/cutscene.c
+++ b/src/st/sel_psp/cutscene.c
@@ -2,6 +2,7 @@
 #include "../sel/sel.h"
 #include <cutscene.h>
 
+extern Dialogue g_Dialogue;
 extern u16 D_psp_092855F0[];
 extern u8 D_psp_09285678[];
 extern u16 D_psp_09285688[];
--- a/src/st/top/cutscene.c
+++ b/src/st/top/cutscene.c
@@ -24,7 +24,6 @@ INCLUDE_RODATA("st/top/nonmatchings/cutscene", D_us_801A89A0);
 
 extern u8 OVL_EXPORT(cutscene_data)[];
 extern bool g_IsCutsceneDone;
-extern Dialogue g_Dialogue;
 extern u32 g_CutsceneFlags;
 extern u8 D_us_80180D38[];
 extern u8 D_us_80180D3C[];


src/st/sel_psp/7D40.c
```diff
@@ -0,0 +1,391 @@
...
+
+void func_psp_0923FE40(char* arg0, s32 arg1) {
+    if (D_psp_09285D08[arg1] != arg0) {
+        func_psp_0923FC08(arg1);
+        D_psp_09285D08[arg1] = arg0;
+        func_psp_0923FCC8(arg0, arg1);
+    }
+}
+
+INCLUDE_ASM("st/sel_psp/nonmatchings/sel_psp/7D40", func_psp_0923FEC0);
```
gamezter, line 200: This has been matched by @Mc-muffin but it uses inline assembly: https://decomp.me/scratch/e8Unw . Would that be fine for this function ?

src/st/sel_psp/cutscene.c
```diff
@@ -0,0 +1,1062 @@
...
+extern u16 D_psp_09285688[];
+extern u8 D_psp_09285710[];
+extern u16 D_psp_09285720[];
+extern u8 D_psp_092857B8[];
+extern u16 D_psp_092857C8[];
+extern u8 D_psp_09285850[];
+extern u16 D_psp_09285860[];
+extern u8 D_psp_092858E8[];
+extern u8 D_psp_09286868[];
+extern Dialogue4 D_psp_09287E60;
```
gamezter, line 17: sel_psp uses 2 different kinds of `g_Dialogue`-like objects.. There's the original `g_Dialogue` which uses `Dialogue`
and `D_psp_09287E60` which uses `Dialogue4` (2 and 3 already exist in `game.h`). I think it may be used for the credits screen ?

src/st/sel_psp/cutscene.c
```diff
@@ -0,0 +1,1062 @@
...
+    /* 0x0A */ s16 nextCharX;         // next char x coord
+    /* 0x0C */ s16 nextLineX;         // next line x coord
+    /* 0x0E */ u16 nextCharY;         // next char y coord
+    /* 0x10 */ u16 portraitAnimTimer; // portrait animation timer
+    /* 0x12 */ u16 unk12;             // unknown
+    /* 0x14 */ u16 clutIndex;         // CLUT index
+    /* 0x16 */ u16 nextCharTimer;     // timer to next character
+    /* 0x18 */ Primitive* prim;       // for dialogue graphics rendering
+    /* 0x1C */ s32 primIndex;         // primIndex: unk
+} Dialogue4;                          // size = 0x20
```
ProjectOblivion: I don't see an immediate way to avoid a new Dialogue struct, but add it to cutscene.h instead of here if it ends up being necessary.
I'm looking at this situation, starting with #2906 .

src/st/sel_psp/7D40.c
JoshSchreuder: I had a bit of a poke around and it doesn't look like MWCC has an equivalent built-in like `__builtin_trap();`, `abort()`, `exit()` etc.  so IMO importing like this would be fine for now, but maybe add a TODO on it in case we can find something in future?

Xeeynamo:
Conflicting after https://github.com/Xeeynamo/sotn-decomp/pull/2906 got merged, but looks good

PR #2909, opened 2025-10-08
Refactor `sel`

gamezter:
Depends on https://github.com/Xeeynamo/sotn-decomp/pull/2907

Sorry for the huge PR...

- Adjust `sel` file splits based off `sel_psp`
- Make `sel` and `sel_psp` match closer by renaming `sel_psp` symbols and adjusting `sel` code structure
- Move some  `sel` .bss symbols to their appropriate file (up to the first `common` bss symbol `g_SkipCutscene`)
- Import some `sel_psp` data

IMO the most *controversial* change in this PR is that I moved a few symbols from `symbols.pspeu.dra.txt` to `symbols.pspeu.txt` since `sel_bss` uses symbols from `dra`

I'm still lost on how to use a `common` bss segment, we spoke about this a [few months ago](https://discord.com/channels/1079389589950705684/1079395288810274857/1384160397648461834) but idk how to make it work with splat


Xeeynamo:
This is conflicting too

Xeeynamo (approved):
Great work!!

Issue #2780, opened 2025-08-01
Investigate `-char signed` on MWCC


sozud:
See https://github.com/Xeeynamo/sotn-decomp/pull/2778#discussion_r2246612038

Xeeynamo:
I started investigating on this. It seems the PS1 version is confirmed to use `-funsigned-char`. All the PS1 SDK headers using `char` do not match with `-fsigned-char`. `MemcardDetectSave` is the primary source of my claim.

Xeeynamo:
`StoreSaveData` on PSP with `-char signed` doesn't fully match either, and requires various casts from `u8*` to `char*` here and there mainly due to `char saveTitle[64]`. I am unsure how to proceed from here.

PR #2910, opened 2025-10-12
Fix mis-alignment between signed and unsigned char

Xeeynamo:
`char` is `unsigned` in our decomp. This forces all compilers to do the same.


sozud:
Are we sure it's the same on psp? https://github.com/Xeeynamo/sotn-decomp/issues/2780

Xeeynamo:
> Are we sure it's the same on psp? #2780

I am unsure about PSP. But I know it mis-matches with PS1. Ideally we want to switch `char` as `signed`, but as you probably know the change is much larger and out of scope for this PR. This change just allows to have menus working again.

PR #2911, opened 2025-10-13
Documented and aligned e_stage_name

ProjectOblivion:
- Documented all three variations of `e_stage_name`
- Moved `gfxBank[]`, `stage_name_gfx[]`, and `LoadStageNameGraphics()` to `e_stage_name_psp.h`
- Adjusted code in places to better align shared code between variations.  They are too different to merge, but a diff between variations should match shared code more accurately
- Renamed `e_stage_name.h` to `e_stage_name_us.h` and `entity_stage_name_popup_jp.h` to `e_stage_name_jp.h` for consistency and association
- Moved the `STAGE_FLAG` definitions to their respective e_stage_name.c files or removed them if they were unnecessary


Xeeynamo:
> Renamed entity_stage_name_popup_jp.h to e_stage_name_psp.h for consistency and association

Keep in mind this might be reverted if we find `e_stage_name_psp.h` is identical to the Japanese PS1 game builds

ProjectOblivion:
My primary issue was that it was too disconnected from `e_stage_name` and our current conventions, though I'm not sure naming these after a version is the best choice since `pspeu` uses two distinct variations and `jp` will potentially follow the same pattern.

In retrospect, I shouldn’t have even named it that to begin with, by my own logic.

PR #2914, opened 2025-10-14
Import `sel_psp` data

gamezter:


src/st/sel_psp/credits_data.h
```diff
@@ -0,0 +1,202 @@
...
+SET_WAIT(9),
+'-','E','x','e','c','u','t','i','v','e',' ','P','r','o','d','u','c','e','r','-',
+SET_WAIT(19),
+'K','A','Z','U','M','I',' ','K','I','T','A','U','E',
+0x1,0x0,
+SET_WAIT(0),
+SET_WAIT(0),
+0x1,0x0,
+0x1,0x3,
+0x9F,0x54,0x82,0x50,0x82,0x58,0x82,0x58,0x82,0x56,' ','K','O','N','A','M','I',
```
gamezter, line 189: Unfortunately the cutscene data importer isnt able to import `credits_data` because it is slightly different from typical cutscene data. Also `0x9F,0x54,0x82,0x50,0x82,0x58,0x82,0x58,0x82,0x56` is "1997" using shift-jis double-wide characters but it seems that somewhere in the build pipeline characters inside .h files aren't considered shift-jis

PR #2915, opened 2025-10-15
Documented `dai` `e_breakable` and cleaned up breakable exts

ProjectOblivion:
While documenting `dai` `e_breakable`, I found that a few of the breakable exts shared a common structure.
Most of them shared a common `breakable` ext, but there were 3 exceptions:
- `DAI_EntityBreakable`, `CHI_EntityBreakable`, and `chi` `EntityBreakableWall` use a common `breakableDebris` ext
- `NO2` is weird and has a prim at `0x84`, unique from the others
- `NO1` is weird and has a `u8` at `0x84`, unique from the others

NO1 and NO2 would have used the common breakable ext if it weren't for those differences, which just further supports that these were likely contained within the files themselves and copied around.

I combined the exts and removed `breakableWall`, `breakableWallDebris`, `breakableWall2`, and `breakableTerrain`.  I also grouped the four remaining breakable exts together.

The breakable wall and floor in NZ0 use the breakable ext structure, so I changed them, but the previous struct was used by much more than just the breakables so I left the previous ext as is.


src/st/dai/e_breakable.c
```diff
@@ -40,14 +86,18 @@ static u8 draw_modes[] = {
     DRAW_TPAGE | DRAW_TPAGE2,
     DRAW_DEFAULT,
     DRAW_DEFAULT};
-static u16 hitbox_offsets[] = {0, 0, -24, -16, 0, 0, 0, 0, 0, 0, 0, 0};
-static s16 debris_offsets[] = {0, 1, 2, 2, 3, 0, 1, 2, 3, 0};
+static u16 hitbox_offsets_y[] = {0, 0, -24, -16, 0, 0, 0, 0, 0, 0, 0, 0};
+/* End of arrays indexed by breakableType */
```
ProjectOblivion, line 90: Is there any way to make this more readable?
I tried using a struct like `ObjInit`/`BackgroundBlockInit`, but I couldn't get it to work.

Xeeynamo: it's fine to keep it like that for now

PR #2919, opened 2025-10-15
Documented `dai` `e_bone_halberd`

ProjectOblivion:


src/st/dai/e_bone_halberd.c
```diff
@@ -1,19 +1,49 @@
...
-static u8 anim_4[] = {
-    5, 1,  5, 2,  5, 7,  6, 8,  2, 9,  2, 10, 2, 11, 2, 12, 2, 13, 2,  10,
-    2, 11, 2, 12, 2, 13, 3, 14, 5, 15, 5, 16, 5, 17, 1, 18, 2, 19, 20, 20,
-    5, 21, 3, 22, 2, 23, 2, 24, 2, 25, 1, 26, 1, 27, 1, 28, 2, 29, -1, 0};
-static u8 anim_5[] = {1, 1, 4, 30, 4, 31, 1, 1, -1, 0, 0, 0};
-static u8 anim_6[] = {1, 1, 4, 30, 6, 31, 4, 30, 1, 1, -1, 0};
-static u16 rotation_interval[] = {256, 128, 72, 32, 64, 16, 32, -32, 32, 0};
-static u8 D_us_80181DD8[] = {
-    48, 32, 20, 12, 24, 16, 64, 48, 48, 0, 0, 0,
+#define EXPLOSION_INIT 0
```
Xeeynamo, line 4: Since `0` is always the init step for every entity, maybe we should make it more generic? We can do it in another PR

ProjectOblivion: I was actually going to do that originally, lol.  I couldn't come up with a compelling reason for having it separate when the enums were still ultimately needed.  I probably went a bit excessive with the namespace isolation, erring on the side of explicit to distinguish between the main entity and the explosion

I did some comparisons of the entities and step init and substep init were always consistent, an idle/ready step was common, and that's about it.

An ENTITY_INIT and SUBSTEP_INIT would probably be useful as global macros/enums somewhere.

Xeeynamo: We could either have a simple `#define something 0` in `game.h` that can be used instead of `EXPLOSION_INIT` ad-hoc and similar. Or frankly even a `step = 0;` is clear enough IMHO.

For sure the option I'd not suggest is to export the Explosion enum to a common header, just to use the first enum field.

Xeeynamo: Merging for now as this matter is trivial

PR #2930, opened 2025-10-20
More sfxIDs added to enum

synamaxmusic:
0x618 looks like it might be a command related to sequenced music?


src/dra/42398.c
```diff
@@ -637,7 +637,7 @@ s32 LoadVabData(void) {
     LOAD_VAB(1, aPbav_0, g_VabAddrs[1], D_8017D350, vab1Len);
     LOAD_VAB(2, aPbav_1, g_VabAddrs[2], D_801A9C80, vab2Len);
     LOAD_VAB(3, aPbav_2, g_VabAddrs[3], D_8018B4E0, vab3Len);
-    func_80131EBC(aPqes, 0x618);
+    func_80131EBC(aPqes, SFX_UNK_618);
```
sozud: As far as I can tell this function is some sort of leftover unused thing and the 0x618 here doesn't represent anything in the current build of the game, so I think this should be left as 0x618

PR #2931, opened 2025-10-21
Decompile `cat` `e_hellfire_beast`

JoshSchreuder:
Codeshared with PSP

`func_us_801CBE14` / `SpawnDeathParts`
// PSX: https://decomp.me/scratch/8sHZB
// PSP: https://decomp.me/scratch/cPuWT

`func_us_801CC2E4` / `EntityHellfireBeast`
// PSX: https://decomp.me/scratch/cIdhf
// PSP: https://decomp.me/scratch/WALGe

`func_us_801CCEF0` / `EntityHellfireBeastThorsHammer`
// PSX: https://decomp.me/scratch/qauph
// PSP: https://decomp.me/scratch/VNrDM

`func_us_801CD614`
// PSX: https://decomp.me/scratch/6lBex 
// PSP: https://decomp.me/scratch/YTUfV

`func_us_801CD728` 
// PSX: https://decomp.me/scratch/9JeKm
// PSP: https://decomp.me/scratch/9liyf

`func_us_801CDB50` / `EntityHellfireBeastFlamePillar`
// PSX: https://decomp.me/scratch/cs5LW
// PSP: https://decomp.me/scratch/BS3VV

`func_us_801CE170` / `EntityHellfireBeastHitbox`
// PSX: https://decomp.me/scratch/xlzVi
// PSP: https://decomp.me/scratch/NZa4Y

### DIFF
--- a/include/sfx.h
+++ b/include/sfx.h
@@ -251,7 +251,6 @@ enum {
 #define JP_VO_SH_SONO_TEIDO 0x530 // Shaft: Sono teido no chikara de tatakai...
 #endif
 
-#define NA_SE_EN_SPITTLEBONE_ACID_SPLAT 0x73C
 #define NA_SE_VO_AXE_KNIGHT_THROW 0x766
 #define NA_SE_VO_AXE_KNIGHT_SCREAM 0x767
 #define NA_SE_EV_WATER_SPLASH 0x7C2
@@ -274,11 +273,7 @@ enum {
 #define BIBLE_SUBWPN_SWOOSH 0x8C3
 
 // STAGE CAT
-#define SFX_LOSSOTH_NAPALM_GRUNT 0x734
-#define SFX_LOSSOTH_DEATH 0x735
-#define SFX_GRAVE_KEEPER_GRAAH 0x754
-#define SFX_GRAVE_KEEPER_HIYAH 0x755
-#define SFX_GRAVE_KEEPER_DEATH 0x756
+
 #define SFX_DISCUS_LORD_HEY 0x763
 #define SFX_DISCUS_LORD_DEATH_VORTEX 0x764
 #define SFX_DISCUS_BUZZ 0x765
@@ -314,8 +309,6 @@ enum {
 #define SE_CASTLE_GATE_RISE 0x7A5
 
 // STAGE NO4
-#define SFX_TOAD_CROAK 0x71A
-#define SFX_FROG_CROAK 0x71B
 #define SFX_OAR_ROW 0x7BF
 #define SFX_WATER_BUBBLE 0x7C4
 
@@ -353,6 +346,12 @@ enum {
 
 // The VAB IDs appear in large chunks so all sounds proceeding
 // a vabid label comment will belong in that VAB group unless noted.
+//
+// Some enemy sfxIDs in vabID 3 can sound different depending on which
+// overlay is loaded.  For example, Grave Keeper has three sound calls
+// in ARE and CAT, but the VH files are different between them so 0x754
+// plays a different sample in ARE.  0x709 is another example; the pitch
+// is higher in NP3, but normal for RCHI and NZ0.
 
 #ifdef VERSION_BETA
 // MAD uses an earlier build and has different sfx IDs
@@ -644,7 +643,7 @@ enum Sfx {
     /* 0x708 */ SFX_RIC_HYDRO_STORM_ATTACK,
 
     // vabid 3
-    /* 0x709 */ SFX_SLOGRA_ROAR,
+    /* 0x709 */ SFX_SLOGRA_ROAR, // different pitch in NP3
     /* 0x70A */ SFX_SLOGRA_ROAR_DEFEAT,
     /* 0x70B */ SFX_SLOGRA_PAIN_A, // Used for Gaibon
     /* 0x70C */ SFX_SLOGRA_PAIN_B,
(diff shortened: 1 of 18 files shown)


src/st/cat/e_hellfire_beast.c
```diff
@@ -0,0 +1,1189 @@
...
+            if (dx > 0x1FF) {
+                dx = 0x1FF;
+            }
+            self->hitboxOffX = -dx / 2;
+            self->hitboxWidth = dx / 2;
+            self->hitboxHeight = 8;
+            self->hitboxOffY = 8;
+
+            // FAKE! PSX requires some trickery here, was not able to resolve
+#ifdef VERSION_US
```
JoshSchreuder, line 820: Thanks to @Mc-muffin for the solve here - I couldn't work out a way to get rid of this on PSX side and make it clean like PSP unfortunately so have marked this as fake for now.

src/st/cat/e_hellfire_beast.c
```diff
@@ -0,0 +1,1189 @@
...
+    case 3:
+        prim = self->ext.hellfireBeast.primThree;
+        prim->drawMode = DRAW_DEFAULT;
+
+        prim = prim->next;
+        prim->drawMode = DRAW_UNK_800;
+
+        // NB. Strange way to write a single loop iteration...
+        // Maybe the number of iterations changed during development?
+        for (i = 0; i <= 0; i++) {
```
JoshSchreuder, line 1138: This is a bit weird but I tried rewriting it in other ways for a single iteration like `do { } while(0)` and it changed the asm. This works at least

src/st/cat/e_hellfire_beast.c
```diff
@@ -0,0 +1,1189 @@
...
+            } else if (GetDistanceToPlayerX() < 64) {
+                // If player is close, use slide punch attack
+                SetStep(PUNCH_ATTACK);
+            } else {
+                // Otherwise prep lightning / flame pillar attack
+                SetStep(ATTACK_INIT);
+            }
+        }
+        break;
+    case UNUSED_STUTTER:
```
JoshSchreuder, line 318: This case appears unused - maybe it was intended as a walk animation at some point but cut
duckstation-qt-x64-ReleaseLTCG_oHQHVi2I2J

src/st/cat/e_hellfire_beast.c
```diff
@@ -0,0 +1,1189 @@
...
+    0x04, 0x24, 0x04, 0x25, 0x06, 0x26, 0x02, 0x27, 0x03, 0x28,
+    0x02, 0x29, 0x10, 0x2A, 0x05, 0x25, 0xFF, 0x00, 0x00, 0x00};
+static u8 anim_punch_hard_stop[] = {
+    0x01, 0x22, 0x03, 0x21, 0x03, 0x20, 0x05, 0x1F, 0x04, 0x19, 0xFF, 0x00};
+static u8 anim_unused_stutter[] = {
+    0x04, 0x24, 0x02, 0x25, 0x02, 0x27, 0x04, 0x19, 0xFF, 0x00, 0x00, 0x00};
+static u8 anim_raise_arms_attack[] = {
+    0x27, 0x19, 0x02, 0x24, 0x02, 0x25, 0x02, 0x28, 0x03, 0x33,
+    0x02, 0x34, 0x02, 0x35, 0x02, 0x34, 0xFF, 0x00, 0x00, 0x00};
+static u8 anim_attack_cast_wiggle[] = {
```
ProjectOblivion: These should be AnimateEntityFrame.  Looks like all of them are terminated with POSE_END except anim_idle_breathe and cast_wiggle, which would be POSE_LOOP(0).

It looks like all of these animations are probably padded for alignment, so the pair of trailing zeros shouldn't be needed.

src/st/cat/e_hellfire_beast.c
```diff
@@ -0,0 +1,1189 @@
...
+    0x01, 0x34, 0x01, 0x36, 0x00, 0x00, 0x00, 0x00};
+
+static Point16 death_parts_positions[6] = {
+    {.x = -32, .y = -26}, {.x = -1, .y = -34}, {.x = -21, .y = -44},
+    {.x = -16, .y = -24}, {.x = -30, .y = 1},  {.x = -13, .y = 1}};
+// Indexed off self->params
+static s16 death_parts_timer[6] = {32, 40, 64, 104, 96, 100};
+// Sets sVec vy for each of 8 death part prims
+static s16 death_parts_yOffsets[] = {
+    0x0000, 0x0200, 0x0380, 0x0600, 0xF800, 0xFE00, 0xF280, 0xFA00};
```
ProjectOblivion: Some of these should be negative?

src/st/cat/e_hellfire_beast.c
```diff
@@ -0,0 +1,1189 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "cat.h"
+
+// Used to set prim->r0,g0,b0 for the Thor's Hammer / Flame Pillar casts
+static u8 attack_rgbs[2][3] = {{2, 1, 7}, {7, 4, 1}};
+// EntityHellfireBeastThorsHammer, EntityHellfireBeastFlamePillar
+static u16 attack_cluts[] = {PAL_DRA(0x16C), PAL_DRA(0x16D)};
```
ProjectOblivion: It doesn't look like CAT has the palette enum, but aren't we trying to get away from using PAL_DRA/PAL_OVL?

src/st/cat/e_hellfire_beast.c
```diff
@@ -0,0 +1,1189 @@
...
+                    }
+                }
+                prim = prim->next;
+            }
+        }
+        break;
+    }
+}
+
+extern EInit g_EInitHellfireBeast;
```
ProjectOblivion: Are we externing these in the overlay .h file or in the .c file as current practice?

src/st/cat/e_hellfire_beast.c
```diff
@@ -0,0 +1,1189 @@
...
+            dy = posY - prim->y2;
+            if (!self->ext.hellfireBeastThorsHammer.unk88) {
+                self->ext.hellfireBeastThorsHammer.unk88 = 4;
+                angle = ratan2(-dy, dx);
+                var_s2 = angle - var_s4;
+                if (var_s2 > 0x800) {
+                    var_s2 = 0x1000 - var_s2;
+                }
+                if (var_s2 < -0x800) {
+                    var_s2 = 0x1000 + var_s2;
```
ProjectOblivion: Since these are angle values, they should be either FLT or ROT.
I'll be renaming ROT to be more indicative of degrees rather than rotation at some point, but these would be ROT(180) and ROT(360) for 0x800 and 0x1000 respectively.

src/st/cat/e_hellfire_beast.c
```diff
@@ -0,0 +1,1189 @@
...
+        }
+        break;
+    case 4:
+        DestroyEntity(self);
+        break;
+    }
+}
+
+// Related entity for the flame pillar
+void func_us_801CD614(Entity* self) {
```
ProjectOblivion: Since you seem to know what this is, should it be renamed so that you can align e_init.c, the E_ID, and the file?
It looks like this is likely the animation and func_us_801CD728 looks to be the prims for the pillar.

src/st/cat/e_hellfire_beast.c
```diff
@@ -0,0 +1,1189 @@
...
+    if (self->poseTimer > 0xF) {
+        DestroyEntity(self);
+    }
+
+#ifdef VERSION_PSP
+    if (self->ext.hellfireBeastThorsHammer.entity != NULL &&
+        !self->ext.hellfireBeastThorsHammer.entity->entityId) {
+#else
+    if (!self->ext.hellfireBeastThorsHammer.entity->entityId) {
+#endif
```
ProjectOblivion, line 888: This can probably be split up into two if statements for psp, not sure.
That would let you ifdef only the first psp if statement, but you would need to ifdef a closing brace so I'm not sure which is the better choice for readability.

src/st/cat/e_hellfire_beast.c
```diff
@@ -0,0 +1,1189 @@
...
+            if (self->ext.hellfireBeastFlamePillar.unk80 < 0) {
+                DestroyEntity(self);
+            }
+        }
+        break;
+    }
+}
+
+extern EInit g_EInitHellfireBeastHitbox;
+void EntityHellfireBeastHitbox(Entity* self) {
```
ProjectOblivion: This isn't the Hellfire Beast hitbox, it is the hitbox for his punch attack.

src/st/cat/e_hellfire_beast.c
Mc-muffin: Worth noting that the `if (0)` has no side-effects in PSP, the special bit is the intermediary variable for `drawMode`

src/st/cat/e_hellfire_beast.c
JoshSchreuder: I don't have strong feelings either way but I think my preference is to have these live with the code that's referencing the extern

src/st/cat/e_hellfire_beast.c
JoshSchreuder: Yeah I considered going down that path but went with this for the reason you mentioned, I think it is slightly easier to read with just one conditional

src/st/cat/e_hellfire_beast.c
ProjectOblivion: I don't have strong feelings either, but I feel the same way.  It's also more in line with the larger direction of putting "things" close to where they're used.

I do think at the top of the file is better, though.  I don't know about anybody else, but I find it somewhat difficult to see these kinds of things when there mixed in with the functions.

include/entity.h
```diff
@@ -3522,6 +3522,45 @@ typedef struct {
...
+    /* 0x98 */ u32 : 32;
+} ET_HellfireBeastThorsHammer;
+
+typedef struct {
+    /* 0x7C */ struct Primitive* prim;
+    /* 0x80 */ s16 unk80;
+    /* 0x82 */ s16 : 16;
+    /* 0x84 */ s16 castTimer;
+} ET_HellfireBeastFlamePillar;
+
```
ProjectOblivion, line 3563: My suggestion was something like this, though I don't know if it is a _good_ suggestion.
```
typedef struct {
    /* 0x7C */ struct Primitive* prim;
    /* 0x80 */ union {
        struct Primitive* primTwo;
        struct Entity* entity;
        s16 unk80;
    } offset80;
    /* 0x84 */ s16 timer;
    /* 0x86 */ s16 deathPartsRotate; // unk86
    /* 0x88 */ union {
        u8 attackChoice;
        s16 unk88;
    } offset88;
    /* 0x89 */ u8 : 8;
    /* 0x8A */ s16 unk8A;
    /* 0x8C */ union {
        struct Entity* attackEntity;
        s32 unk8C;
    } offset8C;
    /* 0x90 */ s32 unk90;
    /* 0x94 */ u8 unk94;
    /* 0x95 */ u8 : 8;
    /* 0x96 */ u16 : 16;
    /* 0x98 */ u32 : 32;
    /* 0x9C */ u32 : 32;
    /* 0xA0 */ u32 : 32;
    /* 0xA4 */ struct Primitive* primThree;
} ET_HellfireBeast;
```

I think it's notable that the offset spacing aligns in such a way that this is plausible.

JoshSchreuder: @sozud @Xeeynamo any thoughts on this approach?

JoshSchreuder: @ProjectOblivion worth noting I don't think this would work for `ET_HellfireBeastFlamePillar` which has
```
    /* 0x80 */ s16 unk80;
    /* 0x82 */ s16 : 16;
```
?

src/st/cat/e_hellfire_beast.c
JoshSchreuder: Thanks, I meant to do that for these animations added since your enum but forgot

include/entity.h
ProjectOblivion: Union uses the largest size of any of the elements, so it should fill up both of those slots (I think).

Xeeynamo: I think it's safer and more readable to keep the two ET Ext separate, despite _some_ fields aligning

src/st/cat/e_hellfire_beast.c
Xeeynamo: Yes, I like the palette approach in DAI: https://github.com/Xeeynamo/sotn-decomp/blob/master/src/st/dai/dai.h#L31 , it would be nice to have the same pattern applied on all overlays.

I suspect the 0x100-0x1FF range should be in `dra.h`, we could always move it there in the future though. Let's just have these `0x16C` and `0x16D` in `cat.h`.

Yes, `PAL_DRA` and `PAL_OVL` were wrong assumptions. It is fairly straightforward to make a regex with a Find&Replace All to remove them any time, I am not worried to have them committed for the time being.

PR #2932, opened 2025-10-23
Import `main_psp`

gamezter:
`main_psp` currently doesn't match, I was hoping making a draft PR could get more eyes on this and help figuring out any issues.

I used `textbin` to include functions that use unrecognized opcodes that make the assembler complain, I had to modify `builds/gen.py` to get those `textbin` linked in.

I also hacked up code to get the `elf` header to be included, but I have no idea if the code actually works. It should be rewritten by someone who knows what they are doing.

There is also the issue that splat / spimdism is mistakingly labeling jump labels as functions, I partially fixed that issue by forcing a size on some functions that were split up.

EDIT:
A lot has happened since the initial commit:

- We still have the issue that `allegrex-as` doesn't recognize some opcodes, but that has been patched for now with `pspas.py`. It will replace unknown opcodes with `.word 0xXXXXXXXX`.
- Some of the functions seem to not be compiled with MWCC (ex: https://decomp.me/scratch/7dPzE). Those would cause exceptions in `mwccgap` so I replaced those functions with `asm` segments
- The build will generate `build/pspeu/main.elf` and `build/pspeu/main.exe` and IDK why. Comparing each of them to `PS.ELF` shows that `main.exe` is a lot more similar. So far most of the binary matches except symbols that reference data past `RODATA` because I currently have the issue that some symbols get stripped out. For instance `.lib.ent.top` is a symbol used by `PS.ELF` header to mark the top of the `.lib.ent` section but it gets stripped out and offsets all symbols that come after.


hohle:
Thank you for starting this. It's been on my backlog for a long time. I found some other examples of elf files, primarily from PS2 decomps.

Xeeynamo:
I am parking here some of my progress using this PR as a base: https://github.com/Xeeynamo/sotn-decomp/tree/psp-main-testing

let me know if you want me to merge this to your PR. Or feel free to cherry-pick it

gamezter:
Please merge it :)

gamezter:
> I am parking here some of my progress using this PR as a base: https://github.com/Xeeynamo/sotn-decomp/tree/psp-main-testing
> 
> let me know if you want me to merge this to your PR. Or feel free to cherry-pick it

I figured out how to merge your branch in myself. Thanks for the help!

config/symbols.pspeu.main.txt
```diff
@@ -0,0 +1,210 @@
...
+sceUmdCheckMedium = 0x0893A258;
+sceUmdWaitDriveStatCB = 0x0893A260;
+sceUmdCancelWaitDriveStat = 0x0893A268;
+sceUmdGetDriveStat = 0x0893A270;
+sceUmdReplaceProhibit = 0x0893A278;
+sceUmdWaitDriveStat = 0x0893A280;
+sceUmdRegisterUMDCallBack = 0x0893A288;
+sceUmdUnRegisterUMDCallBack = 0x0893A290;
+sceUmdActivate = 0x0893A298;
+sceUmdDeactivate = 0x0893A2A0;
```
gamezter, line 420: These were found by pairing up the NIDs found in the `.lib.stub` section with NIDs in this repo : https://github.com/uofw/uofw

From what I understand NIDs are a hash of the function signature so that kernel functions can be dynamically linked ?

src/main_psp/36174.c
```diff
@@ -0,0 +1,610 @@
...
+
+void sceIoLseekAsync(void) {}
+
+void sceIoRename(void) {}
+
+void sceIoClose(void) {}
+
+void sceIoOpenAsync(void) {}
+
+void sceIoReadAsync(void) {}
```
gamezter: These probably get linked at runtime by this decompiled function : https://github.com/uofw/uofw/blob/master/src/kd/loadcore/loadcore.c#L2078

hohle:
Do you think it would make sense to have a symbol file per segment? There are things like the SCE symbols and the Load function symbols that may be better off separate from everything else.

I pulled the `OVL_EXPORT(Load)` symbols from PS a while back. I don't think they can go into a global symbol table since they'll conflict across overlays - https://www.sotn.fun/wiki/Ports/PSP

PR #2933, opened 2025-10-24
Codeshared `cen` and `chi` `e_breakable`

ProjectOblivion:


src/st/cen/cen.h
```diff
@@ -9,6 +9,11 @@
 
 #define OVL_EXPORT(x) CEN_##x
 
+enum OVL_EXPORT(Palette) {
+    PAL_NONE = 0,
+    PAL_BREAKABLE_DEBRIS = 0x159,
+};
+
```
hohle, line 16: Won't all `OVL_DRA` palettes (anything without the top bit set) be shared? This could be moved to `include` since I think it's universal.

src/st/chi_psp/dt_entity.c
```diff
@@ -25,7 +25,7 @@ void EntityDemonSwitchWall(Entity*);
 void EntityDemonSwitch(Entity*);
 void EntityBreakableWall(Entity*);
 void EntityBreakableWallDebris(Entity*);
-void EntityBreakableDebris(Entity*);
+void OVL_EXPORT(EntityBreakableDebris)(Entity*);
 void EntityFallingStairs(Entity*);
 void EntityCerberusGateDebug(Entity*);
 void EntityFallingStep(Entity*);
```
hohle: This file is identical to `src/st/chi/dt_entity.c`.

ProjectOblivion: I have no idea why the file is duplicated in master, but I was only dealing with e_breakable so I didn't look into it.

src/st/cen/cen.h
ProjectOblivion: No, not all non-8XXX palettes will be shared, only at most 0x100-0x1FF.  That being said, there hasn't been a word either way.  I'm not against adjusting this however is best, the enum is just the last word I've gotten..

src/st/e_breakable_with_lighting.h
```diff
@@ -0,0 +1,244 @@
...
+                prim->v0 = prim->v1 = 0;
+                prim->v2 = prim->v3 = 32;
+
+                prim->next->x1 = self->posX.i.hi + 4;
+                prim->next->y0 = self->posY.i.hi - 8;
+                LOH(prim->next->r2) = 32;
+                LOH(prim->next->b2) = 32;
+                prim->next->b3 = 16;
+
+                prim->priority = BREAKABLE_PRIORITY;
```
Xeeynamo:
```suggestion
                prim->priority = BREAKABLE_ZPRIORITY + 2;
```

minor thing: can we remove `BREAKABLE_PRIORITY` and opt for `BREAKABLE_ZPRIORITY + 2`? So there's less mental strain when someone wants to understand the code

Xeeynamo (approved):

src/st/e_breakable_with_lighting.h
ProjectOblivion: That's not a problem.  I struggle with knowing when to combine and when to not combine things like this, so that feedback is helpful.

PR #2934, opened 2025-10-24
Documented `lib` `e_breakable`

ProjectOblivion:


src/st/lib/e_breakable.c
```diff
@@ -1,114 +1,156 @@
...
-    PAL_DRA(0x26C), PAL_DRA(0x26C), PAL_DRA(0x26C), PAL_DRA(0x26C),
-    PAL_DRA(0x26C), PAL_DRA(0x26C)};
-static u16 D_us_80180BFC[] = {
-    ANIMSET_OVL(0x1), ANIMSET_OVL(0x1), ANIMSET_OVL(0x1), ANIMSET_OVL(0x1),
-    ANIMSET_OVL(0xC), ANIMSET_OVL(0xC), ANIMSET_OVL(0xC), ANIMSET_OVL(0xC),
-    ANIMSET_OVL(0xC), ANIMSET_OVL(0xC)};
-static u8 D_us_80180C10[] = {
+#define BREAKABLE_ANIMSET ANIMSET_OVL(12)
+
+enum OVL_EXPORT(BreakableTypes) {
```
ProjectOblivion, line 6: Is this deconfliction necessary?

hohle: It shouldn't be. `enum`s are only a compile time abstraction. they aren't first-class types that have any link-time or run-time information.

The other definition is in `dai`, so by the time these would be linked together in any way, the enums would long since be integers with no other type information.

PR #2935, opened 2025-10-24
Decompiled, imported, and merged `cat` `e_breakable`

ProjectOblivion:


src/st/cat/e_breakable.c
```diff
@@ -1,4 +1,147 @@
...
+        self->drawMode = draw_modes[breakableType];
+        self->hitboxHeight = hitbox_heights[breakableType];
+        self->animSet = anim_sets[breakableType];
+        self->unk5A = unk_5A[breakableType];
+        self->palette = palettes[breakableType];
+#ifdef INVERTED_STAGE
+        self->hitboxOffY = -hitbox_offsets_y[breakableType];
+#else
+        self->hitboxOffY = hitbox_offsets_y[breakableType];
+#endif
```
hohle: This doesn't seem relevant for a single overlay implementation.

src/st/cat/e_breakable.c
```diff
@@ -1,4 +1,147 @@
...
+        case BUST:
+            g_api.PlaySfx(SFX_GLASS_BREAK_E);
+            entity = AllocEntity(&g_Entities[160], &g_Entities[192]);
+            if (entity != NULL) {
+                CreateEntityFromEntity(E_HEART_DROP, self, entity);
+                entity->params = self->params & 0x1FF;
+            }
+            PreventEntityFromRespawning(self);
+            DestroyEntity(self);
+            return;
```
hohle, line 114: These seems bizarre enough to warrant a comment. As written, the sfx will never play.

ProjectOblivion: I believe this may have been a copy paste mistake on my part.  I'll verify it and update it.

src/st/cat/e_breakable.c
ProjectOblivion: RCAT will share this code, which is why I left it in.

ProjectOblivion: I went ahead and removed it for now, easy to add back later.

Xeeynamo: makes me wonder how to handle this in the future. I think having a `INVERTED_STAGE` def is a clever solution though

Xeeynamo (approved):

PR #2939, opened 2025-10-25
Codeshared `e_breakable` with debris

ProjectOblivion:


src/st/e_breakable_with_debris.h
```diff
@@ -0,0 +1,232 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+enum OVL_EXPORT(BreakableTypes) {
```
Xeeynamo: Is `OVL_EXPORT` necessary on enums if they are local to the entity source file? Enums have no exported symbols once the C file is compiled

src/st/e_breakable_with_debris.h
```diff
@@ -0,0 +1,232 @@
...
+            PreventEntityFromRespawning(self);
+            DestroyEntity(self);
+            return;
+        case JUG:
+            g_api.PlaySfx(SFX_GLASS_BREAK_E);
+            entity = AllocEntity(&g_Entities[160], &g_Entities[192]);
+            if (entity != NULL) {
+                CreateEntityFromEntity(E_HEART_DROP, self, entity);
+#ifdef JUG_PARAMS
+                entity->params = JUG_PARAMS;
```
Xeeynamo: What concerns me about this or similar approaches for code sharing is how the alternative code paths are expressed and documented. When I see `JUG_PARAMS`, it doesn’t communicate _why_ `entity->params` is assigned differently, nor does it indicate where that value comes from or what it represents. To understand it, I have to navigate to its definition (if the macro definition is unique) or perform a full search across the codebase, which makes the logic harder to follow.

A clearer approach would be to use something more descriptive, like `#ifdef STAGE_NO1`, and assign the value directly (e.g. `entity->params = 3`). That way, the condition and the effect are both visible at a glance, making it immediately clear which overlay uses the alternate code path and what value is being set.

src/st/e_breakable_with_debris.h
```diff
@@ -0,0 +1,232 @@
...
+            g_api.PlaySfx(CANDELABRA_HIT_SFX);
+            break;
+        case BUST:
+            entity = AllocEntity(&g_Entities[160], &g_Entities[192]);
+            if (entity != NULL) {
+                CreateEntityFromCurrentEntity(E_ID(BREAKABLE_DEBRIS), entity);
+                entity->params = 256;
+            }
+            g_api.PlaySfx(SFX_GLASS_BREAK_E);
+#ifdef PREVENT_BUST_RESPAWN
```
Xeeynamo: Same as my comment for `#ifdef JUG_PARAMS`. When I read the code, I wonder "what is preventing the bust respawn?" to then find out is just logic unique to the bust object unique from TOP/RTOP. The logic is obfuscated, especially if I am reading this coming from another overlay.

src/st/e_breakable_with_debris.h
```diff
@@ -0,0 +1,232 @@
...
+    breakableType = self->params >> 12;
+    if (!self->step) {
+        InitializeEntity(OVL_EXPORT(EInitBreakable));
+        self->zPriority = g_unkGraphicsStruct.g_zEntityCenter - 20;
+        self->drawMode = draw_modes[breakableType];
+        self->hitboxHeight = hitbox_heights[breakableType];
+        self->animSet = anim_sets[breakableType];
+        self->unk5A = unk_5A[breakableType];
+        self->palette = palettes[breakableType];
+#ifdef INVERTED_STAGE
```
Xeeynamo, line 43: This is what I consider a good example of code sharing. It is clear to me the alternate logic applies to all inverted stages, and it makes sense at glance as I read the code with the flipped Y value.

src/st/e_breakable_with_debris.h
ProjectOblivion: So, if we know the purpose it is better to align the macro with the purpose and if we don't know the purpose, to align it with what we do know instead of abstracting it away?
So in this case, params is the index of the item drop.  We would want to do rename JUG_PARAMS to ITEM_DROP_INDEX?
The logic that I'm following is that JUG_PARAMS forces you to start at the very beginning where STAGE_IS_NO1, while not ideal, at least allows you to start at the point where things become unknown.  Is that accurate?

src/st/e_breakable_with_debris.h
ProjectOblivion: So for this, I'm unsure what could make it more clear, `BUST_NO_RESPAWN`?

I would argue that if you look at it from the perspective of the overlay, that macro tells you exactly what is unique about the breakable code - the bust does not respawn.  It is fairly simple to add a comment to say "The bust in TOP/RTOP doesn't respawn".  Inlining the code doesn't solve anything here because it would obfuscate that the only thing unique about TOP/RTOP is the fact that the bust doesn't respawn and it would also obscure the fact that the bust not respawning is unique.

src/st/e_breakable_with_debris.h
Xeeynamo: I'd rather not introduce a ITEM_DROP_INDEX, since it would still hide the actual value `3` behind another layer of indirection. That would create the same issue as JUG_PARAMS:, forcing me to to look up the definition to see what the macro sets to. I think it’s clearer to assign the value directly, maybe followed with a short comment explaining what it represents. That way, both the intent and the value are visible at a glance, without having to chase macro definitions across the codebase.

src/st/e_breakable_with_debris.h
Xeeynamo: What I meant is to use a `#if !defined(STAGE_TOP) and !defined(STAGE_RTOP)` or similar. Then it's clear that both TOP and RTOP are avoiding spawning a heart to then destroy the entity + preventing it to respawn.

src/st/e_breakable_with_debris.h
ProjectOblivion: So we want to be using the STAGE_IS_X macro?

src/st/e_breakable_with_debris.h
ProjectOblivion: Again, this sounds like you're saying to move to the STAGE_IS_X structure you said to get away from.

src/st/e_breakable_with_debris.h
Xeeynamo: In this case it is preferable for the reasons I explained above. Yes.

src/st/e_breakable_with_debris.h
Xeeynamo: My stance is that we should avoid overcomplicating code sharing when the ratio between a function’s length and the number of `#ifdefs` does not justify it. If we have large functions with only small and few differences across overlays, then adding conditional compilation can make sense to avoid duplication.

However, when we do need to branch, using something explicit like STAGE_IS_X is clearer and less abstract than PREVENT_BUST_RESPAWN, since it removes an extra layer of indirection and makes the condition's context immediately understandable.

ProjectOblivion: That makes sense.  I have no issue with that methodology, just trying to make sense of the seemingly opposed guidance.

Xeeynamo: No worries. Apologies if I was unclear.

Xeeynamo (approved):

PR #2940, opened 2025-10-27
Even more sfxIDs

synamaxmusic:


include/sfx.h
```diff
@@ -78,6 +78,15 @@ enum SfxModes {
 #define CD_SOUND_COMMAND_14 14
 #define CD_SOUND_COMMAND_16 16
 
+// SEQ music tracks
+// Only LIB and DAI have sequenced music data.  The IDs are different on PSP.
+enum SeqTracks {
+    MU_SEQ_LIBRARY = 0x202,
+    MU_SEQ_DAI_CHIMES = 0x204,
+    MU_SEQ_LIBRARY_PSP = 0x302,
+    MU_SEQ_DAI_CHIMES_PSP = 0x304
```
JoshSchreuder: Maybe we should push the PSP ifdef up to here instead, and keep just a single enum name for both platforms, eg. `MU_SEQ_LIBRARY`?

synamaxmusic: I thought about that and that's what I was going to do initially, but it looks like the PSP references both IDs?  See lines 7713 and 7740 in lib_psp/e_shop.c

JoshSchreuder: Got it! Just saw your comment on Discord too, that is interesting.

In that case it would be good to drop a comment inline at those places as it seems likely to be a bug on behalf of the developers

ProjectOblivion: I don't think we should categorize 0x302 and 0x304 as seq files, since they're not.
Since MU is specifically XA audio, these should be SEQ_LIBRARIAN_MUSIC (or something along those lines) and SEQ_CONFESSIONAL_CHIME (or something along those lines since that's the terminology I used for the functions).
0x202 is specifically only the music that plays in the librarian's room and I don't think it's necessary to specify the overlay for either of them since they are both unique by their use.

Putting 0x302 and 0x304 in the existing XA music section makes better sense and allows us to identify that the id's represent different music for pspeu vs us.  I don't know if there is a way to confirm the audio for pspeu, but even though we have pretty good indication of what they are, they should probably be commented as unconfirmed until we're able to validate them.

synamaxmusic: Is MU specifically XA audio only?  I thought the MU prefix just meant music.  Because these are sequenced music cues, I went with "MU_SEQ".  I also went with "DAI_CHIMES" because it's shorter than "CONFESSIONAL_CHIME" and it references the original file name of SEQ_DAI.SEQ.

0x302 and 0x304 are already referenced in the music enum as loop points for XA tracks so there's a conflict with the PS1 build.  I'm not sure if it's a good idea to insert more ifdefs in there to fix mismatching PSP stuff since that will make the code look more messy.

Probably the best way to confirm audio in PSP is hex editing the sfxID number for a known music sound call and see what track plays.

synamaxmusic: Looking through the PSP files, the SEQ data has been converted (or recorded) as .AT3 streamed audio files, so that explains why they're missing.

ProjectOblivion: Just above the enum is `// XA music tracks`, but looking at the enum I'm seeing things that aren't MU, so I dunno.
I don't know what the best way to handle this is from an organizational perspective, so I'll leave that to @Xeeynamo .

PR #2941, opened 2025-10-27
Set INCLUDE_ASM and INCLUDE_RODATA as whitespace sensitive

ProjectOblivion:


JoshSchreuder:
Nice! I didn't even know you could do this in clang-format config, having to turn off the formatting via comments always annoyed me when importing a new overlay or renaming some symbols

hohle:
I don't see an issue with this, but this is the underlying issue with splat? I know I've run into it and assumed it had to do with splat determining which functions were `nonmatchings`, but didn't look into it further. If that's the case, I can try to write up an issue.

ProjectOblivion:
This is specifically because mwccgap breaks when INCLUDE_ASM or INCLUDE_RODATA whitespace doesn't match what it expects.

PR #2944, opened 2025-10-28
Name psp `strcmp` symbol

gamezter:
Determined this from looking at the decomp of this function in `main_psp`. It is next to `strcat` and `strcpy` as well so I'm pretty confident it's the right function


src/dra_psp/1DDC0.c
```diff
@@ -550,160 +550,160 @@ s32 func_psp_090FAB30(s32 fileId, SimFileType type, bool arg2) {
     g_SimFile = &D_800A024C[fileId];
     var_s2 = -1;
     if (type == SimFileType_StagePrg) {
-        if (!func_890C9D4(g_StagesLba[g_StageId].ovlName, "SEL")) {
+        if (strcmp(g_StagesLba[g_StageId].ovlName, "SEL") == 0) {
```
ProjectOblivion, line 553: I'm curious, what is the reasoning for switching from `!` to `== 0`?

gamezter: `strcmp` returns 0 for a match, or the difference between the `char`s of the first difference in the string. It's not really a boolean value and `!strcmp` kind of looks like "doesn't match" so I opted to make the test against 0 explicit

ProjectOblivion: That makes absolute sense.  "num different characters == 0"

Issue #2703, opened 2025-06-24
Update the "Add New Overlay" instructions for new build system


bismurphy:
There is an entry at https://github.com/Xeeynamo/sotn-decomp/wiki/Decompilation#add-new-overlay which describes how to add a new overlay to the project. It was made with our old build system and required modifying several files in several places. With our new and improved build system, the process is simpler, and should be documented here.

hohle:
Removed all of the unnecessary steps, but I think`make-config.py` could still do a bit more like generating `{overlay}.h`.

JoshSchreuder:
Have updated the docs to remove some of the stuff which is now automated / removed / changed, such as progress reporting, and using `extract_overlay.py` instead of `make-overlay.py`

Issue #2942, opened 2025-10-28
PSP Overlays Created by `make-config.py` Do Not Follow Current Patterns


hohle:
When creating new stage overlays for PSP, the current convention is to set the source path to `src/st` and prefix segments with either the bare overlay directory for shared code or `${ovl}_psp` for PSP specific code. The `make-config.py` generated splat config, however, sets the source directory to `src/st/${ovl}` and creates a subdirectory named `${ovl}_psp` within that directory for PSP specific code.

`make-config.py` should be updated to create files with the current convention.

RE: https://github.com/Xeeynamo/sotn-decomp/pull/2936

JoshSchreuder:
Looking at recently imported overlays with the `extract_overlay` tool (rather than `make-config`, this appears to no longer be an issue
https://github.com/Xeeynamo/sotn-decomp/pull/3494/changes#diff-0eadfdcdfc5c68c60438d83adfaccefab067eee5573f3005eedacee0ad9491f2

PR #3031, opened 2025-11-28
Add extract_overlay.py and dependency files submodule

ProjectOblivion:


tools/extract_overlay.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python3
+
+import os
+import argparse
+import multiprocessing
+from pathlib import Path
+from sotn_utils import init_logger, extract
```
sozud: What's the rationale behind tools/extract_overlay.py being in this repo vs. the other one?

ProjectOblivion: Really just pathing.  Seems unnecessary to have it be in a subfolder when it's really just a simple entry point that isn't going to be used anywhere else.  Realistically makes very little difference either way.

JoshSchreuder:
@Xeeynamo I think we should consider removing [make-config.py](https://github.com/Xeeynamo/sotn-decomp/blob/master/tools/make-config.py) in favour of this

I have used it for CAT and ARE now and I think it improves upon a lot of the new config adding experience. 
Probably this would close #2703 and #2942

ProjectOblivion:
Beyond what was there for cat and are, it now also automatically imports and merges e_init.c (EntityUpdates and EInits) and header.c (it has generated `<ovl>.h` for a while now).
It uses the dynamic syms process and does not use the previous `git checkout` method for handling symbols, though it does still use `git add` when complete, to prevent the new `config/` files from being wiped by `make clean`, and uses `git clean` to clean the asm directory.  There is no longer a chance for accidental losses that would happen in the past.

Xeeynamo:
I’m a bit surprised to see a project-specific utility recreated and relicensed in a separate repository, then proposed as an external dependency rather than contributing improvements upstream. Any changes to make_config.py or closely related tools should also be coordinated with the existing contributors (CC @hohle).

From a practical perspective, I’m not sure what the advantage is of pulling `sotn_utils` in as a submodule, given that it can already be cloned and used independently. A quick look at the repository also shows several pieces of hard-coded logic that are tightly coupled to our decomp (https://github.com/ProjectOblivion/sotn_utils/blob/main/sotn_config.py, https://github.com/ProjectOblivion/sotn_utils/blob/main/sotn_overlay.py, https://github.com/ProjectOblivion/sotn_utils/blob/main/symbols.py). Ideally, these should be driven by a configuration file rather than embedded directly in the code.

I want to avoid the inevitable situations where updating our project's configuration or symbol definitions require opening PRs in a separate repository, waiting for review, and then syncing the changes back. Such friction will only slow us down and make maintenance more complicated.

ProjectOblivion:
> waiting for review, and then syncing the changes back. Such friction will only slow us down and make maintenance more complicated

This is the primary reason it is in a separate repository, so this doesn't happen to me.  I'm updating it incredibly frequently, on the order of daily at times.  I do not want everything that is currently in the repo to remain in the repo, though, I want some things to be in sotn-decomp and I am in the process of splitting it up, but that takes time.

> Ideally, these should be driven by a configuration file rather than embedded directly in the code.
This is my ultimate intent, but it isn't there yet.  I wanted to get it available and worry about things like that next.

I tried having it just be cloned and used independently, but I first created this tool months ago and the only one who used it aside from myself was @JoshSchreuder .

> should also be coordinated with the existing contributors

I've brought the tool up twice on discord and have gotten little in the way of overall response there.  @JoshSchreuder has been fantastic with feedback and @hohle gave me some good points to adjust (of which I believe I've adjusted all of his concerns).  I don't appreciate being told to coordinate when I've done my absolute best to do just that long _before_ submitting the PR.

tools/extract_overlay.py
ProjectOblivion: Having the script in the sotn-decomp repo also establishes a place to address sotn-decomp specific code in an external repo, I just haven't split it up yet.

Xeeynamo:
> I've brought the tool up twice on discord and have gotten little in the way of overall response there. @JoshSchreuder has been fantastic with feedback and @hohle gave me some good points to adjust (of which I believe I've adjusted all of his concerns). I don't appreciate being told to coordinate when I've done my absolute best to do just that long _before_ submitting the PR.

I wasn't aware you coordinated with the other contributors. A mention of in the pull request description would have helped me. I am simply sharing my feedback based on my data points, which I have none as your pull request description is blank.

hohle:
I've used the tools as well and have commented on that repo. It works well. I'm torn on it being in the repo or not.

I'm not sure what the advantage is to being a submodule vs moving into the repo. The tool can be iterated on separately, but still requires a commit back to `sotn-decomp` for everyone else to get those changes. It doesn't get other contributors changes any faster and makes updates more opaque.

ProjectOblivion:
While much of the code is sotn specific, some of it is not.  I am working toward splitting them apart, but there are a lot of places where things could go sideways.  I don't have any particular desire to have the sotn specific stuff outside of sotn-decomp, but it is particularly frustrating to put time and effort into something that gets forgotten because it isn't in the repo, then essentially being told it shouldn't be a part of the repo.

sozud (approved):
I think this is fine to merge if the conflicts are resolved. Since ProjectOblivion is the primary author of the tool, and they prefer having it in a separate repo, that seems fine to me, assuming changes can be upstreamed from others

sozud:
@ProjectOblivion Do you have any issue with us forking this to update it? Sorry about the slow review process

sozud:
Going to go ahead and merge this before it gets stale, we can iterate on master if we decide there's issues.

PR #3161, opened 2026-02-28
Add cmpgfx matching compression

Xeeynamo:
Implementation of the Castlevania SOTN compression algorithm, which produces 1:1 files from the decompressed assets.

Unlike the previous failed attempt, this version builds the compression dictionary from an array of nibbles and not bytes. This massively simplified how to continue adapting the code to obtain a match.

As observed in many other scenarios, trailing zero bytes were not written in DRA and dirty bytes from a supposedly previous build need to be replicates with `config/dirt.us.json`.


tools/sotn-assets/sotn/deflate.go
```diff
@@ -0,0 +1,279 @@
...
+	writeNibble := func(ch byte) {
+		if writeFlag == 0 {
+			writeFlag = 1
+			out = append(out, ch<<4)
+		} else {
+			writeFlag = 0
+			out[len(out)-1] |= ch & 0xF
+		}
+	}
+
```
hohle, line 36: Is the binary format documented anywhere. It looks like there's a 8 byte header, followed by various encodings for value and run length. Maybe not in this iteration, but future versions may write this as a struct.

Xeeynamo: The first 8 bytes is the dictionary of the most repeated nibbles

sozud (approved):

PR #3176, opened 2026-03-07
Add ./sotn.sh support for Saturn

Xeeynamo:
Continuation of #3172, adding Saturn support and performing a major Makefile clean-up. 

6b670b0 is needed as the GH workflow is the one from `master` and not the one from this PR. After this gets merged, 6b670b0 can be reverted.


sozud (approved):
I don't understand all the details but seems reasonable

PR #3210, opened 2026-04-07
Move `rnz0` gfx to version specific files

JoshSchreuder:


Makefile
```diff
@@ -138,7 +138,7 @@ clean: ##@ clean extracted files, assets, and build artifacts
 clean:
 	@./sotn.sh clean $(VERSION)
 clean_all:
-	@./sotn.sh clean-all $(VERSION)
+	@./sotn.sh clean all
```
JoshSchreuder, line 141: Not related but a minor fix here for a regression in #3172

sotn-assets takes `clean [version|all]` as parameter, without this change `make clean_all` produces

```
❯ make clean_all
unknown command "clean-all" for ""
exit status 1
make: *** [Makefile:141: clean_all] Error 1
```

config/assets.pspeu.yaml
```diff
@@ -741,28 +741,28 @@ files:
           - [0x28EB0, skip]
           - [0x291C0, rooms, rooms]
           - [0x292C8, skip]
-          - [0x29340, cmpgfx, D_80183B54, 128, 128, 4]
```
sozud, line 744: Are all of these actually different between the two versions? Same question for the second set

JoshSchreuder: Not all of them are different, I believe it depends on whether the data sits at a size which would be padded on PSP but not US
Image

These ones are impacted
Image

I can revert it to only be the ones that are different if preferred? Only did this for consistency

sozud: The only difference is the padding? It sounds like the tool ought to be updated to handle this but it doesn't need to be part of this PR. Given it's a tooling issue I think it's fine for now

JoshSchreuder: @sozud you ok to merge this?

sozud: I made an issue for it https://github.com/Xeeynamo/sotn-decomp/issues/3214
Feel free to merge

PR #3212, opened 2026-04-09
Decompile some `main_psp` functions

gamezter:
I discovered that some of this code is from the MSL library, I've based a lot of the work here on this repository : https://github.com/Yotona/twewy/tree/251a0f6b8f59f507a487f6170b4296dc13be0597/libs/MSL


src/main_psp/36174.c
```diff
@@ -5,207 +5,315 @@
...
+void* const D_psp_0893C08C[] = {
+    (void*)"std::bad_alloc",
+    (void*)D_psp_0893C080,
+};
 
-INCLUDE_ASM("main_psp/nonmatchings/main_psp/36174", func_psp_08936828);
+const char D_psp_0893C094[] = "bad_alloc";
 
-INCLUDE_ASM("main_psp/nonmatchings/main_psp/36174", func_psp_08936844);
+const char* func_psp_08934DFC(s32 arg0) { return D_psp_0893C074; }
```
gamezter, line 115: These look like vTables, indicating that this would be C++ code. I think it maps to these 2 classes : https://github.com/Yotona/twewy/blob/251a0f6b8f59f507a487f6170b4296dc13be0597/libs/include/MSL/C%2B%2B/exception.hpp and  https://github.com/Yotona/twewy/blob/251a0f6b8f59f507a487f6170b4296dc13be0597/libs/include/MSL/C%2B%2B/new.hpp

src/main_psp/36E88.c
```diff
@@ -0,0 +1,857 @@
...
+        __DecodeAddress(specptr, (void**)&spectype);
+        if (__throw_catch_compare((u8*)extype, (u8*)spectype, &dummy)) {
+            return true;
+        }
+        specptr += 4;
+    }
+    return false;
+}
+
+INCLUDE_ASM("main_psp/nonmatchings/main_psp/36E88", __unexpected);
```
gamezter, line 772: This function is C++ : https://decomp.me/scratch/WVrIV I think we need to modify the build system to support .cpp

src/main_psp/36E88.c
```diff
@@ -0,0 +1,857 @@
...
+
+#define UNWIND_FRAME_IS_THUMB 0x80
+#define UNWIND_FRAME_IS_DYNAMIC 0x40
+#define UNWIND_FRAME_HAS_FLUSHBACK 0x20
+
+#define __FunctionPointer(info, context, fp) (fp)
+#define __AdjustReturnAddress(info, context, retaddr) (retaddr)
+#define __LocalVariable(context, offset) ((context)->FP + (offset))
+#define __Register(context, regno)                                             \
+    ((context)->target.GPR[regno + 1]) // TODO this should not have + 1
```
gamezter: This differs from https://github.com/Yotona/twewy/blob/251a0f6b8f59f507a487f6170b4296dc13be0597/libs/include/MSL/Runtime/ExceptionHandler.h#L49 . It probably means that either `ThrowContext` or `TargetContext` aren't the same

src/main_psp/36174.c
```diff
@@ -5,207 +5,315 @@
...
+    func_psp_08934DD4(arg0);
+    *arg0 = D_psp_089AD7E8;
+    return arg0;
+}
 
-INCLUDE_ASM("main_psp/nonmatchings/main_psp/36174", func_psp_08936A44);
+const char* func_psp_08934EEC(s32 arg0) { return D_psp_0893C094; }
 
-INCLUDE_ASM("main_psp/nonmatchings/main_psp/36174", func_psp_08936AC8);
+// MSL/Runtime/MWException.cpp
```
sozud, line 129: the plan is to split files later?

sozud (approved):

src/main_psp/36174.c
gamezter: Yes, I'm not 100% sure of the split

PR #3219, opened 2026-04-19
Add comments for some music and movie file functions

binary1230:
No actual code changes, just filling in some comments around:
1. Add descriptions for some of the movie file offsets in game_handlers.c and breadcrumbs for how to analyze with tools like jpsxdec
2. Add comments in various cutscene audio handlers related to "has CD music started playing"

Future work:
A. The cutscene code is duplicated all over the place, I just added the comments in one section only for now. I might take a stab at de-duplicating it later.
B. We could probably safely rename a few things but... I'm not quite sure that's the only meaning of this is related to music (it PROBABLY is). Those renames would be:
- rename func_80131F68() to something like IsMusicPlaybackStarted()
- rename D_8013901C to something like g_IsMusicPlaying    (maybe. probably. just not totally sure)

TODO:
- [ ] fix failed lint checks

### DIFF
--- a/include/cutscene.h
+++ b/include/cutscene.h
@@ -11,9 +11,9 @@ typedef enum {
     CSOP_NEXT_DIALOG,
     CSOP_SET_POS,
     CSOP_CLOSE_DIALOG,
-    CSOP_PLAY_SOUND,
-    CSOP_WAIT_FOR_SOUND,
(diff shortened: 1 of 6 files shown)


binary1230:
woops. i messed up this PR, will fix the lint and do something better with the branch on my own fork and will re-create.

PR #3220, opened 2026-04-19
Add comments for some music and movie file functions

binary1230:
(resubmitted after lint and branch fix on fork)

No actual code changes, just filling in some comments around:
1. Add descriptions for some of the movie file offsets in game_handlers.c and breadcrumbs for how to analyze with tools like jpsxdec
2. Add comments in various cutscene audio handlers related to "has CD music started playing"

Future work:
A. The cutscene code is duplicated all over the place, I just added the comments in one section only for now. I might take a stab at de-duplicating it later.
B. We could probably safely rename a few things but... I'm not quite sure that's the only meaning of this is related to music (it PROBABLY is). Those renames would be:
- rename func_80131F68() to something like IsMusicPlaybackStarted()
- rename D_8013901C to something like g_IsMusicPlaying    (maybe. probably. just not totally sure)

### DIFF
--- a/include/cutscene.h
+++ b/include/cutscene.h
@@ -11,9 +11,16 @@ typedef enum {
     CSOP_NEXT_DIALOG,
     CSOP_SET_POS,
     CSOP_CLOSE_DIALOG,
-    CSOP_PLAY_SOUND,
-    CSOP_WAIT_FOR_SOUND,
-    CSOP_SCRIPT_UNKNOWN_11,
+
+    CSOP_PLAY_SOUND, // play sound, usually (always?) for music.
+                     // (does nothing if cutscene is being skipped)
+
+    CSOP_WAIT_FOR_SOUND, // wait for CD audio playback to begin
+                         // (does nothing if cutscene is being skipped)
+
+    CSOP_SCRIPT_UNKNOWN_11, // wait for CD audio playback to stop
+                            // (does nothing if cutscene is being skipped)
+
     CSOP_SET_END,
     CSOP_SCRIPT_UNKNOWN_13,
     CSOP_SCRIPT_UNKNOWN_14,
@@ -22,7 +29,10 @@ typedef enum {
     CSOP_SET_FLAG,
     CSOP_SCRIPT_UNKNOWN_18,
     CSOP_LOAD_PORTRAIT,
-    CSOP_SCRIPT_UNKNOWN_20,
+
+    CSOP_SCRIPT_UNKNOWN_20, // same as CSOP_PLAY_SOUND, but always runs
+                            // (always runs even if cutscene is being skipped)
+
     CSOP_SCRIPT_UNKNOWN_21,
     CSOP_SCRIPT_UNKNOWN_22,
     CSOP_SCRIPT_UNKNOWN_23,
--- a/src/dra/dra_bss.h
+++ b/src/dra/dra_bss.h
@@ -61,7 +61,7 @@ extern u16 g_CurSfxDistance20_21;
 extern u8 D_80139014;
 extern u8 g_SfxScriptModeCopy[4];
 extern u32 g_DebugCurPal;
-extern s16 D_8013901C;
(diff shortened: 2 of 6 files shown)


src/st/sel/stream_info.c
JoshSchreuder, line 63: Nice work! 
I think we could probably rename these to reference the specific cinematics you found rather than generically named

JoshSchreuder (approved):
Looks good to me, nice research! I think this opens up the door to a number of renames to make the code clearer

include/cutscene.h
```diff
@@ -22,7 +29,10 @@ typedef enum {
     CSOP_SET_FLAG,
     CSOP_SCRIPT_UNKNOWN_18,
     CSOP_LOAD_PORTRAIT,
-    CSOP_SCRIPT_UNKNOWN_20,
+
+    CSOP_SCRIPT_UNKNOWN_20, // same as CSOP_PLAY_SOUND, but always runs
```
JoshSchreuder, line 33: Probably we could rename this something like `CSOP_PLAY_SOUND_IGNORE_SKIP` or similar

include/cutscene.h
```diff
@@ -11,9 +11,16 @@ typedef enum {
...
-    CSOP_WAIT_FOR_SOUND,
-    CSOP_SCRIPT_UNKNOWN_11,
+
+    CSOP_PLAY_SOUND, // play sound, usually (always?) for music.
+                     // (does nothing if cutscene is being skipped)
+
+    CSOP_WAIT_FOR_SOUND, // wait for CD audio playback to begin
+                         // (does nothing if cutscene is being skipped)
+
+    CSOP_SCRIPT_UNKNOWN_11, // wait for CD audio playback to stop
```
JoshSchreuder, line 21: Similarly here we could probably rename this as `CSOP_WAIT_FOR_SOUND_BEGIN` and `CSOP_WAIT_FOR_SOUND_END`

src/dra/dra_bss.h
```diff
@@ -61,7 +61,7 @@ extern u16 g_CurSfxDistance20_21;
 extern u8 D_80139014;
 extern u8 g_SfxScriptModeCopy[4];
 extern u32 g_DebugCurPal;
-extern s16 D_8013901C;
+extern s16 D_8013901C; // likely: "has CD music actually started playing?"
```
sozud, line 64: Let's name it and it can be iterated on later `g_XaPlaying` for example could work

src/dra/sound.c
```diff
@@ -81,6 +81,10 @@ u16 func_80131F38(void) {
     return g_SeqPlayingId | 0x200;
 }
 
+// this is something vaguely like: "is a CD music track currently playing?"
+// seems like: cutscenes and dialogue tend to use this after starting a music
+// track to wait until playback has actually started (or, to wait to know a
+// playing track that was requested to stop has actually ended)
 bool func_80131F68(void) {
```
sozud, line 88: name `IsXaPlaying` or similar?

src/st/sel/stream_info.c
```diff
@@ -1,30 +1,61 @@
...
+//
+// Audio:
+// - Type:XA
+// - File:1, Channel:1, Stereo?:Yes, Sector stride:8
+
+// File: "LOGO15XA.STR" (Konami Logo):
+// Sectors:23113-24544, Size:2932736
+// - Video: Sectors:23113-24543, Frame Count:143, Sector Frames:23113-24534
+// - audio: Sectors:23120-24240
 StreamInfo g_StreamInfo0 = {
```
sozud, line 22: If we know what these are now let's just give them names? `g_LogoStreamInfo` etc.

Xeeynamo:
Waiting for @binary1230 answer to feedback before merging

JoshSchreuder (changes requested):

Xeeynamo:
After 4 months, I assume this PR will not move further. Closing it.

PR #3218, opened 2026-04-19
Fix PSPEU progress report percentage stuck at 100%

Xeeynamo:
The tool Metrowrap always compile the content from `INCLUDE_ASM`, and the reporting toolset cannot distinguish if the PSP source file is fully decompiled or still work in progress. This PR largely depends from https://github.com/ttkb-oss/metrowrap/pull/5

PR #3222, opened 2026-04-21
Update Metrowrap

hohle:
Updating to include @Xeeynamo's flag for emitting empty functions and data for progressing tracking.

Support for transcoding was removed, so `iconv` is used, just like the `gcc` build. The `--src-dir` option was deprecated, so that's been replaced with `-I`.


hohle:
Re: #3218

JoshSchreuder (approved):

PR #3227, opened 2026-04-26
Fix minor warnings + fake symbols

Xeeynamo:
less noise when performing a clean build


src/st/no0/header.c
JoshSchreuder, line 1: Thanks, I saw this and was meaning to fix it up but hadn't got around to it yet 😄

JoshSchreuder (approved):

PR #3228, opened 2026-04-29
PSP - Matched e_spikes.c

xanderhall:
PSP version, 100% match verified using both differ and decomp.me, build successful.

### DIFF
--- a/src/st/rcat_psp/e_spikes.c
+++ b/src/st/rcat_psp/e_spikes.c
@@ -1,9 +1,113 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "../rcat/rcat.h"
 
+enum SpikesSteps {
+    SPIKES_INIT,
+    SPIKES_INTERACT,
+};
+
+enum SpikesPartsSteps {
+    SPIKES_PARTS_INIT,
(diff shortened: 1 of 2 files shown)


src/st/rcat_psp/e_spikes.c
```diff
@@ -1,9 +1,111 @@
...
+        break;
+    case SPIKES_PARTS_MOVE:
+        MoveEntity();
+        self->velocityY += FIX(0.15625);
+        self->rotate += self->ext.spikes.rotate;
+
+        posX = self->posX.i.hi;
+        posY = self->posY.i.hi;
+
+        g_api.CheckCollision(posX, posY, &collider, 0);
```
JoshSchreuder, line 145: This looks fine to me, if you want to bring in PSX side also, then defining `posX` and `posY` as `s16` should match on both.

See same code on CAT side
https://github.com/Xeeynamo/sotn-decomp/blob/master/src/st/e_spikes.h#L74

xanderhall:
I added the rest of the functions, I've confirmed with differ (unless I'm missing something) that they're all 100% match, but the build still fails, so hold off on merging for now

PR #3229, opened 2026-05-02
Matched rcat_psp/e_spikes

xanderhall:
100% match, build successful, func_us_801B1F1C was added by @JoshSchreuder because of data issues


src/st/rcat_psp/e_spikes.c
```diff
@@ -1,14 +1,12 @@
...
-INCLUDE_ASM("st/rcat_psp/nonmatchings/rcat_psp/e_spikes", EntitySpikes);
-
-INCLUDE_ASM("st/rcat_psp/nonmatchings/rcat_psp/e_spikes", EntitySpikesDamage);
+#define HAS_ORIENTATIONS
+#define SPIKES_PARTS_FRAME 18
+#define SPIKES_TILE_WIDTH 48
+#define DAMAGE_ENT_ON_HIT
+#define DAMAGE_ENT_START g_CurrentEntity[1]
+#define DAMAGE_ENT_END g_CurrentEntity[4]
+#define SPIKES_ELEMENT ELEMENT_CUT | ELEMENT_UNK_10 | ELEMENT_UNK_1
```
bismurphy: Generally defines like this should be put in parentheses to avoid any possibility that order of operations will generate invalid parsing when the preprocessor expands the macro.

xanderhall: I didn't see that done anywhere else that SPIKES_ELEMENT is defined, but I can add it!

src/st/rcat_psp/unk_16510.c
```diff
@@ -1,6 +1,50 @@
...
+    case 0:
+        InitializeEntity(g_EInitSpawner);
+        if (self->params) {
+            g_Tilemap.flags &= (u16)~LAYER_SEMI_TRANS;
+            DestroyEntity(self);
+            break;
+        }
+
+        g_Tilemap.flags |= LAYER_SEMI_TRANS;
+        g_api.func_800EA5E4(0x8004);
```
sozud: Can we put ANIMSET_DRA or OVL on these? I forget which one

Xeeynamo: Yes, it should be ` g_api.func_800EA5E4(ANIMSET_OVL(4))`. The same can be applied for all the others ` g_api.func_800EA5E4`

xanderhall: What about these lines?
```
g_api.func_800EA5E4(0xA003);
g_api.func_800EA5E4(0xA005);
```

JoshSchreuder (approved):

JoshSchreuder:
Thanks for your contribution! Hope to see more in future!

PR #3235, opened 2026-05-11
Name a few `main_psp` symbols

gamezter:


Xeeynamo (approved):

include/game_psp.h
```diff
@@ -18,7 +18,7 @@ extern u32 D_psp_08B42050; // psp cross button
 extern u32 D_psp_08B42054; // psp triangle button
 extern s32 g_UserLanguage; // Defined in global address space
```
gamezter, line 19: `g_UserLanguage` is assigned in `main` (1B61C.c line 118) by calling `sceUtilityGetSystemParamInt`. Maybe a better name would be `g_SystemLanguage` ?

JoshSchreuder: Makes sense to me - I did check this back when I renamed this originally and it comes from the PSP system language
https://github.com/Xeeynamo/sotn-decomp/pull/2412/changes#diff-4dbceb2f69232a451b571009bbb93e8c561300b5c08d314cafe8a39f5c49d6d2R2076

PR #3237, opened 2026-05-12
Decompile `t_bat` FindValidTarget

gamezter:
Depends on https://github.com/Xeeynamo/sotn-decomp/pull/3236
Saturn: https://decomp.me/scratch/UFBfy

You'll notice in the scratch that the compiler generates a call to `__ashiftrt_r4_7` when we use the modulo operator. Something to keep in mind if you see weird function calls in the Ghidra decompilation

### DIFF
--- a/src/saturn/t_bat.c
+++ b/src/saturn/t_bat.c
@@ -4,7 +4,25 @@
 
 INCLUDE_ASM("asm/saturn/t_bat/data", d60CF000, d_060CF000);
 INCLUDE_ASM("asm/saturn/t_bat/f_nonmat", f60CF060, func_060CF060);
-INCLUDE_ASM("asm/saturn/t_bat/f_nonmat", f60CF294, func_060CF294);
+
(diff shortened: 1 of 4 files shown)


src/saturn/t_bat.c
```diff
@@ -2,9 +2,122 @@
...
+        }
+        if (entity->posY.i.hi < 0) {
+            continue;
+        }
+        if (entity->hitboxState & 8 &&
+            !g_BatAbilityStats[s_BatStats.level / 10].makeBadAttacks) {
+            continue;
+        }
+        if (ABS(self->posX.i.hi, entity->posX.i.hi) < 64 &&
+            ABS(self->posY.i.hi, entity->posY.i.hi) < 64) {
```
gamezter, line 65: `abs()` would get inlined here, but would not match.

sozud:
Conflicted

sozud (approved):
Looks good once the conflicts are fixed

PR #3242, opened 2026-05-19
Decompile a few 'sat' functions

gamezter:


src/st/st_common.h
```diff
@@ -155,8 +155,8 @@ void MoveEntity(void) {
 }
 
 void FallEntity(void) {
-    if (g_CurrentEntity->velocityY < FALL_TERMINAL_VELOCITY) {
```
sozud, line 158: My opinion is that the documentation aspect here was nice with naming the magic values, but we could just move the constants before the function rather than having them global in a header.

sozud (approved):

PR #3243, opened 2026-05-20
Decompile more `sat` functions

gamezter:


src/dra/dra.h
```diff
@@ -809,7 +809,7 @@ void LearnSpell(s32 spellId);
 void func_800FDE00(void);
 s32 func_800FE3C4(SubweaponDef* subwpn, s32 subweaponId, bool useHearts);
 void GetEquipProperties(s32 handId, Equipment* res, s32 equipId);
-s32 HandleDamage(DamageParam*, s32, s32 amount, s32);
+s32 get_damage_sub(DamageParam*, s32, s32 amount, s32);
```
sozud: Do people think HandleDamage or get_damage_sub is more clear? There's some cases where the original name is worse than what we've come up with so it probably makes sense to add a comment like `original name:` rather than rename.

sozud (approved):

src/dra/dra.h
gamezter: I agree, I'll keep the original name

src/dra/5D5BC.c
```diff
@@ -277,16 +277,16 @@ void LearnSpell(s32 spellId) {
     }
 }
 
-bool func_800FDD44(s32 itemType) {
-    s32 equippedItem = g_Status.equipment[itemType];
+bool reduce_weapon(s32 hand) {
```
JoshSchreuder: Similar to sozud's comment, I think it would be good to stick with `ReduceWeapon` casing here and going forward with the other Saturn functions, for consistency with the rest of the codebase?

src/dra/dra.h
Xeeynamo: I think we need to make a substantial decision here to which naming scheme to stick to.

If we were to have the whole map file with all symbols, we wouldn't probably think to use custom function names. In this mixed situation though I see a bit of a conflict in direction.

Did we consider instead to have original name and custom name flipped, like the following?

```c
// Handle damage, apply resistance of weakness and change player HP
s32 get_damage_sub(DamageParam*, s32, s32 amount, s32) {
    ...
```

29 years after the game release, it is unlikely we will ever find more symbols. Knowing this, let's think of how do we want to approach this problem now.

src/dra/5D5BC.c
Xeeynamo: See https://github.com/Xeeynamo/sotn-decomp/pull/3243#discussion_r3278967297

src/dra/dra.h
sozud: I think some names are especially problematic e.g. `_v_side_hosei` and `_hkyori_search`. Do we want to use Japanese names?

https://github.com/Xeeynamo/sotn-decomp/blob/0c5d9ed1357f6f1ece7d9dec57b7132c40faa876/src/saturn/game.c#L1148

gamezter: I don't mind either way since it's a relatively easy change to make. I do think we should have some way to indicate *this is the original name* vs *this is a custom name* which could just be a comment at the top of the function.

Xeeynamo: Let's stick with `// Original name: _hkyori_search` for this and future PRs then, and use our own names for the actual symbols. Both Sozud and JoshSchreuder suggestions are sound.

Xeeynamo (approved):

PR #3247, opened 2026-05-27
Decompile more `sat`/`game` functions

gamezter:


src/saturn/game.c
```diff
@@ -1853,24 +1900,115 @@ u8 func_0607AC2C(void) { return DAT_06099811; }
...
+    return Ratan2Shifted(dx, dy);
+}
+
+// original name: search_point
+u8 GetAnglePointToEntityShifted(Entity* entity, s16 x, s16 y) {
+    s16 dx = x - entity->posX.i.hi;
+    s16 dy = y - entity->posY.i.hi;
+
+    return Ratan2Shifted(dx, dy);
+}
```
gamezter, line 1948: Why would this function be called `search_point` ?

Xeeynamo: According to the entity origin X,Y and the destination at its X,Y it finds the angle for origin to move towards destination. I think the name refers to searching the destination point that the entity needs to move to reach it. Maybe a `GetAngleToPoint` is a bit more accurate

sozud: I think a lot of the original names are abbreviated to the point that they are misleading, like Xeeynamo said it's probably more like "search_for_an_angle_to_point"

sozud (approved):

PR #3248, opened 2026-05-27
Decompile `sat` GetPlayerCollisionWith

gamezter:
Depends on https://github.com/Xeeynamo/sotn-decomp/pull/3247

`GetPlayerCollisionWith` accesses `D_8009748C` through a pointer to `g_unkGraphicsStruct` on Saturn. This proves that everything between `0x80097448` and `0x80097490` are part of `g_unkGraphicsStruct`


JoshSchreuder:
Nice discovery!

JoshSchreuder (approved):

src/st/cat/e_elevator.c
```diff
@@ -40,7 +40,7 @@ void EntityElevator(Entity* self) {
                 (posY < 0 && ((g_Player.vram_flag ^ TOUCHING_CEILING) &
                               TOUCHING_CEILING))) {
                 player->posY.i.hi += posY;
-                D_80097488.y.i.hi += posY;
+                g_unkGraphicsStruct.D_8009748C.i.hi += posY;
```
sozud: This was a `Pos` previously which documented that it was X and Y positions. If we know that it's X and Y positions we could adjust the name to reflect that

src/dra/8D3E8.c
```diff
@@ -1677,7 +1678,7 @@ void func_801309B4(Entity* self) {
     if (self->poseTimer < 0) {
         if (D_80138448 != 0) {
             D_80138448 -= 1;
-        } else if (*D_80097448 > 0x18) {
+        } else if (g_unkGraphicsStruct.D_80097448 > 0x18) {
```
sozud, line 1681: Being able to get rid of the pointer dereference here seems like some additional evidence that this change is correct, at least up to that point in the struct

sozud (approved):

src/st/cat/e_elevator.c
gamezter: I renamed them `shoveX` / `shoveY` since they seem to be accumulated whenever something in the environment 'shoves' the player. I think it's used to calculate the camera position.

PR #3252, opened 2026-06-02
Decompile `sat` `SPR` and `INT`

gamezter:
Depends on https://github.com/Xeeynamo/sotn-decomp/pull/3251


src/saturn/lib/spr/spr_slv.c
```diff
@@ -0,0 +1,22 @@
...
+void SPR_RunSlaveSH(PARA_RTN* routine, void* parm) {
+    SPR_SlaveParam = (Uint32)parm;
+    routine((void*)SPR_SlaveParam);
+}
+
+// func_06023144
+void SPR_WaitEndSlaveSH(void) {}
+
+// spr_2a.src
+INCLUDE_ASM("asm/saturn/zero/data", d6023150, d_06023150);
```
gamezter, line 22: This is a `.src` file that looks like hand-written assembly. IDK how we could import this

sozud: The steps ought to be something like:
Add spr_2a.src to source tree as is
Add ninja rule to invoke assembler to `tools/builds/saturn.py`
Add spr_2a to lib_srcs
In `add_srcs` conditionally invoke assembler if filename is .src
Change linker script to add the resulting object

Doesn't need to be part of this PR

sozud (approved):

PR #3253, opened 2026-06-03
Decompile some `sat` `PER` functions

gamezter:
Mostly based off this : https://github.com/willll/saturn-docker/blob/915aed1ce5f19f04f225f735c97f448620402778/Resources/sbl6/segalib/per/per_prt_new.c

This code is very different though, I think it belongs to `per_prt.c` instead of `per_prt_new.c` but I haven't found any sources online


sozud (approved):

sozud:
I think this may be one of the libs where they were either not shipping the source or not doing it consistently.

1996-11 CD version
https://gist.github.com/sozud/f1f0c07efb43b761056558f9b6f55397

1995-11 CD version (looks the same)
https://gist.github.com/sozud/8ebaa798c79c728c87027be8429af594

PR #3254, opened 2026-06-03
Decompile `sat` `MTH`

gamezter:

PR #3255, opened 2026-06-03
Split and decompile `sat` `SND`

gamezter:
Depends on https://github.com/Xeeynamo/sotn-decomp/pull/3254

I noticed the Saturn SDK had a `SND_StopPcm` function similar to the `SND_StopPcm2` in the Richter symbols. Turns out that function is exactly the same as `SND_StopPcm`

Here I split off `SND` from `zero.c`. The `SND` module is compiled in -O2 (or -O3) like the rest of `zero.c`.


src/saturn/zero_2.c
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "zero.h"
+#include "inc_asm.h"
+#include "sattypes.h"
+
+INCLUDE_ASM("asm/saturn/zero/f_nonmat", f6019074, func_06019074);
```
gamezter, line 6: Couldn't figure out what this code is, it comes before the original `lib` files and after the new `lib/snd` so maybe it is also part of the Saturn SDK ?

sozud: The 68k sound driver might be nearby? A struct gets passed into SND_Init with the address of the driver so that might reveal something

sozud (approved):

src/saturn/zero_2.c
gamezter: That would be either `0x0603AA7C` or `0x0603AA40` : 
image

sozud: Ok, 0x0603AA7C is the sound driver with a length of 0x67f2 and 0x0603AA40 is the "sound area map" with a length of 0x3a. Thanks for investigating

PR #3256, opened 2026-06-05
Import BO0 for US and PSPEU.

hohle:
Deduped existing shared files and code sharing between platforms.


sozud (approved):

JoshSchreuder:
Needs updates to these files for dups and progress
https://github.com/Xeeynamo/sotn-decomp/blob/3c2673fd8b4b98d9d52a6d50629f9d710bd748f6/tools/dups/src/main.rs#L248-L261

https://github.com/Xeeynamo/sotn-decomp/blob/3c2673fd8b4b98d9d52a6d50629f9d710bd748f6/tools/progress.py#L446

include/castle_flags.h
```diff
@@ -10,6 +10,7 @@ typedef enum {
     /* 0x2 */ NULL_STAGE_FLAG = 2, // When a stage flag isn't defined
     /* 0x2 */ NO0_STAGE_FLAG = 2,
     /* 0x2 */ RARE_STAGE_FLAG = 2,
+    /* 0x2 */ BO0_STAGE_FLAG = 2,
```
JoshSchreuder: Doesn't really matter but I think this is when the stage title doesn't appear on screen? 
In those overlays we should probably just be using `NULL_STAGE_FLAG` or leave it undefined which should be just that

https://github.com/Xeeynamo/sotn-decomp/blob/3c2673fd8b4b98d9d52a6d50629f9d710bd748f6/src/st/e_stage_name_us.h#L5

hohle: I didn't realize I could leave it out. Removed and removed the define in `e_stage_name.c` as well.

PR #3257, opened 2026-06-08
Decompile `sat` `InitStatsAndGear`

gamezter:
I had to split `game.c` because `InitStatsAndGear` expects `AddToInventory` to return `s32` but it returns `void`


src/saturn/game.c
```diff
@@ -1740,7 +611,35 @@ INCLUDE_ASM("asm/saturn/game/f_nonmat", f6078604, func_06078604);
...
+            ret = acc->name;
+        }
+        break;
+    case 3: {
+        RelicDesc* relic = &g_RelicDefs[id];
+        ret = relic->name;
+    } break;
+    case 2: {
+        SpellDef* spell = &g_SpellDefs[id];
+        ret = spell->name;
```
gamezter, line 637: This looks like 2 inlined functions, maybe `GetRelic` and `GetSpell` ?

gamezter:
When I clear all temp files locally, this builds and matches. The logs say `attempt to open build/saturn/game_0.o failed`... Could this be because it's using cached data ?

sozud:
> When I clear all temp files locally, this builds and matches. The logs say attempt to open build/saturn/game_0.o failed... Could this be because it's using cached data ?

I think last time this happened it was due to a header file change? My impression is that the compiler or our build setup doesn't consistently print out errors due to problems in the header files maybe. What happens if you only have the header file change without the rest to check and confirm that's not the issue? If a PR with only the change to sattypes.h passes CI then we know that's not it.

gamezter:
> > When I clear all temp files locally, this builds and matches. The logs say attempt to open build/saturn/game_0.o failed... Could this be because it's using cached data ?
> 
> I think last time this happened it was due to a header file change? My impression is that the compiler or our build setup doesn't consistently print out errors due to problems in the header files maybe. What happens if you only have the header file change without the rest to check and confirm that's not the issue? If a PR with only the change to sattypes.h passes CI then we know that's not it.

It passed : https://github.com/Xeeynamo/sotn-decomp/pull/3258

sozud:
I don't see anything clear in the logs that suggests what the error is so I don't have better ideas than systematically building the PR back up in small commits. Does it pass if all you have is the file split for example? I think I'd go change-by-change until we find the thing that breaks it.

gamezter:
> I don't see anything clear in the logs that suggests what the error is so I don't have better ideas than systematically building the PR back up in small commits. Does it pass if all you have is the file split for example? I think I'd go change-by-change until we find the thing that breaks it.

Just the split fails : https://github.com/Xeeynamo/sotn-decomp/pull/3259

Looks like it's linking `game.o` at step 80 : 
```
[80/118] Linking build/saturn/game.elf from build/saturn/game.o
```
but compiling `game_0` at step 85:
```
[85/118] Building build/saturn/game_0.s from build/saturn/game_0.cpp
```

gamezter:
> > I don't see anything clear in the logs that suggests what the error is so I don't have better ideas than systematically building the PR back up in small commits. Does it pass if all you have is the file split for example? I think I'd go change-by-change until we find the thing that breaks it.
> 
> Just the split fails : #3259
> 
> Looks like it's linking `game.o` at step 80 :
> 
> ```
> [80/118] Linking build/saturn/game.elf from build/saturn/game.o
> ```
> 
> but compiling `game_0` at step 85:
> 
> ```
> [85/118] Building build/saturn/game_0.s from build/saturn/game_0.cpp
> ```

The issue was that the `link_multi` didn't have dependencies on the `sub_objs`.

sozud (approved):

PR #3262, opened 2026-06-10
Decompile some `SAT` `create_entity` functions

gamezter:


src/saturn/game_0.c
```diff
@@ -7,12 +7,192 @@
...
+        tmp = tilemap->scrollX.i.hi;
+        if (DAT_0605c680.g_ScrollDeltaX > 0) {
+            tmp += 0x190;
+            CreateEntitiesToTheRight(tmp);
+        } else {
+            tmp -= 0x50;
+            CreateEntitiesToTheLeft(tmp);
+        }
+    }
+    if (DAT_0605c680.g_ScrollDeltaY != 0) {
```
gamezter, line 185: `g_ScrollDeltaX/Y` are accessed through this pointer on Saturn. I'll need to decompile more Saturn code to find the other struct members to determine if I can do this for other platforms as well

sozud (approved):

PR #3264, opened 2026-06-13
Add RNO3 overlay

bismurphy:
This is the reverse entrance. I'm planning to tackle this moving forward since there are a few interesting enemies to learn about.

Looks like the tool from https://github.com/Xeeynamo/sotn-decomp/pull/3031 did a good job on this. Since it comes from an old version of the master repo, it was a little tricky to make the changes work together, but I think I got it to work and I think all the changes in this PR are restricted to the items of interest.


Xeeynamo (approved):

sozud:
This is missing the progress report and duplicates stuff but otherwise looks good

bismurphy:
I think I added those now, but not 100% sure if I did it right, let me know if there are mistakes

tools/dups/src/main.rs
```diff
@@ -462,6 +462,13 @@ fn do_dups_report(output_file: Option<String>, threshold: f64) {
             include_asm: get_all_include_asm("../../src/st/rare/"),
             path_matcher: "st/rare".to_string(),
         },
+        SrcAsmPair {
+            asm_dir: String::from("../../asm/us/st/rbo3/matchings/"),
+            src_dir: String::from("../../src/st/rbo3/"),
+            overlay_name: String::from("RBO3"),
```
JoshSchreuder: Should be RNO3

tools/progress.py
```diff
@@ -438,6 +438,7 @@ def report_discord(progresses: dict[str, DecompProgressStats]):
     progress["sttop"] = DecompProgressStats("sttop", "st/top")
     progress["stwrp"] = DecompProgressStats("stwrp", "st/wrp")
     progress["strare"] = DecompProgressStats("strare", "st/rare")
+    progress["strbo3"] = DecompProgressStats("strbo3", "st/rbo3")
```
JoshSchreuder: Same here, RNO3

JoshSchreuder (changes requested):

tools/progress.py
bismurphy: oops thank you

JoshSchreuder (approved):

PR #3267, opened 2026-06-14
Align structs in primitive.h

Xeeynamo:
Fixes `EntityStageNamePopup` glitching out

### DIFF
--- a/include/primitive.h
+++ b/include/primitive.h
@@ -350,4 +350,3 @@ STATIC_ASSERT(sizeof(Primitive) == sizeof(AxePrim), "unaligned");
 STATIC_ASSERT(sizeof(Primitive) == sizeof(EntranceCascadePrim), "unaligned");
 STATIC_ASSERT(sizeof(Primitive) == sizeof(FrozenShadePrim), "unaligned");
 STATIC_ASSERT(sizeof(Primitive) == sizeof(NumericPrim), "unaligned");
-STATIC_ASSERT(sizeof(Primitive) == sizeof(NumericPrim), "unaligned");
--- a/include/primitive.h
+++ b/include/primitive.h
@@ -205,9 +205,11 @@ typedef struct EntranceCascadePrim {
     /* 0x1E */ s16 y2;
     /* 0x20 */ f32 velocityY;
     /* 0x24 */ u8 step;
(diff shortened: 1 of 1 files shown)


include/primitive.h
```diff
@@ -340,3 +342,12 @@ typedef struct {
     /* 0x22 */ s16 v2;
     /* 0x24 */ u8 pad[8];
 } PrimWeapon017; /* size = 0x2C */
+
+STATIC_ASSERT(sizeof(Primitive) == sizeof(Primitive2), "unaligned");
```
sozud, line 346: Do we not need to check if one of the members is aligned? e.g. check that the offset of b3 is the same everywhere or something

include/primitive.h
```diff
@@ -340,3 +342,12 @@ typedef struct {
...
+STATIC_ASSERT(sizeof(Primitive) == sizeof(FakePrim), "unaligned");
+STATIC_ASSERT(sizeof(Primitive) == sizeof(PrimLineG2), "unaligned");
+STATIC_ASSERT(sizeof(Primitive) == sizeof(AxePrim), "unaligned");
+STATIC_ASSERT(sizeof(Primitive) == sizeof(EntranceCascadePrim), "unaligned");
+STATIC_ASSERT(sizeof(Primitive) == sizeof(FrozenShadePrim), "unaligned");
+STATIC_ASSERT(sizeof(Primitive) == sizeof(NumericPrim), "unaligned");
```
sozud, line 352: Intentionally have NumericPrim twice?

Xeeynamo: Fixed in [c4d1ec4](https://github.com/Xeeynamo/sotn-decomp/pull/3267/commits/c4d1ec447d6b9b1b4b1d228a6522e3f9477fe453)

include/primitive.h
Xeeynamo: I've been thinking about that, as we do exactly that in `entity.h`.

As @bismurphy discovered a while ago, it seems developers use the `Primitive` struct as blobs of data for different purposes. So comparing field by field is out of the picture. Even `Primitive2` would be quite hard to test field by field. Ultimately I decided to at least ensure all structures has at least equal size.

sozud (approved):

PR #3272, opened 2026-06-18
Decompile some shared files in RNO3

bismurphy:
I have begun decompiling RNO3 on both the US and PSPEU versions.

I am going through this overlay from the top down, following the US file ordering.

The first several files I worked on were files that exist in .h files already. However, most of them required some #ifdef gymnastics. Those are likely the main thing to review here. 

Next file is called "func_us_801b1674" which appears to be unrecognized, so I am stopping here - I like when a PR is purely deduplication or new code, and I want that one's new code to be by itself if possible.


src/st/e_breakable.h
```diff
@@ -18,6 +18,11 @@ void EntityBreakable(Entity* entity) {
         AnimateEntity(g_eBreakableAnimations[breakableType], entity);
         if (entity->hitParams) { // If the candle is destroyed
             Entity* entityDropItem;
+#if defined(STAGE_IS_RNO3)
+            entity->drawFlags = ENTITY_DEFAULT;
+            entity->rotate = 0;
```
JoshSchreuder: `ROT(0)` perhaps but not major

src/st/e_breakable.h
```diff
@@ -38,5 +43,9 @@ void EntityBreakable(Entity* entity) {
         entity->blendMode = blend_modes[breakableType];
         entity->hitboxHeight = g_eBreakableHitboxes[breakableType];
         entity->animSet = g_eBreakableanimSets[breakableType];
+#if defined(STAGE_IS_RNO3)
+        entity->drawFlags = ENTITY_ROTATE;
+        entity->rotate = 0x800;
```
JoshSchreuder: Nit: `ROT(180)`

JoshSchreuder (approved):

sozud (approved):

PR #3275, opened 2026-06-19
RNO3 st_init for PSP

bismurphy:
Critical file that acts as a prereq for many other files since it defines the E_ID entries.

No big stumblers here. I have a script that generates most of the file, might be cool if it was part of the new overlay script, but at the same time we don't really have that many overlays left so the benefit of new utilities is starting to diminish. Good problems to run into!


sozud (approved):

JoshSchreuder:
> I have a script that generates most of the file

I have to do this for PSP CEN still, if you don't mind sharing it might save me a little time at least 😄

bismurphy:
> > I have a script that generates most of the file
> 
> I have to do this for PSP CEN still, if you don't mind sharing it might save me a little time at least 😄

I can't guarantee this will work for other overlays as-is, but here:

```
import re

OVERLAY = "cen"

with open(f"sotn-decomp/src/st/{OVERLAY}/{OVERLAY}.h") as f:
    text = f.read()
print(text)
ents = re.findall(r'(?<=E_).*(?=,)', text)
ents = ents[17:] # first 16 don't go into the init stuff

for ent in ents:
    print(f"s32 E_ID({ent});")

for ent in ents:
    print(f"    E_ID({ent}) = E_{ent};")

with open(f"sotn-decomp/asm/pspeu/st/{OVERLAY}_psp/nonmatchings/{OVERLAY}_psp/st_init/InitEntityIds.s") as f:
    text = f.read()
syms = re.findall(r'(?<=D_pspeu_)[^\)]*(?=\))', text)
addrs = [hex(x)[2:].upper() for x in reversed(sorted(int(x,16) for x in set(syms)))]

for name,num in zip(ents, addrs):
    print(f"D_psp_E_{name} = 0x{num};")
```

Note that this requires you to have InitEntityIds as a named function, which CEN doesn't have, because there is no symbol for it. So start by adding the symbol for that to name the function, then try running this. Note also that the two files it reads start at the `sotn-decomp` directory, so this script should sit in the directory alongside sotn-decomp. Good luck! Message me on discord if you continue to have issues.

PR #3281, opened 2026-06-23
Modify layout/laydef extractor to show room numbers on entities

bismurphy:
This is a rework of the entity layout extraction system. Right now, we have a e_laydef file, which shows, for each room, where its entities are located in the massive array.

image

There is a placeholder entry at the beginning, then we have "Room 0's entities are at offset 95, room 1's entities are at offset 75, room 2's entities are at offset 149", and so on. 

That e_laydef file gives you the indices into the e_layout file, which looks like this:

image

We have annotations at the relevant offset points, so you can see where offset 5, 12, 15, etc are located.

By studying these two files, we can see what entities are in any given room, and conversely, what room contains any relevant entities.

For example, in this block:
```
//55
    0xFFFE, 0xFFFE, E_NONE | 0x0000, 0x0000, 0x0000,
    0x0080, 0x0100, E_BG_LIGHTNING | 0x0000, 0x0080, 0x0100,
    0x0080, 0x0100, E_DEATH_SKY_SWIRL | 0x0000, 0x0090, 0x0000,
    0x0080, 0x01E0, E_BACKGROUND_TREES | 0x0000, 0x0071, 0x0100,
    0x0080, 0x01E0, E_BACKGROUND_TREES | 0x0000, 0x0072, 0x0101,
    0x0080, 0x01E0, E_BACKGROUND_TREES | 0x0000, 0x0073, 0x0102,
    0x0080, 0x01E0, E_BACKGROUND_TREES | 0x0000, 0x0074, 0x0103,
    0x0080, 0x0100, E_CUTSCENE | 0x0000, 0x0088, 0x0000,
    0xFFFF, 0xFFFF, E_NONE | 0x0000, 0x0000, 0x0000,
```

We can see that index 55 contains the Death encounter (including the sky swirl, and the cutscene entity). By careful counting in the e_laydef, we can see this is in room 11, or 0xB. The map at https://www.sotn.fun/wiki/Room_IDs confirms that the Death encounter room is room 0xB. So this is all consistent.

However, it is a bit clunky to need to cross-reference two files (laydef and layout), and to need to carefully count indices (and remember the off-by-one indexing). 

With this PR, the output in e_layout instead looks like this:
image

We pull the room number in, meaning all the information you need is in e_layout.

To make this work, I had to adjust the calling conventions for makeSortedBanks and writeLayoutEntries. I also decided to make writeLayoutEntries call makeSortedBanks itself, rather than doing a nested call of `writeLayoutEntries(&sbData, makeSortedBanks(el.Entities, true), false)`. That meant I had to swap the function orders, so makeSortedBanks would be declared first and therefore be callable from within writeLayoutEntries.

Also while I was in there, I changed sbHeader and sbData to laydefFile and layoutFile, because I kept getting confused on which file was which.

I am definitely not a Go programmer, so please feel free to fix any conventions I am doing wrong. All I was really worried about was whether it would function correctly, not if the code would do things in the preferred way.


JoshSchreuder:
This seems good (and useful!) to me but I am not all that familiar with the assets tool so would be good if you can review this one @Xeeynamo

tools/sotn-assets/assets/layout/layout.go
```diff
@@ -234,6 +211,37 @@ func buildEntityLayouts(fileName, outputDir, subDir string, ovlName string) erro
...
+				return fmt.Errorf("layout entity bank %d needs to have a X:-2 and Y:-2 entry at the beginning", i)
+			}
+			lastEntry := entries[len(entries)-1]
+			if lastEntry.X != -1 || lastEntry.Y != -1 {
+				return fmt.Errorf("layout entity bank %d needs to have a X:-1 and Y:-1 entry at the end", i)
+			}
+			roomNum := 0
+			for indices_num, member := range el.Indices {
+				if member == i {
+					roomNum = indices_num - 1
```
Xeeynamo: I'm unsure about this logic. `indices_num` starts at 0, there's the risk `roomNum` will be `-1`. I think we should loop starting from the second element of `el.Indices`.

tools/sotn-assets/assets/layout/layout.go
```diff
@@ -234,6 +211,37 @@ func buildEntityLayouts(fileName, outputDir, subDir string, ovlName string) erro
...
+		for i, entries := range banks {
+			// do a sanity check on the entries as we do not want to build something that will cause the game to crash
+			if entries[0].X != -2 || entries[0].Y != -2 {
+				return fmt.Errorf("layout entity bank %d needs to have a X:-2 and Y:-2 entry at the beginning", i)
+			}
+			lastEntry := entries[len(entries)-1]
+			if lastEntry.X != -1 || lastEntry.Y != -1 {
+				return fmt.Errorf("layout entity bank %d needs to have a X:-1 and Y:-1 entry at the end", i)
+			}
+			roomNum := 0
```
Xeeynamo: Is this a forced default to silent compiler errors, or is it a genuine default?

tools/sotn-assets/assets/layout/layout.go
```diff
@@ -234,6 +211,37 @@ func buildEntityLayouts(fileName, outputDir, subDir string, ovlName string) erro
...
+			for indices_num, member := range el.Indices {
```
Xeeynamo: It would be simpler to read if extracted to a separate function, so it reads as `roomNum := findRoomNo(...)` or similar

Xeeynamo:
Left a few comments on a possible edge case. I can't test it yet unfortunately. The idea is sound though, and a genuine improvement.

tools/sotn-assets/assets/layout/layout.go
bismurphy: Yes, this - 1 result happens for the offset 0 (the placeholder with two E_NONE). The -1 indicates "not used in any room" essentially. We could add an if-statement to the room printing logic to print a special text for this. I wanted to start with a bare bones approach so for now it just says -1.

tools/sotn-assets/assets/layout/layout.go
bismurphy: Very reasonable, I'll pull this out and call it indexOf, which is a fairly common naming in other languages.

tools/sotn-assets/assets/layout/layout.go
bismurphy: I don't quite understand the question. I think it might be resolved anyway with the indexOf approach. Let me know if further discussion is needed.

tools/sotn-assets/assets/layout/layout.go
bismurphy: Resolved with new commit.

Xeeynamo (approved):

PR #3288, opened 2026-06-27
Decompile RNO3 Jack O Bones

bismurphy:
The jack itself has some really weird PSP ifdef. No idea what the reason is for that.

Otherwise the rest seems pretty normal, nice and readable with all the names on the data.


src/st/rno3/e_jack_o_bones.c
```diff
@@ -1,10 +1,343 @@
...
+        if (self->velocityX > 0) {
+            xVar += 5;
+        } else {
+            xVar -= 5;
+        }
+        g_api.CheckCollision(xVar, yVar, &sp10, 0);
+        if (sp10.effects & EFFECT_SOLID) {
+            self->velocityX = -self->velocityX;
+        }
+        if (self->params)
```
sozud, line 329: My preference is having parens to make the nesting explicit but it's not a big deal

sozud (approved):

src/st/rno3/e_jack_o_bones.c
bismurphy: Yeah normally I absolutely agree, but in this case because it's in the ifdef, the closing bracket would also have to be included there, which gets messy. Leaving it as a single non-bracketed if statement allowed me to skip the closing brace. Otherwise I would have to do a `#else if(1){` or have a full #ifdef block with just a `}` and both of those options are always a bit ugly.

PR #3289, opened 2026-06-28
Fix the build matrix linker

JoshSchreuder:
```
[ 99%] Linking CXX executable sotn
Undefined symbols for architecture arm64:
  "_D_us_801B1664", referenced from:
      _BO4_spriteBanks in libcore.a[343](header.c.o)
  "_D_us_801B2068", referenced from:
      _BO4_spriteBanks in libcore.a[343](header.c.o)
  "_D_us_801B252C", referenced from:
      _BO4_spriteBanks in libcore.a[343](header.c.o)
ld: symbol(s) not found for architecture arm64
clang++: error: linker command failed with exit code 1 (use -v to see invocation)

```

I'm not completely sure if this is a good idea / the right move here.

I think it's because these symbols are externed but never imported anywhere, and the linker is unhappy about it because `BO4_spriteBanks` is imported which references them?

https://github.com/Xeeynamo/sotn-decomp/blob/2226e6fc449c772bde03052d2f66d7b4a2f6cb71/src/boss/bo4/header.c#L32-L41

Correct fix would likely be to import the data


sozud:
I think what's going on is that spriteBanks is supposed to be extern and generated by the tooling? @Xeeynamo would know better probably. The linker got more strict probably which exposed this issue

https://github.com/Xeeynamo/sotn-decomp/blob/0dbb3a9e40e6359d83b01c52f9e7871942b71f74/src/st/nz0/header.c#L8
https://github.com/Xeeynamo/sotn-decomp/blob/0dbb3a9e40e6359d83b01c52f9e7871942b71f74/tools/sotn-assets/assets/spritebanks/spritebanks.go#L18

JoshSchreuder:
@sozud yeah I think you're right now that I look at it further.

There's a few overlays manually importing the spritebanks (MAD, RBO5, BO6, BO4) but none apart from BO4 are in the CMakeLists so don't hit this issue. I'll try and rework the asset import for this rather than tweaking the linker

Issue #2401, opened 2025-04-03
`sotn-assets` palette_def only supports a `OVL_EXPORT(cluts)[]` of size 1


gamezter:
For instance, `st/no1` and `st/lib` have more than 1 element in these arrays: 

![Image](https://github.com/user-attachments/assets/e47529ee-b370-402f-996f-ce47a4013c9d)

![Image](https://github.com/user-attachments/assets/a2743d5b-5282-4261-8be3-9e95a1441eaf)

JoshSchreuder:
CAT has this issue also
<img width="410" height="271" alt="Image" src="https://github.com/user-attachments/assets/e3cee1b2-c0ee-478e-aa0d-8919cdfb714d" />

PR #3291, opened 2026-06-29
Update BO4 data so that it doesn't cause linker issues

JoshSchreuder:


src/boss/bo4/header.c
```diff
@@ -26,51 +29,3 @@ AbbreviatedOverlay2 OVL_EXPORT(Overlay) = {
...
-    PAL_COPY_DATA_(0x20F0, &g_Clut[2][0], 16),
-    PAL_TERMINATE(),
-};
-
-u_long* OVL_EXPORT(cluts)[] = {
-    D_us_80180094,
-    D_us_801800B8,
-    D_us_801800CC,
-    D_us_801800A8,
-};
```
JoshSchreuder, line 73: Wasn't able to import these with the assets tools - there's the existing CLUT issue in #2401 and another oddity with the sprite banks where `D_us_801B0A78`/`D_us_801B159C` don't appear to work as sprites.

The actual amount of data we're losing by doing this is pretty minimal so I think this is ok for now, we can come back and address this later once the assets tool is updated

Xeeynamo (approved):

sozud:
It looks like data was originally imported here: https://github.com/Xeeynamo/sotn-decomp/pull/2666/changes#diff-b4ee1d1301c89d2d974af475537c533deea45742c348b1d15b3b2966cb9d8055R37

@hohle Any issues with this change?

hohle:
I don’t have any concerns.

sozud (approved):

PR #3293, opened 2026-07-01
 Rework GetNormalizedAngle to LimitAngleChange 

bismurphy:
I'm working on the ourobourous enemy, and it uses GetNormalizedAngle. I realized I didn't know what that function's 3 arguments actually do, and when I looked at that function, I saw that it was also very undocumented (all args and local variables were using temp names).

I went through and renamed all variables (locals and args) and commented the function to explain what it is doing. Then I thought a bit more and decided that "normalize" isn't a good description of what this function is doing, so I tried creating a new name for it.

Very open to feedback on this PR. If the language of my comments, or any of my naming could be improved, I am more than happy to make those improvements. But at least this is a starting point which improves the readability of this function (and the things calling it).


src/saturn/game.c
```diff
@@ -1431,7 +1431,7 @@ u16 GetAnglePointToEntity(Entity* entity, s16 x, s16 y) {
     return ratan2(dy, dx);
 }
 
-u16 GetNormalizedAngle(u16 arg0, u16 arg1, u16 arg2) {
+u16 LimitAngleChange(u16 arg0, u16 arg1, u16 arg2) {
```
JoshSchreuder: Worth renaming the args here perhaps

JoshSchreuder:
Seems like a good change to me, will make naming the Entity ext members in places where they are passed to this function simpler too

src/saturn/game.c
bismurphy: Thanks, done

JoshSchreuder (approved):

sozud (approved):

PR #3294, opened 2026-07-01
RNO3 Decompile Ouroborous and Dragon Rider

bismurphy:
This one is fairly complicated, lots of entities doing a complicated dance together.

I decided late into the process to split this file. I think it makes more sense split, since each enemy is self contained, and the data for the Dragon Rider has to come after the Ouroborous's functions due to a literal string in a FntPrint call.

Note: "Ouroborous" is spelled at least 2-3 different ways within the game. There is no single proper spelling the game settles on. We just have to live with that. I arbitrarily picked.


config/symbols.us.ld.txt
JoshSchreuder, line 1: What's this file for?

bismurphy: Oh weird, I don't know what that is. I'm guessing it got created because I did a build of the debug module? I'll remove it from this PR, good catch

bismurphy: ok i think it is removed

Xeeynamo (approved):

sozud (approved):

PR #3297, opened 2026-07-03
Decompile RNO3 Blue Venus Weed

bismurphy:
This one was interesting because we have the Venus Weed (non-blue) already in CHI, but this one is the Blue Venus Weed.

The code is extremely similar. I decided for this one to make a #define BLUE to switch between the two versions. It became a little clunky in places, but I think it worked out okay.


sozud (approved):

config/assets.us.yaml
```diff
@@ -1218,9 +1218,7 @@ files:
       - start: 0
         vram: 0x80180000
         assets:
-          # nb. sprite banks look weird, some non sprite looking data?
           #- [0x34, sprite_banks, sprite_banks]
-          # multiple cluts here are not importing correctly, see issue #2401
```
JoshSchreuder, line 1223: Not a big deal but maybe rebase issue (assume you didn't mean to remove these 😄 )

JoshSchreuder (approved):

config/assets.us.yaml
bismurphy: Thanks, fixed I think

PR #3301, opened 2026-07-04
 Try to fix item drops overlay collsion issue 

sozud:
I'll rebase this once the psyz PR gets merged. This tries to fix a crash when an item drops. The issue was that the prize drops array and HEART_DROP_CASTLE_FLAG from CEN was being used instead of the one for NZ0, which was causing an out of bounds read.

Basically my understanding is
```
        index = OVL_EXPORT(PrizeDrops)[index];
```
If OVL_EXPORT is used inside a function, then the function itself and everything else in the file also needs to use it or you get linking definition issues or something like that. Changing that required a bunch of other cascading changes. I think basically the easiest thing would to mark every entity as OVL_EXPORT probably in a future PR.

image


sozud:
This PR has ended up too big/complicated, I think this needs to be split up. I'll make an issue for this and look at simpler fixes.

PR #3302, opened 2026-07-04
Hard fail on symbol collisions

Xeeynamo:
Currently, we are mass-importing overlays without paying close attention to symbol collisions. Building SOTN as an intermediary library was hiding these competing symbols. As a consequence, stages like NZ0 would accidentally use entities such as `EntityPrizeDrop` from the CEN overlay, causing crashes on NZ0-specific drops. This is just one example of how widespread the problem actually was.

This PR changes the build process so that all project files directly target the final executable. Making this shift exposes all of those conflicting symbols at compilation time, which forces us to be much more explicit with symbol renaming and the use of OVL_EXPORT.

Note: This PR is not yet meant to be merged. Instead, it should be used as a foundation to build upon, allowing us to cherry-pick its changes into separate, smaller PRs. As we continue this process, the number of build errors will steadily decrease until this architectural change is finally ready to be merged. It is also intended to better support ongoing efforts like #3301.


Xeeynamo:
Closing as this is no longer needed after #3335

PR #3303, opened 2026-07-04
Decompile RNO3 water effects

bismurphy:
This one came out a bit messy. There are a whole lot of changes in this version compared to the first castle, so for now I am leaving it as its own file rather than doing the ifdef. That may come in the future. This is a standalone file decompiled on both versions for now.


Xeeynamo (approved):
LGTM. I think it's better to put aside the various `#ifdef` when changes differ too much. I find code quite hard to read when they're too spread.

PR #3311, opened 2026-07-05
Rno3 finishing up pulling in data for both versions

bismurphy:
Draft for now as I seek help with an issue pulling in the different layers on the two versions.


bismurphy:
I believe this one is all done. Everything in the overlay is as documented as I can get it at this stage. Pretty good!

JoshSchreuder (approved):

src/st/rno3/e_orobourous.c
```diff
@@ -3,7 +3,7 @@
 
 // NOTE: Compare this to e_dragon_rider. Lots of similarities.
 
-extern EInit D_us_801809C8;
+extern EInit g_EInitOruburosHead;
```
JoshSchreuder, line 6: Easter egg in the codebase, we even name them differently here too 😛

PR #3314, opened 2026-07-06
decompile BO6_CheckBladeDashInput

flafmg:
decompiled BO6_CheckBladeDashInput, sorry if i may have messed something!

### DIFF
--- a/src/boss/bo6/bo6.h
+++ b/src/boss/bo6/bo6.h
@@ -93,6 +93,4 @@ typedef enum RicSteps {
 
 extern PlayerState g_Ric;
 
-void BO6_CheckBladeDashInput();
-
 #endif // BO6_H
--- a/src/boss/bo6/bo6.h
+++ b/src/boss/bo6/bo6.h
@@ -93,4 +93,6 @@ typedef enum RicSteps {
 
 extern PlayerState g_Ric;
 
+void BO6_CheckBladeDashInput();
+
 #endif // BO6_H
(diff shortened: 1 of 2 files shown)


flafmg:
sorry the 3 commits to fix two stupid issues, now it should be correct!

src/boss/bo6/bo6.h
```diff
@@ -93,4 +93,6 @@ typedef enum RicSteps {
 
 extern PlayerState g_Ric;
 
+void BO6_CheckBladeDashInput();
```
Xeeynamo: No need to put this in the header

src/boss/bo6/richter.c
```diff
@@ -5,7 +5,14 @@ INCLUDE_ASM("boss/bo6/nonmatchings/richter", func_us_801B4BD0);
 
 INCLUDE_ASM("boss/bo6/nonmatchings/richter", func_us_801B4EAC);
 
-INCLUDE_ASM("boss/bo6/nonmatchings/richter", BO6_CheckBladeDashInput);
+void BO6_CheckBladeDashInput() {
```
Xeeynamo: It's preferable to mark functions as `static` when possible. Try to apply it. If the build is still sane then it's safe to commit as static

src/boss/bo6/bo6.h
flafmg: oh sure, thanks! i got a bit confused with it sorry

src/boss/bo6/richter.c
flafmg: thanks for the tips! i applied the changes and everything still compiles

Xeeynamo (approved):
Looks good. Thanks for the contribution!

PR #3316, opened 2026-07-07
Matching decomp for func_us_801B1590

Mamotromico:
[decomp.me scratch](https://decomp.me/scratch/4pLRW)

Unsure if this is adequate for a PR or if its better to wait for more understanding of what this is doing.
I picked this up based on the [function list](https://github.com/Xeeynamo/sotn-decomp/blob/gh-duplicates/functions.md).

Includes the extra Include for the context creation [as was discussed in the discord](https://discord.com/channels/1079389589950705684/1079395108501331990/1524067129463144618)


sozud:
This can be improved more but it's probably something like this https://decomp.me/scratch/VX4Mr

Mamotromico:
> This can be improved more but it's probably something like this https://decomp.me/scratch/VX4Mr

Interesting!
How did you reach the proposed ET_UNK? Is it based on other ET_ structures? I see some similarities with ET_Weapon, but I have no idea how I'd think of this.

src/boss/bo0/2D26C.c
```diff
@@ -31,7 +31,18 @@ INCLUDE_ASM("boss/bo0/nonmatchings/2D26C", func_us_801B0930);
 
 INCLUDE_ASM("boss/bo0/nonmatchings/2D26C", func_us_801B13A8);
 
-INCLUDE_ASM("boss/bo0/nonmatchings/2D26C", func_us_801B1590);
+void func_us_801B1590(s32 step) {
+    // Load pointer stored at Entity+0x80 (inside Ext+0x04)
+    u8* node = (u8*)g_CurrentEntity->ext.ILLEGAL.u32[1];
+
+    g_CurrentEntity->step = (u16)(step & 0xff);
```
bismurphy: This kind of code looks fake. Indeed, if you change the function to take a u8 step, then this can just be =step;

bismurphy:
This looks like a great first PR! Placeholder names are totally fine, especially as you are just starting. There will still be some adjustments before merging but a very solid start. Thank you for contributing! Please update to address the comments from sozud and me and we should be able to merge this soon.

src/boss/bo0/2D26C.c
Mamotromico: > This kind of code looks fake. Indeed, if you change the function to take a u8 step, then this can just be =step;

I think I understand what you mean, the s32 + bitmasking could be a decompilation "artifact" due to type sizes instead of the actual argument?

bismurphy: The decompiler is kind of bad. It often turns instructions into explicit C code that doesn't need to be there. In this case, the & 0xFF was likely not in the original C, and actually just represents the argument of the function being u8. And the cast to u16 is not needed because C automatically converts integer types. So we can remove those transformations which makes the code simpler while still matching.

Mamotromico:
Added the proposed struct as @sozud shown, just renamed the type to ET_B0_Unk to be a bit more specific than just Unk (since this code is from the boss/b0/ section)

Mamotromico:
The `(ET_B0_Unk)*` casts are not necessary to compile according to decomp.me, but avoid some warnings about incompatible pointers.

Seems like I messed up something on the compilation, figuring it out now.

Mamotromico:
Seems like adding `ET_B0_Unk` to `Entity.h` and the `Ext` union causes a checksum failure after build, should that be somewhere else instead?

Mamotromico:
Thanks to @bismurphy its compiling again! 
The suggested struct was too big so it was blowing up the union (and the checksum), it should work fine now.

sozud:
Lets fix it so that the psp version matches too, needs the step and step_s part moved https://decomp.me/scratch/tZoze

JoshSchreuder:
@sozud I think we can merge this and fix up PSP later so this doesn't get stale, what do you think?

sozud:
That's fine

JoshSchreuder:
Thanks for your contribution @Mamotromico!

PR #3317, opened 2026-07-07
Asset tool: fix padding at the end of GfxBanks for PSP

bismurphy:
Here is the data for RNO3_gfxBanks:

```
glabel RNO3_gfxBanks
    /* 2CA68 092640E8 F83F2609 */ .word D_pspeu_09263FF8
    /* 2CA6C 092640EC 10402609 */ .word D_pspeu_09264010
    /* 2CA70 092640F0 A0402609 */ .word D_pspeu_092640A0
    /* 2CA74 092640F4 F83F2609 */ .word D_pspeu_09263FF8
    /* 2CA78 092640F8 F83F2609 */ .word D_pspeu_09263FF8
    /* 2CA7C 092640FC F83F2609 */ .word D_pspeu_09263FF8
    /* 2CA80 09264100 F83F2609 */ .word D_pspeu_09263FF8
    /* 2CA84 09264104 F83F2609 */ .word D_pspeu_09263FF8
    /* 2CA88 09264108 F83F2609 */ .word D_pspeu_09263FF8
    /* 2CA8C 0926410C F83F2609 */ .word D_pspeu_09263FF8
    /* 2CA90 09264110 F83F2609 */ .word D_pspeu_09263FF8
    /* 2CA94 09264114 F83F2609 */ .word D_pspeu_09263FF8
    /* 2CA98 09264118 F83F2609 */ .word D_pspeu_09263FF8
    /* 2CA9C 0926411C F83F2609 */ .word D_pspeu_09263FF8
    /* 2CAA0 09264120 F83F2609 */ .word D_pspeu_09263FF8
    /* 2CAA4 09264124 00000000 */ .word 0x00000000
.size RNO3_gfxBanks, . - RNO3_gfxBanks
```

Notice that at the end there is an empty word. This is not a "true" member of the array. Instead, this is zero-padding because PSP aligns everything to 8 bytes. This way, the array ends at 2CAA8 instead of 2CAA4.

However, the assets tool parses this and produces this array:
```
u_long** OVL_EXPORT(gfxBanks)[] = {
    rno3_gfxbank_0,
    rno3_gfxbank_1,
    rno3_gfxbank_2,
    rno3_gfxbank_0,
    rno3_gfxbank_0,
    rno3_gfxbank_0,
    rno3_gfxbank_0,
    rno3_gfxbank_0,
    rno3_gfxbank_0,
    rno3_gfxbank_0,
    rno3_gfxbank_0,
    rno3_gfxbank_0,
    rno3_gfxbank_0,
    rno3_gfxbank_0,
    rno3_gfxbank_0,
    NULL,
};
```

It puts a NULL at the end, believing it to be real data. When US is built and reads this file, it introduces explicit NULL bytes at the end, but US doesn't need them, so it makes it look like the gfxBanks have different data in the two versions. They don't really. They just look like it, due to the padding.

This patch to the asset tool will not generate the NULL at the end of the array, meaning that the generated .h files are identical. Building still creates the padding bytes - we just no longer specify them in the C code.

There may be better ways to patch this, but I don't know what those might be.


JoshSchreuder:
Nice, likely means we can coalesce back cases where we had to `us/graphics_banks` / `psp/graphics_banks` in the assets files? There's a bunch

tools/sotn-assets/assets/gfxbanks/graphics.go
```diff
@@ -81,6 +81,12 @@ func ReadGraphics(r io.ReadSeeker, ramBase, addr psx.Addr, symbol func(addr psx.
 		}
 		addrGfxBanks = append(addrGfxBanks, addrGfxBankEntry)
 	}
+	// On PSP, the last element might just be padding out to the 8-align. Eliminate that if found.
+	lastBank := addrGfxBanks[len(addrGfxBanks) - 1]
+	if (lastBank == psx.RamNull){
```
JoshSchreuder, line 86: Unsure if maybe a platform check would be good here, see Xeeynamo's earlier PR https://github.com/Xeeynamo/sotn-decomp/pull/3223/changes#diff-68f2008768f387add6c3e5a3c05515d5a24f2d57bf93f5892b8a302e90f10902 in handler.go

bismurphy: No real need for it. GfxBanks never end in a null, so if we find one, we know it must be PSP padding.

bismurphy:
> Nice, likely means we can coalesce back cases where we had to `us/graphics_banks` / `psp/graphics_banks` in the assets files? There's a bunch

Yes, I'm planning to circle back on those and see if any can be un-split in the future. I could add that un-splitting to this PR if desired, just figured I should keep the PR minimal for now? Either way seems good.

JoshSchreuder:
Yep I think that's a good plan, especially since we're not sure how many are affected by the padding vs actually differ

JoshSchreuder (approved):

PR #3323, opened 2026-07-10
maria funcs

sozud:


Xeeynamo (approved):
Not a blocker, but we should change the prefix `Ric` with something else for the following PRs. We already use `Mar` for PSP. Maybe here we can use `Maria`, or maybe we could have `MarS` and `MarP` for the two playable Maria. WDYT?

sozud:
Ok, I'll follow up in #naming

PR #3324, opened 2026-07-10
decompile func_us_8019A98C

flafmg:


flafmg:
i think i shouldve put this on rcen_psp as well since this one comes from decomping the psp one, i forgot sorry!

src/st/rcen/e_shaft.c
```diff
@@ -1,7 +1,33 @@
...
+s16 func_us_8019A98C(s16 arg0, s16 arg1, s16 arg2) {
+    s16 v_s1;
+    s16 v_s0;
+
+    arg1 &= 0xFFF;
+
+    v_s1 = arg2 - arg1;
+    v_s0 = v_s1;
+
+    if (v_s1 > 0x800) {
```
sozud: This function is almost certainly rotation related, could do this
```
    if (v_s1 > ROT(180)) {
        v_s0 = v_s1 - ROT(360);
    }
    if (v_s1 < ROT(-180)) {
        v_s0 = v_s1 + ROT(360);
    }
```
https://decomp.me/scratch/uYEbi

sozud (approved):

src/st/rcen/e_shaft.c
flafmg: i was suspecting it was rotation related
for me it looks like an stepping function that turns one angle into another one by an max step, i did an function that is almost the same a while ago for a litle project, but i got a bit scared of trying and go naming and guessing it wrong lol

i will make these changes thank you!!

flafmg: just a question, ROT is already defined somewhere in the codebase or should i add it in e_shaft header?

JoshSchreuder: It's in our common header so you should be able to use it here with no further includes
https://github.com/Xeeynamo/sotn-decomp/blob/a4267165fb42f5b3c32d6d1cb899e91eccd291b3/include/common.h#L109

PR #3327, opened 2026-07-11
ovl export st_update.c and EntityDamageDisplay

Xeeynamo:
This PR comes with tons of small changes to enable future `OVL_EXPORT` work.

```c
#define OVL_EXPORT(x) MYSTAGE_##x
#include <stage.h>
```

moving the inclusion of `stage.h` after `OVL_EXPORT` allows us to use the macro in the common header, further enforcing rules for new overlays to use the macro.

This is part of a bigger effort to enable #3302

EDIT: it seems updating `UpdateStageEntities` triggered a cascade of failures due to collisions. So I am doing the rest now.


sozud (approved):
I skimmed through this because it's gigantic but seems reasonable

Xeeynamo:
> I skimmed through this because it's gigantic but seems reasonable

Sorry about that. Unfortunately it's a bit tricky to maintain those kind of pull requests small. Thanks for reviewing!

PR #3328, opened 2026-07-11
ovl export EntityExplosion, EntityPrizeDrop and friends

Xeeynamo:
Continuation to #3327


src/pc/sotn.c
```diff
@@ -236,9 +235,6 @@ void InitStrings(void) {
     for (int i = 0; i < LEN(g_EnemyDefs); i++) {
         g_EnemyDefs[i].name = AnsiToSotnMenuString(g_EnemyDefs[i].name);
     }
-    for (int i = 0; i < LEN(g_goldCollectTexts); i++) {
```
sozud, line 239: So we don't actually need to convert these?

sozud (approved):

src/pc/sotn.c
Xeeynamo: We do. but we need to come up with a new way to do so. And it will be broken until then.

Either we defer this to `InitStage*()`, and we only convert the strings once, or we adapt the functions that print these. As we wait to decide, I'm excluding this conversion for the time being.

PR #3331, opened 2026-07-11
e_room_bg for RNO0

bismurphy:
This one is a little weird, just like the original NO0.

I really wish splat would allow using C files that are "up a directory" in the src tree. If we could do that then we could do [0xABC, c, ../no0/e_room_bg]. Of course we can change the src path to go up one, but that's clunky because then we have to actively specify the rno0 overlay on every single file.


src/st/rno0/rno0.h
```diff
@@ -6,6 +6,8 @@
 #define STAGE_IS_RNO0
 #include <stage.h>
 
+#include <stage.h>
+
```
Xeeynamo, line 9:
```suggestion
#include <stage.h>
```

duplicated by accident?

bismurphy: I fixed this myself, and then when I rebased I got your fix another time. Not sure why it didn't show up as a conflict. Fixing.

Xeeynamo:
I changed a few things on a PR that just got merged. You'd need to use `OVL_EXPORT(InitializeEntity)` and `OVL_EXPORT(AnimateEntity)`

src/st/rno0/e_room_bg.c
```diff
@@ -1,4 +1,78 @@
...
+u8 anim4[] = {64, 2, 255};
+u8 anim5[] = {64, 47, 255};
+u8 anim6[] = {4, 34, 0};
+u8 anim7[] = {4, 35, 0};
+u8 anim8[] = {4, 36, 0};
+u8 anim9[] = {4, 37, 0};
+u8 anim10[] = {4, 38, 0};
+u8 anim11[] = {4, 39, 0};
+u8 anim12[] = {4, 40, 0};
+u8 anim13[] = {4, 42, 0};
```
gamezter: It won't always cause a mismatch for alignment reasons, but we typically import animations like these with either a `0, 0` terminator or a `0xFF, 0` terminator. `AnimateEntity` always reads 2 bytes every call so the anim array should have size multiple of 2.

Xeeynamo: We should use `POSE_END` and friends for animation arrays

bismurphy: Okay, pulled those in, thanks

Xeeynamo (approved):

bismurphy:
PSP seems to fail with the pose macros.

```
### mwccpsp.exe Compiler:
#      14: {-1, 0}
#   Error:      ^
#   too many initializers
### mwccpsp.exe Compiler:
#      15: {-1, 0}
#   Error:      ^
#   too many initializers
### mwccpsp.exe Compiler:
#      16: {0, 0}
#   Error:     ^
#   too many initializers
### mwccpsp.exe Compiler:
#      17: {0, 0}
#   Error:     ^
#   too many initializers
```

Am I doing something wrong or is this the first time PSP has used this POSE system and we are discovering it is incompatible?

JoshSchreuder:
@bismurphy change type to AnimationFrame, eg.
`AnimationFrame anim1[] = { {64, 2}, POSE_END};`

bismurphy:
> @bismurphy change type to AnimationFrame, eg. `AnimationFrame anim1[] = { {64, 2}, POSE_END};`

Doesn't work. AnimationFrame has members that are u16, but the data I have here is u8. Using AnimationFrame pads every integer out to 2 bytes. So this data which should be `40 02 FF 00` becomes `40 00 02 00 ff ff 00 00`

JoshSchreuder:
Sorry, `AnimateEntityFrame` should work here as the type?

PR #3333, opened 2026-07-11
ovl export st_common

Xeeynamo:
This should be the last large PR that introduces `OVL_EXPORT` to the hottest areas of the code base.

image

This solves the bug where NZ0 could load CEN entities via the colliding `ReplaceBreakableWithItemDrop`. A cascade of functions in `st_common` had to be renamed to make the change working.


sozud (approved):
Just skimmed again ought to be fine

PR #3335, opened 2026-07-12
pc: add overlays support

Xeeynamo:
I am very excited with the introduction of a new overlay system that allows to compile PS1 overlays into individual libraries, and let them load dynamically when required. Exactly as on PS1, PSP and Saturn.

All recent OVL_EXPORT PRs is also reverted (🪦). I should have re-evaluated this route first before committing to OVL_EXPORT.

Apologise @sozud for the massive PRs. You're been kind to keep an eye on this project while I was fluctuating on and off. And thanks @bismurphy, to steer us back to sanity and prevent the proliferation of silly decisions.

EDIT: For reviewing purposes, I'd suggest to review the individual commits past `Revert OVL_EXPORT`.


sozud:
Is this working on Windows? https://github.com/Xeeynamo/sotn-decomp/pull/995#issuecomment-1902026035

JoshSchreuder:
If you need someone with a Windows machine to try anything out let me know

Xeeynamo:
@sozud @JoshSchreuder still WIP, but I got it working on Windows. Please let me know if you have any problems

cmake/gen_export_def.cmake
```diff
@@ -0,0 +1,79 @@
...
+if(NOT OUT)
+    message(FATAL_ERROR "gen_export_def.cmake: OUT not set")
+endif()
+if(NOT DEFINED INPUTS)
+    message(FATAL_ERROR "gen_export_def.cmake: INPUTS not set")
+endif()
+
+set(_symbols "")
+
+foreach(_input ${INPUTS})
```
Xeeynamo, line 23: BTW I could've never done this alone. For transparency, I copied it straight from a coding agent

cmake/gen_export_def.cmake
```diff
@@ -0,0 +1,79 @@
+# Generates a linker .def file that re-exports every defined external symbol
```
sozud, line 1: What are the pros/cons of this being a python script in `tools`? Cmake doesn't seem to be well-suited for this

CMakeLists.txt
```diff
@@ -9,6 +9,12 @@ if (WIN32)
     set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 endif()
 
+if (MSVC)
+    # HACK: force DLL CRT to avoid heap corruption
+    cmake_policy(SET CMP0091 NEW)
+    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreadedDLL")
```
sozud: What DLL are we using that's mismatched?

Xeeynamo: With `/MT` (aka CRT statically linked), each EXE or DLL will have their own CRT library embedded to it. That means each resource created in one that holds an internal state (malloc heap, file handles) cannot be passed somewhere else. This is a known issue when sharing DLLs on Windows.

By default, CMake produces executables with /MT and DLLs with /MD. We want to compile both EXE and its DLLs all as `/MD` so they can share the CRT state and avoid unexpected crashes. Especially since SOTN overlays have this issue of bidirectional communication with the main executable.

It's also an opportunity to reduce the file size of the executables. For taking out the debugging runtime library it reduces the EXE size from 9.0MB to 2.6MB. The game has almost 100 overlays between playable characters, stages and weapons. And I can't imagine having the risk of statically linking the runtime library plus the cost of the total size.

cmake/gen_export_def.cmake
Xeeynamo: The only Con I can think of is Python not shipped by default on Windows, and it would be a less dependency to think about for users who just want to build the game themselves. Consider, as today, python is not required at all to build the game on PC. Even the asset extraction is done in Go.

The Pro is that CMake never leaves its own environment to perform this additional task. Since I doubt this file will ever get changed, I'd suggest to leave it as it is.

CMakeLists.txt
sozud: My understanding is that if we force MultiThreadedDLL even for debug builds then we lose the crt debug functionality like heap guard bytes / checks and leak detection, which might be useful for catching the numerous memory corruption issues we have

cmake/gen_export_def.cmake
sozud: Go would also be fine if we require Go on windows. It's not a big deal though to leave it as-is

CMakeLists.txt
Xeeynamo: So you suggest `/MT` for Release/Public builds and `/MD` for Debug/Local builds?

sozud: The point is you want to avoid static linking right? So we could use MultiThreadedDLL for release builds and MultiThreadedDebugDLL for debug builds

Xeeynamo: Okay it's clear. I agree with that :+1:

Xeeynamo: solved in bd93e2b

src/st/nz0/e_subweapon_container.c
JoshSchreuder: If we're ifdefing this code anyway I think we should remove this hack, since it was just relying on compiler behaviour to optimise out stuff?

Xeeynamo: I was thinking about it. Thanks for validating that. I went a step further and made this: df5c683

EDIT: it apparently fails, so I had to remove the `USE_PLAYER_MARIA`: 6ec4989

JoshSchreuder:
Have checked on Windows and this works for me, builds the main executable with overlays as separate DLLs
image

Was able to run it and progress through title screen to prologue, and also jumping in directly to CEN overlay with no issues

JoshSchreuder (approved):

Xeeynamo:
I'll wait a second approval (cc @sozud) since this is a quite big change

sozud (approved):
This seems to work on mac, I haven't tested the other platforms. Sound effects seem to be regressed but I was able to load and play wrp/nz0/cen.

PR #3346, opened 2026-07-15
Fix erroneous data split in `st_update` across few overlays

JoshSchreuder:
Something random I noticed while doing the most recent overlay imports


src/st/st_update.h
```diff
@@ -1,13 +1,11 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include <game.h>
 
-#if !defined(STAGE_IS_CAT) && !defined(STAGE_IS_NZ1) &&                        \
-    !defined(STAGE_IS_ARE) && !defined(STAGE_IS_RARE) &&                       \
-    !defined(STAGE_IS_RNO3) && !defined(STAGE_IS_RCEN)
+// NOTE: take care here, this may appear merged with the end of the previous
+// data split but should reside here!
 static u16 unused[] UNUSED = {
```
JoshSchreuder, line 6: Since this was merged with all different splits data depending on the overlay, it's clear it actually should reside here

sozud (approved):

PR #3350, opened 2026-07-15
Extract US NZ1 Clock Tower asset data

bismurphy:
Nothing crazy here, just pulling in data. Purpose was to be able to view the assets for Skull Lord, since there are some mysteries related to the red eyes.

We run into https://github.com/Xeeynamo/sotn-decomp/issues/2401 once again, so it is not possible to finish the data on this one. If it was possible to finish I would probably do PSP as well, but for now I'll just put in what is possible for now.


Xeeynamo (approved):
> We run into #2401 once again,

I'll look into it this week-end.

PR #3353, opened 2026-07-15
flatten USE_ENEMY_3 macro

Xeeynamo:
`USE_ENEMY_3` was only used in two overlays. It was pointless to have it shared in a common file.

### DIFF
--- a/src/st/cen/e_init.c
+++ b/src/st/cen/e_init.c
@@ -61,10 +61,8 @@ PfnEntityUpdate OVL_EXPORT(EntityUpdates)[] = {
 };
 
 EInit OVL_EXPORT(EInitBreakable) = {ANIMSET_OVL(1), 0, 0x00, 0x000, 0};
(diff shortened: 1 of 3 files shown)


src/st/cen/e_init.c
```diff
@@ -61,10 +61,8 @@ PfnEntityUpdate OVL_EXPORT(EntityUpdates)[] = {
 };
 
 EInit OVL_EXPORT(EInitBreakable) = {ANIMSET_OVL(1), 0, 0x00, 0x000, 0};
-#ifndef VERSION_PSP
-#define USE_ENEMY_3
-#endif
 #include "../e_init_common.h"
+EInit g_EInitEnemy3 = {ANIMSET_OVL(1), 0, 0x00, 0x000, 3};
```
Xeeynamo, line 65: unused on PSP, so it gets optimized out automatically

JoshSchreuder (approved):

PR #3357, opened 2026-07-16
Decompile `sat/CheckCollision`

gamezter:


src/saturn/game.c
```diff
@@ -1984,8 +1984,352 @@ void PlaySfxPositional(s32 sfxId) {
...
+            if (res->unk1C + res->unk20 < 0x100000) {
+                res->unk14 = res->unk18 = res->unk1C + res->unk20 - 0xF0000;
+                res->effects = EFFECT_UNK_8000 | EFFECT_SOLID;
+            } else {
+                res->effects = EFFECT_UNK_8000;
+            }
+            break;
+
+        case 0x08:
+            offset = 0x100000;
```
sozud, line 2098: We could use I_TO_FIX on most of the numbers in this function but it's not critical https://github.com/Xeeynamo/sotn-decomp/blob/787470018993684ca40f2c2718517269c4a8fdea/include/common.h#L104

sozud (approved):

PR #3358, opened 2026-07-16
pc: add gameplay recording feature

Xeeynamo:
### TL;DR

Play game. Close it, then do `./build/pc/sotn --replay recording.bin`

### Explanation

After some long testing and thinkering, I am ready to propose this feature.

As the PC port goes public, we will get frequent reports on bugs and crashes we want to easily reproduce. The game will now generated by default a `recording.bin` file (overridable with `--record <file_name>`) that contains the following info:
* Operating system and processor architecture
* Selected parameters via CLI, mainly `--stage` and `--player`
* A dump of `g_Status`, `g_CastleMap`, `g_Settings`, `g_MenuNavigation` and some additional scalars from their own isolated global vars
* A dump of input from PAD0 per frame
* A dump of some game metadata per frame to detect drifting

Each recording file is meant to be reproducible. Invoking sotn with `--replay <file_name>` will set all global variables in the set-up phase, then it will simulate the gamepad to reproduce accurately what the player was doing.

If the game deviates from the recorded metadata, it will drop a warning mentioning the recording has drifted. The recording will continue playing as usual, but it's very likely the remaining recording will be invalid. This can generally happens if either a bug (not crash) was fixed, or the game behavior changes due to a regression. Meaning the drifting detection can also be used as a regression suite.

Replaying can be executed with `--replay-fast`, a flag only valid on a valid replay file. The framerate runs uncapped until the end of the recording. A good couple of minutes of recording can be replayed in less than two seconds (depending of hardware), which is useful for CI runs or frequent replays when hunting a bug or crash.

The flag `--replay-and-exit` can be used on CI. A new exit code `-1` is issued if the recording has drifted. So it's easier to fail on CI when a change in game behavior occurred.

A recording file is always generated, and replaces the existing one. This is important, because an user will want to report the crash when it happens instead of having to replay a second time in an attempt to reproduce it. This is especially useful for bugs difficult to reproduce, and comes at the little cost of `recording.bin` always hanging around.

The recording format is extendible. I've been inspired by the PNG format, where individual chunks containing their own metadata can be self-contained. If later on we decide to include the GIT revision and a memcard dump in the recording, previous recordings will still be replayable. Unrecognized chunk types will be skipped, and a `version` field for now equal to `1` can be used for later revisions and expansions.

Input and drifting info must intersect per frame. We cannot record them in separate chunks, or a game crash will just make nasty to reconstruct the file. The metadata is flushed at every frame, and the recording container is designed to know if a recording was flushed correctly with a clean exit or interrupted by a runtime crash.

Files are pretty big. A two minutes recording is 1MB, but zipped is just 13KB. I am not worried on the weight of sharing or store compressed recording files.


bismurphy:
Seems neat. Only thought I would add is that having only one recording file might not be the best. When I get a crash my first thought is often "was that a random thing, or a reliable failure. Let's try that again and see if it crashes again". But trying again will delete the log from the first run. Of course with knowing this system exists, I would make sure to preserve the log as precious, but not all users will know.

Could make sense to keep 10 recordings with timestamps, and delete the oldest on each run? Either way this is fine.

Xeeynamo:
> Seems neat. Only thought I would add is that having only one recording file might not be the best. When I get a crash my first thought is often "was that a random thing, or a reliable failure. Let's try that again and see if it crashes again". But trying again will delete the log from the first run. Of course with knowing this system exists, I would make sure to preserve the log as precious, but not all users will know.
> 
> Could make sense to keep 10 recordings with timestamps, and delete the oldest on each run? Either way this is fine.

I agree with you, and that's what happened to me a couple of times too. The problem is introducing a reliable recording rotation system that reliably avoid polluting the game folder. Maybe we could have `recording0.bin`, `recording1.bin` until `9`, and at every new run the existing recordings are rotated by +1? Or maybe a simple `recording.bin.bak` with just one backup copy. Not sure. I'm curious about other comments about this.

bismurphy:
Rotating 0-9 seems great if it can work.

JoshSchreuder:
I agree having a rotating recording sounds useful, also putting them in a subdirectory maybe (eg. /recordings)

Xeeynamo:
@bismurphy @JoshSchreuder added a rotation system with max 4 recordings. Plus, I reduced the recording size to 56 bytes per frame. So an hour of gameplay will generate just a 12MB file size.

JoshSchreuder (approved):

PR #3365, opened 2026-07-17
pc: fix st0 crashes with new PRIM_DR_ENV macro

Xeeynamo:
Straighforward fix for a hard problem to track. `LOW(prim->r1)` was reading 32-bit pointers on 64-bit targets, causing a crash. This still doesn't prevent pointer corruption if the same primitive gets `x2` and `y2` replaced, but so far it didn't happen. 

Let's keep an eye close to this pattern for the next coming PRs.

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -84,6 +84,13 @@
 #define HIHU(x) (((u16*)&(x))[1])
 #define LOW(x) (*(s32*)&(x))
 #define LOWU(x) (*(u32*)&(x))
+
+#ifdef VERSION_PC
+// on 64-bit builds, this needs to read 64-bit, not 32-bit with LOW
+#define PRIM_DR_ENV(prim) (*(DR_ENV**)&(prim)->r1)
+#else
+#define PRIM_DR_ENV(prim) ((DR_ENV*)LOW((prim)->r1))
+#endif
 #define F(x) (*(f32*)&(x))
 #define POS(x) (*(Pos*)&(x))
 #define UV(x) (*(uvPair*)&(x))
--- a/src/st/dai/e_bell.c
+++ b/src/st/dai/e_bell.c
@@ -257,7 +257,7 @@ void EntityBell(Entity* self) {
     posX = playerPtr->posX.i.hi - tempPosX;
     posY = tempPosY + 124 - playerPtr->posY.i.hi;
     if ((u16)(posY) < 80 && (abs(posX) < 24)) {
-        drEnv = (DR_ENV*)LOW(prim->r1);
+        drEnv = PRIM_DR_ENV(prim);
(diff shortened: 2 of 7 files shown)


include/common.h
```diff
@@ -84,6 +84,13 @@
 #define HIHU(x) (((u16*)&(x))[1])
 #define LOW(x) (*(s32*)&(x))
 #define LOWU(x) (*(u32*)&(x))
+
+#ifdef VERSION_PC
+// on 64-bit builds, this needs to read 64-bit, not 32-bit with LOW
+#define PRIM_DR_ENV(prim) (*(DR_ENV**)&(prim)->r1)
+#else
+#define PRIM_DR_ENV(prim) ((DR_ENV*)LOW((prim)->r1))
+#endif
```
JoshSchreuder, line 93: Is this the same issue with `g_Player.unk20`?
https://github.com/Xeeynamo/sotn-decomp/blob/5434ab3986f2398ae6429fb611b08377cf052e46/src/maria/pl_subweapons.c#L1135-L1139

Just curious if we need to spot fix this or if there is an alternate way to fix stuff like this via the compilation process

Xeeynamo: Unfortunately it's much trickier here. If we were to apply the same `g_Player.unk20` fix, `Primitive*` should've be expanded by 4 bytes more. But that breaks the alignment with all the other Primitives types and POLY_GT4.

JoshSchreuder: I might be misunderstanding, but perhaps we should have a separate Primitive typedef that replaces r1/g1/b1/p1 with `DR_ENV* foo`, and use it in cases where this is accessed as a DR_ENV* rather than casting? Wouldn't that mean the alignment would work on PC since the struct automatically gets expanded on larger pointer architectures and avoid potential corruption of x2/y2?

Similarly making `g_Player.unk20` an `Entity*` rather than an s32?

Probably a larger change though so I'm ok getting this merged in as a first step

JoshSchreuder (approved):

include/common.h
Xeeynamo: @JoshSchreuder yes, that would be the ideal path. But we cannot create a new `Primitive` struct that has a different byte size than the original `struct Primitive` or the PS1 SDK `struct POLY_GT4`. Everything needs to be carefully aligned and kept in sync.

Expanding `g_Player` is easier because the struct doesn't need to overlap with anything else.

We could make a new struct that replaces r1/g1/b1/p1/x2/y2 with `DR_ENV*` for 64-bit targets, but then it would get mis-aligned with 32-bit targets.

I don't think we can do better than this I am afraid. Agree, let's merge this. But let's keep the discussion open if necessary.

PR #3369, opened 2026-07-17
nz0: fix collision with EntityFloorButton

Xeeynamo:
In nz0, the button that activates and de-activates the floor spikes did not work. What happens is `GetPlayerCollisionWith` returns a `u8`, but since there's no forward declaration the game assumes it returns an `int`. That is fine on PS1, as a `return 0;` essentially does a `addiu $v0, zero, zero`. But on x86_64 it only sets to zero the lowest 8 bits of the returned value. The caller will then receive non-sense and the `collision == 0` condition never occurs.


bismurphy:
I would recommend including a comment on the include explaining why it needs to be there conditionally. Even if it just says "// see PR #3369". Theoretically this is accessible by git blame, but sometimes the chain of commits can be lost when files move or other changes happen. The code file itself is the best way to keep track of these changes.

JoshSchreuder (approved):

Issue #3347, opened 2026-07-15
Idea: auto-lint/document item drops


JoshSchreuder:
I am not sure if it is possible or easy with our lint or assets tooling but it would be nice to have item drops in an overlay self documenting.

For example, rather than
https://github.com/Xeeynamo/sotn-decomp/blob/master/src/st/are/d_prize_drops.c

```c
u16 OVL_EXPORT(PrizeDrops)[] = {
    0x000C, 0x0084, 0x0104, 0x015F, 0x0086, 0x0126, 0x00BA,
#ifndef VERSION_PSP
    0x00F3
#endif
};
```

We could turn this into an item enum like

```c
u16 OVL_EXPORT(PrizeDrops)[] = {
    ITEM_HEART_VESSEL, ITEM_SHIELD_ROD, ITEM_MOON_ROD, 
    ITEM_BLOOD_CLOAK, ITEM_KNIGHT_SHIELD, ITEM_LIBRARY_CARD, 
    ITEM_GREEN_TEA,
#ifndef VERSION_PSP
    ITEM_HOLY_SWORD
#endif
};
```

The algorithm should be roughly speaking (there may be additional rules):

- Take number
- If < 128, `EntityPrizeDrop`
    - If 0 - 1 - heart pickup
    - If 2 - 11 - gold
    - If 12 - heart vessel
    - If 13 - dummy, unused?
    - If 14-22 - subweapon
    - If 23 - life vessel
    - Else no item mapped
- If >= 128 `EntityEquipItemDrop`
    - Subtract 128
    - If < NUM_HAND_ITEMS (169) index into g_EquipDefs
    - If >= NUM_HAND_ITEMS (169), subtract 169 and index into g_AccessoryDefs

As seen in [e_collect.h](https://github.com/Xeeynamo/sotn-decomp/blob/b9d6626ffa085d8e474018652e7d1ba0b2a50eb2/src/st/e_collect.h)

We might get away with simply hardcoding an enum here and using that to document.

This would be useful for modding other items into the game / replacing items in overlays

JoshSchreuder:
`ITEM_MOON_ROD` doesn't appear in Colosseum so it's possible I may have miscalculated something here, or it was originally planned but unused as an item drop? Anyway, I did this just as an example

PR #3374, opened 2026-07-18
Add autolint for item drops and document some item drop usages

JoshSchreuder:
Implements #3347

In doing this I found a few incorrect data splits which I have fixed, and documented some places in the code which drop items

I cross matched the linted prize drops against https://breezewiki.com/castlevania/wiki/Symphony_of_the_Night_Inventory and https://guides.gamercorner.net/sotn/items/ and it looks to be correct. I have put notes against cases where the item drops contain an item that I am pretty sure does not appear in the stage and was likely unused.


config/splat.us.borbo3.yaml
```diff
@@ -47,8 +47,7 @@ segments:
       - [0x4E0, .data, e_red_door_tiles]
       - [0x510, .data, e_lock_camera]
       - [0x590, data] # e_breakable
-      - [0x5F0, .data, d_prize_drops]
-      - [0x5FC, data]
+      - [0x5F0, data, rbo3] # medusa
```
JoshSchreuder, line 50: This one's a very odd one, looks like the PrizeDrops address is pointing to a Medusa animation rather than an empty array in BSS which is what should be there for boss stage. This can be seen in https://github.com/Xeeynamo/sotn-decomp/blob/cffeca0eb371b3bfc11af70f2fbfa5c78ac4e108/src/boss/rbo3/rbo3.c#L185 as well

I've mapped the symbol for now, this will need some handling down the track

Xeeynamo (approved):
Looks good!

PR #3377, opened 2026-07-18
pc: make scratchpad global

Xeeynamo:
Proposed here: https://discord.com/channels/1079389589950705684/1526712290450079744/1528013957884022876


JoshSchreuder:
Think @sozud had some concerns with this on Discord?

https://discord.com/channels/1079389589950705684/1526712290450079744/1528017573315088554

Are we ok to merge this?

sozud:
I think it's just a tradeoff. The scratchpad persists across functions on psx so this is the more authentic behavior. It depends more if we're trying to suss out more memory and uninitialized variable bugs or not. I think the original motivation here is that people were complaining about too many #ifdefs. For me I'd rather have the local variable and the #ifdef so UB sanitizer can work to find unitialized variables. However making this a toggle involves #ifdefs so doesn't solve the original complaint.

JoshSchreuder:
If stale data left in SP is a concern, perhaps we could add code to the overlay load/unload to zero it in between stages?

EDIT: Think I misread, I guess it is still a problem for functions within the same overlay using scratchpad

Xeeynamo:
If we need to investigate a bug in an entity that uses the scratchpad, it would always be possible to temporarily add a `u8 g_Scratchpad[SP_LEN];` in the function to make a new local SP. It's a trade-off yes, but so far I did not see any advantage of always having SP local. In contrast, with this PR it becomes very easy to import any entity, while keeping code clean.

Xeeynamo:
I am still getting vague feedback on this PR. Is there anything that prevents this from being merged? @sozud @JoshSchreuder @bismurphy

sozud (approved):
I disagree with this change because I think leaving `sp` local to each function and uninitialized would help us find bugs, but I'm not going to block a merge if others prefer this approach.

bismurphy:
> I disagree with this change because I think leaving `sp` local to each function and uninitialized would help us find bugs, but I'm not going to block a merge if others prefer this approach.

I don't think there are any known bugs, so we would be talking about trying to find bugs that potentially don't even exist. I also don't know why unrealistically duplicating scratch pad between functions would help find bugs. If there is any uninitialized behavior that could always be found by using dedicated testing code rather than modifying the mainline.

I am supportive of making SP global, but I do not know enough about the PC port to be able to give intelligent approvals to PRs concerning it, so I will abstain from giving an official approval, but for clarity, I have no opposition to this PR and it seems like a good step to me.

PR #3378, opened 2026-07-19
Add some further item drop documentation

JoshSchreuder:
Following on from my PR #3374 I have documented some extra parts of the code


src/config_us.h
JoshSchreuder, line 1: Did this programatically, I did a spot check of various random enemies and their drops appeared to be correct

Xeeynamo (approved):

PR #3380, opened 2026-07-19
Dedupe most PSP `top`

JoshSchreuder:
Just the large cutscene function remaining, and data.

I'll work on importing data like sprites and rooms after this


Xeeynamo (approved):
Thanks!

sozud (approved):

PR #3383, opened 2026-07-19
sat: import more data

sozud:
I know this is enormous, cutting this up into reviewable parts is probably a waste of time because I'd have to roll back saturn-splitter fixes and try to cherry-pick individual edits to the yamls and stuff in the C files etc. A bunch of this is generated with tooling I'm planning to clean up and add later.


Xeeynamo (commented):
Cool if it matches with this large amount of data. Is the tool you're planning designed to import Saturn data specifically?

BTW there's a lot of overlapping data between PS1 and Saturn. I noticed symbols are not sychronized. For example `g_RoomBossTeleports` in Saturn, is `D_800A297C` on PS1. Ideally we'd want to compare one codebase with the other since they're separate, but symbols must be the same.

It's not a change I am asking to do as part of this PR. But I have to ask if it's intended to be done as some point? It's a big PR, and we might forget.

sozud:
Yeah I'd like to merge these better. I'm still figuring out the differences. Some things have changed like how sprites and tilemaps work as well as cutscenes. I made an issue to track the one you mentioned, anything else you saw that needs to be synced can be put here https://github.com/Xeeynamo/sotn-decomp/issues/3384

src/saturn/warp/laydef.c
```diff
@@ -0,0 +1,74 @@
...
+extern LayoutEntity g_WarpVLayout04[];
+extern LayoutEntity g_WarpVLayout05[];
+extern LayoutEntity g_WarpVLayout06[];
+extern LayoutEntity g_WarpVLayout48[];
+extern LayoutEntity g_WarpVLayout49[];
+extern LayoutEntity g_WarpVLayout50[];
+extern LayoutEntity g_WarpVLayout51[];
+extern LayoutEntity g_WarpVLayout52[];
+
+LayoutEntity* g_WarpLayoutVertical[53] = {
```
Xeeynamo, line 53: This stuff is extracted to `assets/` and generated as code in `src/*/gen`. Is the Saturn build going to hardcode these?

sozud: Right, I thought about whether to use the tool or not. My thinking was that I wasn't sure if there would be differences or not. In my mind this PR is just a first pass to establish what patterns are going on in the Saturn port. Now that a large chunk of the data is human-readable we can start polishing it more.

Xeeynamo: I don't anticipate our current tooling can support the Saturn structure at all, due to Big Endian and many more reasons. It makes sense this to be a in-between phase.

Xeeynamo (approved):
Giving my approval. I'd not want to wait for these kind of PRs, but after this gets merged I'd like to also know the thoughts of @gamezter in this. Ideally we'd come up with a nice clean project structure in a near future.

PR #3391, opened 2026-07-20
Decompile `sat/SubDisp`

gamezter:


src/saturn/game.c
```diff
@@ -299,8 +307,208 @@ INCLUDE_ASM("asm/saturn/game/f_nonmat", f6072BCC, func_06072BCC);
...
+    case 2:
+        prevMusicId = currentMusicId;
+        currentMusicId = 0;
+        DAT_0605cd70.unk0 += 3;
+    case 3:
+        func_06064628(g_PlayableCharacter);
+        if (g_PlayableCharacter == 0) {
+            DAT_06085d5c = g_Settings.windowColors[0];
+            DAT_06085d60 = g_Settings.windowColors[1];
+            DAT_06085d64 = g_Settings.windowColors[2];
```
gamezter: This should be 
```
g_StatusPortraitColorOffset[0] = g_Settings.windowColors[0];
g_StatusPortraitColorOffset[1] = g_Settings.windowColors[1];
g_StatusPortraitColorOffset[2] = g_Settings.windowColors[2];
```
but that causes linker errors that I can't figure out

src/saturn/game/miscend.c
```diff
@@ -28,24 +28,14 @@ StatusHudSpriteOffset g_AlternateStatusHudSpriteOffsets[8] = {
     {5, 1}, {17, 0}, {17, 10}, {6, 0}, {6, 0}, {6, 0}, {6, 0}, {0, 0},
 };
 
-u16 g_AlternateStatusHudPrimitiveLayout[34] = {
+u16 g_AlternateStatusHudPrimitiveLayout[33] = {
     0x0002, 0x0018, 0x0004, 0x0074, 0x0019, 0x0074, 0x0010, 0x001D, 0x0018,
     0x001D, 0x0020, 0x001D, 0x0014, 0x0029, 0x002D, 0x0015, 0x0000, 0x1429,
     0x142D, 0x1224, 0x1425, 0x1128, 0x1425, 0x1425, 0x192E, 0x132A, 0x1125,
-    0x1429, 0x0000, 0x0000, 0x0064, 0x0000, 0x0064, 0x0000,
+    0x1429, 0x0000, 0x0000, 0x0064, 0x0000, 0x0064,
```
gamezter, line 35: @sozud this is probably size 29, same as `g_StatusHudPrimitiveLayout`

src/saturn/game.c
sozud: Checked this locally, the linker errors seem to be right, this changes the size of the object. Did you check that change in decomp.me? This is probably a a bad data import and it's supposed to not be an array.

src/saturn/game/miscend.c
sozud: It's used in the function 06076a04 but the ghidra output is hard to interpret. The data import was a rough first pass so it's probable a lot of things need adjustments.

src/saturn/game.c
gamezter: I had not : https://decomp.me/scratch/HQ2t9

I think these might actually be 3 separate `s32`s

sozud (approved):

PR #3394, opened 2026-07-20
no3: fix EntityBoneScimitarParts OOB

Xeeynamo:
Another example where this change is neither related to `VERSION_PC` and it would require too many surgical changes to match the exact PS1 behaviour (which would be different between US, JP, EU and other builds anyway).

We need to agree on a naming scheme of some sort for these. CC @bismurphy

### DIFF
--- a/src/st/e_bone_scimitar.h
+++ b/src/st/e_bone_scimitar.h
@@ -300,7 +300,14 @@ void EntityBoneScimitar(Entity* self) {
 void EntityBoneScimitarParts(Entity* self) {
     if (self->step) {
         if (--self->ext.skeleton.explosionTimer) {
+#if defined(VERSION_PC)
+            // BUG! BONE_SCIMITAR_SPECIAL assigns a self->params value bigger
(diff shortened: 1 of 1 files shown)


sozud (approved):
I can't remember where but I fixed one of these previously by looking up the neighboring values and extending the array. We could come back to it and audit if the behavior change is noticeable

I think I'd probably have `#ifdef FIX_UNINIT_BUGS` and `#ifdef FIX_OOB_BUGS`, classify by the bug type? That would help track these better

bismurphy:
What about FIX_UB? For any situation where we are fixing undefined behavior (which is almost always done by defining the behavior)? This would capture any such necessary changes without needing to flag each one as OOB, uninitialized, etc

Xeeynamo:
> What about FIX_UB? For any situation where we are fixing undefined behavior (which is almost always done by defining the behavior)? This would capture any such necessary changes without needing to flag each one as OOB, uninitialized, etc

I'll make it a `FIX_UB` as a separate PR :+1:

PR #3399, opened 2026-07-21
Add support to extract player sprites

JoshSchreuder:
Importing TOP assets require a change to the assets tool to handle the special player sprite banks which are in a different format.

The TOP spritebanks `- [0x34, sprite_banks, sprite_banks]` has its second bank as the Richter sprite definition in the different format, making the tool barf when trying to import all the other sprite banks. Presumably this is done to support Richter in the boss fight / cutscene inside TOP.

image

These had previously been imported manually into .c files in RIC/MARIA, but I have updated assets config to auto import them instead to further utilise the same logic and prove it's matching.

Most likely this same logic will be required in CEN/RCEN assets on PSP side at least.


tools/sotn-assets/assets/spriteset/spriteparts.go
```diff
@@ -37,13 +44,32 @@ func readSpriteParts(r io.ReadSeeker, baseAddr, addr psx.Addr) (spriteParts, dat
...
-	if err := binary.Read(r, binary.LittleEndian, sprites); err != nil {
-		return nil, datarange.DataRange{}, err
-	}
+
 	alignment := 4
 	if sotn.GetPlatform() == sotn.PlatformPSP {
 		alignment = 8
 	}
+
+	if count >= 0x8000 {
```
JoshSchreuder, line 53: @Xeeynamo I've special cased this though it's possibly a bit brittle. I don't see a reasonable case where `count` would be this high though and no existing sprite banks have an issue with this so it's probably ok?

tools/sotn-assets/assets/spriteset/spriteparts.go
```diff
@@ -37,13 +44,32 @@ func readSpriteParts(r io.ReadSeeker, baseAddr, addr psx.Addr) (spriteParts, dat
...
+		// The value read as count is actually the Flags field.
+		var rest [3]int16
+		if err := binary.Read(r, binary.LittleEndian, &rest); err != nil {
+			return nil, datarange.DataRange{}, err
+		}
+		s := sprite{
+			Flags: count, // includes the 0x8000 bit
+			X:     rest[0],
+			Y:     rest[1],
+			Width: uint16(rest[2]), // Height..Bottom stay 0
```
JoshSchreuder, line 65: I have kind of hacked this into the existing `sprite` struct and only used the first 4 members, it might make sense to implement this another way with a special struct type here but this works?

src/ric/spritebank0.c
```diff
@@ -1,464 +0,0 @@
...
-    D_80153908, D_80153910, D_80153918, D_80153920, D_80153928, D_80153930,
-    D_80153938, D_80153940, D_80153948, D_80153950, D_80153958, D_80153960,
-    D_80153968, D_80153970, D_80153978, D_80153980, D_80153988, D_80153990,
-    D_80153998, D_801539A0, D_801539A8, D_801539B0, D_801539B8, D_801539C0,
-    D_801539C8, D_801539D0, D_801539D8, D_801539E0, D_801539E8, D_801539F0,
-    D_801539F8, D_80153A00, D_80153A08, D_80153A10, D_80153A18, D_80153A20,
-    D_80153A28, D_80153A30, D_80153A38, D_80153A40, D_80153A48, D_80153A50,
-    D_80153A58, D_80153A60, D_80153A68, D_80153A70, D_80153A78, D_80153A80,
-    D_80153A88, D_80153A90, D_80153A98};
-static s16 D_80153400[] = {0x00 | 0x8000, -16, -24, 0};
```
JoshSchreuder, line 253: Generated files output these in the same format as earlier
Image

tools/sotn-assets/info_test.go
```diff
@@ -69,6 +69,7 @@ func TestStagesCompatibility(t *testing.T) {
 		"disks/us/BOSS/BO2/BO2.BIN",
 		"disks/us/BOSS/BO3/BO3.BIN",
 		"disks/us/BOSS/BO5/BO5.BIN",
+		"disks/us/BOSS/BO6/BO6.BIN",
```
JoshSchreuder, line 72: Makes sense this would be working now as well, since this is the Richter boss and probably had the same sprite bank imports issue.

tools/sotn-assets/assets/spriteset/spriteparts.go
Xeeynamo: that's completely fine, it's unlikely such a high amount will ever take place

Xeeynamo (approved):
This is a great addition. Thanks!

PR #3402, opened 2026-07-21
Add FIX_UB macro

Xeeynamo:
Quite an important macro that is intended to separate PC exclusive code paths to undefined behaviour fixes.

I would like to have a review from both @bismurphy on this one, in addition to at least another extra review to ensure we are all on-board on what `FIX_UB` is intended for.

### DIFF
--- a/include/stage.h
+++ b/include/stage.h
@@ -208,12 +208,15 @@ void HitDetection(void);
 s32 Random();
 #ifdef VERSION_PC
 void CreateEntityFromEntity(u16 entityId, Entity* source, Entity* entity);
+#endif
 
+#ifdef FIX_UB
 // BUG! on non-MIPS processors, not adding this forward declaration results to
 // the caller receiving an undefined numbers that can be random and
 // mis-representative of the actual return value.
 u8 GetPlayerCollisionWith(Entity* self, u16 w, u16 h, u16 flags);
 #endif
+
 void DestroyEntity(Entity*);
 void DestroyEntitiesFromIndex(s16 index);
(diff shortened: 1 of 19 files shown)


bismurphy (approved):
This looks great. I particularly like the big comment block in version.h. Nice to have clear boundaries of what does and does not qualify for UB.

I also think most of the uses of FIX_UB should probably have a comment on them saying things like 1. What the undefined behavior is and 2. What the behavior is on the original game (since of course once the code compiles, all behavior is defined). But that is not necessary for this PR. Probably just moving forward. So for example "When you come into the room from the left side, the position overflows, and data is loaded from beyond the end of this array, which pulls from the cloud array that comes after it", or something.

But again, for now this seems like a perfect way to make this work.

I think maybe the "rules" for FIX_UB should include something to suggest a "minimal touch", in the sense of "This is only meant to fix major issues, including 1) Failure to compile 2) Crashes that do not exist in the original game 3) Major divergence between the original game and the ported code". This would make it clear that we are really just fixing undefined behavior in C, to make the ports perform like the original. But I think as long as we understand that as the way things are used, that should be fine.

This PR will go a long way to making the codebase stronger. Thanks for pulling this together.

JoshSchreuder:
LGTM too, I agree with @bismurphy that it would be good to annotate each instance we find with reasoning and what is being fixed for clarity too.

JoshSchreuder (approved):

Xeeynamo:
I'll continue going one by one to document some of these. I'll do it as a follow-up

PR #3404, opened 2026-07-21
Decompile `sat/InitRoomEntities`

gamezter:


src/saturn/game_0.c
```diff
@@ -5,8 +5,8 @@
 
 void PlaySfx(s32 sfxId);
 
-extern LayoutEntity* g_LayoutObjHorizontal;
-extern LayoutEntity* g_LayoutObjVertical;
+extern u16* g_LayoutObjHorizontal;
+extern u16* g_LayoutObjVertical;
```
gamezter, line 7: Couldn't get `InitRoomEntities` to match with these as `LayoutEntity*` after all :(

sozud (approved):

PR #3407, opened 2026-07-22
sat: import data and fix splits

sozud:


src/saturn/lib/mth/mth_fixd.s
JoshSchreuder, line 1: Maybe we should submodule this stuff as an external repo at some point?

JoshSchreuder (approved):

src/saturn/lib/mth/mth_fixd.s
sozud: Yeah made an issue https://github.com/Xeeynamo/sotn-decomp/issues/3412

PR #3408, opened 2026-07-22
Decompile a few `sat/game` functions

gamezter:


src/saturn/sattypes.h
gamezter, line 707: I noticed `Primitive` looks a lot like `SprSpCmd` : https://github.com/Xeeynamo/sotn-decomp/blob/7451138cd0fc06e83abf6564e2286a45db18b223/src/saturn/lib/spr/spr.h#L220-L237

src/saturn/saturn_sprite.h
```diff
@@ -98,7 +98,7 @@ typedef struct {
 typedef struct {
     SaturnSpriteImage* images;
     u16* palettes;
-    s16 allocationIndex;
+    u16 allocationIndex;
     u16 flags;
```
gamezter, line 108: @sozud where do these names come from ? A few functions in this PR use this struct (ex: `func_06066B74`) but the names don't make much sense to me

sozud: Ok the names aren't very good and based on preliminary research. My understanding is it's like this, we can use your new func as an example. Primitive is basically the VDP1 sprite format, used instead of the PS1 primitive types. allocationIndex could be better named but basically it's a slot in VDP1 ram, and each of those slots has additional subslots. vramSlotBase would maybe be a better name.
```

void func_060771D4(Primitive* prim, s32 arg1) {

    SaturnSpriteResource* res = DAT_06086388;
    u16* ptr;

    // DAT_0605aec0 is an array of CMDSRCA and CMDSIZE

    // DAT_06086388->allocationIndex repesents some fixed section probably for
    // certain types of effects or characters, arg1 is the offset into that section
    ptr = DAT_0605aec0[res->allocationIndex + arg1];
    prim->unk8 = ptr[0]; // get CMDSRCA (texture pos in vram)
    prim->unkA = ptr[1]; // get CMDSIZE   (size of texture)
    // palettes operate the same way where flags is a base address
    prim->unk6 = unkFunc(DAT_06086388->allocationIndex + arg1); // get palette 
    prim->unk6 = prim->unk6 & 0x8FFF | 0x4000; 
}
```

func_06066B74 is similar but it sets up a whole VDP1 command rather than the subset in func_060771D4

sozud (approved):

PR #3411, opened 2026-07-22
MAD: import missing assets + add to PC 

Xeeynamo:
I was unable to match https://decomp.me/scratch/hnVHF and siblings, so I am just stubbing the missing required function. Apparently it's responsible for randomly choosing/generating a room. Nothing special.

I managed to import a lot of the missing data and assets too.

image


JoshSchreuder (approved):

JoshSchreuder:
That one outstanding function is so annoying! Wish we had a version without optimisations to look at!

PR #3423, opened 2026-07-25
pc: add cutscene support

Xeeynamo:
I am proud to finally propose a working PS1 cutscene implementation working on PC with the minimum amount of changes possible.

The PS1 cutscene scripts are cursed. They store 20-bit pointers in 4 nibbles decoded by `EntityCutscene`, then jump in them. PSP solve this by changing the script and using a switch table for every resource kind (like portrait data). On PC this is solved by defining a pre-recorded key-value pair, where the key is is the PS1 resolved pointer, and the pair is its PC equivalent address and resource size. This key-value pair needs to be curated per-stage.

Instead of doing a bunch of `#ifdef`, the new `CS_NEXT` and `CS_PTR` macros hide the branching logic between PS1, PSP and PC. It was not possible to do it in all cases, but what I covered is enough to simplify the code.

The `cutscene_data.c` is now directly imported in `src/st/pc/stages/stage_*.c` to have access at its `sizeof(x)`.

The other option I considered over this proposed implementation is to use the PSP version of `cutscene.c`, and re-create the PS1 cutscene scripts ad-hoc. That might come at some point, especially if we desire to give the option of switching between the PS1 and the PSP scripts at runtime. But as of today, I am settling with this approach.

The changes have been tested against these recordings:
* [test_cutscene_nz0.bin.tar.gz](https://github.com/user-attachments/files/30371588/test_cutscene_nz0.bin.tar.gz)
* [test_cutscene_no3.bin.tar.gz](https://github.com/user-attachments/files/30371589/test_cutscene_no3.bin.tar.gz)
* [test_cutscene_cen.bin.tar.gz](https://github.com/user-attachments/files/30371590/test_cutscene_cen.bin.tar.gz)
* [test_cutscene_dre.bin.tar.gz](https://github.com/user-attachments/files/30371591/test_cutscene_dre.bin.tar.gz)
* [test_cutscene_st0.bin.tar.gz](https://github.com/user-attachments/files/30371592/test_cutscene_st0.bin.tar.gz)

My feedback loop to test all of them is:

```
./build/pc/sotn --replay-fast --replay-and-exit --replay test_cutscene_cen.bin
./build/pc/sotn --replay-fast --replay-and-exit --replay test_cutscene_dre.bin
./build/pc/sotn --replay-fast --replay-and-exit --replay test_cutscene_no3.bin
./build/pc/sotn --replay-fast --replay-and-exit --replay test_cutscene_nz0.bin
./build/pc/sotn --replay-fast --replay-and-exit --replay test_cutscene_st0.bin
```

I don't yet feel confident of pushing these recordings on CI. NO3 and NZ0 go out of sync, meaning the recording functionality is not yet fully deterministic.

image


src/pc/stages/stage_dre.c
```diff
@@ -12,10 +13,40 @@ extern LayoutEntity* entityLayoutVertical[];
...
+u8 D_801816C0 = 0;
+u8 D_801816C4[4] = {0};
+
+u8 OVL_EXPORT(cutscene_script)[] = {
+#include "../../st/dre/gen/cutscene_script_psx.h"
+    // unknown, most likely cutscene left-over data from another build
+    0x00, 0x00, 0x00, 0x00, 0x84, 0x21, 0x08, 0x80, 0x08, 0x80, 0x00,
+    0x08, 0x00, 0x85, 0x22, 0x0C, 0xC0, 0x08, 0x83, 0xFF, 0xFF, 0x00,
+    0x00, 0x00, 0x84, 0x23, 0x08, 0x80, 0x08, 0x80, 0x00, 0x08, 0x00,
+    0x85, 0x24, 0x0C, 0xC0, 0x08, 0x83, 0xFF, 0xFF, 0x00, 0x00,
```
Xeeynamo, line 30: I was still not able to identify these

src/pc/cutscene_pc.c
```diff
@@ -0,0 +1,43 @@
...
+                   symbols[i].psxAddr, prevEnd);
+        }
+    }
+}
+
+u8* CutsceneAddrToPc(u32 psxAddr) {
+    u32 addr = PC_ADDR(psxAddr);
+    for (int i = 0; i < _symbolCount; i++) {
+        u32 base = PC_ADDR(_symbols[i].psxAddr);
+        if (addr >= base && addr < base + _symbols[i].size) {
```
Xeeynamo, line 38: This is why the `struct CutsceneSymbol` needs `.size`

src/pc/stages/stage_dre.c
JoshSchreuder: Was this the data that sits after CUTSCENE_END() on some overlays? It's required for the cutscene to function? Or did you just add it here for parity?

Xeeynamo: It is part of the cutscene data. But I'm still unsure to understand it's use. It's not unused tho

Xeeynamo: I plugged a debugger. That trailing data is processed in `CutsceneRun` here: `src/st/cutscene_run.h` and it doesn't look like a usual cutscene script. It very likely need to be handled separately by the asset tool. All it does is some kind of cutscene clean-up phase that is always ran post-cutscene

JoshSchreuder: Some existing overlays are importing this as binary data interleaved with the cutscene splits in the assets yaml, maybe it would be better to do that here as well until proper import can be done? 

But I am also not too fussed about just manually importing the data like you did here

Xeeynamo: I am definitely planning to do that as a separate pull request, as well as more removal of `OVL_EXPORT`. 

There's also the PSP cutscene stuff that I left it untouched as it has its own level of complexity.

I'd propose to freeze the scope of this PR as it is, as it's a healthy improvement

EDIT: I am exploring the change needed to be done to parse the trailing data, and it's a much bigger change than I expected. I'll post a separate draft PR.

Xeeynamo: Done in #3425

src/pc/cutscene_pc.c
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include <game.h>
+#include "pc.h"
+
+static const CutsceneSymbol* _symbols = NULL;
+static int _symbolCount = 0;
+
+// normalize 0x8018xxxx ptr to 0x0018xxxx
+#define PC_ADDR(x) ((x) & 0x0FFFFFFF)
```
sozud: I think NORMALIZE_PSX_ADDR might be more clear

src/pc/pc.h
```diff
@@ -81,4 +81,13 @@ bool FileAsString(bool (*cb)(const struct FileAsString* file),
 bool FileUseContent(bool (*cb)(const struct FileUseContent* file, void* param),
                     const char* filename, void* param);
 
+typedef struct {
```
sozud, line 84: This is the base of the range you look in right? Would this make sense?
```
typedef struct {
    const u8* pcBase;
    u32 psxBase;
    u32 size;
} CutsceneAddressRange;
```

sozud (approved):

src/pc/pc.h
Xeeynamo: done in 162e2ab16

src/pc/cutscene_pc.c
Xeeynamo: done in 162e2ab16

PR #3425, opened 2026-07-25
Add Cutscene Events asset handling

Xeeynamo:
Direct continuation and dependent to the work done in #3423. This directly addresses https://github.com/Xeeynamo/sotn-decomp/pull/3423#discussion_r3650250199 . This can't be merged before #3423 .

Almost every cutscene (CEN being an exception) has a previously unknown trailing data that we were appending as a generic u8 byte array. In reality that is what the function `RunCutsceneEvents` (previously known as `CutsceneRun`) handles, which is own state machine / bytecode that is now separately decoded in its own asset tool handler.

YAML from NO3:

```yaml
events:
  - [0, SPAWN_ENTITY, 0x20, E_DEATH_SCENE_MGR, 128, 128]
  - [0, SPAWN_ENTITY, 0x21, E_DEATH, 160, 120]
  - [END]
```

Generated code:

```c
EVENT_AT(0), EVENT_SPAWN_ENTITY(0x20, E_DEATH_SCENE_MGR, 128, 128),
EVENT_AT(0), EVENT_SPAWN_ENTITY(0x21, E_DEATH, 160, 120),
EVENT_END(),
```

note as the `enum EntityID` is parsed to store in `cutscene_events` what is the exact Entity ID being used in that specific context. This required some normalization done in de24de0f . I decided to go for `EntityID` as it's more C idiomatic than `Entities` or `EntityIDs`.


JoshSchreuder (approved):
This is excellent, great analysis and discovery!

PR #3432, opened 2026-07-26
pc: fix DAI crash, fix ST0 background

Xeeynamo:


src/st/e_clouds.h
```diff
@@ -50,7 +50,7 @@ void EntityClouds(Entity* self) {
     Primitive* primTwo;
     MedusaCloudsUVal* uVals;
     u8* var_s4;
-    s32 var_s3;
+    u_long var_s3;
```
sozud, line 72: Wouldn't we want Primitive* on pc?

Xeeynamo: It wouldn't match. `var_s3` is also used as a integer array index. See at `primTwo = &((Primitive*)(SP(0x20)))[var_s3];` and the `// Possible FAKE: this cast to/from s32/Primitive* is odd` comment

sozud: Right I mean as an #ifdef on pc and have a separate index (also #ifdef)
Does this run on windows? unsigned long is not guaranteed to be == sizeof(void*)
https://learn.microsoft.com/en-us/windows/win32/winprog64/abstract-data-models?utm_source=chatgpt.com

Xeeynamo: As @bismurphy pointed out a while ago, I'd rather prefer to not use `#ifdef VERSION_PC` or any ifdef if possible.

`u_long` is a custom type that is always guaranteed to be 64-bit: https://github.com/Xeeynamo/psyz/blob/3cc13df9a8acec59923d523eb4b51501ab8fc092/psyz/include/psyz/types.h#L20 :

```c
_Static_assert(sizeof(u_long) == sizeof(void*), "u_long type must be pointer aligned");
```

It's the least invasive solution I found to not diverge the code too much.

sozud (approved):

PR #3436, opened 2026-07-27
Add support to extract player sprites

JoshSchreuder:
Same as #3399 but have added a workaround for PSP padding that was causing race condition in build_all, this is a similar fix to #3223


tools/sotn-assets/assets/spriteset/handler.go
```diff
@@ -29,6 +31,12 @@ func (h *handler) Extract(e assets.ExtractArgs) error {
 		if err != nil {
 			return fmt.Errorf("failed to read sprites: %w", err)
 		}
+
+		// HACK: PSP adds zero padding at the end of the file that must
+		// be removed to match 1:1 with PS1
+		if sprites[len(sprites)-1] == nil && e.Version.GetPlatform() == sotn.PlatformPSP {
+			sprites = sprites[:len(sprites)-1]
+		}
```
JoshSchreuder, line 39: Additional fix for race condition

Xeeynamo (approved):
I already reviewed this in your previous PR. I'll accept it outright

PR #3443, opened 2026-07-29
Fix some `top` hidden gfx/palette splits

JoshSchreuder:


config/splat.us.sttop.yaml
```diff
@@ -70,10 +70,10 @@ segments:
       - [0x1BFC, .data, e_clouds]
       - [0x1CFC, .data, e_clock_tower]
       - [0x20AC, .data, gen/e_layout]
-      - [0x2C04, data]
+      - [0x2C04, data] # e_layout padding?
       - [0x2C08, .data, cutscene_data]
-      - [0x2E44, data]
-      - [0x3804, .data, stage_data]
+      - [0x2E44, data, richter_sprites] # import not working
```
JoshSchreuder, line 75: This is a weird one @Xeeynamo, not sure if you have seen this pattern before?
First elements are actual Richter sprites, the last element is just padded out with a reference to an unused JP stage graphic

I don't think it's importable with the current `spritesheet` handler, though the Richter sprites import fine it kinda gets confused by the gfx.

It looks like maybe they just padded out the unused slots with a reference to a gfx for some unknown reason? Number of slots (197) matches the number in RIC, just most are unused.
Image

Xeeynamo: Looks like dummy data. It's fine if we don't import it for the time being

Xeeynamo (approved):

PR #3444, opened 2026-07-29
sat: add reconstructed native compiler as default

sozud:
I recently found the source to the Saturn compiler and put up a repo on how to build an equivalent version for 64-bit Linux. The repo and process is here https://github.com/sozud/saturn-compiler-repro

This adapts the Saturn build to use the native compiler by default but leaves the dos one as an option if it turns out there is an issue.


Xeeynamo (commented):
This is great. I've been wondering if we could've statically recompile CC1 from Saturn, but this is much better.

Does this mean we can soon get rid of the DOS 8.3 paths?

Also, can you add a `bin/*.sha256` file together with the other compilers?

sozud:
> Does this mean we can soon get rid of the DOS 8.3 paths?

Right now I was keeping the original compiler out of caution so we'd have to find a workaround to allow it to use non-8.3 filenames or commit to getting rid of it. I didn't check but the native one ought to just work with normal Linux paths like all the other old gcc variants.

> Also, can you add a bin/*.sha256 file together with the other compilers?

Added

Xeeynamo (approved):
Looks great, thanks! I'll merge it straight away to check eventual problems on CI

PR #3450, opened 2026-07-31
Code-share RNO0 `create_entity` between PS1 and PSP

mathisto:
share

## Summary
Code-shares the complete RNO0 `create_entity` helper family between the US PlayStation and PSPEU builds using the existing implementation in `src/st/create_entity.h`.

This replaces 15 `INCLUDE_ASM` stubs on each platform with one shared source implementation.

| Target | Functions matched | Matching function bytes |
|--------|------------------:|------------------------:|
| US PS1 | 15 / 15 | 2,936 |
| PSPEU | 15 / 15 | 3,980 |

**Total: 30 platform-specific `INCLUDE_ASM` stubs replaced by shared C.**

## Background
RNO0 had separate PlayStation and PSP stub files for the same complete `create_entity` helper family. These routines are already decompiled in the shared `src/st/create_entity.h` implementation used by other overlays.

## Verification
- [x] US PlayStation build matches the target hash
- [x] PSPEU build matches the target hash
- [x] All 15 functions match on both platforms


JoshSchreuder (approved):

Xeeynamo (approved):
Thanks, these are helpful!

PR #3456, opened 2026-07-31
Extract DRA enemy definitions with asset tool

mathisto:
Well met! Enemies decoded!

## Summary
Adds an `enemydefs` asset handler that extracts the complete `EnemyDef` tables into editable YAML and regenerates C initializer headers in the existing compilation units. Introduces a new string handler with complete font tables for US, EU (contains special accent chars the US doesn't), and JP (including support for dakuten and handakuten encoding of the katakana). Though I don't know much Japanese, I did do some research and ensured the characters matched up. Much of the PR's line diff bulk is from the massive amount of deletions that will no longer need to be manually maintained. Go is not my lingua franca but I tried to keep your existing code conventions. Please dont hesitate to rip me apart; it might not be idiomatic.

This covers 2,805 records across seven tables:
- US PlayStation: 400
- Japanese PlayStation/HD: 400
- PSP English, Spanish, Italian, German, and French: 401 each

The handler models the entire structure, including stats, elemental properties, item drops and rates, hitboxes, and flags. Element masks and item IDs use the existing `Elements` and `ItemDrops` enums, while stable IDs and expected record counts protect against missing or reordered entries.

Generated YAML and initializer headers remain ignored. Keeping the generated initializers inside the original source files lets the compiler continue to place and deduplicate the pointed-to strings instead of retaining hardcoded `.rodata` addresses.

## Name encoding
Enemy names use the game's 8x8-font encoding. As mentioned, the shared string handler supports the PSX and PSP character tables, dakuten and handakuten composition, and the PSP European glyphs.

PSX and HD generated sources retain readable `_S(...)` strings. PSP headers contain pre-encoded strings because its string-processing step runs before generated includes are expanded.

The ordinary `go test ./...` reaches the existing unrelated vet failure in `assets/spritesheet/handler.go:168` (`fmt.Errorf` uses `%w` with a string). Anyone have any more info on this failure? I'd love to take a swing at that as well.


tools/sotn-assets/sotn/string.go
```diff
@@ -0,0 +1,209 @@
+package sotn
```
sozud, line 1: Wait until @Xeeynamo responds before changing but my opinion is we shouldn't duplicate sotn_str in the assets tool. You can just call it from go instead

tools/sotn-assets/assets/enemydefs/handler.go
```diff
@@ -0,0 +1,381 @@
...
+
+	entries := make([]enemyModel, count)
+	for i, raw := range rawEntries {
+		name, err := sotn.DecodeString(
+			data, psx.Addr(raw.NameAddr), ramBase, platform)
+		if err != nil {
+			return nil, fmt.Errorf("enemy %#x name: %w", i, err)
+		}
+		entries[i] = enemyModel{
+			ID:                   fmt.Sprintf("0x%03X", i),
```
Xeeynamo: We should not store IDs in the exported asset. They're redundant, and it might confuse a modder who has not much knowledge of the game's internals.

This would also remove the necessity of normalizing an enemy entry.

tools/sotn-assets/sotn/string.go
Xeeynamo: @sozud it's tricky, because `sotn_str` doesn't have anything as of today that decodes and encodes via stdin/stdout. And surely we don't want to make a tool call per enemy entry.

I know @mathisto expressed the intention of keep expanding the tool to support more assets. So a solution must be decided here.

We could have `sotn_str` modified, so a `sotn_str encode` and `sotn_str decode` will just sit and wait, while the Go app would pipe stuff in it and get the output. Just an idea. But I am just wondering if the implementation cost and the consequent fragility of `sotn-assets` would become higher for the sake of duplicating the string de/encoding?

On the other end, if one of the tools will go out of sync, we will know at build time as the game will no longer match. 

I am a bit conflicted here. But I am leaning more towards keeping the scope of this PR as it is.

Xeeynamo (changes requested):
Good stuff so far. There are two comments that needs to be addressed first though.

tools/sotn-assets/sotn/string.go
sozud: @mathisto don't make changes yet but:

I put a post in Discord and what I'm thinking at the moment is:
Change this PR to generate tables like:
```
  {STR_ENEMY_ZOMBIE, 10, 5, ...}
```
And make it consume a yaml per-version like:
```
 # enemy_string_ids.us.yaml
  0x010: STR_ENEMY_ZOMBIE
  0x011: STR_ENEMY_BAT
```
And for the moment we generate a header like: 

```
  // enemy_strings.us.h
  #define STR_ENEMY_ZOMBIE "\x3A\x4F\x4D\x42\x49\x45\xFF"
```
We intentionally don't use the sotn_str encoding/decoding for the moment since the psp build runs sotn_str before expanding includes. This gives us a path towards better localization and modding later.

tools/sotn-assets/assets/enemydefs/handler.go
mathisto: Good call. I went back and revised this to align with your direction. It does clean it up nicely! Thank you 🙇

Xeeynamo (approved):
I think this change is good ✔️ 

There is duplication with the string encoding/decoding, but the advantage this PR introduces outweights the string issue.

I'd be happy to merge this. Let's wait a second review so we know we're all mostly aligned with the decision.

sozud (approved):
text handling needs a deeper look than makes sense for this PR so I think this is fine for the time being

PR #3457, opened 2026-07-31
extract psp sfx, integrate in maria mod

sozud:
This converts the psp format maria samples and implements them for psx. I've left in a maria ps1 sound test and one for psp since I anticipate this may need more debugging. This adds a one-off tool maria-sfx but my eventual plan would be to extract through the assets system for psp then recombine through the psx system. 

This also includes a random fix for cmake where sotn_str that's annoying if you're using containers (e.g. you have a linux container but the host is mac and sotn_str is mutually incompatible). 

https://github.com/user-attachments/assets/78b555b5-0c9d-4200-bc4a-486b9b5ff866


sozud:
Should the maria mod build have access to sd_mar2.spk on CI or should it just be able to build without it? I think it makes sense to provide the file

JoshSchreuder:
Looks fine to me but would be good to get @Xeeynamo eyes on it as someone who knows the mod better

src/maria/pl_main.c
```diff
@@ -839,3 +1076,20 @@ void MarMain(void) {
     }
     D_pspeu_092E5F28 = PadReadPSP();
 }
+
+#else
+
+void MarMain(void) {
+    MarPspSoundTest();
+    MARIA_PSP_NOP_8192
```
Xeeynamo: I think it would be simpler to patch `src/maria/pl_header.c` and swap `MarMain` to the sound test. The sound test itself could get isolated to `mods/mariapsp-forpsx/maria.c` so it gets isolated to let `pl_main.c` remain pristine.

src/maria/pl_main.c
```diff
@@ -442,6 +445,240 @@ static u16 mar_80154574[] = {
...
+#define MARIA_PSP_NOP_16 MARIA_PSP_NOP_8 MARIA_PSP_NOP_8
+#define MARIA_PSP_NOP_32 MARIA_PSP_NOP_16 MARIA_PSP_NOP_16
+#define MARIA_PSP_NOP_64 MARIA_PSP_NOP_32 MARIA_PSP_NOP_32
+#define MARIA_PSP_NOP_128 MARIA_PSP_NOP_64 MARIA_PSP_NOP_64
+#define MARIA_PSP_NOP_256 MARIA_PSP_NOP_128 MARIA_PSP_NOP_128
+#define MARIA_PSP_NOP_512 MARIA_PSP_NOP_256 MARIA_PSP_NOP_256
+#define MARIA_PSP_NOP_1024 MARIA_PSP_NOP_512 MARIA_PSP_NOP_512
+#define MARIA_PSP_NOP_2048 MARIA_PSP_NOP_1024 MARIA_PSP_NOP_1024
+#define MARIA_PSP_NOP_4096 MARIA_PSP_NOP_2048 MARIA_PSP_NOP_2048
+#define MARIA_PSP_NOP_8192 MARIA_PSP_NOP_4096 MARIA_PSP_NOP_4096
```
Xeeynamo: Wouldn't be simpler to have a for loop with a volatile asm nop? Or is timing strict here?

EDIT: depending on what the intention is, you could also fire a bunch of `VSync(0)` to create artificial idle time

Xeeynamo (commented):
This is a great idea. I left some comments to leave the original `src/maria` cleaner from changes

src/maria/pl_main.c
sozud: What this is doing is padding the psp version so that I can drop in the sound test binary without breaking the filesystem

src/maria/pl_main.c
sozud: Ok I moved the psx maria sound test to maria.c

src/maria/pl_main.c
Xeeynamo: I am stil wondering if this is necessary. It's quite an invasive change in the matching PSP code. Shouldn't `sotn-disk inject` work without changing the file system based on the final file size?

sozud: Ok so I looked into this in more detail. I think what the issue is that psp overlays are not relocatable because the mwo_header.bin doesn't get rewritten. So the pointers to data,bss,etc. are in the wrong spots. The nops pad the text section back to the same size as retail

sozud: How about this, I move the psp soundtest to it's own file and leave a comment on how to enable it? I think the proper fix is mwo header writing + correct umd packing

Xeeynamo: > I move the psp soundtest to it's own file and leave a comment on how to enable it? 

I think it would be better, yes. In future it would be cool to also update the MWO header to allow relocations

sozud: Moved now

Xeeynamo (approved):

PR #3460, opened 2026-08-01
Minor Saturn build system 

Xeeynamo:
Previously diffing Saturn locally was pointing to the wrong files. This attempts to fix that.

DIffing locally is still broken, I am unsure since which commit. But this at least fixes the obvious problems.

### DIFF
--- a/diff_settings.py
+++ b/diff_settings.py
@@ -19,8 +19,21 @@ def add_custom_arguments(parser):
     )
 
 
+def apply_saturn(config, args):
+    name = args.overlay
+    config["myimg"] = f"build/saturn/{name}.PRG"
+    config["source_directories"] = [f"src/saturn/"]
+    config["mapfile"] = f"build/saturn/{name.lower()}.map"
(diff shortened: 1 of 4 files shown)


diff_settings.py
```diff
@@ -103,12 +106,9 @@ def apply(config, args):
         f"{name.lower() if version == 'hd' else name}.{binaryExt[version]}",
     )
 
-    config["objdump_executable"] = (
-        "sh-elf-objdump" if version == "saturn" else "mipsel-linux-gnu-objdump"
-    )
+    config["objdump_executable"] = "mipsel-linux-gnu-objdump"
```
sozud, line 109: unless I'm missing something I don't think mipsel objdump is doing to work?

Xeeynamo: That is for the PS1 build. The Saturn one is at line 28, and it uses `sh-elf-objdump`

sozud (approved):

PR #3466, opened 2026-08-01
EntityLockCamera for RNO0

bismurphy:
Required some slight adjustments. There is a set of tiles which are similar across most overlays, but slightly different in some others. This was ifdefed, but once we had 3 different versions, I decided we need to do it differently. Now the .h file has a default which is used if none other is specified, but overlays can specify a particular array if needed.

Also fixed RCHI PSP, which only had this function half incorporated, with a similar preprocessor define.


Xeeynamo (approved):
It's much better extracting these instead of `#ifdef`fing by stage 👍

PR #3467, opened 2026-08-01
pc: add NP3 overlay

Xeeynamo:
`./build/pc/sotn --stage np3`

image

### DIFF
--- a/src/pc/stages/stage_np3.c
+++ b/src/pc/stages/stage_np3.c
@@ -1,0 +1,42 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include <game.h>
+#include <string.h>
+#include "overlay.h"
+#include "../../st/np3/np3.h"
+#include "stage_loader.h"
+
+extern AbbreviatedOverlay OVL_EXPORT(Overlay);
+extern PfnEntityUpdate EntityUpdates[];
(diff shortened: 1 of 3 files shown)


JoshSchreuder (approved):

src/pc/stages/stage_np3.c
```diff
@@ -0,0 +1,42 @@
...
+extern LayoutEntity* entityLayoutVertical[];
+extern GAME_IMPORT PfnEntityUpdate* PfnEntityUpdates;
+extern GAME_IMPORT LayoutEntity** g_pStObjLayoutHorizontal;
+extern GAME_IMPORT LayoutEntity** g_pStObjLayoutVertical;
+
+extern u_long* cluts[];
+extern u16 e_slogra_p[];
+
+static void PatchSlograPalettePc(void) {
+    // On NP3: PAL_BULK_(0x2D90, e_slogra_p, 0x90),
```
sozud, line 20: Is it not cleaner to fix the data at the source?

Xeeynamo: I thought about it. That would mean duplicating `gen/palette_def.h` and use it just for the PC port. We cannot fix the generated file as it would keep re-generating too. Eventually I decided to go with the current approach as it's the simplest solution.

PR #3468, opened 2026-08-01
RNO0 e_breakable

bismurphy:
Very similar to the previous PR, this one involves more ifdef in the function.

Instead of using hardcoded stages, we instead use ifdef based on what features to enable in the function, and then the stages enable those features when #include'ing the function. Much more semantic option.

The function is getting a bit ugly, but for now seems worthwhile to keep in a .h file like this since the differences are still tolerable.


Xeeynamo (commented):
For a 50 lines entity, wouldn't it make more sense to just copy it flat (so without `#ifdef`) to their correspondent source function? It would me much easier to read.

bismurphy:
> For a 50 lines entity, wouldn't it make more sense to just copy it flat (so without `#ifdef`) to their correspondent source function? It would me much easier to read.

Okay, I changed it. I took out all the possible defines. Any overlay that uses any of the "extra" features defines the function directly and does not use the .h, and the .h file only contains the most "basic" form of the function.

Xeeynamo:
> > For a 50 lines entity, wouldn't it make more sense to just copy it flat (so without `#ifdef`) to their correspondent source function? It would me much easier to read.
> 
> Okay, I changed it. I took out all the possible defines. Any overlay that uses any of the "extra" features defines the function directly and does not use the .h, and the .h file only contains the most "basic" form of the function.

Yeah it's a bit tricky, and I do understand the willingness of documenting the code and its differences. But where I am coming from is how isolated the overlays are. When the change ratio between overlays is too high (small function, many variants) I think it's start to not be worth it.

Xeeynamo (approved):
Looks good, thank you!

PR #3482, opened 2026-08-04
import rbo4,rbo6

sozud:


src/boss/rbo6/d_prize_drops.c
```diff
@@ -0,0 +1,23 @@
...
+    0x0503,
+    0x0603,
+    0x0703,
+    0x0803,
+    0x0903,
+    0x0A03,
+    0x0B03,
+    0x0C03,
+    0x0D03,
+    ITEMDROP_GREAT_SWORD,
```
JoshSchreuder, line 21: I suspect these are the same as the incorrect link used in RBO0, and likely this is an animation for one of the entities
https://github.com/Xeeynamo/sotn-decomp/blob/867909c8086085bc67c133429797a7bf61f1bcb9/src/boss/rbo0/e_boss_torch.c#L16

Similarly with the other RBO PrizeDrops that got imported

sozud: Ok I made an issue and I'll investigate this https://github.com/Xeeynamo/sotn-decomp/issues/3484

PR #3494, opened 2026-08-05
add remaining psp stage overlays

sozud:
This adds and partially dedupes several psp overlays at are already 100% on psx.


JoshSchreuder (approved):

src/boss/bo4/create_entity.c
```diff
@@ -1,4 +1,28 @@
...
+
+#if !defined(VERSION_PSP)
+#endif
 #include "../../st/pfn_entity_update.h"
+
+#if defined(VERSION_PSP)
+
+#include <stage.h>
+
+void CreateEntityFromLayout(Entity* entity, LayoutEntity* initDesc) {
```
JoshSchreuder: This seems likely wrong and can be matched on PSP and PSX?

Similarly with the other overlays here

src/boss/rbo5/e_misc.c
```diff
@@ -1,8 +1,134 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "rbo5.h"
 
+#if defined(VERSION_PSP)
```
JoshSchreuder: This should be de-dupeable

src/st/rtop/header.c
```diff
@@ -1,5 +1,34 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "rtop.h"
+#ifdef VERSION_PSP
```
JoshSchreuder: I think these headers can be aligned (same with all overlays in PR)

sozud:
Comments should be resolved now

PR #3499, opened 2026-08-06
Rename RNO0 entities

bismurphy:
I went through the e_layout file and gave every entity ID a name. All UNK_XX values now have real names, and their entity functions also correspond. This will make decompiling easier when we get to these functions since we will know how they relate to other entities.


sozud (approved):

src/st/rno0/rno0.h
```diff
@@ -46,16 +46,16 @@ enum EntityID {
     E_STONE_DOOR,                // EntityStoneDoor
     E_UNK_1F,                    // OVL_EXPORT(Unused801C2338)
     E_DUMMY_20,                  // EntityDummy
-    E_ARMOR_LORD,                // EntityArmorLord
+    E_GUARDIAN,                  // EntityArmorLord
```
JoshSchreuder: Need to rename `EntityArmorLord`?

JoshSchreuder (approved):

PR #3501, opened 2026-08-06
sat: handle map bitmap assets

sozud:
Extracts the castle map. Example:
bitmap


Xeeynamo (approved):
More Saturn stuff is always good 👍

PR #3503, opened 2026-08-07
Code-share 103 functions across RNZ1, RNO0, RNO1, RNO2, RBO6 and BO1/BO3/BO7

eibrunocorrea:
Code-shares 103 functions across 10 overlays by routing them through the existing shared headers in `src/st/`, plus two new shared headers where a matching implementation already existed in another overlay.

### Per overlay

| Overlay | File | Functions |
|---|---|---|
| RNZ1 | `e_collect` + new `d_prize_drops.c` | 16 |
| RNZ1 | `e_misc` | 18 |
| RNZ1 | `e_explosion_puff_opaque` | 2 |
| RNZ1 | `unk_29914` (`func_us_801ABDC8`, hand-decompiled) | 1 |
| NZ1 / RNZ1 | `e_clouds` → new `src/st/e_clock_tower_clouds.h` | 1 |
| RNO0 | `giantbro_helpers` (segment split) + new `src/st/giantbro_helpers_2.h`, also adopted by NO2/NP3 | 22 |
| BO1 / BO3 / BO7 | `e_misc` | 22 |
| RBO6 | `prim_helpers` | 9 |
| RNO1 / RNO2 | `e_misc` | 12 |

### Notable changes

**New shared header `e_clock_tower_clouds.h`** — NZ1 and RNZ1 share the clock tower cloud renderer. It differs only in five constants, so the header is parameterised: initial position, the base Y of the cloud band, the clamp row, and `CLOUDS_INVERTED` (in the reverse tower the band is built downward, and the clamp row sits at the bottom of the screen instead of the top). NZ1 was refactored onto the header and still matches.

**New shared header `giantbro_helpers_2.h`** — the RNO0 `giantbro_helpers` segment holds 22 functions, but `giantbro_helpers.h` only covers the first 15. The remaining 7 already existed as plain C in NO2 (`4966C.c`) and NP3 (`4E04C.c`), so they were extracted into a shared header now used by all three. The RNO0 segment was split at `0x4A320` to mirror the NO2/NP3 layout, and the missing `.bss` entry for `giantbro_helpers` was added to the RNO0 config.

**`e_misc.h` for BO1/BO7** — these two do not carry the trailing `0, 0` in `g_QuadIndices2`. The condition is guarded with `!defined(VERSION_PSP)` so the PSP builds keep their current behaviour.

**`e_valhalla_knight.h`** — RNZ1 uses the wider leash threshold (`posX > 0x70`) on the first check, like ARE.

**Naming** — RNZ1 EInits and entity IDs were given their conventional names (`g_EInitMedusaHeadBlue`/`Yellow`, `g_EInitEnvironment`, `g_EInitValhallaKnightUnk1-3`, `E_VALHALLA_KNIGHT_UNK1`/`2`, `E_EXPLODE_PUFF_OPAQUE`), and `func_us_801C8954`/`801C8AAC` dropped the `_from_are` suffix. The `pspeu` symbol file and `src/st/rnz1_psp/unk_6730.c` were updated to match.

### Verification

Clean `clean` + `extract` + `build` cycle for each version:

```
us     ✅ 111/111
hd     ✅  12/12
pspeu  ✅  63/63
```

`make format` produces no changes on a fresh checkout of this branch.


src/st/rnz1/e_clouds.c
```diff
@@ -1,4 +1,12 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "rnz1.h"
 
-INCLUDE_ASM("st/rnz1/nonmatchings/e_clouds", EntityClouds);
+#define CLOUDS_POS_X 0xB0
+#define CLOUDS_POS_Y 0xD0
+#define CLOUDS_BASE_Y 0x70
+#define CLOUDS_EDGE_Y 0xF0
```
Xeeynamo: These could get folded to a `#ifdef CLOUDS_INVERTED` directly in `e_clock_tower_clouds.h`, so we can reduce the macro definition footprint and keep the code more explicit on what's actually doing.

src/st/nz1/e_clouds.c
```diff
@@ -1,118 +1,9 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "nz1.h"
-#include <scratchpad.h>
 
-extern EInit g_EInitEnvironment;
+#define CLOUDS_POS_X 0x50
+#define CLOUDS_POS_Y 0x30
+#define CLOUDS_BASE_Y 0x90
+#define CLOUDS_EDGE_Y 0x10
```
Xeeynamo: Same comment as https://github.com/Xeeynamo/sotn-decomp/pull/3503/changes#r3732665859

src/st/no2/giantbro_helpers_2.c
```diff
@@ -1,139 +1,4 @@
...
-
-    while (*offsets) {
-        if (*offsets != 0xFF) {
-            entity = g_CurrentEntity + *offsets;
-            polarPlacePart(entity);
-        }
-        offsets++;
-    }
-}
+#include "../giantbro_helpers_2.h"
```
Xeeynamo, line 4: would be nice to have the source file name renamed as `giantbro_helpers_2.c` so we have context at glance.

src/st/np3/giantbro_helpers_2.c
```diff
@@ -1,138 +1,4 @@
...
-
-    while (*offsets) {
-        if (*offsets != 0xFF) {
-            entity = &g_CurrentEntity[*offsets];
-            polarPlacePart(entity);
-        }
-        offsets++;
-    }
-}
+#include "../giantbro_helpers_2.h"
```
Xeeynamo, line 4: Same as https://github.com/Xeeynamo/sotn-decomp/pull/3503/changes#r3732671795

src/st/rnz1_psp/unk_6730.c
```diff
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "../rnz1/rnz1.h"
 
-INCLUDE_ASM("st/rnz1_psp/nonmatchings/rnz1_psp/unk_6730", func_us_801C8954_from_are);
+INCLUDE_ASM("st/rnz1_psp/nonmatchings/rnz1_psp/unk_6730", func_us_801C8954);
 
-INCLUDE_ASM("st/rnz1_psp/nonmatchings/rnz1_psp/unk_6730", func_us_801C8AAC_from_are);
```
Xeeynamo, line 6: These functions are connected to `src/st/e_valhalla_knight.h`. I propose to either revert the name back, or do a code share here too.

src/boss/bo1/bo1.h
```diff
@@ -4,6 +4,8 @@
 
 #include <stage.h>
 
+#define BOSS_IS_BO1
+
 #define STAGE_IS_BO1
```
Xeeynamo: Duplicate macro? We should keep `STAGE_IS_BO1`

src/boss/bo7/bo7.h
```diff
@@ -4,6 +4,8 @@
 
 #include <stage.h>
 
+#define BOSS_IS_BO7
+
 #define STAGE_IS_BO7
```
Xeeynamo: Duplicate macro? Let's keep `STAGE_IS_BO7`

Xeeynamo (changes requested):
Looks a solid start. I left a review to parts of the proposed changes that need attention.

Just a heads-up, while we care a lot about the quality of the output of proposed pull requests to our code base, we will remove any Co-authoring that would associate this project to the branding of an external company (Claude/Anthropic in this case). Let me know if removing the Co-authoring works for you.

eibrunocorrea:
Thanks for the review — all points addressed, and yes, removing the co-authoring is fine. I've rewritten the history so it's gone from every commit.

Each fix was squashed into the commit it belongs to rather than appended, so the history stays readable.

- **Clock tower clouds**: the five per-stage constants are now four `#ifdef CLOUDS_INVERTED` blocks inside `e_clock_tower_clouds.h`. The stage files are down to a single `#define CLOUDS_INVERTED` for RNZ1 and nothing for NZ1.
- **`4966C.c` / `4E04C.c`**: renamed to `giantbro_helpers_2.c`. This also touched `splat.hd.stnp3.yaml` besides the `us` and `pspeu` configs.
- **`rnz1_psp/unk_6730.c`**: reverted to `_from_are`. Digging into it I found the rename was a leftover — I had tried to code-share the Valhalla Knight for RNZ1, found it was a variant and dropped it, but the rename and a now-dead `#if defined(STAGE_IS_ARE) || defined(STAGE_IS_RNZ1)` guard survived. RNZ1 never includes that header, so I removed the guard too. Sharing it properly needs the variant analysis, which I'd rather do in its own PR.
- **`BOSS_IS_BO1` / `BOSS_IS_BO7`**: removed, `e_misc.h` now keys off `STAGE_IS_BO1` / `STAGE_IS_BO7`.

Re-validated from a clean `clean` + `extract` + `build` cycle after the changes: `us` 111/111, `hd` 12/12, `pspeu` 63/63.

Xeeynamo:
@eibrunocorrea I appreciate the time and responsiveness to address my initial feedback. I found a few mistakes in the proposed changes, especially in `e_misc.c` diverging from the implementation of the successfully completed overlays, the order of some splits defined in our `config/`, and a more efficient code sharing approach.

I reviewed the pull request further, and I believe it is better to have the work split to smaller and self-contained chunks. We often use `git blame <file_name>` to look for the reason of certain change, and reach out the original contributors if we need further clarification. This step would be hard to do with the pull request in the current size, especially coming from first-time contributors. It would also be immensely easier for us to review.

My suggestion is to keep reaching us out to Discord. Keep ask questions, and we can find ways to direct you to ways to propose changes and improvements more effectively and according to our rigid standards.

I know it may feel disappointing to see your pull request closed. My initial encouragement on your proposed changes being a good start is still valid. Let's better coordinate the efforts for future proposals, and absorb feedback from the existing collaborators. Thanks!

PR #3509, opened 2026-08-09
pc: Add ARE overlay

JoshSchreuder:
image

~~Plate Lord is still crashing but most of the stage is playable, I was unable to determine the right fix. I think it might be something to do with the Primitive alignment but not certain.~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -282,6 +282,65 @@ set(SOURCE_FILES_STAGE_SEL
     src/st/sel/cutscene_data.c
 )
 
+set(SOURCE_FILES_STAGE_ARE
+    src/pc/stages/stage_are.c
+    src/st/are/header.c
+    src/st/are/sprite_banks.c
+    src/st/are/palette_def.c
+    src/st/are/layers.c
+    src/st/are/graphics_banks.c
+    src/st/are/gen/e_laydef.c
+    src/st/are/e_init.c
+    src/st/are/background_block_init.c
+    src/st/are/e_red_door_tiles.c
+    src/st/are/d_prize_drops.c
+    src/st/are/gen/rooms.c
+    src/st/are/gen/e_layout.c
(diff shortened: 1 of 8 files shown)


sozud (approved):

sozud:
This one got conflicted sorry, should have merged it first. Would you mind rebasing? Thanks

Xeeynamo (approved):
@JoshSchreuder need conflicts resolved. Otherwise good to merge once fixed :+1:

mathisto:
Yo dude, I’ve been playing around with the PC port branch in a private fork, and I’m not 100% confident, but I think I just ran into this same exact issue. On PSX, `ext.prim` is only 4 bytes, but on a 64-bit PC it grows to 8 bytes, so `unk80` and `unk82` can overwrite the upper half of the pointer. Then the next time the game uses it, it dereferences a corrupted pointer and crashes.

Your `ET_PlateLord` change looks right for the main controller. I think there may be one more piece, though: take a looksee at `func_us_801D44A0`. It stores the pointer and then writes through `plateLordUnknown.unk80`, and `e_plate_lord_2.h` seems to have the same pattern. I think `ET_PlatelordUnknown` also needs its own `Primitive* prim`, with those refs changed to `plateLordUnknown.prim`.

https://github.com/user-attachments/assets/005072c7-5872-4807-b89d-decbe32c889e

mathisto (approved):

JoshSchreuder:
> I think there may be one more piece, though: take a looksee at `func_us_801D44A0`. It stores the pointer and then writes through `plateLordUnknown.unk80`, and `e_plate_lord_2.h` seems to have the same pattern. I think `ET_PlatelordUnknown` also needs its own `Primitive* prim`, with those refs changed to `plateLordUnknown.prim`.

Nice, thanks! I've updated that here and it's no longer crashing. Appreciate it!

PR #3514, opened 2026-08-09
Start endian work, tests

sozud:
I've discovered that sotn has a lot of issues with big endian platforms. This is a part of a series of PRs to add tests and fix dra/ric and st0 so that part of the game functions on big endian. The tests get run in 32-bit powerpc and 64-bit s390x to check.

### DIFF
--- a/tests/endian/common/test_types.c
+++ b/tests/endian/common/test_types.c
@@ -1,0 +1,123 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "test_common.h"
+
+_Static_assert(sizeof(s8) == 1, "s8 must be 8 bits");
+_Static_assert(sizeof(u8) == 1, "u8 must be 8 bits");
+_Static_assert(sizeof(s16) == 2, "s16 must be 16 bits");
+_Static_assert(sizeof(u16) == 2, "u16 must be 16 bits");
+_Static_assert(sizeof(s32) == 4, "s32 must be 32 bits");
+_Static_assert(sizeof(u32) == 4, "u32 must be 32 bits");
+_Static_assert(sizeof(f16) == 2, "f16 must be 2 bytes");
+_Static_assert(sizeof(f32) == 4, "f32 must be 4 bytes");
+
+void test_target_byte_order(void) {
+    const uint32_t value = UINT32_C(0x01020304);
+    uint8_t bytes[sizeof(value)];
+
+    memcpy(bytes, &value, sizeof(bytes));
+#if SOTN_BIG_ENDIAN
+    CHECK_EQ_HEX("target.byte_order.big", 0x01, bytes[0]);
+#else
+    CHECK_EQ_HEX("target.byte_order.little", 0x04, bytes[0]);
+#endif
+}
+
+void test_f32_read_views(void) {
+    static const struct {
+        uint32_t val;
+        int16_t hi;
+        uint16_t lo;
+    } cases[] = {
+        {UINT32_C(0x12345678), INT16_C(0x1234), UINT16_C(0x5678)},
+        {UINT32_C(0xFFFF8000), INT16_C(-1), UINT16_C(0x8000)},
+        {UINT32_C(0x80000001), INT16_MIN, UINT16_C(0x0001)},
+        {UINT32_C(0x007B4567), INT16_C(123), UINT16_C(0x4567)},
+        {UINT32_C(0xFFFE4000), INT16_C(-2), UINT16_C(0x4000)},
+    };
+    size_t i;
+
+    for (i = 0; i < sizeof(cases) / sizeof(cases[0]); i++) {
+        f32 value;
+        memcpy(&value.val, &cases[i].val, sizeof(value.val));
+        CHECK_EQ_S32("f32.read.hi", cases[i].hi, value.i.hi);
+        CHECK_EQ_HEX("f32.read.lo", cases[i].lo, s16_bits(value.i.lo));
+    }
+}
+
+void test_f32_write_views(void) {
+    f32 value;
+
+    value.val = (s32)UINT32_C(0x12345678);
+    value.i.hi = (s16)UINT16_C(0x9ABC);
+    CHECK_EQ_HEX("f32.write.hi.value", UINT32_C(0x9ABC5678),
+                 f32_bits(value));
+    CHECK_EQ_HEX("f32.write.hi.preserve_lo", UINT16_C(0x5678),
+                 s16_bits(value.i.lo));
+
+    value.i.lo = (s16)UINT16_C(0xDEF0);
+    CHECK_EQ_HEX("f32.write.lo.value", UINT32_C(0x9ABCDEF0),
+                 f32_bits(value));
+    CHECK_EQ_HEX("f32.write.lo.preserve_hi", UINT16_C(0x9ABC),
+                 s16_bits(value.i.hi));
+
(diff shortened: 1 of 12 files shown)


tests/endian/common/test_types.c
```diff
@@ -0,0 +1,123 @@
...
+#include "test_common.h"
+
+_Static_assert(sizeof(s8) == 1, "s8 must be 8 bits");
+_Static_assert(sizeof(u8) == 1, "u8 must be 8 bits");
+_Static_assert(sizeof(s16) == 2, "s16 must be 16 bits");
+_Static_assert(sizeof(u16) == 2, "u16 must be 16 bits");
+_Static_assert(sizeof(s32) == 4, "s32 must be 32 bits");
+_Static_assert(sizeof(u32) == 4, "u32 must be 32 bits");
+_Static_assert(sizeof(f16) == 2, "f16 must be 2 bytes");
+_Static_assert(sizeof(f32) == 4, "f32 must be 4 bytes");
```
Xeeynamo: Can't we have these in `types.h`?

Xeeynamo (commented):
Long term, how do you envision having big-endian support for the project? And what are the areas that are mostly affected by this issue? Ideally the code should be portable enough for us to not be constraint by endianness 

The Saturn port is big-endian, I assume we could take some inspiration from it. I am asking as you likely know more than me

sozud:
> Long term, how do you envision having big-endian support for the project? And what are the areas that are mostly affected by this issue? Ideally the code should be portable enough for us to not be constraint by endianness

The goal would be to try and make as few invasive changes as possible and only have inline #ifdefs where necessary. I think mostly the introduction of #ifdefs, and some endian-safe helper functions. It's mostly graphics code and positions. Psy-q has endian problems as well as a lot of primitive code.

Examples:

Unions in wrong order
```
typedef union {
    s32 val;
    struct {
#if SOTN_BIG_ENDIAN
        s16 hi;
        s16 lo;
#else
        s16 lo;
        s16 hi;
#endif
    } i;
} f32;
```

Packed GTE results in wrong order
```c
// before
gte_ldv0(&D_801C1BC8[vectorIndex1]);
gte_rtps();
gte_stsxy((long*)&prim->x0);
```
```c
// after
gte_ldv0(&D_801C1BC8[vectorIndex1]);
gte_rtps();
gte_stsxy((long*)SOTN_PACKED_S16_PAIR_OUTPUT(prim->x0, projectedXY));
SOTN_UNPACK_S16_PAIR(prim->x0, prim->y0, projectedXY);
```

More bad packing
```c
// before
LOW(prim->x0) = var_s3[0];
LOW(prim->x1) = var_s3[2];
LOW(prim->x2) = var_s3[1];
LOW(prim->x3) = var_s3[3];
```
```c
// after
SOTN_STORE_PACKED_S16_PAIR(prim->x0, prim->y0, var_s3[0]);
SOTN_STORE_PACKED_S16_PAIR(prim->x1, prim->y1, var_s3[2]);
SOTN_STORE_PACKED_S16_PAIR(prim->x2, prim->y2, var_s3[1]);
SOTN_STORE_PACKED_S16_PAIR(prim->x3, prim->y3, var_s3[3]);
```
+= pointer walk on i.hi

```c
// before
var_s1 = &self->ext.clouds.unk84.i.hi;
...
for (i = 0; i < 2; i++, cloudData++, var_s1 += 4) {
    posX = self->posX.i.hi + *var_s1;
    ...
    sp40 = (u16)var_s1[2];
```
```c
// after
#if !SOTN_BIG_ENDIAN
var_s1 = &self->ext.clouds.unk84.i.hi;
#endif
...
for (i = 0; i < 2; i++, cloudData++
#if !SOTN_BIG_ENDIAN
     , var_s1 += 4
#endif
) {
#if SOTN_BIG_ENDIAN
    if (i == 0) {
        phaseX = self->ext.clouds.unk84.i.hi;
        phaseZ = St0FixedInteger(self->ext.clouds.unk88);
    } else {
        phaseX = St0FixedInteger(self->ext.clouds.unk8C);
        phaseZ = St0FixedInteger(self->ext.clouds.unk90);
    }
    posX = self->posX.i.hi + phaseX;
#else
    posX = self->posX.i.hi + *var_s1;
#endif
    ...
#if SOTN_BIG_ENDIAN
    sp40 = (u16)phaseZ;
#else
    sp40 = (u16)var_s1[2];
#endif
```
Combo of prims / bad macros
```c
// before
F(prim->x2).i.hi = prim->y1;
F(prim->x2).val -= F(prim->x3).val;
prim->y1 = F(prim->x2).i.hi;
```
```c
// after
#if SOTN_BIG_ENDIAN
St0SetPrimitiveFixedInteger(&prim->x2, &prim->y2, prim->y1);
prim->y1 = St0PrimitiveFixedSubtract(
    &prim->x2, &prim->y2, &prim->x3, &prim->y3);
#else
F(prim->x2).i.hi = prim->y1;
F(prim->x2).val -= F(prim->x3).val;
prim->y1 = F(prim->x2).i.hi;
#endif
```


> The Saturn port is big-endian, I assume we could take some inspiration from it. I am asking as you likely know more than me

The main change seen so far is they got rid of the posX/posY union and it's a u32 accessed with FIX/FIX_TO_I etc. We haven't done a lot of entity code so I don't have equivalent examples there but I imagine they just adjusted each spot as they found it. They reworked every entity since I think wrapping psy-q would have been way too slow.

tests/endian/common/test_types.c
sozud: I moved them

Xeeynamo:
For this one specifically, I wonder if would be worth to compile and run the game on CI with qemu-ppc and using the same system established in #3537 ? so we don't need to have tests on the individual structs and macros.

My thoughts is that I'd be preferable to have the game tested end-to-end. Doing unit tests on single macros and structs have the risk of assuming certain decisions would work on a larger scale.

I also wonder if we should wait to know more of the Saturn port, since that uses big endian? The inverted `hi`/`lo` for `f32` is a strong clue, for example

sozud:
> For this one specifically, I wonder if would be worth to compile and run the game on CI with qemu-ppc and using the same system established in https://github.com/Xeeynamo/sotn-decomp/pull/3537 ? so we don't need to have tests on the individual structs and macros.

I think the hard part here is generating all the movies. Like for example if I want to cover every endian-sensitive branch of an enemy I have to spend a lot of time moving back and forth to trigger the right branches.

Also the movies don't self-describe like unit tests. I can look at the test and see what conditions it's setting up and how they are checked, but I couldn't do this for a movie except watching it and then looking at code coverage afterwards.

> My thoughts is that I'd be preferable to have the game tested end-to-end. Doing unit tests on single macros and structs have the risk of assuming certain decisions would work on a larger scale.

Is there anything in particular you're envisioning that would become an issue?

> I also wonder if we should wait to know more of the Saturn port, since that uses big endian? The inverted hi/lo for f32 is a strong clue, for example

I don't think we have any entities done that used a heavy amount of psy-q but an example is https://github.com/Xeeynamo/sotn-decomp/blob/1aeed24eb7c8dbdd173c791647045e55a1ad525e/src/saturn/stage_02.c#L64
https://github.com/Xeeynamo/sotn-decomp/blob/eeeef7562fba266e616fcbb8716b8e459657357d/src/st/nz0/e_nz0_room2.c#L161
I'm pretty sure every entity has adjustments to use their new systems like this one. It's possible further decompilation may show a different strategy

PR #3515, opened 2026-08-09
rno1: wire psp e_paranthropus TU via shared header extraction

eibrunocorrea:
Extracts the fully-decompiled `are/e_paranthropus.c` implementation into
`src/st/e_paranthropus.h` and turns both `are` and `rno1` into thin wrappers,
following the `e_bat.h` / `e_red_door.h` pattern. This wires the whole
`rno1_psp/unk_1028` TU (5 functions: `ParanthropusSetStep`,
`EntityParanthropus`, `EntityParanthropusThrownBone`,
`EntityParanthropusBoneHitbox`, `EntityParanthropusSkull`) from
`INCLUDE_ASM` to shared C; the stub file is deleted.

No defines are needed: both stage headers declare the same `E_PARANTHROPUS*`
enum names, so entity ids resolve per overlay. The only per-overlay mapping
is `g_EInitParanthropusThrownBone -> D_us_801806E8` (the name the sibling
asm objects already use).

- `.data` window `[0x16000]` (sensors + anim tables, verified against the
  retail binary by initializer bytes; ends flush against `st_update`'s data)
- `.rodata` `[0x38300]` is the function's 12-entry jump table
- `are`'s US/HD binaries are unchanged (wrapper preserves bytes)

Validation: `us` 113/113, `hd` 12/12, `pspeu` 64/64 checksums green with
this change alone on top of master; `make format` clean.

Part of a larger PSP effort staged publicly at
`eibrunocorrea/sotn-decomp` (branch `rno0-code-share`) — being upstreamed in
small self-contained PRs like this one. Happy to split or adjust further.


src/st/rno1/e_paranthropus.c
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "rno1.h"
+
+#ifdef VERSION_PSP
+// psp symbols keep the D_us_* name used by sibling asm objects
+#define g_EInitParanthropusThrownBone D_us_801806E8
```
sozud: We ought to be able to get rid of the define with a splat symbol in https://github.com/Xeeynamo/sotn-decomp/blob/master/config/symbols.pspeu.strno1.txt

eibrunocorrea: Done — renamed it in both symbol tables (pspeu and us) and dropped the define. The us side needed the rename too since the compiled e_init.c is what provides the EInit on both versions.

sozud (approved):

PR #3517, opened 2026-08-09
bo7: wire psp boss floor snap as C in mixed TU

eibrunocorrea:
Wires bo7_psp's copy of the boss floor-snap helper (`func_us_80194338`) as C in the mixed TU, with bo7's floor height (`0xD0` vs rbo0's `0xE0`) inlined.

Per review: no shared header — rbo0 is untouched, and the copy lives inline in `bo7_psp/unk_E700.c` under the same symbol name, so a future US/PSP merge just drops the duplicate. 1 file, +16/−1.

Validation: us 113/113, hd 12/12, pspeu 64/64 checksums green on top of master; `make format` clean.


hohle (changes requested):
Shared code in headers like this is typically used to share code across overlays, not across two version of the same overlay. My personal preference would be to leave these inline with the same symbol names and when then PSP and US code is merged at some point, there's no extra file cleanup to be done (or forgotten about).

It's possible that this represents an entire object, but then the file should be split to represent that. If this is only a BO7 function, the include will not be necessary when PSP and US are combined, which is the ultimate goal.

The includes idiom is used to share code that represents entire objects in a way that's compatible with how the splat namespaces for overlays are configured.

src/boss/rbo0/e_boss.c
```diff
@@ -1,20 +1,7 @@
...
-    posY = 0xE0 - posY;
-    if (posY <= 0) {
-        g_CurrentEntity->posY.i.hi += posY;
-        g_CurrentEntity->velocityX = 0;
-        g_CurrentEntity->velocityY = 0;
-        return true;
-    }
-    return false;
-}
+#include "../e_boss_floor_snap.h"
```
JoshSchreuder: Inclined to agree with @hohle here, I don't see much benefit in pulling this one function out into a shared header

JoshSchreuder (changes requested):

src/boss/rbo0/e_boss.c
eibrunocorrea: Agreed — dropped the header entirely. `rbo0/e_boss.c` is back to master untouched, and `bo7_psp` now carries the copy inline as C under the same symbol name with the `0xD0` floor inlined, so the PR is down to the single psp file. Thanks @hohle for spelling out the includes-idiom boundary — applying it going forward.

JoshSchreuder (approved):

PR #3518, opened 2026-08-09
Extract DRA equipment and accessory definitions with asset tool

mathisto:
I'll be taking that equipment

## Summary
Adds separate `equipment` and `accessories` asset handlers that extract the complete `Equipment` and `Accessory` tables into editable YAML and regenerates C initializer headers in the existing compilation units. Extends the shared string handler with Shift-JIS menu-description round trips and PSP European control-code substitutions. 

This covers 2,149 records across fourteen tables:
- US PlayStation: 217 equipment and 90 accessories
- Japanese PlayStation/HD: 217 equipment and 90 accessories
- PSP English, Spanish, Italian, German, and French: 217 equipment and 90 accessories each

The handler models the complete structures, including names, descriptions, prices, categories, icons and palettes, attack and defense values, stat bonuses, elemental and status properties, and flags. Enum and flag values use the existing declarations, while declared sizes, zero padding, boolean bytes, string round trips, and expected record counts protect against malformed or incomplete extraction.

Generated YAML and initializer headers remain ignored. Keeping the generated initializers inside the original source files lets the compiler continue to place and deduplicate the pointed-to strings instead of retaining hardcoded `.rodata` addresses.

## Name and description encoding
Item names use the game's 8x8-font encoding. PSX and HD descriptions use Shift-JIS, while the five PSP European tables use the PSP character mapping and control-code substitutions.

PSX and HD generated sources retain readable names and byte-safe description literals. PSP headers contain pre-encoded names and descriptions because its string-processing step runs before generated includes are expanded.


JoshSchreuder:
This LGTM, any objections @sozud @Xeeynamo ?

JoshSchreuder (approved):

sozud (approved):
Seems fine to me but probably needs a look from Xeeynamo since they know the assets tool the best

Xeeynamo (commented):
Couple of things:
* I noticed the `itemdefs` handler branches internally based on accessory or equipment. Why not creating two separate handlers instead?
* ~We have `blueprints` and you are introducing `accessories`. Can you make `equipment` also plural?~

JoshSchreuder:
> Can you make equipment also plural?

Equipment doesn't have a plural (mass noun)

Xeeynamo:
> > Can you make equipment also plural?
> 
> Equipment doesn't have a plural (mass noun)

T.I.L. 😃

JoshSchreuder (changes requested):
See @Xeeynamo post above

mathisto:
Addressed the handler-split feedback. Equipment and accessories now have independent handlers with no mode dispatch; all 14 table configurations use the new names. Shared encoding/validation helpers remain shared. Kept `equipment` unchanged, as agreed above.

JoshSchreuder (approved):
Changes look good, @Xeeynamo can you please review as well

Xeeynamo (approved):

PR #3519, opened 2026-08-09
rcen: decompile chamber cutscene controllers (us + pspeu)

mathisto:
Controls

## Summary
Decompile the RCEN chamber and cutscene controller family for the US PlayStation and European PSP builds.

This replaces six assembly stubs in each version (12 total) with C and maps the small data tables they consume. The two implementations remain separate where the retail versions differ.


config/splat.pspeu.strcen.yaml
```diff
@@ -85,6 +85,8 @@ segments:
       - [0x31900, data] # background_block_init
       - [0x31A10, .data, rcen/e_lock_camera]
       - [0x31A30, data]
+      - [0x31C20, .data, rcen_psp/unk_1F0D8]
+      - [0x31C58, data]
```
JoshSchreuder, line 89: Can these be merged together with PSX side? (c + data)

mathisto: No problem! I see the virtue in this; nice. I'll capture this pattern in my notes for future work. Thankee, sai.

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,602 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include "rcen.h"
+#include "sfx.h"
 
-INCLUDE_ASM("st/rcen/nonmatchings/unk_1F0D8", func_8018F8EC);
+extern u32 g_CutsceneFlags;
+extern EInit g_EInitCommon;
+extern EInit g_EInitInteractable;
+#ifndef VERSION_PSP
+extern u16 PLAYER_posX_i_hi;
```
JoshSchreuder: Change this to PLAYER.posX.i.hi at the call site and you shouldn't need this

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,602 @@
...
+            g_PauseAllowed = 0;
+            g_unkGraphicsStruct.pauseEnemies = 1;
+#else
+        if ((GetDistanceToPlayerX() < 0x20) &&
+            ((player->posY.i.hi - self->posY.i.hi) < 0x70)) {
+            g_unkGraphicsStruct.pauseEnemies = 1;
+            g_PauseAllowed = 0;
+#endif
+#ifdef VERSION_PSP
+            if (posX > (0x181 - 1)) {
```
JoshSchreuder: posX > 0x180 

and this should also work on PSX I suspect

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,602 @@
...
+                g_Player.padSim = PAD_NONE;
+#else
+                g_Player.padSim = PAD_MIST;
+#endif
+            } else if (g_Player.status & PLAYER_STATUS_WOLF_FORM) {
+                g_Player.padSim = PAD_WOLF;
+            }
+#ifdef VERSION_PSP
+            g_Entities[E_AFTERIMAGE_1].ext.afterImage.disableFlag = 0;
+            g_Player.demo_timer = 1;
```
JoshSchreuder, line 86: Suspect the PSP order is fine here and you only need to ifdef the tilemap call

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,602 @@
...
+#else
+                    g_Player.padSim = PAD_MIST;
+#endif
+                } else if (g_Player.status & PLAYER_STATUS_WOLF_FORM) {
+                    g_Player.padSim = PAD_WOLF;
+                }
+            }
+        } else {
+#ifdef VERSION_PSP
+            if (posX > (0x181 - 1)) {
```
JoshSchreuder: Same here

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,602 @@
...
+
+void func_us_8019F5F0(Entity* self) {
+#ifdef VERSION_PSP
+    Entity* player;
+#endif
+    Primitive* prim;
+#ifdef VERSION_PSP
+    s32 primIndex;
+#else
+    s16 primIndex;
```
JoshSchreuder: primIndex should be s32 on both platforms I think

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,602 @@
...
+    u8 color;
+#endif
+
+    switch (self->step) {
+    case 0:
+#ifdef VERSION_PSP
+        primIndex = (s16)g_api.AllocPrimitives(PRIM_G4, 1);
```
JoshSchreuder: Cast and condition here shouldn't be required?

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,602 @@
...
+    case 2:
+        if (g_CutsceneFlags & 0x100) {
+            g_unkGraphicsStruct.unk20 = 0;
+#ifdef VERSION_PSP
+            self->step++;
+            self->ext.utimer.t = 0;
+#else
+            self->ext.utimer.t = 0;
+            self->step++;
+#endif
```
JoshSchreuder: PSP order should work on PSX with no conditional

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,602 @@
...
+        }
+        g_Player.padSim = PAD_NONE;
+        g_Player.demo_timer = 1;
+        break;
+
+    case 3:
+        self->ext.utimer.t++;
+#ifdef VERSION_PSP
+        if (self->ext.utimer.t > (0x181 - 1)) {
```
JoshSchreuder: Same here

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,602 @@
...
+            prim->b0 = 0;
+            prim->g3 = 0;
+            prim->g2 = 0;
+            prim->g1 = 0;
+            prim->g0 = 0;
+            prim->r3 = 0;
+            prim->r2 = 0;
+            prim->r1 = 0;
+            prim->r0 = 0;
+#endif
```
JoshSchreuder: Conditional can likely be removed

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,602 @@
...
+            if (g_PlayableCharacter != PLAYER_ALUCARD) {
+                D_800978B4 = 4;
+                g_GameState = Game_Ending;
+                g_GameStep = 0;
+            } else {
+#ifdef VERSION_PSP
+                player = &PLAYER;
+                player->posX.i.hi -= 0x200;
+#else
+                PLAYER_posX_i_hi -= 0x200;
```
JoshSchreuder: I suspect the PSP code will work on PSX side here too

JoshSchreuder (changes requested):

mathisto:
Ooof, I have a lot to learn. Ok, I'll get these fixed up and tested to make sure they still hold.

JoshSchreuder:
Left some comments - looking better but needs a bit of work

I would recommend taking the PSP scratch, remove all conditionals so that it's matching completely, then copy that same code into PSX scratch

Usually the PSX side will match exactly with the PSP matching code, negating the need to conditional anything. If there is genuinely platform specific code on PSP or PSX then you can conditional that, but most of what I see here can likely be cleaned up and most conditionals removed

Feel free to post any scratches if you get stuck trying to coalesce them

src/st/rcen/unk_1F0D8.c
JoshSchreuder: To clarify this a little, the compiler generally combines constants like this together, so it is just simpler to write 0x180

In the PSP case you correctly identified it needed a > rather than >= which is a specific PSP optimisation trick to hit the at register than PSX at O2 doesn't require (one of the reasons why using PSP as the baseline target helps us get more "accurate" decomp)

src/st/rcen/unk_1F0D8.c
JoshSchreuder: To clarify these ones, PSX at O2 cares much less about the order of instructions when it determines they don't interact within a block than PSP, so PSP at O0 helps us get the more accurate order of operations

mathisto:
@JoshSchreuder I tried `PLAYER.posX.i.hi` on the PSX side, but the compiler generated diff instructions and broke byte match with RCEN.

JoshSchreuder:
> @JoshSchreuder I tried `PLAYER.posX.i.hi` on the PSX side, but the compiler generated diff instructions and broke byte match with RCEN.

Here's a matching PSX + PSP scratch with all the ifdefs removed. I tried importing it locally and got a green build on both too
https://decomp.me/scratch/pVCaQ
https://decomp.me/scratch/09ZH4

Does that work for you?

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,540 @@
...
 
-INCLUDE_ASM("st/rcen/nonmatchings/unk_1F0D8", func_us_8019F5F0);
+static u8 glassAnimation[] = {
+    0x04, 0x03, 0x04, 0x02, 0x08, 0x01, 0x04, 0x02, 0x04, 0x03, 0x18,
+    0x04, 0x02, 0x03, 0x02, 0x02, 0x02, 0x01, 0x18, 0x05, 0xFF};
 
-INCLUDE_ASM("st/rcen/nonmatchings/unk_1F0D8", func_us_8019F9C0);
+#ifndef VERSION_PSP
+static u8 unused[] = {0x7F, 0x60, 0x58, 0x50, 0x40};
+#endif
```
JoshSchreuder: Ifdef here shouldn't be required, PSP will automatically strip unused data

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,540 @@
...
+    Primitive* prim;
+    s32 primIndex;
+
+    switch (self->step) {
+    case 0:
+        primIndex = g_api.AllocPrimitives(PRIM_G4, 1);
+        if (primIndex != -1) {
+            InitializeEntity(g_EInitInteractable);
+#ifdef VERSION_PSP
```
JoshSchreuder: Apologies as I didn't comment on all these, but my earlier comments apply - I would go through and assess whether these ifdefs are actually required to match by using the PSP version as the "source of truth" for the PSX version.

In this function none of the ifdefs are required
https://decomp.me/scratch/pVCaQ

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,540 @@
...
+
+    case 6:
+    case 7:
+        break;
+    }
+}
+
+void func_us_8019F9C0(Entity* self) {
+    Primitive* prim;
+#ifdef VERSION_PSP
```
JoshSchreuder: Same with all these and the other ifdefs remaining - please go through and check if these are required

mathisto: On it.

mathisto: You were VERY right and I was VERY wrong. I think I understand the sharing mechanism better now. Thank you so much for taking the time to rip this apart and help me learn. This is so outside my normal wheelhouse and I deeply appreciate the effort you are putting into me and other contributors.

I tried to roll in all of your suggestions and am much happier with the result. 🙇

JoshSchreuder: @mathisto no worries! Hopefully my comments didn't come across as ripping you personally apart 😄 

Really appreciate you taking the time to understand the feedback and incorporate it as it's all coming from a place of making the codebase clean and understandable.

src/st/rcen/unk_1F0D8.c
```diff
@@ -1,14 +1,409 @@
...
+
+    switch (self->step) {
+    case 0:
+        InitializeEntity(g_EInitInteractable);
+        func_8018F8EC(0);
+        self->animSet = ANIMSET_OVL(1);
+        self->animCurFrame = 9;
+        self->zPriority = 0x80;
+        self->drawFlags = ENTITY_ROTATE;
+        self->rotate = 0x800;
```
JoshSchreuder: Nitpicky stuff now, this can be `ROT(180)`, there's a few `->rotate` calls in here which can be updated too.

Might write an auto lint for this stuff so we don't have to mention it on PRs and the linter can do it for us.

mathisto: Nitpick away! I will try not to make these same class of errors in the future. I've been building a little patterns and style guide as I learn from y'all.

I converted both the hex constant and the rotate call to `ROT(180)`. 

I love the idea of a linter. I'd be willing to work on that if you have any other patterns in mind.

JoshSchreuder (approved):

sozud (approved):

PR #3521, opened 2026-08-10
bo5: decompile cutscene dialogue helpers via shared headers (us + pspeu)

eibrunocorrea:
Wires the bo5 cutscene dialogue TU to the shared `src/st/` cutscene
helpers on both versions: the US TU swaps its remaining `INCLUDE_ASM`s
for the shared includes, and the psp stub is rebuilt as a composition
file over the same headers with the psp helper order, so the US and PSP
sides land together.

- the shared headers (`cutscene_actor_name.h`, `cutscene_unk4.h`,
  `set_cutscene_script.h`) gain `#ifndef` defaults so their existing
  users are unaffected
- `actor_names`/`g_Dialogue`/`g_CutsceneFlags` named in the us
  symbols config
- validated standalone on current master: us + hd + pspeu checksums
  all green

First slice of the boss-side cutscene family; bo2/rbo2 and
rbo6/mar/bo6 follow on top of this one. Happy to adjust granularity.


src/st/cutscene_unk4.h
```diff
@@ -13,7 +13,7 @@
 #define CUTSCENE_UNK4_V 12
 #endif
 
-#ifndef VERSION_HD
+#if !defined(VERSION_HD) && !defined(CUTSCENE_GLOBAL)
```
sozud: Take a look at https://github.com/Xeeynamo/sotn-decomp/blob/master/src/st/nz0/cutscene.c
I think the issue here is the file is split at the wrong spot

eibrunocorrea: Good catch, and agreed — the helpers only went global because their callers (EntityCutscene and friends) are still raw asm in sibling TUs, so `static` symbols were not reachable from their `jal`s. The right boundary is the one nz0's cutscene.c shows: keep the callers' asm in the same TU as the helpers so everything can stay `static`. I'll rework this slice (and #3522/#3523, which share the recipe) to fold those INCLUDE_ASMs into the composition TU and drop `CUTSCENE_GLOBAL` from the shared headers entirely. Will push the rework here.

eibrunocorrea: Reworked as suggested — the split was indeed the problem. The caller TUs are now folded into `e_cutscene_dialogue` on both platforms (us: former `unk_24A4C` with events/scale decompiled via the shared headers; psp: the cutscene head of the former `unk_16270`, with the non-cutscene tail split into its own TU), so every helper is `static` again and the `CUTSCENE_GLOBAL` knob is gone — the three shared headers are back to master byte for byte. Same pattern as `nz0/cutscene.c` and the bo6 US TU. All three checksums green from a fresh extract.

src/boss/bo5/e_cutscene_dialogue.c
```diff
@@ -1,12 +1,29 @@
...
+#include "../../st/cutscene_unk3.h"
 
-INCLUDE_ASM("boss/bo5/nonmatchings/e_cutscene_dialogue", DrawCutsceneActorName);
+#include "../../st/cutscene_unk4.h"
+
+#include "../../st/cutscene_actor_name.h"
+
+#include "../../st/set_cutscene_events.h"
+
+#define STAGE_IS_DRE
```
JoshSchreuder: This should not be here

eibrunocorrea: Right — replaced with a proper feature knob following the header's existing option pattern: `CUTSCENE_WAIT_FOR_FLAG_REWIND` (documented next to `CUTSCENE_TILEMAP_SCROLL`), and dre now defines it explicitly at its include site instead of relying on its stage identity macro. Checksums green, dre byte-identical.

sozud (approved):

PR #3522, opened 2026-08-10
bo2/rbo2: decompile cutscene dialogue helpers via shared headers (us + pspeu)

eibrunocorrea:
Same recipe as #3521 applied to bo2 and rbo2: both US composition
files swap their `INCLUDE_ASM`s for the shared `src/st/` cutscene
includes and gain the psp helper order under `VERSION_PSP`, replacing
the two standalone psp stubs.

- the two overlays land together because the rbo2 psp stub was shared
  with bo2 (deleting it touches both)
- `actor_names`/`g_Dialogue`/`g_CutsceneFlags` named in both us
  symbol configs
- validated standalone on current master: us + hd + pspeu checksums
  all green

Builds on top of #3521 (the shared-header defaults land there); only
the last commit is new here — will rebase once it lands.

PR #3523, opened 2026-08-10
rbo6/mar/bo6: wire psp cutscene dialogue via shared includes

eibrunocorrea:
Completes the boss-side cutscene dialogue family: the rbo6 US TU swaps
its `INCLUDE_ASM`s for the shared includes, and the three psp stubs
(rbo6, mar, bo6) are rebuilt as composition files over the same shared
headers with the psp helper order.

- per-overlay language actor-name tables (`actor_names_en`..`_ge`)
  named in the pspeu symbol configs
- mar and bo6 need no US-side changes (psp-only wiring)
- validated standalone on current master: us + hd + pspeu checksums
  all green

Builds on top of #3521; only the last commit is new here — will rebase
once it lands.


eibrunocorrea:
Heads-up: force-pushed the same restructure as #3521/#3522 — psp cutscene heads folded into `e_cutscene_dialogue` per overlay (rbo6 us additionally absorbs the four helpers that lived at the tail of `unk_1D690`, and `CutsceneUnk1` is now named in the rbo6 symbol map instead of a define shim). Helpers are all `static`; the shared headers carry no knobs. Checksums green.

src/boss/mar_psp/e_cutscene_dialogue.c
```diff
@@ -1,14 +1,25 @@
...
+#define CUTSCENE_UNK1_UNK17 2
+#include "../../st/cutscene_unk1.h"
+
+#include "../../st/set_cutscene_script.h"
+
+INCLUDE_ASM("boss/mar_psp/nonmatchings/mar_psp/e_cutscene_dialogue", SetCutsceneEvents);
+
+INCLUDE_ASM("boss/mar_psp/nonmatchings/mar_psp/e_cutscene_dialogue", RunCutsceneEvents);
+
+INCLUDE_ASM("boss/mar_psp/nonmatchings/mar_psp/e_cutscene_dialogue", MAR_EntityCutsceneDialogue);
```
JoshSchreuder: `MAR_` prefix should be removed

JoshSchreuder (changes requested):

src/boss/mar_psp/e_cutscene_dialogue.c
eibrunocorrea: Renamed via the symbols file (`MAR_EntityCutsceneDialogue` → `EntityCutsceneDialogue`), same treatment as the bo2/rbo2 ones in #3522. Checksums unchanged.

sozud (approved):

PR #3525, opened 2026-08-10
bo0: name the Olrox drool positional sfx helper

eibrunocorrea:
Repurposed after #3530, which superseded the original change here (the second `PlaySfxPositional` copy is now decompiled inline with its `0x58` max volume in both `3A030.c` twins — the exact shape requested in review).

What's left is the rename suggested in review: the helper's only caller is `EntityRealOlroxDrool`, so the placeholder `func_us_801BA030` becomes `PlayOlroxDroolSfx` on both the us and psp sides (`static` preserved on us). 2 files, 4 lines, byte-neutral.

Validation: us 113/113, hd 12/12, pspeu 64/64 checksums green on top of master; `make format` clean.


src/boss/bo0_psp/2D26C.c
```diff
@@ -19,7 +19,12 @@ INCLUDE_ASM("boss/bo0_psp/nonmatchings/bo0_psp/2D26C", func_pspeu_09256A10);
 
 INCLUDE_ASM("boss/bo0_psp/nonmatchings/bo0_psp/2D26C", func_pspeu_09257F58);
 
-INCLUDE_ASM("boss/bo0_psp/nonmatchings/bo0_psp/2D26C", func_pspeu_09258530);
+// bo0 plays positional sfx quieter than the stages; this is the overlay's
+// second copy of the helper, kept under its placeholder name (the sibling
+// asm object calls it) to avoid clashing with the wired copy at 0x09246A70
+#define PlaySfxPositional func_pspeu_09258530
```
JoshSchreuder: This should be renamed via symbol on PSP, not #define.

This is a common issue with all of your PRs, can you please address this in advance going forwards?

src/boss/bo0_psp/2D26C.c
```diff
@@ -19,7 +19,12 @@ INCLUDE_ASM("boss/bo0_psp/nonmatchings/bo0_psp/2D26C", func_pspeu_09256A10);
...
+#define SFX_POSITIONAL_MAX_VOL 0x58
+#include "../../st/e_play_sfx_positional.h"
```
JoshSchreuder: I don't think this should really be in a shared header either. If it's only in one overlay and only on PSP side then let's just take the inline copy with the specific max volume difference inlined. There is probably some associated entity here which calls this and this can be renamed and tied to that entity accordingly.

JoshSchreuder (changes requested):

JoshSchreuder:
Also this needs rebase

src/boss/bo0_psp/2D26C.c
eibrunocorrea: Agreed on the norm, and it's now applied in advance: I swept the other open PRs and this was the last remaining `#define` shim (the bo2/rbo2/mar prefix removals in #3522/#3523 are done via symbols too).

Meanwhile #3530 landed and superseded this PR — `2D26C.c` is gone and the helper is decompiled inline with the `0x58` volume in both `3A030.c` twins, exactly the shape you describe below. I've repurposed the PR to carry just the rename you suggested: the only caller is `EntityRealOlroxDrool`, so `func_us_801BA030` is now `PlayOlroxDroolSfx` on both sides. Checksums unchanged.

src/boss/bo0_psp/2D26C.c
eibrunocorrea: Gone with the rebase — master's inline copy (via #3530) already has the volume difference inlined, no shared header. This PR is now just the entity-tied rename.

JoshSchreuder:
Let's do renames as part of human analysis, not LLM discovery.

PR #3526, opened 2026-08-10
rno2: decompile azaghal (us + pspeu)

eibrunocorrea:
Splits the azaghal block out of `unk_47A9C` into its own TU: eight
functions (six static helpers plus `EntityAzaghal` and
`EntityAzaghalSwordHitbox`) mirroring the structure of the `rare`
donor, with the psp side compiled from the same file under
`VERSION_PSP` carrying the transplanted static data block.

- no new symbols needed: everything the slice uses already exists in
  the symbol configs (names came with the rare decompilation)
- the tail function `func_us_801CAB8C` stays as asm in a new
  `unk_4AB8C` TU; the karasuman block keeps the `rno2_psp` stub
- the four `PAL_AZAGHAL_*` palette values join the rno2 enum with the
  same values as rare's
- validated standalone on current master with fresh splat extraction:
  us + hd + pspeu checksums all green


src/st/rno2/e_azaghal.c
```diff
@@ -0,0 +1,1216 @@
...
+        g_CurrentEntity->pose++;
+    }
+
+    return true;
+}
+
+// azaghal animation paths: target positions and lerp curves live in the
+// data blob as (SVECTOR*, s16*) pairs; alias each path to its blob label
+// (addresses proven by the relocs in EntityAzaghal)
+#define anim_seek_player D_us_80182140
```
sozud: If we know what these do let's rename the actual data, update the splat syms if needed

JoshSchreuder: This code looks identical to https://github.com/eibrunocorrea/sotn-decomp/blob/9d3779ff89b990c2d9bceac549a9399645451ee2/src/st/rare/e_azaghal.c#L10 besides symbols

Can probably be worked into a shared header with the symbols mapped for rno2 rather than what's done here.

sozud:
Actually I missed that we already have this one decompiled. We should factor out to a header and share between the stages unless there's a lot of differences https://github.com/Xeeynamo/sotn-decomp/blob/95a6d96b666224bcd9773cdaf829fb0453b12c7c/src/st/rare/e_azaghal.c#L256

Edit: posted almost simultaneously with JoshSchreuder

src/st/rno2/e_azaghal.c
eibrunocorrea: Done — dropped the define aliases and renamed the actual data: the blob labels are now named in `symbols.us.strno2.txt` and the psp static mirrors carry the same names directly. `make format` also pruned the symbol entries that became redundant once the functions landed as C. All three checksums still green.

eibrunocorrea:
Agreed — it is the rare code with rno2 symbols (that's where the transplant came from). I'll factor it into a shared `src/st/` header with the per-stage symbols parametrized, same pattern as the dai shortcut, keeping rare byte-identical, and push the rework here.

JoshSchreuder:
> I'll factor it into a shared src/st/ header with the per-stage symbols parametrized, same pattern as the dai shortcut

I might be misunderstanding you, but this should not be required. There should be no additional #DEFINEs necessary to do this. You can map the same symbol names from the shared header in rno2's symbol map

It looks like you've only imported the data on PSP side and not PSX, so importing PSX data would likely solve most of the symbol issues.

eibrunocorrea:
Reworked into the shared header as suggested: `src/st/e_azaghal.h` now carries the whole entity (data and code), and both stage files reduce to a two-line include (`rare.h`/`rno2.h` + the shared header).

No defines involved. For psx rno2 the header declares the data as externs (under `STAGE_IS_RNO2 && !VERSION_PSP`) using the same names as rare's statics, and rno2's symbol map resolves them into the stage data (`sensors_ground`, the sword vectors and `indices` join the animation labels that were already named there). rare psx keeps its statics untouched, and the PSP side of both stages compiles the same static branch as before. us/hd/pspeu checksums all green from a fresh extract.

sozud (approved):

src/st/e_azaghal.h
```diff
@@ -0,0 +1,1197 @@
...
+#include <scratchpad.h>
+
+extern EInit g_EInitAzaghal;
+
+typedef struct {
+    SVECTOR* targetPosition;
+    s16* lerpVals;
+} AzaghalPosition;
+
+#if defined(STAGE_IS_RNO2) && !defined(VERSION_PSP)
```
sozud: Sorry I missed this part. Why can't we import the data?

eibrunocorrea: No reason not to, turns out — imported. The rno2 data segment is now carved as `e_azaghal`'s own `.data` window and the header compiles the same statics for both stages on both platforms, so the extern branch (and the symbol-map names) are gone entirely. The only conditionals left in the header are the two real retail differences (rno2 sets `zPriority` and initializes the hitbox fields in the opposite order). All three checksums green.

PR #3529, opened 2026-08-10
rbo1: decompile Beelzebub encounter controller

mathisto:
## Summary
Decompile `func_us_80194C50`, the RBO1 Beelzebub encounter controller.

This controller creates and sequences the room blocks, gates the fight and doors, records time attack, manages music, and spawns the reward. Its C, data, and rodata are split at existing binary boundaries so the jump table should retain the retail layout; the adjacent `EntityBossRoomBlock` remains in assembly.

The function uses one local fixed-register constraint to reproduce the retail compiler allocation. Without it, the generated instructions are otherwise equivalent but exchange two saved registers.


config/splat.us.borbo1.yaml
```diff
@@ -71,6 +74,8 @@ segments:
       - [0x11FD8, c, unk_11FD8]
       - [0x120B8, c, e_lock_camera]
       - [0x12274, c, unk_12274]
+      - [0x14C50, c, func_us_80194C50]
+      - [0x14FCC, c, entity_boss_room_block]
```
JoshSchreuder, line 78: This order holds on PSP so likely these were part of the same file

We would probably just merge this back together later as the two functions look related, so I would just keep these two together here, and not attempt to merge the PSP and PSX for now until someone has decomp'd the whole file

I know this is a little counter to what I wrote in the other PR so apologies. PSX and PSP can generally be combined once the full file is decompiled but for cases like this we should assess the function/data order and context to try to determine if functions should live together in a file or not

src/boss/rbo1/func_us_80194C50.c
```diff
@@ -0,0 +1,201 @@
...
+#define RBO1_BOSS_FLAGS D_pspeu_09256DC8
+#define RBO1_E_BOSS_ROOM_BLOCK D_pspeu_09256D50
+#define RBO1_E_LIFE_UPSPAWN D_pspeu_09256D48
+#define RBO1_E_UNK_16 D_pspeu_09256D98
+#else
+extern s32 D_us_801A1F1C;
+#define RBO1_BOSS_FLAGS D_us_801A1F1C
+#define RBO1_E_BOSS_ROOM_BLOCK E_BOSS_ROOM_BLOCK
+#define RBO1_E_LIFE_UPSPAWN E_LIFE_UPSPAWN
+#define RBO1_E_UNK_16 E_UNK_16
```
JoshSchreuder, line 37: These can be `E_ID(UNK_16)` etc. at the call site, and symbols mapped on PSP side to mostly remove the ifdefs here

PSP would need `extern s32 E_ID(UNK_16)`

src/boss/rbo1/func_us_80194C50.c
```diff
@@ -0,0 +1,201 @@
...
+extern s32 D_pspeu_09256D50;
+extern s32 D_pspeu_09256D98;
+extern s32 D_pspeu_09256DC8;
+#define RBO1_BOSS_FLAGS D_pspeu_09256DC8
+#define RBO1_E_BOSS_ROOM_BLOCK D_pspeu_09256D50
+#define RBO1_E_LIFE_UPSPAWN D_pspeu_09256D48
+#define RBO1_E_UNK_16 D_pspeu_09256D98
+#else
+extern s32 D_us_801A1F1C;
+#define RBO1_BOSS_FLAGS D_us_801A1F1C
```
JoshSchreuder, line 34: Instead of a define here, you could map this as a symbol (eg. `bossFlags`) in both the PSX and PSP symbol map to remove the ifdef.

I'm guessing this probably lives in BSS and we can import it later once we work out which file would be the obvious owner

src/boss/rbo1/func_us_80194C50.c
```diff
@@ -0,0 +1,201 @@
...
+#define RBO1_E_BOSS_ROOM_BLOCK E_BOSS_ROOM_BLOCK
+#define RBO1_E_LIFE_UPSPAWN E_LIFE_UPSPAWN
+#define RBO1_E_UNK_16 E_UNK_16
+#endif
+
+void func_us_80194C50(Entity* self) {
+    Entity* entity;
+#ifndef VERSION_PSP
+    u16* placementY;
+    register Entity* entityToCreate __asm__("$17");
```
JoshSchreuder, line 44: Register usage is almost unseen in overlays, did you have a scratch we can poke at to try to work this out?

JoshSchreuder (changes requested):

src/boss/rbo1/func_us_80194C50.c
mathisto: Hi Josh! I might need a hand with this one in that case. I hate to bother you but I tried to find a way around the register but failed multiple times trying to make it work. I have [my working scratch here](https://decomp.me/scratch/Yul7S).

JoshSchreuder: @mathisto here you go, yours was pretty close.

As always I recommend using the PSP as a starting point and copying your match over to PSX side. 95% of the time it will match out of the box, otherwise with a few small tweaks here and there. PSP with O0 optimisations is far far easier to decode than PSX side 😄

Mostly it's easier to see what variables are real and which aren't, which led me to remove your `entityToCreate` and `scroll`.

// PSX: https://decomp.me/scratch/qOSG2
// PSP: https://decomp.me/scratch/TA6Wi

src/boss/rbo1/entity_boss_room_block.c
```diff
@@ -0,0 +1,5 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#include "rbo1.h"
+
+INCLUDE_ASM(
+    "boss/rbo1/nonmatchings/entity_boss_room_block", EntityBossRoomBlock);
```
Xeeynamo, line 5:
```suggestion
// clang-format off
INCLUDE_ASM(
    "boss/rbo1/nonmatchings/entity_boss_room_block", EntityBossRoomBlock);
```

The formatting here will break a bunch of automated tools that expect `INCLUDE_ASM` to be one line. Let's temporarily disable formatting, or rename the file shorter as `e_boss_room_block`. It could be even shorter, because we know the context of the overlay is `boss_room` 🤔

src/boss/rbo1/func_us_80194C50.c
```diff
@@ -0,0 +1,201 @@
...
+#define RBO1_BOSS_FLAG_DEFEATED (1 << 1)
+#define RBO1_BOSS_FLAG_REWARD_READY (1 << 2)
+#define RBO1_BOSS_FLAG_DOORS_CLOSE (1 << 3)
+#define RBO1_BOSS_FLAG_DOORS_OPEN (1 << 4)
+
+typedef struct {
+    s16 x;
+    s16 y;
+    s16 params;
+} Rbo1BlockPlacement;
```
Xeeynamo, line 14:
```suggestion
} Rbo1TilePlacement;
```

I propose to rename `Block` to `Tile`

src/boss/rbo1/func_us_80194C50.c
```diff
@@ -0,0 +1,201 @@
...
+    {-8, 0x168, 0},    {-8, 0x178, 0},    {-8, 0x188, 0},    {-8, 0x198, 0},
+};
+
+extern EInit g_EInitInteractable;
+#ifdef VERSION_PSP
+extern s32 D_pspeu_09256D48;
+extern s32 D_pspeu_09256D50;
+extern s32 D_pspeu_09256D98;
+extern s32 D_pspeu_09256DC8;
+#define RBO1_BOSS_FLAGS D_pspeu_09256DC8
```
Xeeynamo, line 28: I noticed you're introducing aliases for PSP and US to share the code. Most of them can be achieved seamlessly by tweaking the correspondent `config/symbols.*.txt`, which will both eliminate a level of code redirection and name the correct symbol in `build/*/borbo1.map`

Xeeynamo (dismissed):
Requesting some changes.

I believe too many `#ifdef VERSION_PSP` are hiding the real code code that could match both versions. Let's think of that in a second review phase.

JoshSchreuder (approved):
Feedback looks addressed, thanks!

Going to merge this as it's been sitting up for quite a while

PR #3530, opened 2026-08-10
Aligning BO0 US and PSP, Additional Matches, Some New Functions

hohle:
Dedupes several files from NO2. Any named files are new matches from existing code. Some additional decompiled functions that were used to determine where various segments were. Also `st_init_psp`.


src/boss/bo0/st_init_psp.c
```diff
@@ -0,0 +1,138 @@
...
+s32 D_pspeu_092903C0;
+s32 D_pspeu_092903B8;
+s32 D_psp_E_SPIKES_DAMAGE;
+s32 D_psp_E_BOSS_DOORS;
+s32 D_pspeu_092903A0;
+s32 D_pspeu_09290398;
+
+void InitEntityIds(void) {
+    D_pspeu_09290568 = 0x11;
+    D_pspeu_09290560 = 0x12;
```
Xeeynamo:
```suggestion
    E_ID(BACKGROUND_BLOCK) = E_BACKGROUND_BLOCK;
    E_ID(LOCK_CAMERA) = E_LOCK_CAMERA;
```
can we use the same pattern used in other overlays?

hohle: Absolutely.

PR #3531, opened 2026-08-10
sotn-disk: Preserve Timestamps During Export

hohle:
Preserves the ISO-9660 timestamp when exporting files from a disk image.


Xeeynamo (approved):
Thanks!

PR #3533, opened 2026-08-10
Weapon psp small

hohle:
The PSP version of w0_000 mostly works for `us` without the need for
conditional blocks.


Xeeynamo (approved):
Clean!

PR #3537, opened 2026-08-11
Test demos on CI

Xeeynamo:
* Unify asset extraction job with the other jobs to save time.
* Tests succeed if SOTN PC port has exit code 0 for every executed demo

### DIFF
--- a/.github/workflows/picci.yaml
+++ b/.github/workflows/picci.yaml
@@ -24,14 +24,10 @@ on:
   workflow_dispatch:
 
 env:
-  SDL_VIDEODRIVER: offscreen
-  SDL_AUDIODRIVER: dummy
   DEMOS: "0 6"
 
 jobs:
   build-linux:
-    # Building and testing cannot work if the repository owner is not Xeeynamo
-    # due to the missing secrets to clone the CI dependencies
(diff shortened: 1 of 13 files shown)


sozud:
Could psy-z have true headless mode? The pc port used to have this

Xeeynamo:
> Could psy-z have true headless mode? The pc port used to have this

I am a bit surprised the offscreen doesn't work on Windows out of the box. I think for Windows build I may need to use the SDL3 GL backend, where Windows offers a software rasterizer that works headless.

I need to look carefully how the SDL3 project is doing this. `SDL_VIDEODRIVER=offscreen` works flawlessly on Linux for example. I'd not want to install Vulkan on Windows CI just for this, I agree. I'd also like to not have to introduce a null backend too, if we find a decent solution.

I'll keep looking.

Xeeynamo:
Turns out the CI failure on Windows caught a real crash at runtime when running on Windows!

The trick of running the game headless is here: https://github.com/Xeeynamo/sotn-decomp/pull/3537/changes/e6b7dc557427a8132f6ab63c178c4fe003ad5a96

There's a way to use D3D12 instead of Vulkan, but I need to compile and sign the shaders on PsyZ. Then we can remove the need of installing Vulkan on Windows.

PR #3538, opened 2026-08-11
docs: document EntityHeartDrop mechanism (#3005)

thekiss1864:
### Summary
Documents EntityHeartDrop in src/st/e_collect.h as requested in #3005.

- Explains persistent pickup checking via HEART_DROP_CASTLE_FLAG in g_CastleFlags.
- Documents item ID lookup in PrizeDrops[].
- Details delegation to EntityPrizeDrop or EntityEquipItemDrop.

### Verification
- [x] Verified binary compiles 1:1 matching with 
inja.


src/st/e_collect.h
```diff
@@ -1474,6 +1474,10 @@ void EntityRelicOrb(Entity* self) {
 // defined in d_prize_drops.c
 extern u16 PrizeDrops[];
 
+// EntityHeartDrop: Handles persistent room item drops (hearts, relics, items, equipment).
+// Checks g_CastleFlags via HEART_DROP_CASTLE_FLAG to prevent re-spawning already collected items.
+// Looks up item ID in PrizeDrops[] and delegates entity update behavior to EntityPrizeDrop
+// or EntityEquipItemDrop depending on item index range (< 128 vs >= 128).
 // params: Local index of this drop
 void EntityHeartDrop(Entity* self) {
```
JoshSchreuder, line 1483: Think we should probably rename this accordingly. 
`EntityPersistentItemDrop` or `EntityRoomDrop` perhaps @bismurphy @sozud @Xeeynamo ?

thekiss1864: Both EntityPersistentItemDrop and EntityRoomDrop sound clearer than EntityHeartDrop! Happy to update the function name in this PR once you guys settle on a preference.

sozud: EntityPersistentItemDrop or EntityPersistentDrop both seem reasonable to me. Note that renaming this is a relatively large diff since this gets used all over but it makes sense to fix the name if we know better now

thekiss1864: Thanks both! I'll go with EntityPersistentItemDrop. The rename touches a lot of files so I'll do it as a separate PR, and keep this one to just the docs and the formatting fix. Pushed the fix, linter should be green now.

Issue #3292, opened 2026-07-01
Document Entity's "hitboxState" variable


bismurphy:
This is a "study the decompiled code and maybe even improve it" task, rather than anything that involves decompiling new code.

We have a hitboxState value that is used in many entities. I noticed that when I search the codebase, it is usually set to 0, 1, 2, or 3. This makes me think it might be a bit field, where the 1 bit means something and the 2 bit means something else. It would be helpful if someone studied how this is used and accessed to allow us to make the code more readable by creating enums for those two bit values.

JoshSchreuder:
Had a quick glance, equipment may be a good starting point for documentation.
https://github.com/Xeeynamo/sotn-decomp/blob/022fe58de4d023412d996970df1fa9033f9b5ac5/src/weapon/shared.h#L98

https://github.com/Xeeynamo/sotn-decomp/blob/022fe58de4d023412d996970df1fa9033f9b5ac5/include/game.h#L1555

There are some values like 4 (mostly shields?), 6 (Dark Shield / Medusa Shield), 2 (swords and other offensive items), 0 (consumables like food),

Values 1, 3, 5 don't appear used for equipment at least
https://github.com/Xeeynamo/sotn-decomp/blob/022fe58de4d023412d996970df1fa9033f9b5ac5/src/config_us.h#L21-L190

Looks like the rando calls this "range"?
https://github.com/sotnrando/sotnrando/blob/master/src/stats.js#L108

PR #3540, opened 2026-08-11
docs(entity): document hitboxState bitfield and add HitboxStates enum (#3292)

thekiss1864:
committed by mistake


include/game.h
```diff
@@ -886,6 +886,14 @@ typedef struct {
 
 #define ANIM_FRAME_LOAD 0x8000
 
+typedef enum {
+    HITBOX_INACTIVE     = 0,
+    HITBOX_ACTIVE       = (1 << 0), // 0x01: Active hit detection
+    HITBOX_SOLID        = (1 << 1), // 0x02: Solid object / platform collision
+    HITBOX_WEAPON_HIT   = (1 << 2), // 0x04: Weapon hit state
+    HITBOX_INVULNERABLE = (1 << 7), // 0x80: Ignore hit detection / invulnerable
+} HitboxStates;
```
JoshSchreuder, line 895: How did you determine these?

sozud:
I think if we're going to name these we want to lint them and make sure they're fixed everywhere. These are LLM PRs right? If you were writing by hand I wouldn't ask but an LLM can easily knock out a linter here https://github.com/Xeeynamo/sotn-decomp/tree/master/tools/lints/sotn-lint

include/game.h
JoshSchreuder: Reason I ask is there is another commit linked to the issue (think LLM generated) which seems to contradict your 0x4
https://github.com/kazuyette/sotn-decomp/commit/9cbe9539068d6a4fb96e13b81311d7a899ee7524

It has it as
```
HITBOX_STATE_NO_DAMAGE = 0x04, // collision.h: DealDamage result forced to
                                   // 0 even though attack is set
```

So I am curious if you have checked which is correct here and know why that other comment is there.

Maybe you can analyse collision.h in addition to st0_hit_detection.h

JoshSchreuder: Actually it looks like what you analysed likely only applies to the prologue and should not be applied wholesale to the rest of the game. 

https://github.com/Xeeynamo/sotn-decomp/blob/dc80608397c8a3b13f66653861b10f9e935020a3/src/st/st0_hit_detection.h#L101-L103

I encourage you to investigate this further using PSX debugger etc. to determine the correct names.

JoshSchreuder (changes requested):

bismurphy:
Thank you for your contribution. Unfortunately, this PR does not appear to be up to our standards. It just creates an enum, without justification for the values seen in it, and without applying the enum values to anything in the code.

If you have done the analysis to back up these values, you can feel free to open a new PR showing that analysis. If this is LLM output, then it is unlikely to produce trustworthy analysis. Decomp projects like this are not well suited to LLM analysis because the types of reverse engineering tasks we are doing are not typical software and therefore it is a bit of a blind spot for these systems.

Closing for now, but this is not final - creating new PRs is free :). But next time, we will be aiming for a stronger level of justification for the changes you're making.

thekiss1864:
Appreciate you leaving the door open :D just didn't have the time back then, might look into it again sometime.

PR #3542, opened 2026-08-11
RNO0 Guardian enemy

bismurphy:
It turns out the Guardian shares the vast majority of his code with the Armor Lord. It's effectively a reskinned enemy.

I updated the file to be as documented as possible - named animations (using the new animation frame system), named steps, named entries in the Ext struct, etc. Seems nicely improved now.


Xeeynamo (approved):
Nice one! There are some conflicts, very likely due to the removal of `OVL_EXPORT`.

bismurphy:
Rebase went smoothly, should be all good now

PR #3543, opened 2026-08-11
Decompile `sat/DecompressLzss`

gamezter:
Mostly based on this : https://github.com/Piker-Alpha/RevoBoot/blob/613d72be66f09d9cec27913625a9fba63d197bfc/i386/boot2/lzss.c#L57


Xeeynamo (approved):
nice one!

PR #3546, opened 2026-08-11
weapon: finish w_035

Xeeynamo:
left a comment in the sotn asset tool on what was required to make the Red Rust working on PC.


tools/sotn-assets/assets/spriteset/handler.go
```diff
@@ -6,13 +6,18 @@ import (
...
 	"strings"
 
 	"github.com/xeeynamo/sotn-decomp/tools/sotn-assets/assets"
 	"github.com/xeeynamo/sotn-decomp/tools/sotn-assets/sotn"
 	"github.com/xeeynamo/sotn-decomp/tools/sotn-assets/util"
 )
 
+// HACK: only weapon spritesets are static. This is required for PC, as
+// two-handed weapons are both loaded at once, and symbols must not collide.
+var weaponAnimsetSymbol = regexp.MustCompile(`^w_\d{3}_\d+$`)
```
sozud, line 19: Is the idea to expand the pc overlay system later to include weapon0/1?

Xeeynamo: It's already done! But there's still the symbol collision issue if two overlays are loaded and have the same symbols in them.

`w0_035` and `w1_035` are both loaded, hence why this problem happens. It can be replicated with two more weapons of the same kind loaded to each hand.

JoshSchreuder (approved):

PR #3548, opened 2026-08-11
assets: support non-contiguous entity layout blocks

### LINKED ISSUE #3508: assets: NO0 layout / laydef issue
`- [0x778, layout, entity_layouts]` cannot be currently imported in NO0

This is due to some unreferenced entity layout information which does not have a corresponding entry in laydef, leading to an error in the assets tool:
```
❯ go run ./tools/sotn-assets info build/us/NO0.BIN
panic: gap between data detected: 0x801839C2 != 0x80183A64

goroutine 1 [running]:
github.com/xeeynamo/sotn-decomp/tools/sotn-assets/datarange.MergeDataRanges(...)
        /home/josh/dev/sotn-decomp/tools/sotn-assets/datarange/data_range.go:48
github.com/xeeynamo/sotn-decomp/tools/sotn-assets/assets/layout.readEntityLayout({0x9235e8, 0x18ae09ba3d70}, {0x8e94bb, 0x5}, 0x8018084c, 0x80180000, 0x35, 0x0)
        /home/josh/dev/sotn-decomp/tools/sotn-assets/assets/layout/layout.go:185 +0xde7
github.com/xeeynamo/sotn-decomp/tools/sotn-assets/assets/layout.readEntityLayout({0x9235e8, 0x18ae09ba3d70}, {0x8e94bb, 0x5}, 0x80180778, 0x80180000, 0x35, 0x1)
```

It is possible to update assets to handle such data, however it is quite an invasive change, and NO0 appears to be the only overlay that has this problem, so we should assess if we want to support it in sotn-assets or just manually handle this case.

Image

mathisto:
## Summary
This PR addresses the layout gaps identified by @JoshSchreuder in #3508.

`NO0` contains four gaps between referenced entity-layout blocks. Previously, the layout handler attempted to merge every block into one contiguous range, causing `sotn-assets info` and asset extraction to fail.

The changes:
- preserve individual entity-layout block ranges during decoding
- consolidate only adjacent ranges when producing `info` output
- leave unexplained gaps unclaimed while continuing to reject overlaps
- propagate layout decoding errors during extraction
- enable US `NO0` entity-layout extraction
- move `NO0` into the passing stage compatibility list

The decoded layout data is unchanged. A fresh extraction produces 29 unique blocks referenced by 53 layout-table entries.

## Validation
- focused layout and data-range tests pass
- the complete stage compatibility test passes
- forced clean `NO0` extraction succeeds 
- US, HD, and PSP-EU builds still byte match the pinned retail bins

Closes #3508

### DIFF
--- a/tools/sotn-assets/assets/layout/layout.go
+++ b/tools/sotn-assets/assets/layout/layout.go
@@ -175,14 +175,20 @@ func readEntityLayout(r io.ReadSeeker, ovlName string, off, baseAddr psx.Addr, c
 		if err := hydrateYOrderFields(l, yLayouts); err != nil {
 			return layouts{}, nil, fmt.Errorf("unable to populate YOrder field: %w", err)
 		}
-		xMerged := datarange.MergeDataRanges(xRanges)
-		yMerged := yRanges[1]
-		return l, []datarange.DataRange{
-			datarange.MergeDataRanges([]datarange.DataRange{datarange.New(off, endOfArray), yRanges[0]}),
-			datarange.MergeDataRanges([]datarange.DataRange{xMerged, yMerged}),
-		}, nil
+		// Keep block ranges separate here. Info consolidates adjacent runs while
+		// leaving gaps unclaimed; Extract does not consume the ranges.
+		laydefRange, err := datarange.Merge([]datarange.DataRange{
+			datarange.New(off, endOfArray), yRanges[0]})
+		if err != nil {
+			return layouts{}, nil, err
+		}
+		layoutRanges := append(
+			[]datarange.DataRange{laydefRange}, xRanges...)
+		return l, append(layoutRanges, yRanges[1:]...), nil
 	} else {
-		return l, []datarange.DataRange{datarange.New(off, endOfArray), datarange.MergeDataRanges(xRanges)}, nil
+		return l, append(
+			[]datarange.DataRange{datarange.New(off, endOfArray)},
+			xRanges...), nil
 	}
 }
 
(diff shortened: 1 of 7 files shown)


JoshSchreuder:
Looks good to me, quite a bit simpler than my hacked together prototype for the bug ticket which is nice 😄 
See what @Xeeynamo says about incorporating asset changes for single overlay but I think this seems fine

sozud (approved):
Seems fine to me but needs a review from Xeeynamo

Xeeynamo (approved):

tools/sotn-assets/assets/layout/layout.go
```diff
@@ -175,14 +175,20 @@ func readEntityLayout(r io.ReadSeeker, ovlName string, off, baseAddr psx.Addr, c
...
+			return layouts{}, nil, err
+		}
+		layoutRanges := append(
+			[]datarange.DataRange{laydefRange}, xRanges...)
+		return l, append(layoutRanges, yRanges[1:]...), nil
 	} else {
-		return l, []datarange.DataRange{datarange.New(off, endOfArray), datarange.MergeDataRanges(xRanges)}, nil
+		return l, append(
+			[]datarange.DataRange{datarange.New(off, endOfArray)},
+			xRanges...), nil
```
JoshSchreuder, line 191: @mathisto my bad as I didn't verify this fully end to end when reviewing, but this is missing the extraction of the unused pointer elements, which leads to a mismatch when trying to mark data imported.

You can verify this by updating `config/splat.us.stno0.yaml` with the following
```
- [0x778, .data, gen/e_laydef]
...
- [0x291C, .data, gen/e_layout]
- [0x461C, data]
```

Is this something you want to try and fix up?

mathisto: Yeah, dawg, I"d love to take a crack at this. Great catch sir. I'll prep a small follow up patch to address this. 🤜 🤛

PR #3550, opened 2026-08-12
RNO0 Floortrap

bismurphy:
Mostly just copied from the RNZ0 floor trap, but had to be pulled out into a .h file.

As I usually like to do when pulling in .h files, I went through and spruced things up a bit with readability - step enums, animation frame updates, etc.

One weird data quirk, there is a sequence of 8 bytes that appears to be some sensors or something, but they only exist in the RNZ0 version, not RNO0. They also don't seem to go to the next file in the overlay (ctulhu, which is already in other overlays), so not sure what's up with that. SOTN nonsense I guess.


src/st/e_floor_trap.h
```diff
@@ -0,0 +1,193 @@
...
+} FloorTrapSteps;
+
+// tools/play_animation.py indicates that this animation displays the wheel
+// spinning.
+static AnimateEntityFrame anim[] = {
+    {1, 3}, {1, 4}, {1, 5}, {1, 6}, {1, 7}, {1, 8}, POSE_LOOP(0)};
+
+// Needs more study - might not be part of this file. Probably stripped on PSP
+#ifdef STAGE_IS_RNZ0
+static u16 unused[] = {0, 40, 8, 0};
```
JoshSchreuder, line 20: Stone rose has these unused too
Given Armor Lord's usage I suspect they are unused sensors

sozud (approved):

PR #3554, opened 2026-08-12
RNO0 corpseweed/thornweed 

bismurphy:
Changes necessary for RNO0 itself were quite minimal, but as usual I went ahead and made some changes to make the shared .h file more closely aligned with how we usually do things


src/st/e_thornweed_corpseweed.h
JoshSchreuder: Would you mind checking out the prim usages here while you're in this file?
Otherwise I can go through and do it separately

bismurphy: on it

bismurphy: only used by corpseweed so it was easy

bismurphy: Actually hang on, these should be a single ext

JoshSchreuder (approved):

src/st/e_thornweed_corpseweed.h
JoshSchreuder: Thanks!

sozud (approved):

PR #3562, opened 2026-08-14
pc: add stage15

sozud:
This uses the previously added converters to allow exploring stage15 on pc. There's no entities except I added red door and the save room. The saturn map format is 3 layers of 8x8 tiles which results in lot of GT4 commands, so I had to expand the buffer. This also requires expanding the buffers in psyz, which I'm planning to send upstream. This adds some tests that walk Alucard around and capture pngs since the room transitions were finicky to get right. This also contains a change to tools/sotn_str/src/main.rs where since call it twice, it always regenerates the deps files and causes a full rebuild every time you run cmake --build. Now we get proper incremental builds. 

The stage doesn't connect to anything right now. Usage like `ENABLE_STAGE15=ON` and `./sotn --stage stage15`.

image


Xeeynamo:
image

I am giving it a try now. A few thoughts:
1. You could use `SPRT_8` for each tile, which should add less stress at rendering time.
2. It's using a custom renderer that bypasses `RenderTilemap`. I wonder if it would be simpler to partition the exported PNG map into 16x16, de-dupe them and generate tileset+tilemap from it on the fly, so we can rely on the idiomatic SOTN way of displaying stages. I am thinking about this also if we are able to bring these stages on PS1, not only on PC.
3. Graphics is rough. 320x240 from Saturn, scaled to 256x240 for the PS1 look, then scaled it back up to 320x240 by PsyZ. The Saturn stages seem to have a predictable Nearest Neighbour scaling pattern, where I believe we can predictably skip certain column of pixels when scaling those tiles from 20x16 to 16x16. Alternatively we could re-use most of the 16x16 tiles from existing PS1 stages, since the exclusive Saturn stages seem to recycle most graphics.
4. I'm greatly in favour of adding debugging capabilities that favour automation, such as the `--capture <path>` or `--transition-test` flags. The current implementation is fine for testing, but I don't believe they belong to the right place. PCSX Redux has a Web Debugger that exposes HTTP endpoints to do similar stuff. I'll take note and add similar features to PsyZ so we can keep SOTN PC clean.

Great works so far though! I had a feeling you were aiming for this, and I was curious how you were planning to pull this off. There are different ways to achieve this. Let's keep evaluating pro and cons for each approach.

sozud:
> I wonder if it would be simpler to partition the exported PNG map into 16x16, de-dupe them and generate tileset+tilemap from it on the fly, so we can rely on the idiomatic SOTN way of displaying stages. I am thinking about this also if we are able to bring these stages on PS1, not only on PC.

I actually tried this and as far as I can tell it's actually authored at 8x8 unlike psx. The best packing I was able to find reduced draw calls by like 20% at the cost of wasting ram on more baked tiles. I don't remember the exact amount but baking all 16x16 tile variants resulted in a huge number of extra tiles that wouldn't fit on psx. I could post the packing code if others want to see if there's a better solution

> Graphics is rough. 320x240 from Saturn, scaled to 256x240 for the PS1 look, then scaled it back up to 320x240 by PsyZ. The Saturn stages seem to have a predictable Nearest Neighbour scaling pattern, where I believe we can predictably skip certain column of pixels when scaling those tiles from 20x16 to 16x16. Alternatively we could re-use most of the 16x16 tiles from existing PS1 stages, since the exclusive Saturn stages seem to recycle most graphics.

Which tiles are you thinking of? Some of it is reused like the karasuman room is basically the skeleton leader room. This could be investigated in more detail. Probably could do a search of every stage and find the original tiles and objects they copied and resized.

I tried removing different columns and the pattern is not totally consistent, I think it's a mix of hand reized stuff, auto-resized. e.g. at least for what I tried you can't just do a 4/5 conversion and get the original psx tiles back. It's possible this could be improved further with more investigation. I think probably for the saturn specific levels the best solution is change the resolution to 320 width and stretch alucard/hud etc. to match. Then you don't get the ugly resizing. As far as I can tell the new graphics are actually drawn at 320 width which is why they look bad downsized. The CD hallway looks bad and ought to use the psx graphics but hadn't gotten around to it yet.

> I'm greatly in favour of adding debugging capabilities that favour automation, such as the --capture <path> or --transition-test flags. The current implementation is fine for testing, but I don't believe they belong to the right place. PCSX Redux has a Web Debugger that exposes HTTP endpoints to do similar stuff. I'll take note and add similar features to PsyZ so we can keep SOTN PC clean.

Is the request here to remove the new tests for the time being until this is figured out?

sozud:
I set up a search of each 8x8 4bbp tile in the game and about 25% of the nonempty tiles in stage15 come directly from elsewhere. They are highlighted in white in the picture. This checks all 4 hv flips, and is done on the raw color #s before a palette is applied. There's tiles that are borrowed like the columns in the boss room but they are apparently hand modified.
level-any-identified
From what I see the new art is drawn at 320 width and not upscaled. I think the options are something like:
Downscale all saturn art to 256: Saturn 320-native tiles look ugly
Upscale psx art to 320: Waste of vram, psx art looks ugly
Keep tile metadata and selectively use original psx tile or saturn tile depending on the whether it's a saturn unique tile: probably the best solution but more complex 
Entities are drawn at 320 width so eventually we would want to change the viewport to 320 and stretch the hud/player etc. to preserve the entity graphics in my opinion

Xeeynamo:
> Entities are drawn at 320 width so eventually we would want to change the viewport to 320 and stretch the hud/player etc. to preserve the entity graphics in my opinion

It's an interesting challenge. From a quick look I see the Maria sprite of Saturn is also correctly drawn for the 320x240 resolution. We should get this right at this stage.

The key, critical question is: do we want Saturn stages to be backported to PlayStation 1? Or do we want it to be a PC-exclusive feature? That alone will likely decide what kind of approach we'd like to take.

If we decide to stick with 256x224 for stages, 320-width entities should be easy to handle. We have a `entity->scaleX` field we can tweak to adjust their aspect ratio. On 1x internal resolution they will look downscaled. But at 2x or more they will look much cleaner. Same thing once we are able to port Maria Saturn.

Stages are truly the biggest hurdle. Those original PS1 tiles could've be carefully cherry-picked in a custom tilemap that is created on a custom tilemap we can add in the repo as a creative effort, and a tool that creates the tileset from the existing extracted PS1 assets. The map will still be rendered via `RenderTilemap()` as usual. But the 320-width exclusive tiles from Saturn could they be rendered as an overlay tilemap to add on top with a custom renderer baked in the overlay? I am yet unsure how the final result would look like with this hybrid pixel art. TL;DR I am inclined with your proposed hybrid approach.

Alternatively, we'd recreate the pixel art to fit the original PS1 aspect ratio. But that's a whole effort outside the skillset of any team member.

> Is the request here to remove the new tests for the time being until this is figured out?

@sozud Just an update to let you know I am actively working on offering a better interface directly on PsyZ. So we don't have to put `--capture` and similar to in SOTN. Hopefully it will help maintaining this repo codebase lean and clean.

image

sozud:
> The key, critical question is: do we want Saturn stages to be backported to PlayStation 1?

I was planning to look into it. I think this ought to be doable but I'm not 100% about how expensive drawing all the 8x8 tiles is. The hybrid renderer that takes tiles from both versions might be too much though. If we change the draw width to 320 that gives up a fair amount of VRAM. I don't see an issue with supporting multiple drawing styles (256 vs. 320) on psx vs. pc if necessary.

> Those original PS1 tiles could've be carefully cherry-picked in a custom tilemap that is created on a custom tilemap we can add in the repo as a creative effort, and a tool that creates the tileset from the existing extracted PS1 assets. 

I think if we have to specify some tiles manually that's not an issue.

> But the 320-width exclusive tiles from Saturn could they be rendered as an overlay tilemap to add on top with a custom renderer baked in the overlay? I am yet unsure how the final result would look like with this hybrid pixel art. 

I can look and see how complicated mixed-width rendering and what this would look like in a few days.

> So we don't have to put --capture and similar to in SOTN.

Ok I can adjust the PR to use that when it is ready.

sozud:
> So we don't have to put --capture and similar to in SOTN.

I looked at the debug server and I think using it is going to make the sorts of tests I'm writing harder to write and debug. I think the debug server makes sense if we want to do e2e testing that prevents inspection of the internal state of the game since it's based around controller inputs and screenshots. 

I think being able to test things directly like "teleport to this spot, check what room we are in, and check if the door spawned and ready" is more exact than "play this movie and check if this screenshot matches". I can test the full set of transitions rather than just check-pointing a few screenshots. Generating and keeping gold screenshots is a task as well.

Xeeynamo:
> > So we don't have to put --capture and similar to in SOTN.
> 
> I looked at the debug server and I think using it is going to make the sorts of tests I'm writing harder to write and debug. I think the debug server makes sense if we want to do e2e testing that prevents inspection of the internal state of the game since it's based around controller inputs and screenshots.
> 
> I think being able to test things directly like "teleport to this spot, check what room we are in, and check if the door spawned and ready" is more exact than "play this movie and check if this screenshot matches". I can test the full set of transitions rather than just check-pointing a few screenshots. Generating and keeping gold screenshots is a task as well.

Okay got it. I'll re-review this PR again later in the day. I do not want to risk slowing down the momentum.

src/dra/4CE2C.c
Xeeynamo:
```suggestion
#ifdef VERSION_PC
SPRT_16 _sp16;
TilemapRenderer _r;
SPRT_16* sp16 = &_sp16;
TilemapRenderer* r = &_r;

if (g_PcStageLayerRenderer != NULL && g_PcStageLayerRenderer()) {
    return;
}
```

can we merge these two?

src/dra/5087C.c
```diff
@@ -1253,7 +1253,13 @@ void RunMainEngine(void) {
 #if defined(VERSION_PSP)
         D_psp_091CE578 = 1;
 #endif
+#if defined(VERSION_PC) && defined(ENABLE_STAGE15)
+        if (g_StageId != STAGE_SATURN_15) {
+            g_StageId = func_800F16D0();
+        }
+#else
```
Xeeynamo: This breaks the teleportation to the librarian. Why is it needed for STAGE15 specifically?

src/pc/stages/stage_saturn15.c
```diff
@@ -0,0 +1,774 @@
...
+        u0 += 7;
+        u1 = u0 - (blockWidth * 8 - 1);
+    }
+    if (pnd & 0x0800) {
+        v0 += 7;
+        v1 = v0 - (blockHeight * 8 - 1);
+    }
+
+    poly = &g_CurrentBuffer->polyGT4[g_GpuUsage.gt4++];
+    setPolyGT4(poly);
```
Xeeynamo:
```suggestion
    setPolyFT4(poly);
```
We should switch to FT4, as the RGB for each corner is always 0x80

src/pc/stages/stage_saturn15.c
```diff
@@ -0,0 +1,774 @@
...
+    if (g_GpuUsage.g4 >= MAX_POLY_G4_COUNT) {
+        return;
+    }
+    poly = &g_CurrentBuffer->polyG4[g_GpuUsage.g4++];
+    setPolyG4(poly);
+    setSemiTrans(poly, false);
+    setRGB0(poly, 0x31, 0x00, 0x29);
+    setRGB1(poly, 0x31, 0x00, 0x29);
+    setRGB2(poly, 0x31, 0x00, 0x29);
+    setRGB3(poly, 0x31, 0x00, 0x29);
```
Xeeynamo: What do these values stand for?

src/pc/stages/stage_saturn15.c
```diff
@@ -0,0 +1,774 @@
...
+}
+
+static void Stage15DrawBackScreen(void) {
+    POLY_G4* poly;
+
+    if (g_GpuUsage.g4 >= MAX_POLY_G4_COUNT) {
+        return;
+    }
+    poly = &g_CurrentBuffer->polyG4[g_GpuUsage.g4++];
+    setPolyG4(poly);
```
Xeeynamo: Since this rectangle is flat, you can use `TILE` instead of `G4`

tools/sotn_str/src/main.rs
```diff
@@ -622,6 +622,31 @@ fn preprocess_args(args: &[String]) -> Vec<String> {
     out
 }
 
+fn strip_dependency_args(args: &[String]) -> Vec<String> {
```
Xeeynamo: Why there is need to do this?

src/dra/dra.h
```diff
@@ -582,7 +582,12 @@ extern u16 g_JosephsCloakColors[4];
 extern GfxBank** g_GfxSharedBank[];
 extern s16** D_800A3B70[18];
 extern u_long* D_800A3BB8[];
-extern Lba g_StagesLba[80];
+#if defined(VERSION_PC) && defined(ENABLE_STAGE15)
+#define NUM_STAGES 81
+#else
+#define NUM_STAGES 80
+#endif
+extern Lba g_StagesLba[NUM_STAGES];
```
Xeeynamo:
```suggestion
extern Lba g_StagesLba[];
```
We never check for the `g_StagesLba` boundaries, we can just remove the fixed array size altogether.

Xeeynamo (commented):
A few changes here and there.

I am still not sure how I feel about `--capture` and `--transition-test`. They seem to be required specifically to help iterate on this PR's changes. Testing those flags on any other stage will crash the game.

Ideally what I would like to see is the least amount of code exceptions possible to make this work. I realize adding Saturn stages is already quite challenging on its own. But I'd like this piece of work to land as clean as possible.

A few things I had in mind we should enable to better support this work, and any future mods:
* We should reserve `0x4C` and `0x6C` for Stage15 and Reverse Stage 15, then `0x4D` and `0x6D` for the other Saturn-exclusive stage. Then slot `0x51` to `0x5F` and `0x71` to `0x7F` for user-specific mods.
* We should not edit the `Stages` enum. A new stage overlay should be able to slot itself in without the game engine being aware.
* We should let the stage overlay init sequence load the VB and VH files, so we don't need to edit `lba_stage.c`. I am still not sure about selecting `MU_STAGE15` cleanly. 
* We should probably get rid of `allowed_stages` and let the game error if the specified stage overlay is not found.
* Pathing via `sim_pc` is painful. We should find a better way to do this. You suggested copying files from `disks/us` to the build folder. It's probably a good idea to have a single source of data. Especially with the various PsyZ backend ports.
* The `ENABLE_STAGE15` part in the `CMakeLists.txt` could point to an ad-hoc cmake file, for example `cmake/stage15.cmake`, so it remains easily pluggable and unpluggable.

Almost all of my suggestions are mostly to allow new stages to be plugged in like Lego blocks, instead of doing code surgery. I realize it's quite a way to reach that state. Ultimately I am excited to see this PR take shape as an example of how to create custom stages.

sozud:
> I am still not sure how I feel about --capture and --transition-test. 

Ok I removed them

> We should reserve 0x4C and 0x6C for Stage15 and Reverse Stage 15, then 0x4D and 0x6D for the other Saturn-exclusive stage. Then slot 0x51 to 0x5F and 0x71 to 0x7F for user-specific mods.

Stage15 uses that ID now

> We should not edit the Stages enum. A new stage overlay should be able to slot itself in without the game engine being aware.

Edits removed

> We should let the stage overlay init sequence load the VB and VH files, so we don't need to edit lba_stage.c. I am still not sure about selecting MU_STAGE15 cleanly.

Edits to lba_stage removed

> We should probably get rid of allowed_stages and let the game error if the specified stage overlay is not found.

ok got rid of it

> Pathing via sim_pc is painful. We should find a better way to do this. You suggested copying files from disks/us to the build folder. It's probably a good idea to have a single source of data. Especially with the various PsyZ backend ports.

ok moved the data

> The ENABLE_STAGE15 part in the CMakeLists.txt could point to an ad-hoc cmake file, for example cmake/stage15.cmake, so it remains easily pluggable and unpluggable.

ok moved to it's own file

src/dra/dra.h
sozud: Done

tools/sotn_str/src/main.rs
sozud: this get run 2x during the build so we write out the deps files, then write them out again, causing a full rebuild every time. This slows down iterating on PC a lot

src/pc/stages/stage_saturn15.c
sozud: Done

src/pc/stages/stage_saturn15.c
sozud: It's the VDP2 BACK layer color, I added a define

src/pc/stages/stage_saturn15.c
sozud: Done

src/dra/5087C.c
sozud: I forget why this was here, removed

src/dra/4CE2C.c
sozud: Ok diff reduced

tools/sotn_str/src/main.rs
Xeeynamo: Makes sense. I propose to detach this change and put it as a separate PR. It would also help other devs to iterate faster on Windows

sozud: reverted it

PR #3563, opened 2026-08-14
pc: preemptive fix various Primitive pointer struct alignments

JoshSchreuder:
As discussed on Discord, I went through and looked at existing `ext.prim` usages and fixed up cases where there was an existing struct which was not being used correctly and likely to cause alignment on larger pointer architectures


Xeeynamo (approved):
I was anticipating new `ET` were to be created. But the change you made came very clean!

PR #3564, opened 2026-08-14
BO1: decompile codeshare most functions

Xeeynamo:
`src/boss/bo1/unk_21C80.c` is the one that is still a bit of work in progress. I'll park the changes here and resume this on the week-end.


config/symbols.pspeu.txt
```diff
@@ -245,7 +245,7 @@ D_8003C708 = 0x091E1308; // size:0x8
 g_BgLayers = 0x091E1328; // size:0x300
 g_Tilemap = 0x091E1628; // size:0x50
 g_CurrentEntity = 0x091E1678;
-g_Entities = 0x091E1680;
+g_Entities = 0x091E1680; // size:0xBC00
```
JoshSchreuder, line 248: Does this work for you? Last time I tried this Splat essentially ignores it because we stub in specific elements from the g_Entities array like `PLAYER`

Build outputs an error
```
Range check triggered: .bss symbol (name: g_Entities, address: 0x091E1680):
    User declared size (0xBC00) does not match the .space that will be emitted (0x2).
    Try checking the size again or look for symbols which overlaps this region
```

And it doesn't correctly output the `g_Entities + 0xF00` in the ASM either like it would if this was working correctly.

Xeeynamo: Not on US, no. But it works on PSPEU now that I pushed fe29544:

```s
    /* 10AD0 09248150 1E09043C */  lui        $a0, %hi(g_Entities + 0x2F00)
    /* 10AD4 09248154 80458424 */  addiu      $a0, $a0, %lo(g_Entities + 0x2F00)
    /* 10AD8 09248158 1F09053C */  lui        $a1, %hi(D_80097C98)
    /* 10ADC 0924815C 80D2A524 */  addiu      $a1, $a1, %lo(D_80097C98)
    /* 10AE0 09248160 2812490E */  jal        AllocEntity
```

JoshSchreuder: Yeah unfortunately then we lose PLAYER getting picked up automatically in the scratches (they'll just come in as g_Entities or g_Entities[0])

Pity splat can't handle overlapping length with symbol aliases, oh well!

sozud (approved):

config/symbols.pspeu.txt
Xeeynamo: Let's merge and see if it's going to cause problems

PR #3567, opened 2026-08-15
RNO0 Subweapon Container

bismurphy:
Pulled from NZ0, with a bit of renaming and other cleanup.


src/st/e_subweapon_container.h
```diff
@@ -0,0 +1,333 @@
...
+        g_api.FreePrimitives(self->primIndex);
+        self->flags &= ~FLAG_HAS_PRIMS;
+        g_api.PlaySfx(SFX_SUBWEAPON_CONTAINER_BREAK);
+        glassPieceTBL = glassPiecesParams;
+        for (i = 0; i < LEN(glassPiecesParams); i++, glassPieceTBL++) {
+            newEntity = AllocEntity(&g_Entities[224], &g_Entities[256]);
+            if (newEntity != NULL) {
+                CreateEntityFromEntity(E_FALLING_GLASS, self, newEntity);
+                newEntity->posX.i.hi += glassPieceTBL->velX;
+                newEntity->posY.i.hi += glassPieceTBL->velY;
```
JoshSchreuder, line 111: I know this was in the original code but while we're here... is `velX`/`velY` the right naming here? These look like positions rather than velocity

src/st/e_subweapon_container.h
```diff
@@ -0,0 +1,333 @@
...
+        break;
+
+    case SUBWPNCONT_DEBUG:
+        /**
+         * Debug: Press SQUARE / CIRCLE on the second controller
+         * to advance/rewind current animation frame
+         * NOTE: THIS IS DIFFERENT FROM NORMAL CASES OF THIS
+         * We are incrementing newEntity's animationFrame, not self's!
+         */
+        FntPrint("charal %x\n", self->animCurFrame);
```
JoshSchreuder, line 145: Think this can probably pull in `pad2_anim_debug.h`?

bismurphy: Nope, look at the comment right above :)

JoshSchreuder: Whoops, never mind, just saw the comment above NOTE 😄 
Disregard!

src/st/e_subweapon_container.h
bismurphy: It's essentially an initial position offset which represents one "step" of velocity.

Because there is also:

`newEntity->ext.subwpnContGlass.velX = glassPieceTBL->velX;`
and then in the glass
`        self->velocityX = self->ext.subwpnContGlass.velX * 0x1000;`

so velX is definitely accurate, and therefore velY should probably match the same.

JoshSchreuder (approved):

PR #3568, opened 2026-08-15
Rename EntityHeartDrop to EntityPersistentItemDrop

thekiss1864:
Rename EntityHeartDrop to EntityPersistentItemDrop, as discussed in #3538.

PR #3570, opened 2026-08-15
Rename heartDrop field to persistentDrop

thekiss1864:
Rename the heartDrop field to persistentDrop, matching the
EntityPersistentItemDrop rename in #3568.


thekiss1864:
The struct rename is already merged in #3569.

PR #3571, opened 2026-08-15
pc: enable remaining NO stage overlays

sozud:
This is prep work for the psx->saturn stage stitching.


sozud:
I think this is pretty straightforward so going to merge to enable next steps

PR #3576, opened 2026-08-15
tools: add Saturn stage stitching

sozud:
This adds the assets tooling for grafting the saturn stage entrances onto the existing psx stages. The steps are like: get saturn stage collision/layout/entities, overwrite or add the relevant part, emit that as patched json/tilemap for the regular assets tool. The integration PR will come later.


Xeeynamo (approved):
Should be fine. I did not try the new tool logic as it's not fully wired. It would be nice to see it working end to end in a later PR.

PR #3577, opened 2026-08-16
Decompile remaining `mad` functions

JoshSchreuder:
Some wacky pointer arithmetic to achieve the room randomisation but finally got the last 3 funcs matched
https://decomp.me/scratch/tZ94s
https://decomp.me/scratch/p6Ukv
https://decomp.me/scratch/dMinl

Also added a hook to allow the PC version to support the debug room which extends #3411!
image

image

### DIFF
--- a/src/pc/pl_maria.c
+++ b/src/pc/pl_maria.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include <game.h>
-#include "../../maria/maria.h"
+#include "../maria/maria.h"
 
 // To use Maria as a playable character, add -DUSE_PLAYER_MARIA=1 in cmake
 
--- a/src/st/mad/D8C8.c
+++ b/src/st/mad/D8C8.c
@@ -2,11 +2,260 @@
 #include "mad.h"
 #include "sfx.h"
 
-void func_8018D8C8(u16* tilemap);
-INCLUDE_ASM("asm/us/st/mad/nonmatchings/D8C8", func_8018D8C8);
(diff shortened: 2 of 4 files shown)


src/st/mad/D8C8.c
```diff
@@ -2,11 +2,260 @@
...
+    for (i = 0; i < LEN(heights); i++) {
+        heights[i] = 0;
+    }
+
+    // make some kind of mountain pattern?
+    for (i = 0; i < 20; i++) {
+    loop:
+
+        // pick an offset in range [2, 12]
+        idx = 2 + (rand() % 11);
```
JoshSchreuder, line 36: @Xeeynamo is rand() not implemented properly on PC side or am I doing something wrong? 
I seem to get the same deterministic room layout on every execution

JoshSchreuder: Actually I think this is expected, if we boot right into MAD then rand() never gets called beforehand, and it's always going to start at the same seed value?

I have pushed a PC hook function which randomises the seed at startup in MAD to allow for proper randomisations when you load in the debug room.

src/pc/pl_maria.c
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 #include <game.h>
-#include "../../maria/maria.h"
+#include "../maria/maria.h"
```
JoshSchreuder, line 3: Unrelated, fixes player maria flag on PC build

Xeeynamo (approved):
Wow you did it. Congrats!

PR #3580, opened 2026-08-16
Dedupe `rno2` Ctulhu

JoshSchreuder:


src/st/e_floor_trap.h
```diff
@@ -15,11 +15,6 @@ typedef enum {
 static AnimateEntityFrame anim[] = {
     {1, 3}, {1, 4}, {1, 5}, {1, 6}, {1, 7}, {1, 8}, POSE_LOOP(0)};
 
-// Needs more study - might not be part of this file. Probably stripped on PSP
-#ifdef STAGE_IS_RNZ0
-static u16 unused[] = {0, 40, 8, 0};
-#endif
```
JoshSchreuder, line 21: @bismurphy turns out this was from e_ctulhu, this + stone rose were prior splits to e_ctulhu, and had the same unused sensor that was present on RNO2 (which had no prior split).

src/st/e_ctulhu.h
```diff
@@ -620,10 +621,10 @@ void EntityCtulhuFireball(Entity* self) {
         self->primIndex = primIndex;
         prim = &g_PrimBuf[primIndex];
         self->ext.ctulhu.prim = prim;
-#ifdef STAGE_IS_RNZ0
-        prim->tpage = 0x12;
-#else
+#ifdef STAGE_IS_NO0
```
sozud: we can use a `#define TEXTURE_PAGE`?

sozud (approved):

src/st/e_ctulhu.h
JoshSchreuder: Done

PR #3578, opened 2026-08-16
pc: graft saturn doors

sozud:
This turns on the level additions to visit the new stages. No new stages are present so the CD rooms lead to nowhere. This adds some functionality to teleport to the areas, see the usage below. To avoid contaminating the regular build the modified headers and assets are put in the cmake build folder, so some of the header paths have been adjusted.
Usage like:
```
  cmake -S . -B build -G Ninja \
    -DENABLE_SATURN_STITCHES=ON

  ./sotn --spawn-point np3-stage15-door
  ./sotn --spawn-point reverse-trapdoor-underground

  ./sotn --spawn-point marble-soul
  ./sotn --spawn-point reverse-marble-soul

  ./sotn --spawn-point caverns-soul
  ./sotn --spawn-point reverse-caverns-soul
```

PR #3581, opened 2026-08-16
pc: multiple NO4 fixes

Xeeynamo:
I was reviewing #3578 to test the Saturn entrance from NO4. And I noticed there were crashes happening at multiple spots. With the current PR, I played almost all NO4 with no major issues.

There are two remaining bugs that this PR doesn't fix:
* the `e_frozen_shade` shards are white, it seems transparency is not applied well (PsyZ problem maybe)
* The platform on the water has wrong hitboxes. It might be a signed issue.


src/st/no4/e_toad.c
```diff
@@ -209,6 +209,14 @@ void EntityToad(Entity* self) {
         break;
     case 6:
         index = self->animCurFrame - 9;
+#ifdef FIX_UB
+        // The tongue entity is spawned with step=6 and animCurFrame=0, but
+        // AnimateEntity() only starts on the 9-15 frame range once the parent
+        // reaches step 4/stepm_s 2. Avoid reading hitbox with a negative index.
+        if (index < 0) {
```
JoshSchreuder: If you want a one liner CLAMP_MIN macro should work for these cases

JoshSchreuder (approved):

src/st/no4/e_toad.c
Xeeynamo: I just saw the `e_owl_knight` fix using CLAMP_MIN. Thanks for the suggestion. Done :+1:

PR #3591, opened 2026-08-18
sat: decompile 23 function(s)

sozud:


src/saturn/alucard.c
```diff
@@ -176,7 +176,34 @@ INCLUDE_ASM("asm/saturn/alucard/f_nonmat", f60AF4B8, func_060AF4B8);
...
 INCLUDE_ASM("asm/saturn/alucard/f_nonmat", f60AF7F0, func_060AF7F0);
-INCLUDE_ASM("asm/saturn/alucard/f_nonmat", f60AF8E0, func_060AF8E0);
+typedef enum {
+    TELEPORT_CHECK_NONE = 0,
+    TELEPORT_CHECK_TO_RTOP = 2,
+    TELEPORT_CHECK_TO_TOP = 4
+} TeleportCheck;
+
+extern s32 g_PlayerX;
+extern s32 g_PlayerY;
```
Xeeynamo, line 186: Let's move these to `sattypes.h`

EDIT: I found them here too https://github.com/Xeeynamo/sotn-decomp/pull/3593/changes#diff-f9bc5b2a0f57b4999519fef8f1c73681660e12bfdb52095550e758ebdced9ac9R70

These externs are from ZERO, right? I am thinking we should move the common functions and global vars there to better differentiate them with what comes exclusively from the overlay.

src/saturn/maria.c
```diff
@@ -3,7 +3,34 @@
...
 
-INCLUDE_ASM("asm/saturn/maria/f_nonmat", f60A5060, func_060A5060);
+typedef enum {
+    TELEPORT_CHECK_NONE = 0,
+    TELEPORT_CHECK_TO_RTOP = 2,
+    TELEPORT_CHECK_TO_TOP = 4
+} TeleportCheck;
+
+extern s32 g_PlayerX;
+extern s32 g_PlayerY;
```
Xeeynamo, line 13: Same here

Xeeynamo (commented):
Looks good. Left a couple of minor comments

src/saturn/alucard.c
sozud: Ok I'll merge this for now so master is up to date then I will do a cleanup of this problem in general

PR #3593, opened 2026-08-18
Decompile `sat/RunMainEngine`

gamezter:
Depends on https://github.com/Xeeynamo/sotn-decomp/pull/3592

Thanks again to @sozud for the final mismatches


sozud (approved):
Merged the other one, seems good after a rebase

PR #3595, opened 2026-08-18
matching CheckHighJumpInput

christianpaez:
https://decomp.me/scratch/o5gQi


src/boss/bo6/richter.c
```diff
@@ -114,7 +114,17 @@ static void CheckBladeDashInput() {
     }
 }
 
-INCLUDE_ASM("boss/bo6/nonmatchings/richter", CheckHighJumpInput);
+extern u16 RIC_step;
+extern s32 RIC_velocityY;
```
JoshSchreuder: These should be using `RIC.*` (eg. `RIC.step`), not the alias symbols.
See rest of functions in this file like `func_us_801B4BD0` for example

JoshSchreuder (changes requested):

src/boss/bo6/richter.c
christianpaez: done, matching went from 100% to 99,44% but i think thats expected due to using the struct

JoshSchreuder: As long as it matches on local build that is the important thing

JoshSchreuder (approved):

PR #3596, opened 2026-08-19
sat: move most declarations to appropriate headers

sozud:
This moves most extern and forward declarations to the headers of the overlays that own them. This was done with a tool I'll put up a PR for later.


sozud:
This is going to cause a lot of conflicts so going to go ahead and merge

PR #3598, opened 2026-08-19
Decompile `sat/func_0606BEE4`

gamezter:


src/saturn/game_3a.c
```diff
@@ -5,18 +5,123 @@
...
+    case 4:
+        if (func_06012D88() != 0) {
+            return 1;
+        }
+        break;
+    }
+    return 0;
+}
+
+void PlaySfx(s32 sfxId);
```
gamezter, line 123: Had to move it here because `func_0606C088` mismatches if `void PlaySfx(s32 sfxId);` is above

sozud:
Conflicts

PR #3600, opened 2026-08-20
sat: decompile some funcs

sozud:


src/saturn/alucard.c
```diff
@@ -30,15 +56,68 @@ void func_060A5594(s32 param_1) {
...
+        g_Entities->velocityX -= arg0;
+        if (g_Entities->velocityX < 0) {
+            g_Entities->velocityX = 0;
+        }
+    }
+}
 // DecelerateY (PLAYER-specialized Saturn version)
-INCLUDE_ASM("asm/saturn/alucard/f_nonmat", f60A55E4, func_060A55E4);
+void func_060A55E4(s32 arg0) {
+    if (g_Entities->velocityY < 0) {
```
JoshSchreuder: Doesn't have to be in this PR, but is it worth mapping `g_Entities[0]` -> `PLAYER` for parity with PSX at this point or nah?

src/saturn/alucard.c
```diff
@@ -5,7 +5,33 @@
...
 // CheckMoveDirection
-INCLUDE_ASM("asm/saturn/alucard/f_nonmat", f60A5060, func_060A5060);
+s32 func_060A5060(void) {
+    s32 result = 0;
+
+    if (g_Player.unk44 & 2) {
+        return 0;
+    }
+
+    if (g_Entities->facingLeft == 0) {
```
Xeeynamo:
```suggestion
    if (!g_Entities->facingLeft) {
```
Does it match using `!` instead of `==0`? While they're interchangeable on PS1 with -O2, I noticed on PSP the codegen differs between the two when the field type is 8 or 16 bitt. `facingLeft` specifically rarely use `== 0`. It would be nice to reconstruct the code closer to what it might have been.

src/saturn/alucard.c
```diff
@@ -5,7 +5,33 @@
...
+        if (g_Player.padPressed & 0x4000) {
```
Xeeynamo:
```suggestion
        if (g_Player.padPressed & PAD_LEFT) {
```
and a few other `g_Player.padPressed`

src/saturn/alucard.c
```diff
@@ -105,9 +286,36 @@ INCLUDE_ASM("asm/saturn/alucard/f_nonmat", f60A8DF4, func_060A8DF4);
 INCLUDE_ASM("asm/saturn/alucard/f_nonmat", f60A9658, func_060A9658);
 INCLUDE_ASM("asm/saturn/alucard/f_nonmat", f60A96A0, func_060A96A0);
 INCLUDE_ASM("asm/saturn/alucard/f_nonmat", f60A972C, func_060A972C);
-INCLUDE_ASM("asm/saturn/alucard/f_nonmat", f60A9D90, func_060A9D90);
+void func_060A9D90(u16 arg0) {
+    s32 xOffset;
+
+    xOffset = 4;
+    if (g_Entities->facingLeft != 0) {
```
Xeeynamo: Same comment applies here

src/saturn/alucard.c
```diff
@@ -105,9 +286,36 @@ INCLUDE_ASM("asm/saturn/alucard/f_nonmat", f60A8DF4, func_060A8DF4);
...
+        xOffset = -4;
+    }
+    g_Entities->posY.val -= 0x160000;
```
Xeeynamo: FIX?

src/saturn/alucard.c
sozud: Ok fixed these

src/saturn/alucard.c
sozud: changed to !

src/saturn/alucard.c
sozud: Added the macros

src/saturn/alucard.c
sozud: Added

sozud:
Comments resolved going to merge to put up another batch

PR #3606, opened 2026-08-20
assets: emit tiledef with predictable known names

### LINKED ISSUE #3602: Asset tool needs to be able to pull in tiledefs with swapped orders
In RNO0 we have these two different layers files:

Image

They appear to be referencing the same data, it's just that they have the tiledefs in two different orders (what is 1 on US is 2 on PSP, and vice versa). We need some solution for this. Probably the solution is to have the ability to name the tiledefs, and then we can #include them in the opposite order in the C files. So we would have:

- `bargle` : custom name for current rno0_tiledef_0
- `nawdle` : custom name for current rno0_tiledef_1
- `zouss` : custom name for current rno0_tiledef_2

And then we would have, in `tilemaps.c`:
```
#include "gen/bargle.h"
#ifdef VERSION_US
#include "gen/nawdle.h"
#include "gen/zouss.h"
#else
#include "gen/zouss.h"
#include "gen/nawdle.h"
#endif
```
In the image above, most of US uses tiledef_1 and most of PSP uses tiledef_2. In this adjusted setup, both versions would mostly use `nawdle`, and the only difference is that in US, `nawdle` comes first while in PSP `nawdle` comes last.

There may be issues with this approach; you may need to explore other options.

Xeeynamo:
Fixes #3602

All overlays have between one to three tiledefs. One is always the one representing the stage graphics, then two optional ones for the load and save room. As their names are predictable, I decided to use `tiledef_{STAGE_NAME}`, `tiledef_save` and `tiledef_load`, instead of the generic `tiledef_X` where X was a number between 0 to 2.

Their identification is done by hashing the data blob of the tiledef. I chosen SHA1 as all today's hardware have hardware acceleration for it.

All "unused" tiledefs were either from the load room or from the save room.

An idea for the future might be to just export, share and compile the tiledefs save and load once. It's one of the few parts of the codebase I think we could only compile once and link it everywhere.


src/st/rno0/tilemaps.c
```diff
@@ -50,6 +50,12 @@
 #include "gen/rno0_tilemap_46.h"
 #include "gen/rno0_tilemap_47.h"
 #include "gen/rno0_tilemap_48.h"
-#include "gen/rno0_tiledef_0.h"
-#include "gen/rno0_tiledef_1.h"
-#include "gen/rno0_tiledef_2.h"
+
+#include "gen/rno0_tiledef_rload.h"
+#ifdef VERSION_PSP
+#include "gen/rno0_tiledef_rsave.h"
```
sozud, line 56: This makes me think these were separate C files

sozud (approved):

src/st/rno0/tilemaps.c
Xeeynamo: I am thinking the same thing. I'd propose to save this change once I go ahead with compiling these separately, and share save/rsave/load/rload properly

PR #3614, opened 2026-08-22
pc: add CAT overlay

JoshSchreuder:
image

Played through, mostly issue free except for a few out of bounds access and something else I'll leave drop a note on for @Xeeynamo to double check

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -336,6 +336,66 @@ set(SOURCE_FILES_STAGE_ARE
     src/st/are/bss.c
 )
 
+set(SOURCE_FILES_STAGE_CAT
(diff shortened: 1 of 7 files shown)


JoshSchreuder:
@Xeeynamo There seems to be some background texture issues in the lava room. 

https://github.com/user-attachments/assets/8e0fdb74-e939-4702-b841-2703b39ff4d6

image

The black triangles and weird pop-in thing on the right hand side of the background

I suspect likely a bug on PSYZ or something not implemented, but not certain

sozud (approved):

Xeeynamo (approved):
Should be fine. I did not check on my computer yet. I'll investigate on the background bug though

PR #3620, opened 2026-08-24
Decompile `sat/func_06075838`

gamezter:


src/saturn/game_2a.c
```diff
@@ -381,8 +379,169 @@ typedef struct {
...
+        x2p = &prim->x2;
+        x3p = &prim->x3;
+        prim->x0 = prim->x1 = (i * 2) + 0x110;
+        *x2p = *x3p = prim->x0 + 1;
+
+        y0p = &prim->y0;
+        y3p = &prim->y3;
+        prim->y1 = prim->y2 = 0x16;
+        *y0p = *y3p = 0x75;
+
```
gamezter, line 533: @sozud thanks for help figuring this out. Still not satisfied with this though, it looks like it might be an inlined function but I havent figured out a way to make it work

sozud: Might be worth checking the SDK to see if there's some standard inline funcs or macros for VDP1. I think it's fine for now though, we can continue iterating

sozud (approved):

PR #3624, opened 2026-08-25
Decompile `sat/func_06077D88`

gamezter:


src/saturn/game_2a.c
```diff
@@ -1031,13 +1031,14 @@ void SetYH(Primitive* prim, s32 y, s32 h) {
     prim->y2 = prim->y3 = y + h - 1;
 }
 
+static void SetUnk(Primitive* prim, u16* arg1) {
+    prim->unk8 = arg1[0];
+    prim->unkA = arg1[1];
+}
+
```
gamezter, line 1038: I tried using this in higher functions but it started causing mismatches

sozud: These may just be macros

sozud (approved):

PR #3630, opened 2026-08-27
pc: use build cache

Xeeynamo:
Inspired by @sozud work in this commit: https://github.com/Xeeynamo/sotn-decomp/pull/3562/changes/06c8864fcd4d466e23da96c1f4fb054ff821d017 it will avoid rebuilding the whole PC executable at every build, speeding up iterating with it


sozud (approved):
Looks like the core part is the same as the other commit, seems fine

PR #3640, opened 2026-08-30
RNZ1 boss fight manager

bismurphy:
Most of the "new" code for RNZ1 is just the Darkwing Bat fight. This starts the work on that.

I also went back to the Karasuman version and updated/renamed some stuff, to be more in line with the boss fight with Slogra and Gaibon.

Could still use some work on documenting enum steps and such, but for now I think this is okay.

Git shows 59 files changed but most of that is just because symbols have been piling up across the different overlays without having people `make format` them.


Xeeynamo (approved):

src/boss/bo1/st_init_psp.c
```diff
@@ -49,7 +49,7 @@ void InitEntityIds(void) {
     SET_E_ID(EXPLOSION_FLAME);
     SET_E_ID(UNK_23);
     SET_E_ID(HOMING_LASER);
-    SET_E_ID(LIFE_UPSPAWN);
```
JoshSchreuder, line 52: This reminds me a lot of updog 😜

PR #3641, opened 2026-08-30
pc: warp point functionality

sozud:
I'm breaking down the stalled Saturn content PRs into pieces to try and get them upstreamed. This is the warp functionality to spawn where the new doors will be added. Original is https://github.com/Xeeynamo/sotn-decomp/pull/3578 Usage like

 `--spawn-point np3-stage15-door` etc.


Xeeynamo (commented):
We should get `warp.c` renamed to `spawn_point.c` to both be coherent with the cmdline, and to not confuse it with the actual warp overlay.

I am not a fan of `allowed_spawn_points` in `main.c`, or the verbose `--spawn-point` help. What I'd suggest is to move the allowed list of spawn  points directly to the file responsible for it.

Since the code is doing a `O(n)` scrolling the list of allowed spawn points anyway, the `s_SpawnPoints` could just become something like this:

```c
{STAGE_NP3, 1, 192, 870, "np3-stage15-door"},
{STAGE_RNO3, 1, 96, 104, "reverse-trapdoor-underground"},
{STAGE_NO0, 0, 1440, 688, "marble-soul"},
{STAGE_NO4, 5, 176, 396, "caverns-soul"},
{STAGE_RNO0, 0, 64, 132, "reverse-marble-soul"},
{STAGE_RNO4, 5, 208, 132, "reverse-caverns-soul"},
```

That would allow the code to have a single source of truth, instead of needing to maintain and keep in sync `allowed_spawn_points`, `PC_SPAWN_COUNT`, the `s_SpawnPoints` array, and the macros at `warp.h`.

Not for this PR, but at some point it would be nice to see this expanded to something more dynamic. On top of the `--spawn-point <name>`, we could also parse something like `--stage <name> --spawn-point <room_no>` that always spawns at x:128,y:128, or optionally specifying the coordinates with ``--stage <name> --spawn-point <room_no>,<player_x>,<player_y>`. This is to not having to needlessly expand the spawn list for testing rooms that only need a few local iterations.

src/pc/pc.h
```diff
@@ -38,6 +38,7 @@ struct InitGameParams {
     int scale;
     const char* recordPath;
     const char* replayPath;
+    const char* spawnPoint;
```
Xeeynamo, line 41: is this unused?

Xeeynamo (approved):
Left a new minor comment. Cool implementation, this will be very helpful and pay off dividends when debugging. Thanks!

src/pc/pc.h
sozud: It's used unless I'm missing something

% grep -r spawnPoint
./pc.h:    const char* spawnPoint;
./spawn_point.c:    s_Spawn = FindSpawnPoint(params->spawnPoint);
./main.c:    outParams->spawnPoint = NULL;
./main.c:            outParams->spawnPoint = argv[++i];
./main.c:            if (SpawnPoint_StageForName(outParams->spawnPoint) < 0) {
./main.c:    if (params.stage < 0 && params.spawnPoint != NULL) {
./main.c:        params.stage = SpawnPoint_StageForName(params.spawnPoint);

PR #3642, opened 2026-08-30
tools: isolate stage asset packing

sozud:
This is part of the Saturn stage tooling. The point is to be able to go:
```
stitched/unpacked assets -> pack-stage -> build/saturn_stitches/src/<stage>/gen/*
```
And keep the altered psx stage data away from the canonical build.

It also writes packed.stamp so Cmake can track it.


Xeeynamo (commented):
Code looks good as it is.

Conceptually I am still not eager to implement the following in `stage_data.c`:

```c
#ifdef ENABLE_SATURN_STAGES
#include <gen/np3_tilemap_0.h>
```

I am still wondering if it would be best to:
* Surgically patch at runtime from `src/pc/stages/stage_np3.c` and similar. Could end up being fragile.
* Create a `assets.{mod_name}.yaml` that has the sole purpose of generating custom content via `./sotn.sh build-assets`. But then matching build must re-create verbatim files.
* Keep current approach, but only use `ENABLE_SATURN_STAGES` strictly for the necessary generated files. Invasive, but requires no code major code changes.

sozud:
What if we just change the generators to use brackets all the time? I think allowing people to redirect headers helps with modding cleanly. Then ENABLE_SATURN_STAGES is not necessary to adjust the headers.

Xeeynamo:
> What if we just change the generators to use brackets all the time? I think allowing people to redirect headers helps with modding cleanly. Then ENABLE_SATURN_STAGES is not necessary to adjust the headers.

I think it's a great idea. It could also remove the necessity of having `#define GEN_VERSION(file)`, and add to the compiler `-Isrc/st/{STAGE}/gen/{VERSION}` directly. I am also open to move `gen/` folders altogether from `src/` if it helps making everything easier to connect.

sozud:
Now that we have the brackets are there any followups needed here?

Xeeynamo (approved):

PR #3649, opened 2026-09-01
Collision/Effect docs

Xeeynamo:
I am testing a map editor, and I got to test all the various collision types with the game running. I documented all the high confidence effects, paired with some static analysis with the extracted stages to understand how they're actually getting used.


include/game.h
```diff
@@ -1462,28 +1462,29 @@ typedef struct {
 
 typedef enum {
     EFFECT_NONE = 0,
-    EFFECT_SOLID = 1 << 0,
-    EFFECT_UNK_0002 = 1 << 1,
-    EFFECT_QUICKSAND = 1 << 2,
+    EFFECT_SOLID = 1 << 0, // Test floor and ceiling sensors
+    EFFECT_SIDE = 1 << 1,  // Test wall sensors
+    EFFECT_SINK = 1 << 2,  // Lower y by 1px every 4 frames, unused
```
JoshSchreuder, line 1467: This does appear used in the aggregate and also individually in DRA?

include/game.h
```diff
@@ -1462,28 +1462,29 @@ typedef struct {
...
+    EFFECT_UNK_0800 = 1 << 11, // slope related
+    EFFECT_UNK_1000 = 1 << 12, // slope related
+    EFFECT_UNK_2000 = 1 << 13, // slope related
+    EFFECT_UNK_4000 = 1 << 14, // slope related
+    EFFECT_UNK_8000 = 1 << 15, // slope related
 
     // Aggregate helpers below:
-    EFFECT_NOTHROUGH = EFFECT_SOLID | EFFECT_QUICKSAND,
-    EFFECT_NOTHROUGH_PLUS = EFFECT_SOLID | EFFECT_UNK_0002 | EFFECT_QUICKSAND,
+    EFFECT_FULL_SOLID = EFFECT_SOLID | EFFECT_SIDE,
```
JoshSchreuder: This is solid on left right top bottom? Maybe EFFECT_ENCLOSED or SURROUNDED or similar is slightly more descriptive?

JoshSchreuder (approved):
Nice to have these documented!

include/game.h
Xeeynamo: I did not see it referenced in any of the `tiledef` in our code base. But the effect indeed work.

Is it actually used in-game somewhere and I just missed it?

include/game.h
Xeeynamo: You're right about using a better name. May I propose `EFFECT_BLOCK`? And yes, it represents solidity from all sides, and blocks passing through it.

include/game.h
JoshSchreuder: Ah think I misunderstood, I was referring to codebase usage and you to tilemap usage. Guess it's an implemented but unused (in stages) effect

I can't think of anywhere in game with a quicksand like effect

include/game.h
JoshSchreuder: Sounds good 😌

include/game.h
Xeeynamo: This gives me the idea of using `unreferenced` for this, and `unimplemented` for the others. I'll change it before getting this PR in

include/game.h
```diff
@@ -1462,28 +1462,29 @@ typedef struct {
 
 typedef enum {
     EFFECT_NONE = 0,
-    EFFECT_SOLID = 1 << 0,
-    EFFECT_UNK_0002 = 1 << 1,
-    EFFECT_QUICKSAND = 1 << 2,
+    EFFECT_SOLID = 1 << 0, // Test floor and ceiling sensors
+    EFFECT_SIDE = 1 << 1,  // Test wall sensors
```
gamezter, line 1466: Why not `EFFECT_SOLID_X` and `EFFECT_SOLID_Y` or `EFFECT_SOLID_H` and `EFFECT_SOLID_V` for these ? They seem to go together

src/dra/71830.c
```diff
@@ -158,16 +158,16 @@ bool func_80111DE8(bool mistReset) {
...
-        filtered_effects == EFFECT_UNK_4000 + EFFECT_UNK_0800 +
-                                EFFECT_UNK_0002 + EFFECT_SOLID ||
-        filtered_effects == EFFECT_UNK_8000 + EFFECT_UNK_0002 + EFFECT_SOLID ||
-        filtered_effects == EFFECT_UNK_0800 + EFFECT_UNK_0002 + EFFECT_SOLID ||
-        filtered_effects == EFFECT_UNK_0002 + EFFECT_SOLID) {
+        filtered_effects ==
+            EFFECT_UNK_4000 + EFFECT_UNK_0800 + EFFECT_SIDE + EFFECT_SOLID ||
+        filtered_effects == EFFECT_UNK_8000 + EFFECT_SIDE + EFFECT_SOLID ||
+        filtered_effects == EFFECT_UNK_0800 + EFFECT_SIDE + EFFECT_SOLID ||
+        filtered_effects == EFFECT_SIDE + EFFECT_SOLID) {
```
gamezter: can these `EFFECT_... + EFFECT_... ` be converted to  `EFFECT_... | EFFECT_... ` ? If so you could replace `EFFECT_SIDE + EFFECT_SOLID` with `EFFECT_FULL_SOLID`

include/game.h
Xeeynamo: Yes, I was thinking to have `EFFECT_SOLID = EFFECT_SOLID_H + EFFECT_SOLID_V` initially. But I was not fully confident of renaming `EFFECT_SOLID`. For this PR I decided to leave it as it is, but keep the doors open for an eventual rename later. The main reason is the `EFFECT_SOLID` rename would be quite invasive.

For what I observed, it is indeed used to verify the solidity of the floor, but only after a slope or staircase. It basically ignores the sensors on the bottom left and bottom right of the player, so the player would not stumble on an invisible wall when climbing up a steep 45'' slope.

src/dra/71830.c
Xeeynamo: Changing from `+` to `|` changes the codegen on PSP. But weirdly enough, I could replace `EFFECT_SOLID + EFFECT_SIDE` with `EFFECT_BLOCK` without any consequence in the codegen.

Xeeynamo:
Merging this. I'm happy to follow-up with any further changes.

PR #3651, opened 2026-09-03
RNZ1 Darkwing Bat boss

bismurphy:
A fresh new file, instead of the copying from normal NZ1.

This covers the core of the Darkwing Bat. I think his wings are in the next file. Kind of weird.

As usual, did what I could in terms of documenting steps (but not substeps), animation naming, etc, but there is always more progress to be made in documenting these complicated entities.


src/st/rnz1/e_darkwing_bat.c
```diff
@@ -0,0 +1,806 @@
...
+        *rot = target;
+        return 1;
+    }
+    if (*rot > target) {
+        *rot -= increment;
+    }
+    if (*rot < target) {
+        *rot += increment;
+    }
+    return 0;
```
JoshSchreuder: If we're returning `bool` here probably these are clearer as true/false

src/st/rnz1/e_darkwing_bat.c
```diff
@@ -0,0 +1,806 @@
...
+        }
+        xVar = F(other->posX).i.hi + F(other->velocityX).i.hi;
+        yVar = other->posY.i.hi;
+        g_api.CheckCollision(xVar, yVar, &sp2C, 0);
+        if (sp2C.effects & (EFFECT_UNK_0002 | EFFECT_SOLID)) {
+            var_s5 = 1;
+        }
+        if (var_s5 == 0) {
+#else
+        if (var_s5 == 0 && g_PlayableCharacter == PLAYER_ALUCARD) {
```
JoshSchreuder, line 144: Curious what PSX does differently if you're playing as Richter?

src/st/rnz1/e_darkwing_bat.c
```diff
@@ -0,0 +1,806 @@
...
+                self->velocityX -= 0xC00;
+                if (self->velocityX <= FIX(-1.25)) {
+                    self->velocityX = FIX(-1.25);
+                }
+            }
+            other = &PLAYER;
+            yVar = other->posY.i.hi - self->posY.i.hi;
+            yVar += yOffsets[self->ext.darkwing.randTo3];
+            if (yVar < -8) {
+                self->velocityY -= 0x600;
```
JoshSchreuder: FIX

src/st/rnz1/e_darkwing_bat.c
```diff
@@ -0,0 +1,806 @@
...
+                if (self->velocityY <= FIX(-5.0 / 8)) {
+                    self->velocityY = FIX(-0.625);
+                }
+            }
+            if (yVar > 8) {
+                self->velocityY += 0xC00;
```
JoshSchreuder: FIX

src/st/rnz1/e_darkwing_bat.c
```diff
@@ -0,0 +1,806 @@
...
+        case 1:
+            if (!--self->ext.darkwing.timer) {
+                self->ext.darkwing.unk84 = 0;
+                self->animCurFrame = 0xB;
+                other = &PLAYER;
+                xVar = other->posX.i.hi - self->posX.i.hi;
+                yVar = other->posY.i.hi - self->posY.i.hi;
+                angle = ratan2(yVar, xVar);
+                self->velocityX = (rcos(angle) * 0x28000) >> 0xC;
+                self->velocityY = (rsin(angle) * 0x28000) >> 0xC;
```
JoshSchreuder: Perhaps FIX on 0x28000 ?

src/st/rnz1/e_darkwing_bat.c
```diff
@@ -0,0 +1,806 @@
...
+                self->step_s++;
+            }
+            break;
+        case 7:
+            self->animCurFrame = 0x25;
+            self->ext.darkwing.unk84 = 1;
+            self->ext.darkwing.unk85 = 1;
+            self->ext.darkwing.unk86 = 1;
+            self->drawFlags = ENTITY_ROTATE;
+            self->rotate = 0x180;
```
JoshSchreuder, line 563: Seems like a kind of arbitrary value? Perhaps this was a mistype of 180 deg instead

src/st/rnz1/e_darkwing_bat.c
```diff
@@ -0,0 +1,806 @@
...
+
+extern EInit D_us_80180BC4;
+// When you first walk into the room, Darkwing Bat is hanging from a golden
+// perch at the top of the screen. It sticks out, compared to the large
+// structure in the center. When DB is idle before the fight, the perch is
+// part of its sprite, but once the fight starts, we create a static entity
+// that sits there to keep showing the perch.
+void EntityDarkwingBatPerch(Entity* self) {
+    if (!self->step) {
+        InitializeEntity(D_us_80180BC4);
```
JoshSchreuder, line 734: Rename `D_us_80180BC4` `g_EInitDarkwingPerch` or similar?

src/st/rnz1/e_darkwing_bat.c
```diff
@@ -0,0 +1,806 @@
...
+void EntityDarkwingWindDust(Entity* self) {
+    s32 yShift;
+
+    if (!self->step) {
+        InitializeEntity(g_EInitParticle);
+        self->drawFlags = ENTITY_ROTATE | ENTITY_SCALEY | ENTITY_SCALEX;
+        self->palette = 0x8170;
+        self->animSet = 5;
+        self->animCurFrame = 1;
+        self->palette = 0x8195;
```
JoshSchreuder: These two can be `PAL_FLAG` (:747 + :750)

src/st/rnz1/e_darkwing_bat.c
```diff
@@ -0,0 +1,806 @@
...
+    switch (self->step) {
+    case 0:
+        InitializeEntity(g_EInitParticle);
+        self->animSet = 0xE;
+        self->unk5A = 0x79;
+        self->drawFlags = ENTITY_OPACITY;
+        self->blendMode = BLEND_ADD | BLEND_TRANSP;
+        self->palette = 0x2E4;
+        self->facingLeft = Random() & 1;
+        self->velocityX = (Random() << 8) - 0x8000;
```
JoshSchreuder: FIX maybe

src/st/rnz1/e_darkwing_bat.c
bismurphy: It is also used as the init for func_us_801ABDE4 so I don't want to rename it until we know what that is and what it can be named across both. Might just be darkwing2 or something, but we'll rename it once all the uses are decompiled

src/st/rnz1/e_darkwing_bat.c
bismurphy: Perhaps this should be an auto lint

src/st/rnz1/e_darkwing_bat.c
bismurphy: Thanks for all these comments! There's always something I miss so I appreciate the second set of eyes. Should be all good now I believe :)

JoshSchreuder (approved):

PR #3653, opened 2026-09-05
Add autolint for palette flag and definitions

JoshSchreuder:
Suggested by @bismurphy 
https://github.com/Xeeynamo/sotn-decomp/pull/3651#discussion_r3929837237

Added an autofix lint for palettes, both the 0x8000 -> PAL_FLAG macros, and also automapping the `PaletteIndices` enum


include/palettes.h
JoshSchreuder, line 1: Moved these to a small header to simplify the enum loading, otherwise it has to try and preprocess `game.h` which was a bit of a pain

bismurphy: Seems like a great idea, file is nice and self contained

Xeeynamo (approved):

PR #3655, opened 2026-09-05
adding tt_005 and tt_006 overlays

flafmg:
Im not entirely sure if i did everything correctly, sorry if i messed up something im still a bit lost with the tooling

i plan on decompiling both of these in the following weeks

anyways sorry if i messed up something and thank you!!


JoshSchreuder (approved):
Looks good, didn't spot anything obviously wrong.

Would be good to also import these for PSP side too and I recommend using PSP as your basis for matches because O0 is far easier to get a "correct" match on, but can be done in a separate PR if you would like

flafmg:
ok, i can do in this same pr later today (ts already 2AM over here so i need to rest)

Xeeynamo:
Yeah it would be good to have the PSP counterpart as part of this PR as well. @flafmg LMK if you're comfortable to do that, or if you prefer we do it as a follow-up to this merge.

flafmg:
i had some trouble with the configuration but i think i made it work correctly, sorry it took a while i was trying to figure out how to make it work

i hope its all correct, lmk if something is wrong!

Xeeynamo (approved):

PR #3656, opened 2026-09-05
PlayStation Format Angle Transformer & ILLEGAL Lint

hohle:
Adds a transformer for `Entity.rotation` and constant arguments to several SDK functions (`rsin`, `rcos`, etc.) that take PlayStation format angles (4096 = 360°).

Adds a failing linter for the `ILLEGAL` Entity extension.

### DIFF
--- a/src/boss/bo0/e_secrets.c
+++ b/src/boss/bo0/e_secrets.c
@@ -487,7 +487,7 @@ void func_us_801AC894(Entity* self) {
         InitializeEntity(g_EInitEnvironment);
         self->animCurFrame = 0;
         self->drawFlags |= ENTITY_ROTATE;
-        self->rotate = ROT(270.0);
+        self->rotate = ROT(270);
         if (g_CastleFlags[NO2_SECRET_CEILING_OPEN]) {
             for (i = 0; i < 8; i++) {
                 tileIdx = D_us_80180E14[i];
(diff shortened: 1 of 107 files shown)


JoshSchreuder:
Thanks! A ROT lint was going to be next on my list after #3653

bismurphy (approved):
Wow - I had always assumed we had a linter for ILLEGAL. I guess we've just been quite good at staying on top of that one through manual reviews, so good job everyone!

I notice a few of these have things like: 
```
            if (DOPPLEGANGER.velocityX > 0) {
                self->rotate = ROT(11.25);
            } else {
                self->rotate = ROT(348.75);
            }
```

It seems like that second one would be more natural as -11.25. Would it be possible to adjust to do that? Perhaps make all ROT() macros use -180 to +180 rather than 0 to 360? I'm guessing not due to limitations of the preprocessor, but just asking in case.

hohle:
> I notice a few of these have things like:
> 
> ```
>             if (DOPPLEGANGER.velocityX > 0) {
>                 self->rotate = ROT(11.25);
>             } else {
>                 self->rotate = ROT(348.75);
>             }
> ```
> 
> It seems like that second one would be more natural as -11.25. Would it be possible to adjust to do that? Perhaps make all ROT() macros use -180 to +180 rather than 0 to 360? I'm guessing not due to limitations of the preprocessor, but just asking in case.

In polar coordinates those are equivalent, but in the PlayStation format angle representation, they are not. 0x1000 (360°), 0x2000 (720°), etc. have different binary representation. Likewise, -11.25°/`ROT(-11.25)` has a different encoding (0xFFFFFF80) than 348.75°/`ROT(348.75)` (0xF80). I'm not sure how a macro could know the intent to facilitate the correct encoding for a binary match.

hohle:
Rebased with fixes for `master`.

Xeeynamo (approved):
It's important for the codebase to have a high bar for quality and readability. Great work on putting more emphasis and automation around this goal!

PR #3658, opened 2026-09-06
RNZ1 decompile enemies from prior overlays

bismurphy:
This covers everything remaining in RNZ1, except the crusher platforms, which are going to be new code and which I am therefore holding off on.

- water effects uses the existing .h file, no real changes
- Cloaked knight was not in a .h file before, it is now
- Medusa heads needed a slight modification (some extra lines of code to override their spawn behavior), just added into the .h file
- Valhalla Knight requires some tiny append to existing ifdefs, all done

So nothing huge in this one, but means RNZ1 only has one file left!


src/st/e_cloaked_knight.h
```diff
@@ -0,0 +1,445 @@
...
+#ifdef VERSION_PSP
+            angle = self->hitFlags & 3;
+            if (angle) {
+                SetStep(5);
+            }
+#else
+            if (self->hitFlags & 3) {
+                SetStep(5);
+            }
+#endif
```
Xeeynamo: That's sad. I assume you exhausted your tries to unify this?

Xeeynamo (approved):
Great work!

src/st/e_cloaked_knight.h
bismurphy: This was directly copied from the existing cloaked knight, I did not even look at the code, I just moved it to the .h

bismurphy: That said... I have now looked, and i have no ideas. Matching scratches for the two versions:

https://decomp.me/scratch/qq2v1
https://decomp.me/scratch/FjjML

JoshSchreuder: Think this matches both, needed another angle usage further down on PSP
https://decomp.me/scratch/rNhV4
https://decomp.me/scratch/qmpjh

bismurphy: Legendary! Pulled it in here now.

PR #3664, opened 2026-09-09
Decomp four small bo5 functions

Xm3Ga:
Hello! First time contributor here.

Here is a PR decompiling four small bo5 functions: func_us_801A1BA0, func_us_801A4430, CutsceneCameraPan and BO5_RicSetDeadPrologue

Also, I propose changing the names of the two placeholder functions:

func_us_801A1BA0 -> **CheckHippogryphFloorCollision**:

This function handles the Hippogryph enemy not landing, enforcing a fallback: : at world-space coords Y ≥ 184, it sets Y to 183 and zeros both velocities.

func_us_801A4430 -> **EntityFireDamageHitbox**

This function sets dimensions of 24 and 4, with offsets of −48 and −2, and its initializer selects enemy-definition index 0x12E, which deals fire damage.

bo5 matches byte-for-byte, US checks and formatting pass. PSP wasn’t tested locally


bismurphy (approved):
Looks good as a first contribution!

Since BO5 does not contain Richter, the function name is definitely going to need changing, but that can happen once we learn what its true purpose is.

Issue #3667, opened 2026-09-10
Clut extraction for NZ1 does not work


bismurphy:
The assets for NZ1 have a comment that says "Issue #2401" (https://github.com/Xeeynamo/sotn-decomp/blob/915cdd3027eb2c3101947aa2b1d058180c801f9d/config/assets.us.yaml#L970 and https://github.com/Xeeynamo/sotn-decomp/blob/915cdd3027eb2c3101947aa2b1d058180c801f9d/config/splat.us.stnz1.yaml#L43)

Given that #2401 has been resolved I was hoping this would work, but it appears to still be broken. I uncommented A0 in the two yaml files (assets and the main splat for nz1) and got the following error:

`unable to extract asset "palette_def" in "assets/st/nz1": cluts entry at offset 0x12C points to 0xA4000000, which is not a known palette def`

I do not know where this issue is coming from, but wanted to create an issue so it does not get lost.

I am currently working on RNZ1 assets which has the same issue; I will update the comments with this issue number for the future. Not an urgent matter, but wanted to make sure it is not forgotten.

bismurphy:
Same issue exists in RNO1

PR #3668, opened 2026-09-11
Rnz1 assets

bismurphy:
Handles all remaining data for US and PSP, except the cluts/palette_def assets, since those are stuck with #3667 

Otherwise this overlay is mostly done and I will prepare to move to the next one.


bismurphy:
Ah, psp needs ifdef on the unused cmpgfx, will fix

JoshSchreuder (approved):

PR #3670, opened 2026-09-12
bo4: converge some PSP code with PS1

Xeeynamo:


src/boss/bo4/doors.c
```diff
@@ -49,16 +62,12 @@ void EntityUnkId18(Entity* self) {
 #ifdef VERSION_PSP
                 g_Player.padSim = PAD_NONE;
```
Xeeynamo, line 63: I couldn't use `PAD_MIST` here. I am not sure why on PSP it's using `PAD_NONE` 🤷

JoshSchreuder: This ifdef is pretty common across a lot of overlays and scenarios, I am unsure why also.

JoshSchreuder (approved):

PR #3671, opened 2026-09-12
Document DRA camera shaking system, and organize Wolf file

bismurphy:
When I did the crusher in RNZ1, I found a call to a g_api function that shakes the camera vertically when the crusher hits the floor.

There is a small subsystem that handles this. I studied that subsystem and documented it here.

One place this subsystem is used is in the wolf files - if the wolf is sprinting and crashes into a wall, the camera shakes horizontally. I noticed that there was a fake file split in the middle of the various wolf functions, so I decided to un-split those files, and name them usefully to wolfForm.c.

The camera shake function takes a numerical arg that picks one of 7 behaviors; I made an enum for this and changed all calls to use that enum.


JoshSchreuder (approved):
This is great! Makes a number of parts of the code much clearer, great work documenting

PR #3677, opened 2026-09-16
Decompile `sat/func_06005A90`

gamezter:


src/saturn/zero.c
```diff
@@ -1066,25 +1232,25 @@ void func_06007D54(void) {
...
     cmd.control = 0x100A;
     LOW(cmd.ax) = DAT_0600E23C;
     if (SpMstCmdPos < 0x278) {
         SPR_2Cmd(0, &cmd);
-        d_0605AEAC += 0x20;
+        d_0605AEAC++;
     }
 
-    d_0605AEAC = (s32)&DAT_06050684[SpMstCmdPos];
+    d_0605AEAC = &DAT_06050684[SpMstCmdPos];
```
gamezter, line 1319: I noticed `d_0605AEAC` is actually a `SprSpCmd*` here. That means all the `d_0605AEAC += 0x20` were actually `d_0605AEAC++`

src/saturn/zero.c
```diff
@@ -704,7 +704,172 @@ void func_060059F4(void) {
...
+
+const char DAT_06005CA4[] = "S011.CHR";
+
+const char DAT_06005CB0[] = "S011.PTN";
+
+const char DAT_06005CBC[] = "S160.CHR";
+
+const char DAT_06005CC8[] = "S160.PTN";
+
+const char DAT_06005CD4[] = "SWATA.CHR";
```
gamezter: These are weird. They are referenced by an array that is in the data segment far from here. There is probably a function around here that uses that array. Probably not good that the CPU has to jump around memory so much.

JoshSchreuder (approved):

src/saturn/zero.c
```diff
@@ -704,9 +701,245 @@ void func_060059F4(void) {
...
-INCLUDE_ASM("asm/saturn/zero/f_nonmat", f6005E3C, func_06005E3C);
-INCLUDE_ASM("asm/saturn/zero/f_nonmat", f600607C, func_0600607C);
+extern s32 DAT_060389E0[][2];
+extern s32 DAT_060389F4[];
+
+s32 func_06005A90(void) {
+    func_06019FE4(0);
+    switch (DAT_0605D770.unk8) {
+    case 0:
+        ((s32(*)(s32))PlaySfx)(DAT_060389E0[g_PlayableCharacter][0]);
```
JoshSchreuder, line 711: What's the deal with this kinda stuff

src/saturn/zero.c
```diff
@@ -704,9 +701,245 @@ void func_060059F4(void) {
...
+            return "S030.CHR";
+        } else {
+            return "S030.PTN";
+        }
+    }
+    return NULL;
+}
+
+bool func_0600607C(u16 arg0) {
+    if (g_PlayableCharacter != 0 && g_CurrentRoom.stageID == 0x9) {
```
JoshSchreuder, line 909: Do we have eg. PLAYER_ALUCARD / stage ID defines on Saturn?

JoshSchreuder (approved):

src/saturn/zero.c
gamezter: Not yet. I eventually plan to make a PR for this

src/saturn/zero.c
gamezter: the default function signature on saturn returns a `s32`. Here the real signature for `PlaySfx` is defined at the top of the file. Eventually we'll split `zero.c` so that this function does not see the real signature. In the mean time it's a good hint to figure out where the split is

PR #3679, opened 2026-09-20
Big cleanup on RNO1

bismurphy:
This overlay had a lot of weirdness. There were a lot of files (such as for Nova Skeleton, as a random example) where the file contains a few functions, and the file would have a few decompiled, and a few not decompiled, despite the fact that it should have been a simple `#include "../e_nova_skeleton.h"`. This was creating a lot of frustration.

I have gone through and cleaned up what I could. PSP is using US code for all but 4 files (3 if you ignore the standard st_init). Those files represent all the unique entities for the overlay (many of which are directly pulled from the regular NO1).

I did very little touching of the actual code here - mostly just a lot of shuffling things around and putting them in the right places. Not sure how these files got mixed up. I thought we had a pretty standardized process for pulling in overlays that makes it straightforward, but this one had a lot of random unusual behaviors scattered around. It's not all resolved yet (the remaining unk_whatever files still have a hard-to-navigate scattering of decompiled and non-decompiled functions), but I will focus on doing those functions themselves in my next PR. For now, this brings US and PSP into alignment, and uses existing .h files everywhere possible.


src/st/rno1/e_room_bg.c
```diff
@@ -9,15 +9,16 @@ typedef struct {
...
 
 static AnimateEntityFrame anim1[] = {{64, 2}, POSE_END};
 static AnimateEntityFrame anim2[] = {{64, 59}, POSE_END};
 static AnimateEntityFrame anim3[] = {{64, 60}, POSE_END};
 static AnimateEntityFrame anim4[] = {{1, 72}, POSE_END};
-static AnimateEntityFrame anim5[] = {{8, 68}, POSE_END};
-static AnimateEntityFrame anim6_unused[] = {{8, 69}, {8, 70}, POSE_LOOP(0)};
+// Looks like a mistake? Kind of weird here
+static AnimateEntityFrame anim5[] = {
+    {8, 68}, POSE_END, {8, 69}, {8, 70}, POSE_LOOP(0)};
```
JoshSchreuder, line 21: Looks like two separate animations, one unused. `anim6` is notably missing from `BackgroundBlockInit` below

Is this data stripped on PSP?

JoshSchreuder (approved):

src/st/rno1/e_room_bg.c
bismurphy: It is not stripped on PSP. That was what perplexed me. When I left it as before (with anim6_unused), PSP was stripping it. Maybe all the anims are one array or something, hard to say.

JoshSchreuder: I guess most likely it's an array of `AnimateEntityFrame[]` then, and `BackgroundBlockInit` does eg. `anims[0]`, `anims[1]` etc.

Interesting find

src/st/rno1/unk_29A8C.c
```diff
@@ -0,0 +1,60 @@
...
+        }
+    case 1:
+    default:
+        break;
+    }
+}
+
+extern u16 D_us_80180778;
+
+void func_us_80198A18_from_rbo4(Entity* self) {
```
Xeeynamo, line 36: at some point we need to rename these to have more appropriate namings

Xeeynamo (approved):
lgtm

src/st/rno1/unk_29A8C.c
bismurphy: Yes, I intend to search the overlay for `func_` and rename everything, once it is decompiled and understood

